Issue #5, opened 2025-05-29
[MCP SDK] Complete MCP protocol


chr-hertel:
There are a few things we should add support for

 - [ ] Completion
 - [x] Batch requests https://github.com/symfony/ai/pull/8
 - [x] Pagination https://github.com/symfony/ai/pull/6
 - [ ] Resource Templates
 - [ ] Authentication
 - [ ] [Resumability and Redelivery](https://modelcontextprotocol.io/specification/draft/basic/transports#resumability-and-redelivery)
 - [ ] Client
   - [ ] Sampling
   - [ ] Roots

Please announce yourself before starting to work on one of these.

ineersa:
Resource Templates

I'm planning to work on this one during this week, will open PR little bit later.

mjakubowski:
We might want to add "Elicitation" to the Client capabilities, as it was added to the specification in the meantime :)

https://modelcontextprotocol.io/docs/learn/client-concepts#elicitation
https://modelcontextprotocol.io/specification/2025-06-18/client/elicitation

chr-hertel:
Closed due to moving this MCP SDK to https://github.com/modelcontextprotocol/php-sdk.

PR #6, opened 2025-05-29
Add support for pagination

Nyholm:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | https://github.com/symfony/ai/issues/5
| License       | MIT

Just following the protocol. 

Note that there is no real benefit with pagination for our standard Collection implementation

### DIFF
--- a/src/mcp-sdk/src/Capability/Prompt/CollectionInterface.php
+++ b/src/mcp-sdk/src/Capability/Prompt/CollectionInterface.php
@@ -13,10 +13,16 @@ declare(strict_types=1);
 
 namespace Symfony\AI\McpSdk\Capability\Prompt;
 
+use Symfony\AI\McpSdk\Exception\InvalidCursorException;
+
 interface CollectionInterface
 {
     /**
-     * @return MetadataInterface[]
+     * @param int $count the number of metadata items to return
+     *
+     * @return iterable<MetadataInterface>
+     *
+     * @throws InvalidCursorException if no item with $lastIdentifier was found
      */
-    public function getMetadata(): array;
+    public function getMetadata(int $count, ?string $lastIdentifier = null): iterable;
 }
--- a/src/mcp-sdk/src/Capability/ResourceChain.php
+++ b/src/mcp-sdk/src/Capability/ResourceChain.php
@@ -19,6 +19,7 @@ use Symfony\AI\McpSdk\Capability\Resource\MetadataInterface;
 use Symfony\AI\McpSdk\Capability\Resource\ResourceRead;
 use Symfony\AI\McpSdk\Capability\Resource\ResourceReaderInterface;
 use Symfony\AI\McpSdk\Capability\Resource\ResourceReadResult;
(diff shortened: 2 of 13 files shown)


chr-hertel:
> Note that there is no real benefit with pagination for our standard Collection implementation

you mean because everything is still in the collection anyways?

src/mcp-sdk/src/Capability/ResourceChain.php
```diff
@@ -35,9 +35,26 @@ public function __construct(
...
+                $found = true;
+                continue;
+            }
+
+            if ($found) {
+                yield $item;
+                if (--$count <= 0) {
+                    break;
+                }
+            }
```
chr-hertel: how about

```suggestion
            if (false === $found) {
                $found = $item->getUri() === $lastIdentifier;
                continue;
            }

            yield $item;

            if (--$count <= 0) {
                break;
            }
```

Nyholm: Thank you. 

I know I write this almost 3 times. I plan to refactor when I add support for resource templates

chr-hertel: i'm also a fan of abstracting late :+1:

Nyholm:
> you mean because everything is still in the collection anyways?

Yes, But another implementation might have 1000 items and read them from disk or database. They could benefit.

src/mcp-sdk/src/Server/RequestHandler/ToolListHandler.php
```diff
@@ -21,26 +21,37 @@
 final class ToolListHandler extends BaseRequestHandler
 {
     public function __construct(
-        private readonly CollectionInterface $toolCollection,
+        private readonly CollectionInterface $collection,
+        private readonly int $pageSize = 20,
```
chr-hertel, line 25: that introduces paging by default without the possibility to opt out, right?
well, one could set `PHP_INT_MAX` here, but what about allowing null to disable it?

anyhow, totally minor thing - valid to merge anyways

Nyholm: Good point. Let's remember that we should fix this. 

(I created an issue)

src/mcp-sdk/src/Capability/Prompt/CollectionInterface.php
```diff
@@ -13,10 +13,16 @@
 
 namespace Symfony\AI\McpSdk\Capability\Prompt;
```
OskarStark, line 14: Off topic, but what about using Mcp\Sdk as namespace?

Nyholm: No "Symfony" prefix?

That would be a first for Symfony organization. 

I do see they are long an ugly, but it follows Symfony UX' pattern

PR #8, opened 2025-05-29
Support batching

Nyholm:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | #5 
| License       | MIT

We need to add a test or two to this.

### DIFF
--- a/src/mcp-sdk/src/Exception/InvalidInputMessageException.php
+++ b/src/mcp-sdk/src/Exception/InvalidInputMessageException.php
@@ -11,6 +11,6 @@
 
 namespace Symfony\AI\McpSdk\Exception;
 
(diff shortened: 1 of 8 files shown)


src/mcp-sdk/src/Exception/InvalidInputMessageException.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+declare(strict_types=1);
```
OskarStark: Needs to be removed, should be adjusted in the PHP-CS Fixer config

Nyholm: Sorry. 

A rebase will fix the issue.

PR #11, opened 2025-05-29
Catch more exceptions

Nyholm:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | 
| License       | MIT

We should return `-32002` if resource is not found

### DIFF
--- a/src/mcp-sdk/src/Server/JsonRpcHandler.php
+++ b/src/mcp-sdk/src/Server/JsonRpcHandler.php
@@ -89,11 +89,15 @@ readonly class JsonRpcHandler
             } catch (NotFoundExceptionInterface $e) {
                 $this->logger->warning(\sprintf('Failed to create response: %s', $e->getMessage()), ['exception' => $e]);
 
-                yield $this->encodeResponse(Error::methodNotFound($message->id ?? 0, $e->getMessage()));
+                yield $this->encodeResponse(Error::methodNotFound($message->id, $e->getMessage()));
             } catch (\InvalidArgumentException $e) {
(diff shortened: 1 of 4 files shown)


src/mcp-sdk/src/Server/JsonRpcHandler.php
```diff
@@ -77,6 +78,10 @@ public function process(string $message): ?string
                 : $this->encodeResponse($this->handleRequest($message));
         } catch (\DomainException) {
             return null;
+        } catch (ResourceNotFoundException $e) {
+            $this->logger->warning(\sprintf('Failed to find resource: %s', $e->getMessage()), ['exception' => $e]);
+
+            return $this->encodeResponse(new Error($message->id ?? 0, Error::RESOURCE_NOT_FOUND, $e->getMessage()));
```
chr-hertel: from a design point of view this is too high for this specific exception - i would argue that it should be the responsibility of a resource specific handler to convert to an `Error` and the `JsonRpcHandler`  should know nothing about specific capabilities

chr-hertel: at least the `RequestHandlerInterface` should support that by signature

```php
    public function createResponse(Request $message): Response|Error;
``` 

but we could also think of `ErrorResponse`  or sth :thinking:

Nyholm: I agree. 

Thank you for pointing it out.

Nyholm: PR updated

Nyholm:
I also made some classes with public properties "read-only"

PR #12, opened 2025-05-29
Add some more tests and docs

Nyholm:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | 
| License       | MIT

There is one new feature here: A notification can be handled by multiple notification handlers.

### DIFF
--- a/src/mcp-sdk/tests/Server/JsonRpcHandlerTest.php
+++ b/src/mcp-sdk/tests/Server/JsonRpcHandlerTest.php
@@ -1,0 +1,88 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\McpSdk\Tests\Server;
+
+use PHPUnit\Framework\Attributes\CoversClass;
+use PHPUnit\Framework\Attributes\Small;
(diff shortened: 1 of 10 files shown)


src/mcp-sdk/tests/Server/JsonRpcHandlerTest.php
```diff
@@ -0,0 +1,91 @@
...
+        $handlerC->expects($this->once())->method('handle');
+
+        $jsonRpc = new JsonRpcHandler(new Factory(), [], [$handlerA, $handlerB, $handlerC], new NullLogger());
+        $jsonRpc->process(
+            '{"jsonrpc": "2.0", "id": 1, "method": "notifications/foobar"}'
+        );
+    }
+
+    /**
+     * Make sure a single request can NOT be handled by multiple handlers.
```
OskarStark: What about using testdox here?

Nyholm: What do you mean? Just adding this into the method name?

OskarStark: Use

@testdox A single request must ....

I am currently on a phone 📱

chr-hertel:
usually I'd vote for not having the `@throws` tag - only on an interface with explicit description of what it should express.

but that's rather opinionated and style - i just disable that phpstorm inspection because i don't see the value while reading the code and in CI we don't check for it

PR #13, opened 2025-05-30
chore: removing irrelevant or default cs config

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | 
| License       | MIT

* removing strict_types
* PHP 7.1 and PHPUnit 7.5 migration should be irrelevant
* ` .php-cs-fixer.cache`  is the default AFAIK


Nyholm:
Thank you @chr-hertel.

PR #14, opened 2025-06-01
Kicking off Platform, Store and Agent components + integration bundle

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | not yet
| Issues        | 
| License       | MIT

This basically brings in LLM Chain + bundle - split into three components.
**this PR not done yet and the code hasn't seen a runtime**

- [ ] sort dependencies
- [x] bring in examples
- [ ] bring in readme

PR #15, opened 2025-06-06
Integration of LLM Chain

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | 
| License       | MIT

Replaces #14 since this is coming with the git history.

- [x] Splits LLM Chain into three components as packages
- [x] Renames `Chain` component to `Agent` 
- [x] Renames `PhpLlm\LlmChain` into `Symfony\AI`


chr-hertel:
Yes agree, in this case it makes sense not to squash and have a merge commit to preserve the history.

Would you still use `gh` or just hit the button?

fabpot:
Just rebase first to get rid of the merge commits and gh will work fine.

chr-hertel:
the reason for the merge commit here is that I integrated llm-chain with `merge --unrelated-history` on top of what we have here.

another option i see is to use the llm-chain history as base and rebase the history of `symfony/ai` on top of that - at least that would be easier than rebasing those 300 commits that conflict in structure :see_no_evil:

chr-hertel:
@Nyholm any objections from your end if i wipe `main` for now and rebase the current status on top of this here?
at least i'd give it a try - should be easier than the other way around
not sure i'm able to do a force push on main tho :D

fabpot:
If it’s too complicated, don’t bother and just hit the merge button.

chr-hertel:
reworded the big library merge, squashed the Symfony AI changes on top and merged with no commit on main. the best way out for the history without too much of a stress.

done here :)

PR #34, opened 2025-06-22
fix: remove llm chain remains in bundle

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | 
| License       | MIT

Fixes some left overs discovered while integrating the demo app.


OskarStark:
Thank you @chr-hertel.

PR #38, opened 2025-06-27
chore: allow mongodb/mongodb 2.0 (#341)

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | 
| License       | MIT

Cherry picking https://github.com/php-llm/llm-chain/pull/341

### DIFF
--- a/src/store/composer.json
+++ b/src/store/composer.json
@@ -29,12 +29,9 @@
         "symfony/uid": "^6.4 || ^7.1",
         "webmozart/assert": "^1.11"
     },
-    "conflict": {
-        "mongodb/mongodb": "<1.21"
-    },
     "require-dev": {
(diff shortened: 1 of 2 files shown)


src/store/composer.json
```diff
@@ -29,12 +29,9 @@
         "symfony/uid": "^6.4 || ^7.1",
         "webmozart/assert": "^1.11"
     },
-    "conflict": {
```
xabbuh, line 32: Is there a reason not to keep it?

OskarStark: What do you mean, we now support `^1.21 and ^2.0`, so both work.

xabbuh: I mean the fact that we don’t test with MongoDB < 1.21, but we don’t forbid to install it anymore.

Issue #41, opened 2025-06-28
Ideas & Plans for this Project


chr-hertel:
## Idea

This collection of libraries is meant to provide multiple things
* A agent/multi agent framework => Agent Component
* An abstraction layer for AI model (not only LLMs) and inference provider or libraries => Platform Component
* A store integration for RAG and similar pattern => Store Component
* Generalized tooling for chatbots => Chat Component
* A low level MCP library for server and client => MCP SDK

By design these components are intended to be framework and vendor agnostic - even if using Symfony components.
Meaning that specific integrations for models, inference provider or stores are moved to bridges, and additional bundles integrate those into the framework => AI Bundle and MCP Bundle

## Current Status

The main focus currently is to settle public interfaces of this library and polish extension points, with the goal to have a first tagged release, see #877. Extension points to third-party services have been identified and bridges are mostly in place in Platform, Agent, and Store component. To be extracted to standalone packages.
Most work is rather stabilizing the components, leveling up the AI bundle, and incorporating feedback.

### Known Problems (high-level instead of issue-level

* **Platform Component**
  * Redundancy around `ModelClient` and `ResultConverter` implementations
  * Weak adoption of `Capability` concept
  * Error Handling
  * Dynamic model catalog
* **Store Component**
  * Missing abstractions in querying for documents - only vague options
  * Doctrine ORM compatibility
* **Agent Component**
  * `AgentInterface` is hardly providing more than `PlatformInterface`
  * Insights into agent runtime, e.g. feedback on reasoning or progress
  * Tool calling loops

## Future Plans

The AI development is moving fast, and adopting most common and promising trends is important.

### Bigger Picture

**Autonomous Agent**
Agents are not about only executing task, but controlling workflows autonomously. Enabling that planning/reasoning layer in PHP apps with pre-defined pattern and APIs while providing transparency to that runtime, is quite a powerful vision.

**Real-time & Multi-Modal**
Voice and image/video in and out is great for modern and interactive UIs, but currently the support is quite limited and looking at protocols like WebRTC would enable more efficient interactions.

**Generative UI**
[...]

PR #43, opened 2025-06-28
fix: Async response not instanceof ImageResponse in example

smnandre:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no 
| Issues        | Fix #... 
| License       | MIT

In [examples/openai/image-output-dall-e-3.php](https://github.com/smnandre/ai/blob/96c9a47e3e912b4f495d45ee7b6e4a01dd43ec34/examples/openai/image-output-dall-e-3.php#L35)

```php
assert($response instanceof ImageResponse);   //  ❌
```

The response is an `AsyncResponse` that wraps a `ImageResponse` (and forwards methods and properties via __call and __get) but the assertion fails.

Not sure how to explain this simply in the example code.

### DIFF
--- a/examples/openai/image-output-dall-e-3.php
+++ b/examples/openai/image-output-dall-e-3.php
@@ -12,6 +12,7 @@
 use Symfony\AI\Platform\Bridge\OpenAI\DallE;
 use Symfony\AI\Platform\Bridge\OpenAI\DallE\ImageResponse;
 use Symfony\AI\Platform\Bridge\OpenAI\PlatformFactory;
+use Symfony\AI\Platform\Response\AsyncResponse;
 use Symfony\Component\Dotenv\Dotenv;
 
 require_once dirname(__DIR__).'/vendor/autoload.php';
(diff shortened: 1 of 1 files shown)


chr-hertel:
Yes, that's a valid concern - and with some spoilers, it will happen over and over when working with the lib in user land. it's a bit annoying how the `AsyncResponse` is currently wrapped around the others. but i guess i was too lazy in https://github.com/php-llm/llm-chain/pull/141

what i'm doing from time to time:
```php
$response = $platform->request(...);
assert($response instance of AsyncResponse);
$response = $response->unwrap();
assert($response instance of ImageResponse);
```
which is not great DX at all

so your finding here is only the tip of the ice berg and what i listed in #41 as
> Better type support for user land (.e.g. $platform->requestText(...) or $platform->request(...)->asText())

what do you think about the option to call the `Platform` like this to make sure to have the actual `ImageResponse` instead of the `AsyncResponse`:
**a)**
```php
$response = $platform->requestImage(
    model: new DallE(name: DallE::DALL_E_3),
    input: 'A cartoon-style elephant with a long trunk and large ears.',
    options: [
        'response_format' => 'url', // Generate response as URL
    ],
);
``` 

**b)**
```php
$response = $platform->request(
    model: new DallE(name: DallE::DALL_E_3),
    input: 'A cartoon-style elephant with a long trunk and large ears.',
    options: [
        'response_format' => 'url', // Generate response as URL
    ],
)->asImage();
```

smnandre:
```php
$response = $platform->requestImage(
```

Readable and coherent today... but would that scale in the next few months ? 


```php
)->asImage();
```

Same problem in a way ? 

--

I have not enough vision on all the work done here, and still ~~some~~ code and articles/resources to read to have any valid opinion i think..  

...but as you asked 😅 
 my two cents would be: i like the Process api a lot, because one can explicitely call in async or sync.

So what about $platform->asyncRequest(...)  / request(...)

(i do not like the "name" asyncRequest but you see the point i guess :) )

One would return an ImageResponse, the other a Promise<ImageResponse> in a way..

chr-hertel:
the `$platform->asyncRequest(...) / request(...)` wouldn't solve that i can't expect the type on user side.

but you're also right that the other options have a scaling issue ... :thinking: 

WDYT about https://github.com/php-llm/llm-chain/pull/376

chr-hertel:
Closed in favor of #73 - not needed anymore

PR #44, opened 2025-06-29
feat: add Postgres store

smnandre:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Docs?         |  no
| Issues        | Fix #... 
| License       | MIT

Here is a first/naive attempt to add a Postgres store (via pgvector extension)

Very inspired by the [MariaDB store ](https://github.com/symfony/ai/pull/39) PR (and existing examples).

Trying to digg in the repository, please tell me if anything is not in the spirit/style of what you guys building here..

### DIFF
--- a/examples/.env
+++ b/examples/.env
@@ -59,6 +59,9 @@ MONGODB_URI=
 PINECONE_API_KEY=
 PINECONE_HOST=
 
+# For using Postgres (store)
+POSTGRES_URI=pdo-pgsql://postgres:postgres@127.0.1:5432/my_database
+
 # Some examples are expensive to run, so we disable them by default
 RUN_EXPENSIVE_EXAMPLES=false
 
--- a/src/store/src/Bridge/Postgres/Store.php
+++ b/src/store/src/Bridge/Postgres/Store.php
@@ -1,0 +1,162 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Store\Bridge\Postgres;
+
+use Doctrine\DBAL\Connection;
+use Symfony\AI\Platform\Vector\Vector;
+use Symfony\AI\Platform\Vector\VectorInterface;
+use Symfony\AI\Store\Document\Metadata;
+use Symfony\AI\Store\Document\VectorDocument;
+use Symfony\AI\Store\Exception\InvalidArgumentException;
+use Symfony\AI\Store\InitializableStoreInterface;
+use Symfony\AI\Store\VectorStoreInterface;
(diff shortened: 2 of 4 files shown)


src/store/src/Bridge/Postgres/Store.php
```diff
@@ -0,0 +1,165 @@
...
+
+    public function add(VectorDocument ...$documents): void
+    {
+        $statement = $this->connection->prepare(
+            \sprintf(
+                'INSERT INTO %s (id, metadata, %s)
+                 VALUES (:id, :metadata, :vector)
+                 ON CONFLICT (id) DO UPDATE SET metadata = EXCLUDED.metadata, %s = EXCLUDED.%s',
+                $this->tableName,
+                $this->vectorFieldName,
```
OskarStark, line 64: Can we use $2s to not add it 3 times?

src/store/src/Bridge/Postgres/Store.php
```diff
@@ -0,0 +1,165 @@
...
+
+    /**
+     * @param array<string, mixed> $options
+     * @param float|null           $minScore Minimum score to filter results (optional)
+     *
+     * @return VectorDocument[]
+     */
+    public function query(Vector $vector, array $options = [], ?float $minScore = null): array
+    {
+        $limit = $options['limit'] ?? 5;
```
OskarStark: No need for a one time variable

chr-hertel:
Looks really good - thanks @smnandre! Needs a rebase and Oskar's comments - good to merge than

src/store/src/Bridge/Postgres/Store.php
smnandre: TIL :)

smnandre:
Done :)

OskarStark:
to keep the flow straight forward, I would love, if you could squash your commits and create a PR on the original llm-chain lib. We can cherry-pick then afterwards

smnandre:
Wait, so .. i must have two separated forks, work on AI, rename paths but also concepts from one to the other, open the PR on LLM, and then it's done here ? :| 

Could we ease that process a bit?

smnandre:
(rebased and squashed here)

OskarStark:
A right, you must rename the namespaces and so on, didn't thought about this. Ok then nvm, let's merge it here

examples/.env
```diff
@@ -59,6 +59,9 @@ MONGODB_URI=
 PINECONE_API_KEY=
 PINECONE_HOST=
 
+# For using Postgres (store)
+POSTGRES_URI=
```
chr-hertel: I guess we can just add a default value here - nothing secret here :)

```suggestion
POSTGRES_URI=pdo-pgsql://postgres:postgres@127.0.1:5432/my_database
```

chr-hertel:
This is great - thanks for adding that @smnandre :pray:

chr-hertel:
Thanks @smnandre.

PR #45, opened 2025-06-30
chore: Allow arbitrary vector dimensions for MariaDB store

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | 
| Issues        | 
| License       | MIT

Cherry picking https://github.com/php-llm/llm-chain/pull/348

> MariaDB insists the dimensions of the vector being inserted match exactly the dimensions of the vector defined in the schema.

Since the store knows nothing about the Model & it's expected vector dimensions, let's allow passing vector dimensions as a parameter to `initialize()` leave the problem to the user.


OskarStark:
Thank you @chr-hertel.

PR #47, opened 2025-06-30
fix: recursively remove `additionalProperties` from Gemini requests

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | 
| License       | MIT

Cherry picking https://github.com/php-llm/llm-chain/pull/351

> Tool parameters with nested objects cause 400 bad request from Gemini due to only removing the unsupported `additionalProperties` on top-level.

```
Invalid JSON payload received. Unknown name "additionalProperties" at 'tools[0].function_declarations[0].parameters.properties[0].value': Cannot find field.
```


OskarStark:
Thank you @chr-hertel.

PR #52, opened 2025-06-30
feat: Google Gemini server tools

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | 
| License       | MIT

Cherry picking https://github.com/php-llm/llm-chain/pull/356

### DIFF
--- a/examples/google/server-tools.php
+++ b/examples/google/server-tools.php
@@ -1,0 +1,49 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use PhpLlm\LlmChain\Chain\Chain;
+use PhpLlm\LlmChain\Chain\Toolbox\ChainProcessor;
+use PhpLlm\LlmChain\Chain\Toolbox\Tool\Clock;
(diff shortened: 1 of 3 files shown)


examples/google/server-tools.php
```diff
@@ -0,0 +1,49 @@
...
+
+require_once dirname(__DIR__, 2).'/vendor/autoload.php';
+(new Dotenv())->loadEnv(dirname(__DIR__, 2).'/.env');
+
+if (empty($_ENV['GOOGLE_API_KEY'])) {
+    echo 'Please set the GOOGLE_API_KEY environment variable.'.\PHP_EOL;
+    exit(1);
+}
+
+$platform = PlatformFactory::create($_ENV['GOOGLE_API_KEY']);
```
smnandre, line 30: As Google seems to brand Gemini a bit differently than it does for other cloud services, should we use `GEMINI_API_KEY` here ? 


https://github.com/google-gemini/gemini-cli/tree/main?tab=readme-ov-file#quickstart

image


(even if they mention the GOOGLE_API_KEY too here, as "master key" : https://ai.google.dev/gemini-api/docs/api-key?hl=fr )

OskarStark: Will create a PR on LLM chain

OskarStark:
Thank you @chr-hertel.

PR #79, opened 2025-07-07
fix example scripts for docker, class names and env var names

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  |no
| Docs?         | yes
| Issues        | 
| License       | MIT

basically fixing some left overs after cherry picking and merging


examples/composer.json
```diff
@@ -22,7 +22,8 @@
         "symfony/filesystem": "^6.4|^7.0",
         "symfony/finder": "^6.4|^7.0",
         "symfony/process": "^6.4|^7.0",
-        "symfony/var-dumper": "^6.4|^7.0"
+        "symfony/var-dumper": "^6.4|^7.0",
+        "mrmysql/youtube-transcript": "^0.0.5"
```
OskarStark: Should be alpha sorted

examples/README.md
```diff
@@ -25,6 +25,15 @@ corresponding example you want to run.
 
 _Now you can run examples standalone or via the example runner._
 
+#### Store with Docker
+
+Some of the store examples require locally running services, meaning that you need to have Docker installed and running
+to test these examples.
+
+```bash
+docker compose up -d
```
xabbuh, line 41:
```suggestion
cd examples
docker compose up -d
```

chr-hertel: Thanks, yes, we should be explicit here.
I added a section above since this is also true for the `composer install` and example execution part.

OskarStark:
Thank you @chr-hertel.

PR #86, opened 2025-07-11
[AI Store] Remove `webmozart/assert` library

Spomky:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Docs?         | no 
| Issues        | none
| License       | MIT

Hi,

I am not sure w need the `webmozart/assert` dependency for just one verification.
Let me squash this PR before any action (I am not with my usual setup at the moment).

Regards.

### DIFF
--- a/src/platform/src/Bridge/Azure/OpenAI/EmbeddingsModelClient.php
+++ b/src/platform/src/Bridge/Azure/OpenAI/EmbeddingsModelClient.php
@@ -12,12 +12,12 @@
 namespace Symfony\AI\Platform\Bridge\Azure\OpenAI;
 
 use Symfony\AI\Platform\Bridge\OpenAI\Embeddings;
+use Symfony\AI\Platform\Exception\InvalidArgumentException;
 use Symfony\AI\Platform\Model;
 use Symfony\AI\Platform\ModelClientInterface;
 use Symfony\Component\HttpClient\EventSourceHttpClient;
 use Symfony\Contracts\HttpClient\HttpClientInterface;
 use Symfony\Contracts\HttpClient\ResponseInterface;
-use Webmozart\Assert\Assert;
 
 /**
  * @author Christopher Hertel <mail@christopher-hertel.de>
@@ -34,11 +34,11 @@ final readonly class EmbeddingsModelClient implements ModelClientInterface
         #[\SensitiveParameter] private string $apiKey,
     ) {
         $this->httpClient = $httpClient instanceof EventSourceHttpClient ? $httpClient : new EventSourceHttpClient($httpClient);
-        Assert::notStartsWith($baseUrl, 'http://', 'The base URL must not contain the protocol.');
-        Assert::notStartsWith($baseUrl, 'https://', 'The base URL must not contain the protocol.');
-        Assert::stringNotEmpty($deployment, 'The deployment must not be empty.');
-        Assert::stringNotEmpty($apiVersion, 'The API version must not be empty.');
-        Assert::stringNotEmpty($apiKey, 'The API key must not be empty.');
+        !str_starts_with($this->baseUrl, 'http://') || throw new InvalidArgumentException('The base URL must not contain the protocol.');
+        !str_starts_with($this->baseUrl, 'https://') || throw new InvalidArgumentException('The base URL must not contain the protocol.');
+        '' !== $deployment || throw new InvalidArgumentException('The deployment must not be empty.');
+        '' !== $apiVersion || throw new InvalidArgumentException('The API version must not be empty.');
+        '' !== $apiKey || throw new InvalidArgumentException('The API key must not be empty.');
     }
 
     public function supports(Model $model): bool
--- a/src/store/src/Document/TextDocument.php
+++ b/src/store/src/Document/TextDocument.php
@@ -11,8 +11,8 @@
 
 namespace Symfony\AI\Store\Document;
 
+use Symfony\AI\Store\Exception\InvalidArgumentException;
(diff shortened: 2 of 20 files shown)


src/store/src/Document/TextDocument.php
```diff
@@ -24,6 +23,6 @@ public function __construct(
         public string $content,
         public Metadata $metadata = new Metadata(),
     ) {
-        Assert::stringNotEmpty(trim($this->content));
+        '' !== trim($this->content) || throw new \InvalidArgumentException('The content shall not be an empty string');
```
Nyholm: Let's be more explicit and do 

```
if ('' === trim($this->content)) {
  throw new \InvalidArgumentException('The content shall not be an empty string');
}
```

Nyholm (changes requested):
Thank you. I agree

chr-hertel:
yes, agree as well, but for the Platform component you'd need to replace more code - or keep it for now and only take action on store

Spomky:
Yes indeed. I will continue this PR when I get back home on Thursday.

src/store/src/Document/TextDocument.php
Spomky: Hi @Nyholm,

You're absolutely right the `if` notation is more explicit. However in this case I opted for the shorter expression to maintain consistency across classes that perform many inline validations, such as the [With](src/platform/src/Contract/JsonSchema/Attribute/With.php) class.

My goal was to keep the constructor concise and uniform, even if slightly less explicit.
That said, if you feel strongly about it, I will witch to the `if` version for clarity.

src/platform/src/Bridge/Azure/OpenAI/EmbeddingsModelClient.php
```diff
@@ -34,11 +33,11 @@ public function __construct(
...
-        Assert::notStartsWith($baseUrl, 'http://', 'The base URL must not contain the protocol.');
-        Assert::notStartsWith($baseUrl, 'https://', 'The base URL must not contain the protocol.');
-        Assert::stringNotEmpty($deployment, 'The deployment must not be empty.');
-        Assert::stringNotEmpty($apiVersion, 'The API version must not be empty.');
-        Assert::stringNotEmpty($apiKey, 'The API key must not be empty.');
+        !str_starts_with($this->baseUrl, 'http://') || throw new \InvalidArgumentException('The base URL must not contain the protocol.');
+        !str_starts_with($this->baseUrl, 'https://') || throw new \InvalidArgumentException('The base URL must not contain the protocol.');
+        '' !== $deployment || throw new \InvalidArgumentException('The deployment must not be empty.');
+        '' !== $apiVersion || throw new \InvalidArgumentException('The API version must not be empty.');
+        '' !== $apiKey || throw new \InvalidArgumentException('The API key must not be empty.');
```
chr-hertel: Let's us the component specific exception here please: `Symfony\AI\Platform\Exception\InvalidArgumentException`

chr-hertel: same for the other places as well

Spomky: Indeed I had missed this class.
I have updated it wherever necessary.

src/platform/src/Bridge/Azure/OpenAI/EmbeddingsModelClient.php
```diff
@@ -34,11 +34,11 @@ public function __construct(
         #[\SensitiveParameter] private string $apiKey,
     ) {
         $this->httpClient = $httpClient instanceof EventSourceHttpClient ? $httpClient : new EventSourceHttpClient($httpClient);
-        Assert::notStartsWith($baseUrl, 'http://', 'The base URL must not contain the protocol.');
-        Assert::notStartsWith($baseUrl, 'https://', 'The base URL must not contain the protocol.');
-        Assert::stringNotEmpty($deployment, 'The deployment must not be empty.');
-        Assert::stringNotEmpty($apiVersion, 'The API version must not be empty.');
-        Assert::stringNotEmpty($apiKey, 'The API key must not be empty.');
+        !str_starts_with($this->baseUrl, 'http://') || throw new InvalidArgumentException('The base URL must not contain the protocol.');
+        !str_starts_with($this->baseUrl, 'https://') || throw new InvalidArgumentException('The base URL must not contain the protocol.');
```
OskarStark, line 38: we can use an `OR` here

OskarStark:
I will prepare a PR to add unit tests for the code parts using the lib. So we can merge my PR first, then rebase and check the CI 👍

OskarStark:
Here you go:
* https://github.com/symfony/ai/pull/126

chr-hertel:
Merged Oskar's tests, rebase please :)

Spomky:
Thanks. Rebased, squashed and tests are ok.
I noted there are `InvalidArgumentException` classes in `ai-platform`, `agent` and `store`.
As `agent` and `store` depend `on ai-platform`, do you think interesting to remove redundant classes? (not here, but another PR)

chr-hertel:
From a design point of view my understanding was that we want exceptions clustered per package - rather a best practice than tho.
how does `symfony/symfony` handle this? throwing exceptions of a dependency feels a bit odd to me - it's an extra dependency and not explicitly design for that :thinking:

OskarStark:
Yes per package, imagine you are just using one package, the exception will not be available.

chr-hertel:
Thank you @Spomky.

Spomky:
You are welcome.
Noted for the design. Makes sense.

Issue #59, opened 2025-07-01
[AiBundle] OpenAI Embeddings model used no matter the platform defined on the indexed


welcoMattic:
Given this user configuration:

```yaml
ai:
    indexer:
        default:
            platform: 'symfony_ai.platform.mistral'
            model:
                name: 'Embeddings'
                version: 'mistral-embed'
```

If the embeddings model name is "Embeddings" (as is the case for [OpenAI](https://github.com/symfony/ai/blob/main/src/platform/src/Bridge/OpenAI/Embeddings.php), [Mistral](https://github.com/symfony/ai/blob/main/src/platform/src/Bridge/Mistral/Embeddings.php) and [Google](https://github.com/symfony/ai/blob/main/src/platform/src/Bridge/Google/Embeddings.php)), the model object injected in `Embedder` object will always be `Symfony\AI\Platform\Bridge\OpenAI\Embeddings` due to https://github.com/symfony/ai/blob/main/src/ai-bundle/src/DependencyInjection/AIExtension.php#L474

I see 3 solutions to fix this issue:

- Force to (prefix|suffix) Embeddings model name with the platform name (e.g. `MistralEmbeddings`, `OpenAIEmbeddings`)
- Check the platform name to determine the Embeddings class to set in Definition
- Change the `name` config entry to `className` and allow users to pass `Symfony/AI\Platform\Bridge\Mistral\Embeddings` as value. In `AIExtension` we could check the class existence and if it extends `Symfony\AI\Platform\Model`.

@chr-hertel you probably have a clear idea of what is the best solution here, maybe a third one?

---

Mirror issue of https://github.com/php-llm/llm-chain-bundle/issues/98

welcoMattic:
Repost original comment from @chr-hertel (https://github.com/php-llm/llm-chain-bundle/pull/99#issuecomment-3037211345):

> Yup, that first implementation was def too stupid, true. And I guess the className thing is something that should be supported - especially when you thin about user land model classes for example.
> 
> However, looking at this config:
> 
> ```yaml
> llm_chain:
>     embedder:
>         default:
>             platform: 'llm_chain.platform.mistral'
>             model:
>                 name: 'PhpLlm\LlmChain\Platform\Bridge\Mistral\Embeddings'
>                 version: 'mistral-embed'
> ```
> 
[...]

PR #88, opened 2025-07-11
[AiBundle] Use model FQCN for indexer config, to allow any `Symfony\AI\Platform\Model` child class

welcoMattic:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Docs?         | no 
| Issues        | Fix https://github.com/symfony/ai/issues/59 
| License       | MIT

Before:

```yaml
ai:
    indexer:
        default:
            platform: 'symfony_ai.platform.mistral'
            model:
                name: 'Embeddings'
                version: 'mistral-embed'

```

After

```yaml
ai:
    indexer:
        default:
            platform: 'symfony_ai.platform.mistral'
            model:
                class: 'Symfony\AI\Platform\Bridge\Mistral\Embeddings'
                name: !php/const Symfony\AI\Platform\Bridge\Mistral\Embeddings::MISTRAL_EMBED
```

This way, any class extending `Symfony\AI\Platform\Model` can be used as embedder model, even one written by users themselves. 
Embeddings model of providers (OpenAI, Mistral, Gemini, ...) can be named `Embeddings` with no issue.

---

Replace https://github.com/php-llm/llm-chain-bundle/pull/99


welcoMattic:
Repost original comment from @chr-hertel (https://github.com/php-llm/llm-chain-bundle/pull/99#issuecomment-3037211345):

> Yup, that first implementation was def too stupid, true. And I guess the className thing is something that should be supported - especially when you thin about user land model classes for example.
> 
> However, looking at this config:
> 
> ```yaml
> llm_chain:
>     embedder:
>         default:
>             platform: 'llm_chain.platform.mistral'
>             model:
>                 name: 'PhpLlm\LlmChain\Platform\Bridge\Mistral\Embeddings'
>                 version: 'mistral-embed'
> ```
> 
> This is a bit redundant, right? the service `llm_chain.platform.mistral` would anyways only support one embeddings class - I think we could also solve that by the `Platform` exposing which models (and their names) it would support. so that we would basically only chose from a subset of those classes.
> 
> WDYT?

chr-hertel:
Yes, let's do this to unblock the usage for now - we can work on a slimmer config as a next step.

chr-hertel (changes requested):
Let's break it once for both parameters and also the agent config - same issue.

What do you think about?
* `class`  instead of `name` or `className`
* `name` instead of `version` (also rather deprecated)

and I guess the docs and demo need an update as well here

chr-hertel:
Conflicts after merging #94, but rebase should do the trick hopefully

chr-hertel (changes requested):
Two things open from my side:

1. adopting the config of the model in `agent` section, so it is consistent:

https://github.com/symfony/ai/blob/f7e72fae7e047866a03cbe712b10b498f250b63e/src/ai-bundle/config/options.php#L73-L74

2. adopting the change in the demo's configuration, e.g.:

https://github.com/symfony/ai/blob/f7e72fae7e047866a03cbe712b10b498f250b63e/demo/config/packages/ai.yaml#L50-L51

welcoMattic:
@chr-hertel comment addressed, I made the same changes on Agent configuration.
Docs and demo are also fixed.

chr-hertel:
Follow up ideas:
- Generic model classes should be possible
- Is the model a service? feels a bit odd since it looks like a DTO but also carries configuration
- A Platform usually defines already a set of models - why bother to configure it explicitly!?

chr-hertel:
Thank you @welcoMattic.

PR #94, opened 2025-07-11
[AiBundle][McpBundle] Upgrade bundle structure

yceruto:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | Fix #91 
| License       | MIT

This pull request introduces significant changes to the configuration system for AI-related bundles. It replaces the `Configuration.php` files with new `options.php` files, streamlining the definition of configuration options. Additionally, it removes exclusions from PHPStan analysis for the deleted `Configuration.php` files.

### Configuration System Overhaul:
* [`src/ai-bundle/config/options.php`](diffhunk://#diff-cc69af4e9e9e5b26760c9b97f61ea6557fa5545e06d89c17eb8e95c3475e10d9R1-R211): Introduced a new configuration file to define options for AI platforms, agents, stores, and indexers. This replaces the previous `Configuration.php` file with a more concise and flexible structure.
* [`src/mcp-bundle/config/options.php`](diffhunk://#diff-1a31327f5d505b06684dbd1329edca53f16fe3f3535d3ed7cb3a31d6a5f9044dR1-R65): Added a new configuration file for MCP bundle options, including client transport settings and application metadata.

### Removal of Legacy Files:
* [`src/ai-bundle/src/DependencyInjection/Configuration.php`](diffhunk://#diff-6b3b818564647842e4bf70729631c74910a4bdd62890e2acecf07428978ac087L1-L225): Removed the old configuration file that defined options for AI platforms, agents, stores, and indexers. Its functionality has been replaced by the new `options.php` file.

### PHPStan Analysis Updates:
* [`src/ai-bundle/phpstan.dist.neon`](diffhunk://#diff-b99fef75a571df1766ed883ed6618ff528a41b65f4748dd58588e7968610c226L5-L7): Removed the exclusion of `Configuration.php` from PHPStan analysis, as the file has been deleted.
* [`src/mcp-bundle/phpstan.dist.neon`](diffhunk://#diff-b99fef75a571df1766ed883ed6618ff528a41b65f4748dd58588e7968610c226L5-L7): Similarly, removed the exclusion of `Configuration.php` from PHPStan analysis for the MCP bundle.

PR #102, opened 2025-07-12
[AIBundle] Add `#[IsGrantedTool]` for tool access control

valtzu:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | Fix https://github.com/php-llm/llm-chain/issues/360
| License       | MIT

Add `#[IsGrantedTool]` attribute for tool access control with similar behavior as `#[IsGranted]` in `symfony/security-http`.

Moved from https://github.com/php-llm/llm-chain/pull/382

### DIFF
--- a/src/ai-bundle/doc/index.rst
+++ b/src/ai-bundle/doc/index.rst
@@ -180,6 +180,27 @@ To inject only specific tools, list them in the configuration:
                 tools:
                     - 'Symfony\AI\Agent\Toolbox\Tool\SimilaritySearch'
 
+To restrict the access to a tool, you can use the ``IsGrantedTool`` attribute, which
+works similar to ``IsGranted`` attribute in `symfony/security-http`. For this to work,
+make sure you have `symfony/security-core` installed in your project.
+
+::
+
+    use Symfony\AI\Agent\Attribute\IsGrantedTool;
+
+    #[IsGrantedTool('ROLE_ADMIN')]
+    #[AsTool('company_name', 'Provides the name of your company')]
+    final class CompanyName
+    {
+        public function __invoke(): string
+        {
+            return 'ACME Corp.';
+        }
+    }
+The attribute ``IsGrantedTool`` can be added on class- or method-level - even multiple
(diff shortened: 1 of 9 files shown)


chr-hertel (changes requested):
Thanks for bringing it to Symfony AI and switching to the bundle - I think it makes more sense :+1: 

Besides the pipeline, this needs a rebase after #94, and a brief section in the docs of the bundle - it's a powerful new feature users should know about :)

valtzu:
Rebased & added docs.

About PHPStan error, not sure what to do with it, it's about this 3rd parameter: https://github.com/symfony/symfony/blob/7.4/src/Symfony/Component/Security/Core/Authorization/AuthorizationCheckerInterface.php#L27

src/ai-bundle/doc/index.rst
```diff
@@ -180,6 +180,24 @@ To inject only specific tools, list them in the configuration:
...
+    #[IsGrantedTool('ROLE_ADMIN')]
+    #[AsTool('company_name', 'Provides the name of your company')]
+    final class CompanyName
+    {
+        public function __invoke(): string
+        {
+            return 'ACME Corp.';
+        }
+    }
+
```
chr-hertel, line 203: maybe to clarify, we could add:
```suggestion
The attribute ``IsGrantedTool`` can be added on class- or method-level - even multiple
times. If multiple attributes apply to one tool call, a logical AND is used and all access
decisions have to grant access.

```

or similar - if I even got it right

chr-hertel:
> About PHPStan error, not sure what to do with it, it's about this 3rd parameter

my first thought here would be to ignore the error in the `phpstan.dist.neon` with some kind of comment that we can drop it one day - when dropping Symfony 7.4 support?

valtzu:
Thanks, adjusted based on your feedback, pipe finally green :relieved:

chr-hertel (approved):
Great work & great feature! :raised_hands:

chr-hertel:
Thank you @valtzu.

Issue #90, opened 2025-07-11
AIBundle and Mcpbundle should configure services explicitly and expose only meaningful autowirable aliases


stof:
Currently, the bundles do not follow the rules applied in Symfony core (McpBundle does a mix):

- their services are all register using the FQCN as id. This makes any of those concrete implementation available for autowiring (which also means that adding traceable decorators implementing an interface to integrate in the profiler would become a BC break btw).
- they use autowiring and autoconfiguration, which means that the actual service definitions of the bundle can be impacted by autoconfiguration rules defined in projects, which will be a nightmare for people providing support in support channels (which includes myself)

Services should be registered under dedicated IDs (starting with `ai.` or `mcp.` to follow the best practices for shared bundles of using the bundle-specific prefix) and only supported autowirable types should have aliases registered using FQCN.
Services should also be wired explicitly to make maintenance and support easier by being sure that the actual service definition is the one we maintain.

chr-hertel:
Good one, thanks! 👍

OskarStark:
See #165

chr-hertel:
Covered with #163 and #165

PR #103, opened 2025-07-12
[AIBundle] Configure services explicitly

valtzu:
Related to #90

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Issues        | Fix a part of #90
| License       | MIT

Stop using autowiring in the bundle & use named services instead of fqcn.


valtzu:
Hmm, looks like I missed a lot of cases inside `AiExtension`, will need to dig deeper a bit later,

Closing for now not to pollute open PRs

PR #105, opened 2025-07-12
[Store] Add Meilisearch

### LINKED ISSUE #93: Support Meilisearch, at least as a Vector Store
I prefer Meilisearch to ChromaDB for a vector store, super easy to deploy and like ChromaDB, it is open source and free to use.

Meilisearch also supports Semantic Search, and has some experimental features like generative AI (conversational) and just this week they announced support for image similiarity.  In both cases you need an OPEN_AI key for creating the embeddings, but a lot of work is built into the engine.

Regardless of the experimental features, replacing ChromaDB with the Meilisearch Vector Store would be great. 

Thanks for your consideration.

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Docs?         | yes
| Issues        | Fix #93 
| License       | MIT

Hi 👋🏻 

As explained in #93, this PR aim to add `meilisearch` as a vector store, the full integration is not finished (tests to be validated) but the main implementation is ready.

### DIFF
--- a/examples/store/meilisearch-similarity-search.php
+++ b/examples/store/meilisearch-similarity-search.php
@@ -1,0 +1,85 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use Symfony\AI\Agent\Agent;
+use Symfony\AI\Agent\Toolbox\AgentProcessor;
+use Symfony\AI\Agent\Toolbox\Tool\SimilaritySearch;
+use Symfony\AI\Agent\Toolbox\Toolbox;
+use Symfony\AI\Platform\Bridge\OpenAI\Embeddings;
+use Symfony\AI\Platform\Bridge\OpenAI\GPT;
+use Symfony\AI\Platform\Bridge\OpenAI\PlatformFactory;
+use Symfony\AI\Platform\Message\Message;
+use Symfony\AI\Platform\Message\MessageBag;
+use Symfony\AI\Store\Bridge\Meilisearch\Store;
+use Symfony\AI\Store\Document\Metadata;
+use Symfony\AI\Store\Document\TextDocument;
+use Symfony\AI\Store\Document\Vectorizer;
+use Symfony\AI\Store\Indexer;
+use Symfony\Component\Dotenv\Dotenv;
+use Symfony\Component\HttpClient\HttpClient;
+use Symfony\Component\Uid\Uuid;
+
+require_once dirname(__DIR__).'/vendor/autoload.php';
+(new Dotenv())->loadEnv(dirname(__DIR__).'/.env');
+
+if (!isset($_SERVER['OPENAI_API_KEY'], $_SERVER['MEILISEARCH_HOST'], $_SERVER['MEILISEARCH_API_KEY'])) {
+    echo 'Please set OPENAI_API_KEY, MEILISEARCH_API_KEY and MEILISEARCH_HOST environment variables.'.\PHP_EOL;
+    exit(1);
+}
+
+// initialize the store
+$store = new Store(
+    httpClient: HttpClient::create(),
+    endpointUrl: $_SERVER['MEILISEARCH_HOST'],
+    apiKey: $_SERVER['MEILISEARCH_API_KEY'],
+    indexName: 'movies',
+);
+
+// our data
+$movies = [
+    ['title' => 'Inception', 'description' => 'A skilled thief is given a chance at redemption if he can successfully perform inception, the act of planting an idea in someone\'s subconscious.', 'director' => 'Christopher Nolan'],
+    ['title' => 'The Matrix', 'description' => 'A hacker discovers the world he lives in is a simulated reality and joins a rebellion to overthrow its controllers.', 'director' => 'The Wachowskis'],
+    ['title' => 'The Godfather', 'description' => 'The aging patriarch of an organized crime dynasty transfers control of his empire to his reluctant son.', 'director' => 'Francis Ford Coppola'],
+];
+
+// create embeddings and documents
+$documents = [];
+foreach ($movies as $i => $movie) {
+    $documents[] = new TextDocument(
+        id: Uuid::v4(),
+        content: 'Title: '.$movie['title'].\PHP_EOL.'Director: '.$movie['director'].\PHP_EOL.'Description: '.$movie['description'],
+        metadata: new Metadata($movie),
+    );
+}
+
+// initialize the index
+$store->initialize();
+
+// create embeddings for documents
+$platform = PlatformFactory::create($_SERVER['OPENAI_API_KEY']);
+$vectorizer = new Vectorizer($platform, $embeddings = new Embeddings());
+$indexer = new Indexer($vectorizer, $store);
+$indexer->index($documents);
+
+$model = new GPT(GPT::GPT_4O_MINI);
+
+$similaritySearch = new SimilaritySearch($platform, $embeddings, $store);
+$toolbox = Toolbox::create($similaritySearch);
+$processor = new AgentProcessor($toolbox);
+$agent = new Agent($platform, $model, [$processor], [$processor]);
+
+$messages = new MessageBag(
+    Message::forSystem('Please answer all user questions only using SimilaritySearch function.'),
+    Message::ofUser('Which movie fits the theme of technology?')
+);
+$response = $agent->call($messages);
+
+echo $response->getContent().\PHP_EOL;
--- a/src/store/doc/index.rst
+++ b/src/store/doc/index.rst
@@ -40,6 +40,7 @@ You can find more advanced usage in combination with an Agent using the store fo
 * `Similarity Search with MariaDB (RAG)`_
 * `Similarity Search with MongoDB (RAG)`_
 * `Similarity Search with Pinecone (RAG)`_
+* `Similarity Search with Meilisearch (RAG)`_
 
 Supported Stores
 ----------------
@@ -50,6 +51,7 @@ Supported Stores
 * `MongoDB Atlas`_ (requires `mongodb/mongodb` as additional dependency)
 * `Pinecone`_ (requires `probots-io/pinecone-php` as additional dependency)
 * `Postgres`_ (requires `ext-pdo`)
+* `Meilisearch`_
 
 .. note::
 
@@ -86,10 +88,12 @@ This leads to a store implementing two methods::
 .. _`Similarity Search with MariaDB (RAG)`: https://github.com/symfony/ai/blob/main/examples/store/mariadb-similarity-search.php
 .. _`Similarity Search with MongoDB (RAG)`: https://github.com/symfony/ai/blob/main/examples/store/mongodb-similarity-search.php
 .. _`Similarity Search with Pinecone (RAG)`: https://github.com/symfony/ai/blob/main/examples/store/pinecone-similarity-search.php
+.. _`Similarity Search with Meilisearch (RAG)`: https://github.com/symfony/ai/blob/main/examples/store/meilisearch-similarity-search.php
 .. _`Azure AI Search`: https://azure.microsoft.com/products/ai-services/ai-search
 .. _`Chroma`: https://www.trychroma.com/
 .. _`MariaDB`: https://mariadb.org/projects/mariadb-vector/
 .. _`MongoDB Atlas`: https://www.mongodb.com/atlas
 .. _`Pinecone`: https://www.pinecone.io/
 .. _`Postgres`: https://www.postgresql.org/about/news/pgvector-070-released-2852/
+.. _`Meilisearch`: https://www.meilisearch.com/
 .. _`GitHub`: https://github.com/symfony/ai/issues/16
--- a/src/store/src/Bridge/Meilisearch/Store.php
+++ b/src/store/src/Bridge/Meilisearch/Store.php
@@ -1,0 +1,134 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Store\Bridge\Meilisearch;
+
+use Symfony\AI\Platform\Vector\NullVector;
+use Symfony\AI\Platform\Vector\Vector;
+use Symfony\AI\Store\Document\Metadata;
+use Symfony\AI\Store\Document\VectorDocument;
+use Symfony\AI\Store\Exception\InvalidArgumentException;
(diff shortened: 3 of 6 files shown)


examples/store/meilisearch-similarity-search.php
```diff
@@ -0,0 +1,85 @@
...
+(new Dotenv())->loadEnv(dirname(__DIR__).'/.env');
+
+if (!isset($_SERVER['OPENAI_API_KEY'], $_SERVER['MEILISEARCH_HOST'], $_SERVER['MEILISEARCH_API_KEY'])) {
+    echo 'Please set OPENAI_API_KEY, MEILISEARCH_API_KEY and MEILISEARCH_HOST environment variables.'.\PHP_EOL;
+    exit(1);
+}
+
+// initialize the store
+$store = new Store(
+    new NativeHttpClient(),
```
chr-hertel: Could be just `HttpClient::create()` and let the factory decide which implementation to use

src/store/src/Bridge/Meilisearch/Store.php
```diff
@@ -0,0 +1,132 @@
...
+use Symfony\AI\Platform\Vector\Vector;
+use Symfony\AI\Store\Document\Metadata;
+use Symfony\AI\Store\Document\VectorDocument;
+use Symfony\AI\Store\Exception\InvalidArgumentException;
+use Symfony\AI\Store\InitializableStoreInterface;
+use Symfony\AI\Store\VectorStoreInterface;
+use Symfony\Component\Uid\Uuid;
+use Symfony\Contracts\HttpClient\HttpClientInterface;
+
+final readonly class Store implements InitializableStoreInterface, VectorStoreInterface
```
chr-hertel, line 27: you could add

```suggestion
/**
 * @author Guillaume Loulier <guillaume.loulier@hotmail.fr>
 */
final readonly class Store implements InitializableStoreInterface, VectorStoreInterface
```

src/store/src/Bridge/Meilisearch/Store.php
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    public function add(VectorDocument ...$documents): void
+    {
+        $response = $this->request('PUT', 'documents', array_map(
+            fn (VectorDocument $document): array => [$this, 'convertToIndexableArray'], $documents)
+        );
+
+        if (202 !== $response['status']) {
+            throw new \RuntimeException(\sprintf('An error occurred while adding documents to Meilisearch: %s', $response['error']));
```
chr-hertel: please use the `Symfony\AI\Store\Exception\RuntimeException`  instead of the natvie one to enable users to explicitly catch all store component exceptions

src/store/src/Bridge/Meilisearch/Store.php
```diff
@@ -0,0 +1,132 @@
...
+    {
+        $response = $this->request('POST', 'search', [
+            'vector' => $vector->getData(),
+            'hybrid' => [
+                'embedder' => $this->embedder,
+            ],
+        ]);
+
+        if (200 !== $response['status']) {
+            throw new \RuntimeException(\sprintf('An error occurred while querying Meilisearch: %s', $response['error']));
```
chr-hertel: Same here

src/store/src/Bridge/Meilisearch/Store.php
```diff
@@ -0,0 +1,132 @@
...
+        }
+
+        return array_map([$this, 'convertToVectorDocument'], $response['hits']);
```
chr-hertel: could we use this instead?
```suggestion
        return array_map($this->convertToVectorDocument(...), $response['hits']);
```

src/store/src/Bridge/Meilisearch/Store.php
```diff
@@ -0,0 +1,132 @@
...
+
+        $indexCreationResponse = $this->httpClient->request('POST', 'indexes', [
+            'json' => [
+                'uid' => $this->indexName,
+                'primaryKey' => 'id',
+            ],
+        ]);
+
+        if (202 !== $indexCreationResponse->getStatusCode()) {
+            throw new \RuntimeException(\sprintf('An error occurred while creating Meilisearch index: %s', $indexCreationResponse->getContent(false)));
```
chr-hertel: Please use `Symfony\AI\Store\Exception\RuntimeException`

src/store/src/Bridge/Meilisearch/Store.php
```diff
@@ -0,0 +1,132 @@
...
+        private string $embedder = 'default',
+        private string $vectorFieldName = '_vectors',
+    ) {
+    }
+
+    public function add(VectorDocument ...$documents): void
+    {
+        $response = $this->request('PUT', 'documents', array_map(
+            fn (VectorDocument $document): array => [$this, 'convertToIndexableArray'], $documents)
+        );
```
chr-hertel: isn't it the same like?´
```suggestion
        $response = $this->request('PUT', 'documents', array_map($this->convertToIndexableArray(...), documents));
```

src/store/src/Bridge/Meilisearch/Store.php
```diff
@@ -0,0 +1,132 @@
...
+    public function query(Vector $vector, array $options = [], ?float $minScore = null): array
+    {
+        $response = $this->request('POST', 'search', [
+            'vector' => $vector->getData(),
+            'hybrid' => [
+                'embedder' => $this->embedder,
+            ],
+        ]);
+
+        if (200 !== $response['status']) {
```
chr-hertel: maybe I'm a bit confused, but `$response` is basically the `toArray()` of the response body - does the API return a status field as part of the json?
if you'd have a pointer to the api docs, that would help as well

examples/store/meilisearch-similarity-search.php
```diff
@@ -0,0 +1,85 @@
...
+use Symfony\AI\Store\Document\Vectorizer;
+use Symfony\AI\Store\Indexer;
+use Symfony\Component\Dotenv\Dotenv;
+use Symfony\Component\HttpClient\NativeHttpClient;
+use Symfony\Component\Uid\Uuid;
+
+require_once dirname(__DIR__).'/vendor/autoload.php';
+(new Dotenv())->loadEnv(dirname(__DIR__).'/.env');
+
+if (!isset($_SERVER['OPENAI_API_KEY'], $_SERVER['MEILISEARCH_HOST'], $_SERVER['MEILISEARCH_API_KEY'])) {
```
chr-hertel, line 33: the new vars `MEILISEARCH_HOST` and `MEILISEARCH_API_KEY` should go to the `.env`

chr-hertel (changes requested):
Hey @Guikingone, thanks for jumping on that and providing the first commits :) :+1: 

In general it looks good already and i think going with the API directly is a good and lightweight idea :+1: 

I left a couple of comments and was wondering, if we can ease the usage of the example here. were you using the cloud or a docker container of meilisearch?

I think, if reasonable to you as well, it would be great to extend the docker compose setup of the example to bring in meilisearch there.

and please add docs and a test for the Store class

src/store/src/Bridge/Meilisearch/Store.php
Guikingone: You're right, I fixed this issue 🙂

Guikingone:
Hi @chr-hertel 👋🏻 

The PR is fixed, the CI fails due to a zip error (not on unit tests side), the store is improved and tested 🙂

src/store/doc/index.rst
```diff
@@ -50,8 +51,9 @@ Supported Stores
 * `MongoDB Atlas`_ (requires `mongodb/mongodb` as additional dependency)
 * `Pinecone`_ (requires `probots-io/pinecone-php` as additional dependency)
 * `Postgres`_ (requires `ext-pdo`)
+* `Meilisearch`_
 
-.. note::
+.. note::s
```
chr-hertel: looks like a small glitch here
```suggestion
.. note::
```

chr-hertel:
Thanks for the update and especially the docker changes - now I can easily run it, but I get this error while executing the example:
```
php store/meilisearch-similarity-search.php
PHP Fatal error:  Uncaught Symfony\Component\HttpClient\Exception\InvalidArgumentException: Invalid URL: scheme is missing in "indexes". Did you forget to add "http(s)://"? in /home/christopher/Projects/SymfonyAI/repo/examples/vendor/symfony/http-client/HttpClientTrait.php:580
Stack trace:
#0 /home/christopher/Projects/SymfonyAI/repo/examples/vendor/symfony/http-client/HttpClientTrait.php(178): Symfony\Component\HttpClient\CurlHttpClient::resolveUrl()
#1 /home/christopher/Projects/SymfonyAI/repo/examples/vendor/symfony/http-client/CurlHttpClient.php(100): Symfony\Component\HttpClient\CurlHttpClient::prepareRequest()
#2 /home/christopher/Projects/SymfonyAI/repo/src/store/src/Bridge/Meilisearch/Store.php(72): Symfony\Component\HttpClient\CurlHttpClient->request()
#3 /home/christopher/Projects/SymfonyAI/repo/examples/store/meilisearch-similarity-search.php(64): Symfony\AI\Store\Bridge\Meilisearch\Store->initialize()
#4 {main}
  thrown in /home/christopher/Projects/SymfonyAI/repo/examples/vendor/symfony/http-client/HttpClientTrait.php on line 580
```

i'm just using the default values you provided with the `.env` file

Guikingone:
Ah yes, forgot to call `request` method during the index initialization, the clients wasn't fully configured and the error occurred when you've called `initialize`, it should be better now 🙂

chr-hertel:
Yes, that works better - now I'm getting this error:
> Invalid value type at `.hybrid.semanticRatio`: expected a number, but found a string: `"1.0"`

which refers to this: https://github.com/symfony/ai/pull/105/files#diff-969e7ab9dcae7946fdb6f82fcc866bb86e4389e23e3d40b752e3976de997dd09R58

after fixing this, i get the next one:
> Cannot find embedder with name `default`.

Looks like I'm doing something wrong while running the example or is it also failing for you?

Guikingone:
Ah yes, forgot about the initialization of an embedder in `initialize`, I added the dimensions as constructor argument (required as we don't know which model is used by the end user).

chr-hertel:
@Guikingone so the provided example runs locally for you and is functional?

Guikingone:
Yes, works locally (I failed to see the indexes embeddings as my local indexes were already configured) 🙂 

However, I can't run the script in `examples`, I need to create a new file 🤔 

> Thanks for the update and especially the docker changes - now I can easily run it, but I get this error while executing the example:
> 
> ```
> php store/meilisearch-similarity-search.php
> PHP Fatal error:  Uncaught Symfony\Component\HttpClient\Exception\InvalidArgumentException: Invalid URL: scheme is missing in "indexes". Did you forget to add "http(s)://"? in /home/christopher/Projects/SymfonyAI/repo/examples/vendor/symfony/http-client/HttpClientTrait.php:580
> Stack trace:
> #0 /home/christopher/Projects/SymfonyAI/repo/examples/vendor/symfony/http-client/HttpClientTrait.php(178): Symfony\Component\HttpClient\CurlHttpClient::resolveUrl()
> #1 /home/christopher/Projects/SymfonyAI/repo/examples/vendor/symfony/http-client/CurlHttpClient.php(100): Symfony\Component\HttpClient\CurlHttpClient::prepareRequest()
> #2 /home/christopher/Projects/SymfonyAI/repo/src/store/src/Bridge/Meilisearch/Store.php(72): Symfony\Component\HttpClient\CurlHttpClient->request()
> #3 /home/christopher/Projects/SymfonyAI/repo/examples/store/meilisearch-similarity-search.php(64): Symfony\AI\Store\Bridge\Meilisearch\Store->initialize()
> #4 {main}
>   thrown in /home/christopher/Projects/SymfonyAI/repo/examples/vendor/symfony/http-client/HttpClientTrait.php on line 580
> ```
> 
> i'm just using the default values you provided with the `.env` file

How are you running this file? Any error on the autoload? 🤔

chr-hertel:
like this:
image

And the underlying issue that OpenAI returns is
> This model's maximum context length is 128000 tokens. However, your messages resulted in 281690 tokens (281645 in the messages, 45 in the functions). Please reduce the length of the messages or functions.

Due to the fact that the vectors are part of the payload. Might be rather a design flaw of the `SimiliaritySearch` feature than your implementation tho :thinking:

Was just wondering if it's only me, or if the example works for you

Guikingone:
Strange, if I try to run the script from `examples`, here's my error: 

Capture d’écran 2025-07-13 à 19 52 01

Regarding the issue with OpenAI, that seems strange, I'll take a look, maybe I missed something 🤔

chr-hertel:
Ah, I know that one, that should've been documented, sorry - see #112 

The issue is that Composer is installing the dependencies from main currently and therefore your new class is not available. Meaning, after `composer install` you can run `../link`  to bring in your locally changed code

chr-hertel:
Oh, and that's blocked by #108 :see_no_evil: - a bit rough here still :D

src/store/tests/Bridge/Meilisearch/StoreTest.php
```diff
@@ -0,0 +1,208 @@
...
+use Symfony\AI\Store\Document\VectorDocument;
+use Symfony\Component\HttpClient\Exception\ClientException;
+use Symfony\Component\HttpClient\MockHttpClient;
+use Symfony\Component\HttpClient\Response\MockResponse;
+use Symfony\Component\Uid\Uuid;
+
+#[CoversClass(Store::class)]
+final class StoreTest extends TestCase
+{
+    public function testStoreCannotInitializeOnInvalidResponse(): void
```
OskarStark, line 27: Remove all `void` return types in tests please

chr-hertel: but why @OskarStark?

it looks like it's common in `symfony/symfony` but that also supports other versions of PHP and PHPUnit I guess

Guikingone: Yes, not to mention that PHPStan doesn't like when return types are not specified 😄

OskarStark: It's just, that it does not add any value in tests.

chr-hertel: Well, I get that there is no functional value, but same is true for the `public` keyword for test methods.

I get that in combination with supporting various versions and having merge conflicts for nothing but coding best practices is annoying for `symfony/symfony` - but we don't have that struggle here. so for me adding return type declarations is a best practice to opt-out for reasons, but i see no reason here.

i'd vote for keeping them in

OskarStark: Let's do it

Guikingone:
Hi @chr-hertel 👋🏻 

Regarding the issue you mentioned yesterday, here's my local output: 

Capture d’écran 2025-07-14 à 10 31 31

It seems to work "as expected", I got some errors on the Meilisearch instance but this is due to the `initialize` call that it's called each time (if the index is already in place, MS trigger an error in the logs).

chr-hertel:
Thank you @Guikingone.

PR #109, opened 2025-07-13
[Store] Add InMemory

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | None
| License       | MIT

Hi 👋🏻 

This PR aim to introduce an `InMemoryStore` to store vectors (mostly used for tests environments and/or POCs), the implementation is based around `cosine` similarity (others search algorithms can be used but will requires a dedicated interface like `SearchStrategyInterface` or other).

This store also allows to retrieve only a specific amount of items if required.

This store don't use `initialize` as everything is stored in an array.

PS: An optimized approach would be to use `fpow` but it requires `8.4` so ... Well, maybe for later 😄

### DIFF
--- a/examples/store/memory-similarity-search.php
+++ b/examples/store/memory-similarity-search.php
@@ -1,0 +1,75 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use Symfony\AI\Agent\Agent;
+use Symfony\AI\Agent\Toolbox\AgentProcessor;
+use Symfony\AI\Agent\Toolbox\Tool\SimilaritySearch;
+use Symfony\AI\Agent\Toolbox\Toolbox;
+use Symfony\AI\Platform\Bridge\OpenAI\Embeddings;
+use Symfony\AI\Platform\Bridge\OpenAI\GPT;
+use Symfony\AI\Platform\Bridge\OpenAI\PlatformFactory;
+use Symfony\AI\Platform\Message\Message;
+use Symfony\AI\Platform\Message\MessageBag;
+use Symfony\AI\Store\Document\Metadata;
+use Symfony\AI\Store\Document\TextDocument;
+use Symfony\AI\Store\Document\Vectorizer;
+use Symfony\AI\Store\Indexer;
+use Symfony\AI\Store\InMemoryStore;
+use Symfony\Component\Dotenv\Dotenv;
+use Symfony\Component\Uid\Uuid;
+
+require_once dirname(__DIR__).'/vendor/autoload.php';
+(new Dotenv())->loadEnv(dirname(__DIR__).'/.env');
+
+if (!isset($_SERVER['OPENAI_API_KEY'])) {
+    echo 'Please set OPENAI_API_KEY environment variable.'.\PHP_EOL;
+    exit(1);
+}
+
+// initialize the store
+$store = new InMemoryStore();
+
+// our data
+$movies = [
+    ['title' => 'Inception', 'description' => 'A skilled thief is given a chance at redemption if he can successfully perform inception, the act of planting an idea in someone\'s subconscious.', 'director' => 'Christopher Nolan'],
+    ['title' => 'The Matrix', 'description' => 'A hacker discovers the world he lives in is a simulated reality and joins a rebellion to overthrow its controllers.', 'director' => 'The Wachowskis'],
+    ['title' => 'The Godfather', 'description' => 'The aging patriarch of an organized crime dynasty transfers control of his empire to his reluctant son.', 'director' => 'Francis Ford Coppola'],
+];
+
+// create embeddings and documents
+foreach ($movies as $i => $movie) {
+    $documents[] = new TextDocument(
+        id: Uuid::v4(),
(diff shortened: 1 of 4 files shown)


src/store/src/InMemoryStore.php
```diff
@@ -0,0 +1,76 @@
...
+final class InMemoryStore implements VectorStoreInterface
+{
+    /**
+     * @var VectorDocument[]
+     */
+    private array $documents;
+
+    public function add(VectorDocument ...$documents): void
+    {
+        $this->documents = array_values($documents);
```
Nyholm: Are you meant to overwrite the existing documents?

src/store/src/InMemoryStore.php
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    public function query(Vector $vector, array $options = [], ?float $minScore = null): array
```
Nyholm, line 49: Let's document the supported arguments in $options

src/store/src/InMemoryStore.php
Guikingone: You're right, wrong refactoring

src/store/src/InMemoryStore.php
Guikingone: Fixed, should be better 🙂

examples/store/memory-similarity-search.php
```diff
@@ -0,0 +1,75 @@
...
+        id: Uuid::v4(),
+        content: 'Title: '.$movie['title'].\PHP_EOL.'Director: '.$movie['director'].\PHP_EOL.'Description: '.$movie['description'],
+        metadata: new Metadata($movie),
+    );
+}
+
+// create embeddings for documents
+$platform = PlatformFactory::create($_SERVER['OPENAI_API_KEY']);
+$vectorizer = new Vectorizer($platform, $embeddings = new Embeddings());
+$indexer = new Indexer($vectorizer, $store);
```
OskarStark, line 59: No need for a store variable, we can initialize here, right?

OskarStark (approved):

examples/store/memory-similarity-search.php
Guikingone: Actually, the variable is used at line `64` in the constructor of `SimilaritySearch` 🤔

src/store/src/InMemoryStore.php
```diff
@@ -0,0 +1,83 @@
...
+    /**
+     * @var VectorDocument[]
+     */
+    private array $documents;
+
+    public function add(VectorDocument ...$documents): void
+    {
+        foreach ($documents as $document) {
+            $this->documents[] = $document;
+        }
```
chr-hertel: I think we could use `array_push` instead of iterating over all documents to add:
```suggestion
        array_push($this->documents, $documents);
```
see https://www.php.net/array_push

src/store/src/InMemoryStore.php
```diff
@@ -0,0 +1,83 @@
...
+     * } $options If maxItems is provided, only the top N results will be returned
+     */
+    public function query(Vector $vector, array $options = [], ?float $minScore = null): array
+    {
+        $currentEmbeddings = array_map(
+            callback: fn (VectorDocument $vectorDocument): array => [
+                'distance' => $this->cosineSimilarity($vectorDocument, $vector),
+                'document' => $vectorDocument,
+            ],
+            array: $this->documents,
```
chr-hertel: named arguments don't add value here or do I miss something?

```suggestion
            fn (VectorDocument $vectorDocument): array => [
                'distance' => $this->cosineSimilarity($vectorDocument, $vector),
                'document' => $vectorDocument,
            ],
            $this->documents,
```

chr-hertel: same for the others down below

chr-hertel (changes requested):
This is super cool - found only smaller items :+1:

Thanks already @Guikingone!

Guikingone:
Hi @chr-hertel 

This PR is fixed with your reviews, the `add` method has been improved with `array_push` (better than a loop, I agree) 🙂

src/store/src/InMemoryStore.php
Nyholm: Thank you. Please add a test for this.

Guikingone: This part has been refactored to: 

```php
    public function add(VectorDocument ...$documents): void
    {
        array_push($this->documents, ...$documents);
    }
```

The test exist in `InMemoryStoreTest`.

chr-hertel (approved):
Looks good to me, but needs a rebase :+1:

Guikingone:
Rebase done 👍🏻

Guikingone:
The unit tests are failing due to zip archive, nothing to do with the code.

Guikingone:
@OskarStark @chr-hertel Is it okay for you if we delay the approval/merge of this PR? 

I might have an idea to support more sorting algorithms without adding complexity to the API.

Guikingone:
Ok, should be better (thanks to `match` 😄 ), ready to review when you have time @chr-hertel @OskarStark.

PS: Sorry for the previous ping.

OskarStark:
> PS: Sorry for the previous ping.

No need to be sorry 😃 thanks for your dedication to this project 🙏

OskarStark (approved):

chr-hertel (approved):

chr-hertel:

PR #111, opened 2025-07-13
feat(store): qdrant store started

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | None
| License       | MIT

Hi 👋🏻 

Alright, last one for today, here's the implementation for `Qdrant`, this one's tricky as the default search route is deprecated, the implementation is based on https://qdrant.tech/documentation/quickstart/#run-a-query, the whole store use the HTTP API along with the API key.

Some options are supported (not all as some depends on the query you're building), more can be added if needed.


Guikingone:
Oops, forgot about testing the extra options, more tests to come (works locally but not added in tests suite).

src/store/src/Bridge/Qdrant/Store.php
```diff
@@ -0,0 +1,162 @@
...
+     *     offset?: positive-int,
+     *     with_vectors?: bool,
+     *     with_payload?: bool,
+     * } $options
+     */
+    public function query(
+        Vector $vector,
+        array $options = [],
+        ?float $minScore = null,
+    ): array {
```
OskarStark: one line please

src/store/src/Bridge/Qdrant/Store.php
```diff
@@ -0,0 +1,162 @@
...
+    /**
+     * @param array<string, mixed> $payload
+     *
+     * @return array<string, mixed>
+     */
+    private function request(
+        string $method,
+        string $endpoint,
+        array $payload,
+    ): array {
```
OskarStark: one line please

src/store/tests/Bridge/Qdrant/StoreTest.php
```diff
@@ -0,0 +1,173 @@
...
+use Symfony\AI\Store\Document\VectorDocument;
+use Symfony\Component\HttpClient\Exception\ClientException;
+use Symfony\Component\HttpClient\MockHttpClient;
+use Symfony\Component\HttpClient\Response\MockResponse;
+use Symfony\Component\Uid\Uuid;
+
+#[CoversClass(Store::class)]
+final class StoreTest extends TestCase
+{
+    public function testStoreCannotInitializeOnInvalidResponse(): void
```
OskarStark, line 27:
```suggestion
    public function testStoreCannotInitializeOnInvalidResponse()
```

no `void`, same for the other testcases, as Symfony does not use it

src/store/src/Bridge/Qdrant/Store.php
Guikingone: Style fixed

src/store/src/Bridge/Qdrant/Store.php
```diff
@@ -0,0 +1,137 @@
...
+        return array_merge([
+            'id' => $document->id->toRfc4122(),
+            'vector' => $document->vector->getData(),
+        ], $document->metadata->getArrayCopy());
+    }
+
+    /**
+     * @param array<string, mixed> $data
+     */
+    private function convertToVectorDocument(array $data): VectorDocument
```
chr-hertel, line 127: the `$data` array by default contains `id`, `version` and `score`, right?

I'd say we shouldn't map everything of it to metadata - just saw that we did the same with #105 - but this potentially bloats what we send to the LLM with the similarity search.

Metadata should only be a bag of additional data attributes, not all - and not those which we can map to explicit properties of the `VectorDocument`

Guikingone:
Hum, strange, here's the output I get locally: 

Capture d’écran 2025-07-15 à 10 16 22

I'll take a look at the metadata 🤔

Guikingone:
Hi @chr-hertel 👋🏻 

I found the issue, I forgot to add the `with_vector` as a hard option in the query, the vectors weren't retrieved properly (my screenshot was on my stashed version 😄 ).

I also added an additional check on the collection existence to prevent double HTTP call, here's my local output:

Capture d’écran 2025-07-15 à 10 16 22

The tests have been improved too.

examples/compose.yaml
```diff
@@ -22,3 +22,10 @@ services:
           MEILI_MASTER_KEY: "${MEILISEARCH_MASTER_KEY:-changeMe}"
       ports:
           - "7700:7700"
+
+  qdrant:
+      image: qdrant/qdrant
+      environment:
+          QDRANT__SERVICE__API_KEY: '${QDRAT__SERVICE__API_KEY:-changeMe}'
```
welcoMattic:
```suggestion
          QDRANT__SERVICE__API_KEY: '${QDRANT__SERVICE__API_KEY:-changeMe}'
```

Why double underscore on "our" env var side?

Guikingone: The idea was to respect the "format" from Qdrant, I'm not locked to respect it if that's the question 😄

OskarStark: No need for double underscores 🙌

Guikingone: You're right, fixed 🙂

src/store/src/Bridge/Qdrant/Store.php
chr-hertel: Unfortunately this is not resolved, but my comment said "we should map everything of it to metadata" ... but I meant "should**n't**" --- my bad, sorry - fixed it!

please have a look at https://github.com/symfony/ai/pull/137 for reference.

examples/store/qdrant-similarity-search.php
```diff
@@ -0,0 +1,84 @@
...
+
+// our data
+$movies = [
+    ['title' => 'Inception', 'description' => 'A skilled thief is given a chance at redemption if he can successfully perform inception, the act of planting an idea in someone\'s subconscious.', 'director' => 'Christopher Nolan'],
+    ['title' => 'The Matrix', 'description' => 'A hacker discovers the world he lives in is a simulated reality and joins a rebellion to overthrow its controllers.', 'director' => 'The Wachowskis'],
+    ['title' => 'The Godfather', 'description' => 'The aging patriarch of an organized crime dynasty transfers control of his empire to his reluctant son.', 'director' => 'Francis Ford Coppola'],
+];
+
+// create embeddings and documents
+foreach ($movies as $movie) {
```
chr-hertel: please switch to the new `Movies::all()` fixture method

src/store/tests/Bridge/Qdrant/StoreTest.php
```diff
@@ -0,0 +1,220 @@
...
+    {
+        $httpClient = new MockHttpClient([
+            new MockResponse(json_encode([
```
OskarStark: JsonMockResponse?

src/store/tests/Bridge/Qdrant/StoreTest.php
```diff
@@ -0,0 +1,220 @@
...
+        self::expectException(ClientException::class);
+        self::expectExceptionMessage('HTTP 400 returned for "http://localhost:6333/collections/test".');
+        self::expectExceptionCode(400);
+        $store->initialize();
+    }
+
+    public function testStoreCannotInitializeOnExistingCollection(): void
+    {
+        $httpClient = new MockHttpClient([
+            new MockResponse(json_encode([
```
OskarStark: same here

src/store/tests/Bridge/Qdrant/StoreTest.php
```diff
@@ -0,0 +1,220 @@
...
+
+        $store->initialize();
+
+        self::assertSame(1, $httpClient->getRequestsCount());
+    }
+
+    public function testStoreCanInitialize(): void
+    {
+        $httpClient = new MockHttpClient([
+            new MockResponse(json_encode([
```
OskarStark: same here

src/store/tests/Bridge/Qdrant/StoreTest.php
```diff
@@ -0,0 +1,220 @@
...
+
+        $store->initialize();
+
+        self::assertSame(2, $httpClient->getRequestsCount());
+    }
+
+    public function testStoreCannotAddOnInvalidResponse(): void
+    {
+        $httpClient = new MockHttpClient([
+            new MockResponse(json_encode([]), [
```
OskarStark: same here

src/store/tests/Bridge/Qdrant/StoreTest.php
```diff
@@ -0,0 +1,220 @@
...
+        self::expectException(ClientException::class);
+        self::expectExceptionMessage('HTTP 400 returned for "http://localhost:6333/collections/test/points".');
+        self::expectExceptionCode(400);
+        $store->add(new VectorDocument(Uuid::v4(), new Vector([0.1, 0.2, 0.3])));
+    }
+
+    public function testStoreCanAdd(): void
+    {
+        $httpClient = new MockHttpClient([
+            new MockResponse(json_encode([
```
OskarStark: same here

src/store/tests/Bridge/Qdrant/StoreTest.php
```diff
@@ -0,0 +1,220 @@
...
+
+        $store->add(new VectorDocument(Uuid::v4(), new Vector([0.1, 0.2, 0.3])));
+
+        self::assertSame(1, $httpClient->getRequestsCount());
+    }
+
+    public function testStoreCannotQueryOnInvalidResponse(): void
+    {
+        $httpClient = new MockHttpClient([
+            new MockResponse(json_encode([]), [
```
OskarStark: same here

src/store/tests/Bridge/Qdrant/StoreTest.php
```diff
@@ -0,0 +1,220 @@
...
+        self::expectException(ClientException::class);
+        self::expectExceptionMessage('HTTP 400 returned for "http://localhost:6333/collections/test/points/query".');
+        self::expectExceptionCode(400);
+        $store->query(new Vector([0.1, 0.2, 0.3]));
+    }
+
+    public function testStoreCanQuery(): void
+    {
+        $httpClient = new MockHttpClient([
+            new MockResponse(json_encode([
```
OskarStark: same here

src/store/tests/Bridge/Qdrant/StoreTest.php
```diff
@@ -0,0 +1,220 @@
...
+                'http_code' => 200,
+            ]),
+        ], 'http://localhost:6333');
+
+        $store = new Store(
+            $httpClient,
+            'http://localhost:6333',
+            'test',
+            'test',
+        );
```
OskarStark: one line, please also for the others, thanks

examples/compose.yaml
OskarStark: I still se two or is it only on my end?

Guikingone: Yes, forgot about the one in `compose.yaml`

examples/store/qdrant-similarity-search.php
Guikingone: Fixed

src/store/tests/Bridge/Qdrant/StoreTest.php
Guikingone: Agree, didn't thought about it 😄

Guikingone:
Hi @chr-hertel 

Regarding https://github.com/symfony/ai/pull/111#discussion_r2208958011, I see the problem, I'll solve it (sorry, can't answer on your comment 🙁 ).

Guikingone:
Alright, here's the fix for the metadata, I'm facing an issue with them:

Capture d’écran 2025-07-16 à 11 25 01

If I remove the `vectors` and the `id` from the metadata, the model cannot formulate an answer (the first screenshot is without the score, second without vectors and the score) 🤔 

@chr-hertel If you have an idea, I'm open to discuss about it, the behavior seems to be different from MS.

OskarStark:

Thanks

src/store/src/Bridge/Qdrant/Store.php
```diff
@@ -0,0 +1,145 @@
...
+
+    /**
+     * @return array<string, mixed>
+     */
+    private function convertToIndexableArray(VectorDocument $document): array
+    {
+        return array_merge([
+            'id' => $document->id->toRfc4122(),
+            'vector' => $document->vector->getData(),
+        ], $document->metadata->getArrayCopy());
```
chr-hertel: In the end we need the metadata to be looped through the store, if that's not possible with the engine, we would need to use the `id` to hydrate the record from another store. Luckily Qdrant has the concept of `payload`, that we can use here:
```suggestion
        return [
            'id' => $document->id->toRfc4122(),
            'vector' => $document->vector->getData(),
            'payload' => $document->metadata->getArrayCopy(),
        ];
```
See https://api.qdrant.tech/api-reference/points/upsert-points

src/store/src/Bridge/Qdrant/Store.php
```diff
@@ -0,0 +1,145 @@
...
+            : new Vector($data['vector']);
+
+        $score = $data['score'] ?? null;
+
+        unset($data['score']);
+
+        return new VectorDocument(
+            id: Uuid::fromString($id),
+            vector: $vector,
+            metadata: new Metadata($data),
```
chr-hertel: And we can use that `payload` here to populate the metadata again to the document -> later used by the `SimilaritySearch` to provide information to the LLM:
```suggestion
            metadata: new Metadata($data['payload']),
```

(the `unset($data['score'])` is not needed anymore then)

chr-hertel (changes requested):
Had a look and found the issue I guess, see comments below - hope that helps :)

src/store/src/Bridge/Qdrant/Store.php
Guikingone: Ah, clever, I saw the `payload` option but thought it was related to the "content" of the document.

src/store/src/Bridge/Qdrant/Store.php
Guikingone: Improved, thanks for the information about `payload` 🙂

chr-hertel (approved):

chr-hertel:
Thanks @Guikingone for working on this feature, this is much appreciated.

PR #113, opened 2025-07-13
refactor: replace composer suggests with runtime checks

### LINKED ISSUE #76: No suggestions in composer.json
We don't do that in Symfony

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  |no
| Docs?         | no
| Issues        | Fix #76 
| License       | MIT

Cherry picking https://github.com/php-llm/llm-chain/pull/385

> Remove the `suggest` section from composer.json and add explicit runtime checks
that throw exceptions with helpful composer require commands when optional
dependencies are missing.
> 
> This provides better developer experience by:
> - Giving clear error messages when optional dependencies are needed
> - Providing the exact composer command to install missing packages
> - Following Symfony's pattern for handling optional dependencies
> 
> Optional dependencies now throw exceptions instead of silently failing when the required packages are not installed.

### DIFF
--- a/src/agent/composer.json
+++ b/src/agent/composer.json
@@ -43,14 +43,6 @@
         "symfony/event-dispatcher": "^6.4 || ^7.1",
         "symfony/http-foundation": "^6.4 || ^7.1"
     },
-    "suggest": {
-        "mrmysql/youtube-transcript": "For using the YouTube transcription tool.",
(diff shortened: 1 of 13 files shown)


OskarStark (approved):

OskarStark:
Thank you @chr-hertel.

Issue #16, opened 2025-06-07
Models, Platforms & Stores Support


chr-hertel:
# Models, Platforms & Stores
Potential targets for supported models, platforms and stores:

## Language Models
* [x] [GPT](https://platform.openai.com/docs/models/overview)
* [x] [Anthropic Claude](https://www.anthropic.com/claude)
* [x] [Google Gemini](https://cloud.google.com/vertex-ai/generative-ai/docs/learn/models#gemini-models)
* [ ] [Google Gemma](https://cloud.google.com/vertex-ai/generative-ai/docs/learn/models#gemma-models)
* [x] [Meta Llama](https://www.llama.com/)
* [x] [Mistral](https://docs.mistral.ai/getting-started/models/models_overview/)
* [ ] [Microsoft Phi](https://azure.microsoft.com/en-us/products/phi)
* [x] [DeepSeek](https://www.deepseek.com/)
* [x] [Perplexity Sonar](https://docs.perplexity.ai/getting-started/overview)
* [ ] [Cohere](https://docs.cohere.com/docs/models)

## Embeddings Models
* [x] [OpenAI Text Embeddings](https://platform.openai.com/docs/guides/embeddings/embedding-models)
* [x] [Voyage Embeddings](https://docs.voyageai.com/docs/embeddings)
* [ ] [Gecko](https://cloud.google.com/vertex-ai/generative-ai/docs/learn/models#models)
* [x] [Mistral Embed](https://docs.mistral.ai/capabilities/embeddings/)

## Platforms
* [x] [OpenAI](https://platform.openai.com/docs/overview)
* [x] [Anthropic](https://www.anthropic.com/)
* [x] [Voyage](https://www.voyageai.com/) 
* [x] [Azure](https://azure.microsoft.com/en-gb/solutions/ai/) (only GPT right now)
* [ ] [GitHub Marketplace](https://github.com/marketplace/models) (basically Azure?)
* [x] [Google Cloud](https://cloud.google.com/model-garden)
* [x] [Amazon Bedrock](https://aws.amazon.com/bedrock/)
* [x] [Ollama](https://ollama.com/)
* [x] [Huggingface](https://huggingface.co/)
* [x] [Replicate](https://replicate.com/explore)
* [x] [OpenRouter](https://openrouter.ai/)
* [x] [Perplexity](https://www.perplexity.ai/)
* [ ] [Cohere](https://docs.cohere.com/reference/chat)
[...]

PR #115, opened 2025-07-14
[Platform] Add Groq support

dave1010:
Contributes to #16

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | Contributes to #16
| License       | MIT

### DIFF
--- a/examples/groq/chat.php
+++ b/examples/groq/chat.php
@@ -33,4 +33,4 @@ $response = $agent->call($messages, [
     'max_tokens' => 500, // specific options just for this call
 ]);
 
-echo $response->getContent().PHP_EOL;
+echo $response->getContent().\PHP_EOL;
--- a/examples/groq/chat.php
+++ b/examples/groq/chat.php
@@ -1,0 +1,36 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use Symfony\AI\Agent\Agent;
+use Symfony\AI\Platform\Bridge\Groq\Llama;
+use Symfony\AI\Platform\Bridge\Groq\PlatformFactory;
+use Symfony\AI\Platform\Message\Message;
+use Symfony\AI\Platform\Message\MessageBag;
+use Symfony\Component\Dotenv\Dotenv;
+
+require_once dirname(__DIR__, 2).'/vendor/autoload.php';
+(new Dotenv())->loadEnv(dirname(__DIR__).'/.env');
+
+$platform = PlatformFactory::create($_SERVER['GROQ_API_KEY']);
+$model = new Llama(Llama::LLAMA3_70B, [
+    'temperature' => 0.5, // default options for the model
+]);
+
+$agent = new Agent($platform, $model);
+$messages = new MessageBag(
+    Message::forSystem('You are a pirate and you write funny.'),
+    Message::ofUser('What is the Symfony framework?'),
+);
+$response = $agent->call($messages, [
+    'max_tokens' => 500, // specific options just for this call
+]);
+
+echo $response->getContent().PHP_EOL;
--- a/src/platform/doc/index.rst
+++ b/src/platform/doc/index.rst
@@ -72,8 +72,16 @@ usually defined by the specific models and their documentation.
 * **Language Models**
     * `OpenAI's GPT`_ with `OpenAI`_ and `Azure`_ as Platform
     * `Anthropic's Claude`_ with `Anthropic`_ and `AWS Bedrock`_ as Platform
-    * `Meta's Llama`_ with `Azure`_, `Ollama`_, `Replicate`_ and `AWS Bedrock`_ as Platform
+    * `Meta's Llama`_ with `Azure`_, `Ollama`_, `Replicate`_, `AWS Bedrock`_ and `Groq`_ as Platform
+    * `Google's Gemma`_ with `Groq`_ as Platform
+    * `Mistral's Mixtral`_ with `Groq`_ as Platform
+    * `Alibaba's Qwen`_ with `Groq`_ as Platform
+
+.. note::
+
+    While Groq supports various models, specific API requirements or limitations might apply. Refer to Groq's official documentation for details.
     * `Google's Gemini`_ with `Google`_ and `OpenRouter`_ as Platform
+    * `Groq's Llama`_ with `Groq`_ as Platform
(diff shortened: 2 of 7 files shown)


examples/groq/chat.php
```diff
@@ -0,0 +1,36 @@
...
+use Symfony\AI\Platform\Message\Message;
+use Symfony\AI\Platform\Message\MessageBag;
+use Symfony\Component\Dotenv\Dotenv;
+
+require_once dirname(__DIR__, 2).'/vendor/autoload.php';
+(new Dotenv())->loadEnv(dirname(__DIR__).'/.env');
+
+$platform = PlatformFactory::create($_SERVER['GROQ_API_KEY']);
+$model = new Llama(Llama::LLAMA3_70B, [
+    'temperature' => 0.5, // default options for the model
```
OskarStark, line 24: as this is the default, can we remove it in the example?

src/platform/doc/index.rst
```diff
@@ -72,8 +72,16 @@ usually defined by the specific models and their documentation.
 * **Language Models**
     * `OpenAI's GPT`_ with `OpenAI`_ and `Azure`_ as Platform
     * `Anthropic's Claude`_ with `Anthropic`_ and `AWS Bedrock`_ as Platform
-    * `Meta's Llama`_ with `Azure`_, `Ollama`_, `Replicate`_ and `AWS Bedrock`_ as Platform
+    * `Meta's Llama`_ with `Azure`_, `Ollama`_, `Replicate`_, `AWS Bedrock`_ and `Groq`_ as Platform
+    * `Google's Gemma`_ with `Groq`_ as Platform
+    * `Mistral's Mixtral`_ with `Groq`_ as Platform
+    * `Alibaba's Qwen`_ with `Groq`_ as Platform
```
OskarStark, line 78: Maybe we should use some kind of table for this?

src/platform/src/Bridge/Groq/Llama.php
```diff
@@ -0,0 +1,46 @@
...
+namespace Symfony\AI\Platform\Bridge\Groq;
+
+use Symfony\AI\Platform\Capability;
+use Symfony\AI\Platform\Model;
+
+/**
+ * @author Christopher Hertel <mail@christopher-hertel.de>
+ * @author Dave Hulbert <dave1010@gmail.com>
+ */
+class Llama extends Model
```
chr-hertel, line 21: could be something more neutral like
```suggestion
class Model extends BaseModel
```

since this is not only Llama it's a bit confusing - it's rather a collection of models available on Groq

examples/groq/chat.php
```diff
@@ -0,0 +1,36 @@
...
+use Symfony\AI\Platform\Bridge\Groq\Llama;
+use Symfony\AI\Platform\Bridge\Groq\PlatformFactory;
+use Symfony\AI\Platform\Message\Message;
+use Symfony\AI\Platform\Message\MessageBag;
+use Symfony\Component\Dotenv\Dotenv;
+
+require_once dirname(__DIR__, 2).'/vendor/autoload.php';
+(new Dotenv())->loadEnv(dirname(__DIR__).'/.env');
+
+$platform = PlatformFactory::create($_SERVER['GROQ_API_KEY']);
```
chr-hertel, line 22: when introducing a new env var, please add it to `.env` file (without value in this case) and check if the variable is set before - see other examples

examples/groq/chat.php
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+use Symfony\AI\Agent\Agent;
+use Symfony\AI\Platform\Bridge\Groq\Llama;
+use Symfony\AI\Platform\Bridge\Groq\PlatformFactory;
+use Symfony\AI\Platform\Message\Message;
+use Symfony\AI\Platform\Message\MessageBag;
+use Symfony\Component\Dotenv\Dotenv;
+
+require_once dirname(__DIR__, 2).'/vendor/autoload.php';
```
chr-hertel, line 19: The 2 can be removed here:

```suggestion
require_once dirname(__DIR__).'/vendor/autoload.php';
```

chr-hertel: after this change the example works for me :+1:

src/platform/doc/index.rst
chr-hertel: I agree, this doesn't scale well - not necessarily scope of this PR, but I'm in need for better ideas - even feels like a three dimensional table :laughing:

src/platform/tests/Bridge/Groq/LlamaTest.php
chr-hertel, line 1: Please split the test class per tested class

src/platform/tests/Bridge/Groq/LlamaTest.php
```diff
@@ -0,0 +1,74 @@
...
+use Symfony\AI\Platform\Capability;
+use Symfony\AI\Platform\Model;
+use Symfony\Component\HttpClient\MockHttpClient;
+use Symfony\Component\HttpClient\Response\MockResponse;
+
+final class LlamaTest extends TestCase
+{
+    /**
+     * @covers \Symfony\AI\Platform\Bridge\Groq\Llama
+     */
```
chr-hertel, line 27: please use `#[CoversClass(Llama::class)]` attribute instead

chr-hertel (changes requested):
Welcome to this repository @dave1010 - and thanks a ton for working on this! New bridges are always welcome!

@OskarStark and me added some comments, but I think mostly minor stuff - so great work already :) :+1: 

Please always make sure to run provided example classes yourself locally - to make sure they are functional. You can find some pointers here: https://github.com/symfony/ai/blob/main/examples/README.md

examples/groq/chat.php
dave1010: Ah, thanks. Somehow I got in a bit of a mess with doing a composer install in the monorepo!

src/platform/src/Bridge/Groq/Llama.php
dave1010: Thanks. Yeah, I'll update. Started with Llama, then I realised they have others now.

OskarStark:

Thanks

dave1010:
Will do. Thanks.

Sorry, haven't had a chance to get back to this yet.

OskarStark:
Would you like to finish this PR?

chr-hertel:
Hi @dave1010, do you plan to continue with this?

dave1010:
Hi @OskarStark @chr-hertel 

Apologies, I was on holiday then completely forgot about this.

It looks like the project has moved on, so I'll close the PR. It looks like it'd be easy enough to just change the connection host name anyway, rather than needing a specific provider class.

PR #116, opened 2025-07-14
[Platform] Add lmstudio

lubiana:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | 
| License       | MIT

Add Support for [LM Studio](https://lmstudio.ai/docs/app) as Platform. As it is mostly OpenAI compatible I picked the parts from that platform and copied them into the LM Studio namespace with some adjustments.

What I am unsure about is if this is the best way to implement this. Basically LM Studio ist mostly API compatible with the OpenAI API, so instead of copying the relevant parts from the OpenAI platform into the LM Studio one another approach could be to to allow setting a different url for the OpenAI platform. But in that case the model classes there might need to be not final, so a user can extend those and add capabilities for those models.

On the other hand it might be better to keep this approach, as the OpenAI and LM Studio APis might diverge further in the future.

As this is my first contribution to symfony i would be happy to get some feedback. I am especially unsure if my approach of wrapping the GPT Responseconverter in `src/platform/src/Bridge/LMStudio/Completions/ResponseConverter.php` is a wise choice.


examples/lmstudio/image-input-binary.php
```diff
@@ -0,0 +1,41 @@
...
+
+require_once dirname(__DIR__).'/vendor/autoload.php';
+
+$platform = PlatformFactory::create();
+$model = new Completions(
+    name: 'gemma-3-4b-it-qat',
+    capabilities: [
+        ...Completions::DEFAULT_CAPABILITIES,
+        Capability::INPUT_IMAGE,
+    ]
```
OskarStark: one line like in the other examples please

OskarStark:
I would not add all these examples, I think one is enough

lubiana:
> I would not add all these examples, I think one is enough

I have removed all the examples except for chat and image input as the later shows how to add additional capabilites.

src/platform/src/Bridge/LMStudio/Completions.php
```diff
@@ -0,0 +1,37 @@
+<?php
+
+declare(strict_types=1);
+
```
chr-hertel: Please remove this

src/platform/doc/index.rst
```diff
@@ -85,6 +85,7 @@ usually defined by the specific models and their documentation.
 * **Other Models**
     * `OpenAI's Dall·E`_ with `OpenAI`_ as Platform
     * `OpenAI's Whisper`_ with `OpenAI`_ and `Azure`_ as Platform
+    * Most open source Models with `LM Studio`_ as Platform
```
chr-hertel: we could link to the model catalog of LM Studio here:
https://lmstudio.ai/models

examples/lmstudio/image-input-binary.php
```diff
@@ -0,0 +1,38 @@
...
+use Symfony\AI\Platform\Message\Content\Image;
+use Symfony\AI\Platform\Message\Message;
+use Symfony\AI\Platform\Message\MessageBag;
+
+require_once dirname(__DIR__).'/vendor/autoload.php';
+
+$platform = PlatformFactory::create();
+$model = new Completions(
+    name: 'gemma-3-4b-it-qat',
+    capabilities: [...Completions::DEFAULT_CAPABILITIES, Capability::INPUT_IMAGE]
```
chr-hertel, line 32: nice idea :+1:

examples/lmstudio/image-input-binary.php
```diff
@@ -0,0 +1,38 @@
...
+use Symfony\AI\Platform\Bridge\LMStudio\Completions;
+use Symfony\AI\Platform\Bridge\LMStudio\PlatformFactory;
+use Symfony\AI\Platform\Capability;
+use Symfony\AI\Platform\Message\Content\Image;
+use Symfony\AI\Platform\Message\Message;
+use Symfony\AI\Platform\Message\MessageBag;
+
+require_once dirname(__DIR__).'/vendor/autoload.php';
+
+$platform = PlatformFactory::create();
```
chr-hertel: let's please bring in a env var for the hostUrl, similar to ollama example.

you could add in `.env` then:
```
LMSTUDIO_HOST_URL=http://localhost:1234
```

like that it would be easy to configure a different port for me locally when running the examples :+1:

chr-hertel (changes requested):
Hi @lubiana, thanks for driving this - adding support for LM Studio would be awesome!

I added some smaller remarks, but for now I need to rely on you to test the examples locally - didn't set up LM Studio yet

src/platform/doc/index.rst
lubiana: As LM Studio allows to fetch models from their own catalog as well as from huggingface i linked both of them now.

examples/lmstudio/image-input-binary.php
lubiana: Added the env var. Should i also add lmstudio in processPlatformConfig in the AIBundle as well as the models in the agent and indexer config processing part there, or should that be left for a seperate pr?

chr-hertel: so the beauty of mono repo is that we can do both - but it would be more complete if you'd add this here right away :)

lubiana: great. i  have added the service configuration in the AIBundle component

src/ai-bundle/config/options.php
```diff
@@ -57,6 +57,11 @@
                             ->scalarNode('api_key')->isRequired()->end()
                         ->end()
                     ->end()
+                    ->arrayNode('lmstudio')
+                        ->children()
+                            ->scalarNode('host_url')->end()
```
chr-hertel: we should also have the default here:
```suggestion
                            ->scalarNode('host_url')->defaultValue('http://127.0.0.1:1234')->end()
```

src/ai-bundle/src/AIBundle.php
```diff
@@ -255,6 +256,24 @@ private function processPlatformConfig(string $type, array $platform, ContainerB
...
+        if ('lmstudio' === $type) {
+            $platformId = 'symfony_ai.platform.lmstudio';
+            $definition = (new Definition(Platform::class))
+            ->setFactory(LMStudioPlatformFactory::class.'::create')
+                ->setAutowired(true)
+                ->setLazy(true)
+                ->addTag('proxy', ['interface' => PlatformInterface::class])
+                ->addTag('symfony_ai.platform');
+
+            if (isset($platform['host_url'])) {
```
chr-hertel: with a default this will always be set

OskarStark:
Can you please add some tests? Thanks

lubiana:
> Can you please add some tests? Thanks

I have added unit tests for the modelclients and responseconverters

chr-hertel (approved):
This looks like it's done to me - good one! :+1: 

Like I said before, I rely on @lubiana here for the functionality of the examples - cannot set LM Studio up right now.

OskarStark:

Thanks

chr-hertel:
Alright @lubiana, one last rebase + addition in changelog, than we merge it in :+1:

lubiana:
i have added lmstudio to the platforms CHANGELOG.md. Thanks for the reviews and support so far

src/platform/tests/Bridge/LMStudio/Embeddings/ResponseConverterTest.php
```diff
@@ -0,0 +1,92 @@
...
+#[UsesClass(Embeddings::class)]
+class ResponseConverterTest extends TestCase
+{
+    #[Test]
+    public function itConvertsAResponseToAVectorResponse(): void
+    {
+        $response = $this->createStub(ResponseInterface::class);
+        $response
+            ->method('toArray')
+            ->willReturn(json_decode($this->getEmbeddingStub(), true));
```
OskarStark: No need need for a method here, just use the json here

chr-hertel:
Thank you @lubiana.

PR #117, opened 2025-07-14
[Agent] Add basic setup for memory injections to system prompt

DZunke:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        |  
| License       | MIT

See original contribution at https://github.com/php-llm/llm-chain/pull/387

> This PR introduces a flexible memory system that allows the LLM to recall contextual information that are permantent for the conversation. In difference to tools it can be always utilized, when use_memory option is not disabled. It would be possible to fetch memory by a tool with a system instruction like always call the tool foo_bar but, for me, this feels like a bad design to always force the model to do a tool call without further need.
>
> Currently i have added just two memory providers to show what my idea is. I could also think about a write layer to fill a memory, together with a read layer, this could, for example, be working good with a graph database and tools for memory handling. But these are ideas for the future.
>
> So for this first throw i hope you get what i was thinking about to reach. I decided to inject the memory to the system prompt, when it is available, instead of adding a second system prompt to the message bag. But this was just a 50/50 thinking. I tried both seem to be working equally, at least for open ai. I am open to change it back again.
>
> What do you think?

### DIFF
--- a/examples/store/mariadb-chat-memory.php
+++ b/examples/store/mariadb-chat-memory.php
@@ -1,0 +1,80 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use Doctrine\DBAL\DriverManager;
+use Doctrine\DBAL\Tools\DsnParser;
+use Symfony\AI\Agent\Agent;
+use Symfony\AI\Agent\Memory\EmbeddingProvider;
+use Symfony\AI\Agent\Memory\MemoryInputProcessor;
+use Symfony\AI\Platform\Bridge\OpenAI\Embeddings;
+use Symfony\AI\Platform\Bridge\OpenAI\GPT;
+use Symfony\AI\Platform\Bridge\OpenAI\PlatformFactory;
+use Symfony\AI\Platform\Message\Message;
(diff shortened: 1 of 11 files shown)


examples/store/mariadb-chat-memory.php
```diff
@@ -0,0 +1,80 @@
...
+use Symfony\AI\Store\Bridge\MariaDB\Store;
+use Symfony\AI\Store\Document\Metadata;
+use Symfony\AI\Store\Document\TextDocument;
+use Symfony\AI\Store\Document\Vectorizer;
+use Symfony\AI\Store\Indexer;
+use Symfony\Component\Dotenv\Dotenv;
+use Symfony\Component\Uid\Uuid;
+
+require_once dirname(__DIR__, 2).'/vendor/autoload.php';
+(new Dotenv())->loadEnv(dirname(__DIR__, 2).'/.env');
```
chr-hertel: small change here missing:
```suggestion
require_once dirname(__DIR__).'/vendor/autoload.php';
(new Dotenv())->loadEnv(dirname(__DIR__).'/.env');
```

chr-hertel (changes requested):
Thanks for moving this - that saves me some time :pray: 

Found one small left over in the example, and please remove `declare(strict_types=1);` - then we're good to merge this :)

OskarStark:
> please remove declare(strict_types=1); - then we're good to merge this :)

Can we update the PHP-CS-Fixer config?

chr-hertel:
Thank you @DZunke.

PR #122, opened 2025-07-15
[Platform] Add `TokenOutputProcessor` for `Mistral`

Renrhaf:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| License       | MIT

Add a TokenOutputProcessor for the Mistral platform, like it exists for the OpenAI one.

### DIFF
--- a/src/platform/src/Bridge/Mistral/TokenOutputProcessor.php
+++ b/src/platform/src/Bridge/Mistral/TokenOutputProcessor.php
@@ -1,0 +1,58 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Platform\Bridge\Mistral;
+
+use Symfony\AI\Agent\Output;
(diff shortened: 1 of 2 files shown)


OskarStark:
please add unit tests for the new processor, thanks

Renrhaf:
Hi @OskarStark, tests are added, thanks !

src/platform/src/Capability.php
```diff
@@ -33,4 +33,7 @@ class Capability
 
     // FUNCTIONALITY
     public const TOOL_CALLING = 'tool-calling';
+
+    // METADATA
+    public const METADATA_USAGE = 'metadata-usage';
```
chr-hertel: what's your reasoning/motivation to add it as capability?

at least it's not used at this point, not sure about it and wonder if that's really a model capability.
I'd say we remove this for now - if you don't have something concrete in mind

Renrhaf: Hi @chr-hertel it was a first step towards the integration in the AI Bundle as I saw some code was relying on capabilities.
I'll remove it as a first step, and we'll see later for that !

chr-hertel (approved):

chr-hertel:
Thank you @Renrhaf.

PR #125, opened 2025-07-15
Add PHPStan rule forbidding native exceptions and switch remaining ones

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | 
| License       | MIT

This is only a POC by Claude, but wanted to check how easy it is since I'm already catching this as recurring pattern while reviewing incoming PRs.

WDYT?

### DIFF
--- a/.phpstan/ForbidNativeExceptionRule.php
+++ b/.phpstan/ForbidNativeExceptionRule.php
@@ -1,0 +1,135 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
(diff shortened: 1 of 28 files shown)


valtzu:
I wonder if it'd be possible to use cs fixer instead, to auto-replace `\LogicException` with `Symfony\AI\Agent\Exception\LogicException` under `src/agent` etc

Regardless, definitely a good idea to add it to the pipe :100:

OskarStark:
It is possible with rector from what I know

PR #129, opened 2025-07-15
[Platform] Add missing Anthropic models

### LINKED ISSUE #128: New Anthropic models needs to be added
Based on:
https://docs.anthropic.com/en/docs/about-claude/models/overview#model-aliases

OskarStark:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | Fix #128 
| License       | MIT

### DIFF
--- a/src/platform/src/Bridge/Anthropic/Claude.php
+++ b/src/platform/src/Bridge/Anthropic/Claude.php
@@ -20,12 +20,15 @@ use Symfony\AI\Platform\Model;
 class Claude extends Model
 {
     public const HAIKU_3 = 'claude-3-haiku-20240307';
-    public const HAIKU_35 = 'claude-3-5-haiku-20241022';
+    public const HAIKU_35 = 'claude-3-5-haiku-latest';
     public const SONNET_3 = 'claude-3-sonnet-20240229';
-    public const SONNET_35 = 'claude-3-5-sonnet-20240620';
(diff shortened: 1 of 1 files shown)


src/platform/src/Bridge/Anthropic/Claude.php
```diff
@@ -21,17 +21,24 @@ class Claude extends Model
 {
     public const HAIKU_3 = 'claude-3-haiku-20240307';
     public const HAIKU_35 = 'claude-3-5-haiku-20241022';
+    public const HAIKU_35_LATEST = 'claude-3-5-haiku-latest';
```
OskarStark: I would propose to use the model with `-latest` suffix as `HAIKU_35` and maybe remove the `public const HAIKU_35 = 'claude-3-5-haiku-20241022';` entirely then? 

WDYT @chr-hertel

chr-hertel: good one, was thinking about how to handle subversions of models as well, but if we'd use `HAIKU_35 = 'claude-3-5-haiku-latest'`  we wouldn't need to care about changes here: if Anthropic changes were the pointer is going, we would go with that change. that's simple and honest. i like that.
users can pin a specific version themselves.

OskarStark:
Good to merge from my side

chr-hertel:
Thank you @OskarStark.

PR #130, opened 2025-07-15
[Platform] Do not throw exception if tools are not used

Renrhaf:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| License       | MIT

When using a model which does not supports tools, an exception is thrown even when I don't intend to use any.

### DIFF
--- a/src/agent/src/Toolbox/AgentProcessor.php
+++ b/src/agent/src/Toolbox/AgentProcessor.php
@@ -46,6 +46,11 @@ final class AgentProcessor implements InputProcessorInterface, OutputProcessorIn
 
     public function processInput(Input $input): void
     {
+        $options = $input->getOptions();
+        if (!isset($options['tools'])) {
+            return;
+        }
+
         if (!$input->model->supports(Capability::TOOL_CALLING)) {
             throw MissingModelSupportException::forToolCalling($input->model::class);
         }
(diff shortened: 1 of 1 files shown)


Renrhaf:
The behavior of throwing an exception blocks requests when we do not intend to use tools, is it really wanted ? 
Ignoring the processor by returning earlier seems more logical to me but I might be wrong, what do you think ?

src/agent/src/Toolbox/AgentProcessor.php
```diff
@@ -47,7 +47,7 @@ public function __construct(
     public function processInput(Input $input): void
     {
         if (!$input->model->supports(Capability::TOOL_CALLING)) {
```
OskarStark, line 49: This exception basically says: "If you are providing tools and the model you are using is not capable of using tools, we should inform the developer."

To me your change looks not correct

valtzu: I agree, if you don't intend to use tools and are using the bundle, just add `tools: false` to the chain. We could consider adding a build-time check into the bundle for that?

If you're not using the bundle, then just don't use this `Toolbox\AgentProcessor` when no tools are needed.

Renrhaf: OK I'll try that, thanks for your feedback. Closing for now !

PR #134, opened 2025-07-15
[AIBundle] Remove security listener when missing dependencies

### LINKED ISSUE #133: Is the security bundle missing in demos ?
Hi,

It seems the security bundle is missing from the composer.json file in the demos. The following error is thrown when running composer install:

"Cannot autowire service symfony_ai.security.is_granted_attribute_listener: argument $authChecker of method Symfony\AI\AIBundle\Security\EventListener\IsGrantedToolAttributeListener::__construct() has type Symfony\Component\Security\Core\Authorization\AuthorizationCheckerInterface, but this class was not found."

I fixed it by running composer require symfony/security-bundle.

valtzu:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | Fix #133
| License       | MIT

Make `security-core` optional + also register the attribute for autoconfiguration & throw to make sure no one tries to use the attribute without the listener (which would make one think that you have some access control when you don't).


chr-hertel:
That was crazy fast, thanks @valtzu!

PR #136, opened 2025-07-15
[Platform] Use JSON Path to convert responses

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | 
| License       | MIT

This approach combines two ideas:
1. having a central openai-based result converter which removes the need of duplication for some of our implementations
2. introducing a jsonpath based abstraction for configuring the mechanism of data extraction

And of course those two things don't have to be coupled, but one thing led to the other somehow ...

- [x] This concept basically still needs proof for the second part by adopting it with Anthropic and Gemini.
- [ ] Tests for streaming examples and customized text and tool call extraction


alexandre-daubois:
I really like the idea! Know that if you manage to get a resource for the response, JsonPath uses JsonStreamer to optimize the part to actually decode. Just in case 🙂

Given that JsonPath is experimental, I'd suggest changing the composer constraint to 7.3.*. That's what we do in core because experimental components may not be compatible between their minor versions. I have no plan to change classes or signatures in the JsonPath component until it's stable release as I managed [to make it RFC compliant](https://github.com/symfony/symfony/pull/61132) lately with the exact same API. But better be safe than sorry!

valtzu:
Interesting approach!

Do we even need the platform-specific `ResponseConverter`s if there's no platform-specific logic apart a custom json path? This path could be injected in `PlatformFactory`.

---

EDIT: ok hmm I guess the idea was to encapsulate the "protocol" in the converter to be reused in different platforms

chr-hertel:
> I'd suggest changing the composer constraint to 7.3.*

Thanks for that!

> encapsulate the "protocol" in the converter to be reused in different platforms

exactly

OskarStark:

Thanks

src/platform/src/Contract/JsonPathConverter/ResultExtractorInterface.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+declare(strict_types=1);
```
fabpot: To be removed ;)

chr-hertel: you're faster reviewing than myself - need to change my file template for this project

OskarStark: Let's add rector and if not available, create an extra rule

chr-hertel:
how do I deal with that kind of stuff?
image

alexandre-daubois:
Running CS fixer locally doesn't help?

chr-hertel:
we need a centralized result conversion heavily-inspired by the OpenAI contract - but we don't have to use JSON Path here.

what do you think? clever thing to do or too artistic?

@OskarStark @DZunke @alexandre-daubois @lyrixx @derrabus @xabbuh @fabpot @valtzu 

we end up having customized result conversion with extraction paths like this:
https://github.com/symfony/ai/blob/0a97792d800487efac2a8b884aea90c2249ca70e/src/platform/src/Bridge/Anthropic/PlatformFactory.php#L39-L52

alexandre-daubois:
> heavily-inspired by the OpenAI contract

I'm not sure to get what it means?

Also about your last comment, I'm not sure to understand the problem, but I'd be happy to help if you can elaborate on it 🙂

chr-hertel:
To far away from current code base - even if i still like the idea for extractors with json paths. might revisit later.

PR #139, opened 2025-07-16
[Store] Add SurrealDB

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | None
| License       | MIT

Hi 👋🏻 

This PR aim to introduce `SurrealDB` as a store for embeddings, the store use the `HTTP` API as the PHP SDK is ugly to use (and does not allow mocks), the tests needs to be improved / completed but the main implementation is ready.

PS: By default, we can't send more than `1275` embeddings during the insertion phase, the SDK can bypass this limitation but again, testing issues, API and more so, the road to follow is up to the team.


OskarStark:

Thanks

examples/store/surrealdb-similarity-search.php
```diff
@@ -0,0 +1,82 @@
...
+    $documents[] = new TextDocument(
+        id: Uuid::v4(),
+        content: 'Title: '.$movie['title'].\PHP_EOL.'Director: '.$movie['director'].\PHP_EOL.'Description: '.$movie['description'],
+        metadata: new Metadata($movie),
+    );
+}
+
+// create embeddings for documents
+$platform = PlatformFactory::create($_SERVER['OPENAI_API_KEY']);
+$vectorizer = new Vectorizer($platform, $embeddings = new Embeddings());
```
chr-hertel, line 58: I guess we need a different embedding model than:
image

chr-hertel (changes requested):
You're really on a great streak here to level up the store bridges - thanks for that! :pray:

Somehow the example doesn't work for me again - see below

examples/store/surrealdb-similarity-search.php
Guikingone: Yes, that's a limitation from the HTTP API, if we exceed `1275`, a `Payload too large` is triggered 🙁 

I'll take a look at it.

chr-hertel: It's even triggered when I use Mistral embeddings with only `1024` dimensions - is it maybe a config issue of the docker setup?
was just looking at this: https://surrealdb.com/docs/surrealdb/cli/env#surrealdb-environment-variables

chr-hertel: if you configure `SURREAL_HTTP_MAX_KEY_BODY_SIZE` in the env vars of the docker container with a higher value than the default, it works - also with more than 1275.

so can remove that validation from `query` and `add` methods, and use 1536 as default embedding dimensions

src/store/src/Bridge/SurrealDB/Store.php
```diff
@@ -0,0 +1,194 @@
...
+            : new Vector($data[$this->vectorFieldName]);
+
+        unset($data['_metadata']['_id']);
+
+        return new VectorDocument(
+            id: Uuid::fromString($id),
+            vector: $vector,
+            metadata: new Metadata(array_merge($data['_metadata'], [
+                $this->vectorFieldName => $data[$this->vectorFieldName],
+            ])),
```
chr-hertel: Let's not add the vector in the metadata, similar to #111 - metadata is only used for additional, qualitative data for the LLM to read
```suggestion
            metadata: new Metadata($data['_metadata']),
```

examples/store/surrealdb-similarity-search.php
Guikingone: Ah, nice catch, totally forgot to check the env variables 🙁 

I'll remove the check and improve tests, thanks for the feedback 🚀

src/store/src/Bridge/SurrealDB/Store.php
Guikingone: I removed the embeddings from the metadata 🙂

examples/store/surrealdb-similarity-search.php
Guikingone: Fixed

Guikingone:
@OskarStark 👋🏻 

I updated the `CHANGELOG.md` file for `store` 🙂

OskarStark:
Needs a rebase please

src/store/src/Bridge/SurrealDB/Store.php
```diff
@@ -0,0 +1,182 @@
...
+                'body' => $payload,
+            ];
+        }
+
+        $response = $this->httpClient->request($method, $url, array_merge($finalPayload, [
+            'headers' => array_merge($extraHeaders, [
+                'Accept' => 'application/json',
+                'Content-Type' => 'application/json',
+            ]),
+        ]));
```
chr-hertel: what about changing the `array $extraHeaders` argument to `string $authenticationToek` and moving everything to here?

```suggestion
        $response = $this->httpClient->request($method, $url, array_merge($finalPayload, [
            'headers' => [
                'Accept' => 'application/json',
                'Content-Type' => 'application/json',
                'Surreal-NS' => $this->namespace,
                'Surreal-DB' => $this->database,
                'Authorization' => \sprintf('Bearer %s', $authenticationToken),
            ],
        ]));
```

src/store/src/Bridge/SurrealDB/Store.php
```diff
@@ -0,0 +1,182 @@
...
+            metadata: new Metadata($data['_metadata']),
+        );
+    }
+
+    /**
+     * @param array{
+     *     namespacedUser?: bool
+     * } $options The namespacedUser option is used to determine if the user is root or not, if not, both the namespace and database must be specified
+     */
+    private function authenticate(array $options): string
```
chr-hertel: if we only use that one specific option here, we could replace the blurry `array $options` by `bool $namespacedUser`

in general i wonder why that is an option and not part of the contructor arguments like `$user` and `$password` - isn't it directly connected?

chr-hertel (changes requested):
small things about the inner contract of the Store implementation here

src/store/src/Bridge/SurrealDB/Store.php
Guikingone: `namespacedUser` is used if you're not `root`, this force the user to specify a user with its namespace, I can define the options as `false` by default.

src/store/src/Bridge/SurrealDB/Store.php
Guikingone: The authentication token is not always required, same for `Surreal-NS` and `Surreal-DB`, those headers are not required during `signin`

chr-hertel: okay, missed that, than please let's decouple those two types of requests.

meaning, instead of using the `->request(...)` in authenticate - just use the plain httpClient.

this will also remove that `is_string ? body : json` handling in the `request` method.

src/store/src/Bridge/SurrealDB/Store.php
```diff
@@ -0,0 +1,182 @@
...
+        }
+
+        $authenticationResponse = $this->request('POST', 'signin', $authenticationPayload);
+
+        if (!\array_key_exists('token', $authenticationResponse)) {
+            throw new RuntimeException('The SurrealDB authentication response does not contain a token.');
+        }
+
+        return $authenticationResponse['token'];
+    }
```
chr-hertel: what do you think about a local cache var to make sure we authenticate only once per request/command/message?

```php
    {
        if (isset($this->authenticationToken) {
            return $this->authenticationToken;
        }
    
        $authenticationPayload = ['user' => $this->user, 'pass' => $this->password];

        if ($this->isNamespacedUser) {
            $authenticationPayload['ns'] = $this->namespace;
            $authenticationPayload['db'] = $this->database;
        }

        $authenticationResponse = $this->httpClient->request('POST', sprintf('%s/signin'), $this->endpointUrl, $authenticationPayload)->toArray();

        if (!\array_key_exists('token', $authenticationResponse)) {
            throw new RuntimeException('The SurrealDB authentication response does not contain a token.');
        }

        return $this->authenticationToken = $authenticationResponse['token'];
    }
```

src/store/src/Bridge/SurrealDB/Store.php
```diff
@@ -0,0 +1,182 @@
...
+/**
+ * @author Guillaume Loulier <personal@guillaumeloulier.fr>
+ */
+final class Store implements InitializableStoreInterface, VectorStoreInterface
+{
+    public function __construct(
+        private readonly HttpClientInterface $httpClient,
+        private readonly string $endpointUrl,
+        #[\SensitiveParameter]
+        private readonly string $user,
```
OskarStark:
```suggestion
        #[\SensitiveParameter] private readonly string $user,
```
same for the others

src/store/src/Bridge/SurrealDB/Store.php
Guikingone: Agree, I'll work on it tomorrow 🙂

Guikingone: Fixed for both authentication and namespaced user.

src/store/src/Bridge/SurrealDB/Store.php
Guikingone: The check on `string` vs `array` must stay, the SQL query is sent using `body` and not `json` 🙂

chr-hertel (approved):

chr-hertel:
Thank you @Guikingone.

PR #143, opened 2025-07-17
[MCP SDK] Use foreach instead of `array_map` because `array_map` does not accept generators

dorrogeray:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | N/A
| License       | MIT

Use foreach instead of array_map because array_map does not accept generators and add test coverage.

### DIFF
--- a/src/mcp-sdk/tests/Server/RequestHandler/PromptListHandlerTest.php
+++ b/src/mcp-sdk/tests/Server/RequestHandler/PromptListHandlerTest.php
@@ -1,0 +1,80 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\McpSdk\Tests\Server\RequestHandler;
+
+use PHPUnit\Framework\Attributes\CoversClass;
+use PHPUnit\Framework\Attributes\Small;
+use PHPUnit\Framework\TestCase;
+use Symfony\AI\McpSdk\Capability\Prompt\MetadataInterface;
+use Symfony\AI\McpSdk\Capability\PromptChain;
+use Symfony\AI\McpSdk\Message\Request;
+use Symfony\AI\McpSdk\Server\RequestHandler\PromptListHandler;
+
+#[Small]
+#[CoversClass(PromptListHandler::class)]
+class PromptListHandlerTest extends TestCase
+{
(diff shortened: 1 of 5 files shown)


src/mcp-sdk/tests/Server/RequestHandler/PromptListHandlerTest.php
```diff
@@ -0,0 +1,113 @@
...
+#[CoversClass(PromptListHandler::class)]
+class PromptListHandlerTest extends TestCase
+{
+    public function testHandleEmpty(): void
+    {
+        $collection = $this->getMockBuilder(CollectionInterface::class)
+            ->disableOriginalConstructor()
+            ->onlyMethods(['getMetadata'])
+            ->getMock();
+        $collection->expects($this->once())->method('getMetadata')->willReturn([]);
```
chr-hertel: For me it would be totally fine to go with a real instance instead:
```suggestion
        $collection = new PromptChain([]);
```

src/mcp-sdk/tests/Server/RequestHandler/PromptListHandlerTest.php
```diff
@@ -0,0 +1,113 @@
...
+     * @param iterable<MetadataInterface> $metadataList
+     */
+    #[DataProvider('metadataProvider')]
+    public function testHandleReturnAll(iterable $metadataList): void
+    {
+        $collection = $this->getMockBuilder(CollectionInterface::class)
+            ->disableOriginalConstructor()
+            ->onlyMethods(['getMetadata'])
+            ->getMock();
+        $collection->expects($this->once())->method('getMetadata')->willReturn($metadataList);
```
chr-hertel: Same here:
```suggestion
        $collection = new PromptChain($metadataList);
```

src/mcp-sdk/tests/Server/RequestHandler/PromptListHandlerTest.php
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    public function testHandlePagination(): void
+    {
+        $item = self::createMetadataItem();
+        $collection = $this->getMockBuilder(CollectionInterface::class)
+            ->disableOriginalConstructor()
+            ->onlyMethods(['getMetadata'])
+            ->getMock();
+        $collection->expects($this->once())->method('getMetadata')->willReturn([$item, $item]);
```
chr-hertel: And here:
```suggestion
        $collection = new PromptChain([$item, $item]);
```

chr-hertel:
Thank you @dorrogeray.

PR #147, opened 2025-07-17
Prefix all bundle services with `ai` instead of `symfony_ai`

llupa:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | N/A
| License       | MIT

I am opening this PR as a proposal / RFC after a short conversation in official Symfony Slack, open to do more changes or close it 😄  

> This is just a question and more for my own interest, but I noticed that Symfony services from Symfony UX are prefixed as `ux.` while Symfony AI has `symfony_ai.` Why not just use `ai.` ? 
> @chr-hertel Oh, tbh, I didn't check how UX is doing it - valid issue I think
> @OskarStark es let’s go with „ai.“ Open for a PR ?

### DIFF
--- a/src/ai-bundle/src/AIBundle.php
+++ b/src/ai-bundle/src/AIBundle.php
@@ -79,7 +79,7 @@ final class AIBundle extends AbstractBundle
         foreach ($config['platform'] ?? [] as $type => $platform) {
             $this->processPlatformConfig($type, $platform, $builder);
         }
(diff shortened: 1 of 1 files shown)


OskarStark:
I created a PR including all demo files etc. in:
* https://github.com/symfony/ai/pull/148

Thanks @llupa

chr-hertel:
Thanks @llupa for catching that inconsistency and raising it :+1:

PR #149, opened 2025-07-17
[RFC]: OpenAI Responses API

glengemann:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Docs?         | no 
| Issues        | Fix #... 
| License       | MIT

Hi! I've been playing around with the new OpenAI Responses API for a project. Just wondering if the changes in this branch are moving in the right direction.

### DIFF
--- a/examples/openai/responses-stream.php
+++ b/examples/openai/responses-stream.php
@@ -1,0 +1,43 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use Symfony\AI\Agent\Agent;
+use Symfony\AI\Platform\Bridge\OpenAI\GPT;
+use Symfony\AI\Platform\Bridge\OpenAI\PlatformFactory;
+use Symfony\AI\Platform\Bridge\OpenAI\Responses;
+use Symfony\AI\Platform\Message\Message;
+use Symfony\AI\Platform\Message\MessageBag;
+use Symfony\Component\Dotenv\Dotenv;
+
+require_once dirname(__DIR__).'/vendor/autoload.php';
+(new Dotenv())->loadEnv(dirname(__DIR__).'/.env');
+
+if (!isset($_SERVER['OPENAI_API_KEY'])) {
+    echo 'Please set the OPENAI_API_KEY environment variable.'.\PHP_EOL;
+    exit(1);
+}
+
+$platform = PlatformFactory::create($_SERVER['OPENAI_API_KEY']);
+$model = new Responses(Responses::GPT_4O_MINI);
+
+$agent = new Agent($platform, $model);
+$messages = new MessageBag(
+    Message::forSystem('You are a thoughtful philosopher.'),
+    Message::ofUser('What is the purpose of an ant?'),
+);
+$response = $agent->call($messages, [
+    'stream' => true, // enable streaming of response text
+]);
+
+foreach ($response->getContent() as $word) {
+    echo $word;
+}
+echo \PHP_EOL;
--- a/examples/openai/responses-structured-output-clock.php
+++ b/examples/openai/responses-structured-output-clock.php
@@ -1,0 +1,61 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use Symfony\AI\Agent\Agent;
+use Symfony\AI\Agent\StructuredOutput\Responses\ResponsesAgentProcessor;
+use Symfony\AI\Platform\Bridge\OpenAI\PlatformFactory;
+use Symfony\AI\Platform\Bridge\OpenAI\Responses;
+use Symfony\AI\Platform\Message\Message;
+use Symfony\AI\Platform\Message\MessageBag;
+use Symfony\Component\Dotenv\Dotenv;
+
+require_once dirname(__DIR__).'/vendor/autoload.php';
+(new Dotenv())->loadEnv(dirname(__DIR__) . '/.env');
+
+if (!isset($_SERVER['OPENAI_API_KEY'])) {
+    echo 'Please set the OPENAI_API_KEY environment variable.' . \PHP_EOL;
+    exit(1);
+}
+
+$platform = PlatformFactory::create($_SERVER['OPENAI_API_KEY']);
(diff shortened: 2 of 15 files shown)


chr-hertel:
Hey @glengemann, thanks for working on this and thanks for checking in early for feedback :+1:

I notice some things here:
* in general I'd like to avoid different endpoint or models having impact to the Agent component - it might not always be possible and you found an unflexible part of the code there. I would rather try to bring in some extension points or contract handling - like with the tool schema - instead of just duplicating the `AgentProcessor` and `ResponseFormatFactory`
* also with the changes within the Platform component there is a lot of duplication going on, e.g. 'GPT` and `Responses` classes are basically the same - we just need it here because we can't differentiate the other extension points, like the `InputBagNormalizer` in a different way - that's unfortunate.

what I'm basically saying here, I get why you checked in early - somehow this duplication feels odd, but the current extension points, like those supports methods, don't provide more flexibility.

In those cases I'd like to start from the user land code: how would we want the user to select the different apis? do they care at this point?

let's image we would use this:
```php
use Symfony\AI\Platform\Bridge\OpenAI;

$platform = OpenAI\PlatformFactory($apiKey);
$model = new OpenAI\GPT();

$response = $platform->request($model, $messages);
```

vs.

```php
use Symfony\AI\Platform\Bridge\OpenAI;

$platform = OpenAI\PlatformFactory($apiKey);
$model = new OpenAI\Responses();

$response = $platform->request($model, $messages);
```

So, my questions:
* What would be the difference from a user point of view? Which features do we gain by adopting the Responses API?
* If there is a benefit, could we go with something like this?
    ```php
    use Symfony\AI\Platform\Bridge\OpenAI;

    $platform = OpenAI\PlatformFactory($apiKey);
    $model = new OpenAI\GPT();

    $response = $platform->request($model, $messages, ['api' => 'responses']); // with default ['api => 'completions']
    ```

If we'd go with an option, we would need to find a way to extend the supports methods to check for that - and the `OpenAI\ModelClient` would need to interpret that to choose the right endpoint.

chr-hertel:
@glengemann do you have some thoughts or plans with this or should we close this for now?

glengemann:
Hi @chr-hertel! Let me work on this feature a little more.

MartkCz:
@chr-hertel I was trying to finish the PR because I now need to use the responses API (the websearch function for GPT-5 is only available through the responses API, not the completions API anymore).

But I hit a problem here:
https://github.com/symfony/ai/blob/main/src/agent/src/StructuredOutput/AgentProcessor.php#L68

The response_format key for structured output is only required by the completions API. The responses API, on the other hand, expects it under the text key.

It looks like Ollama also doesn’t support response_format, see:
https://github.com/ollama/ollama/issues/3102
https://ollama.com/blog/structured-outputs

So the question is: how do we deal with this?

The only solution I can think of is to remove ResponseFormatFactory and let the model itself decide how the output should look—since it’s the only one that actually knows the proper structure. Then in the agent, call something like:

```php
$responseClass = $options['output_structure'];
$structuredOutputName = u($responseClass)->afterLast('\\')->toString();
$schema = $this->schemaFactory->buildProperties($responseClass);
$input->model->addStructuredOutputToInput($input, $structuredOutputName, $schema, $responseClass);
```

It might not even be necessary to pass that many parameters.

chr-hertel:
@MartkCz i guess this should cover the issue, doesn't it?
https://github.com/symfony/ai/blob/899ca129323d9d2e68c212b4ca9d5d44f2d96e6e/src/agent/src/StructuredOutput/AgentProcessor.php#L60-L62

basically if you use a model that doesn't support that feature, you can't use it.

MartkCz:
@chr-hertel The key difference is that the Responses API still supports structured output, but not under response_format anymore. Instead, it expects it in a slightly different structure under text.

Current implementation:

```php
$options['response_format'] = [
    'type' => 'json_schema',
    'json_schema' => [
        'name' => u($responseClass)->afterLast('\\')->toString(),
        'schema' => $this->schemaFactory->buildProperties($responseClass),
        'strict' => true,
    ],
];
```

Responses API:
```php
$options['text'] = [
    'format' => [
        'type' => 'json_schema',
        'name' => u($responseClass)->afterLast('\\')->toString(),
        'schema' => $this->schemaFactory->buildProperties($responseClass),
        'strict' => true,
    ],
];
```
---
Current implementation:
https://github.com/symfony/ai/blob/899ca129323d9d2e68c212b4ca9d5d44f2d96e6e/src/agent/src/StructuredOutput/AgentProcessor.php#L68
What we need (Completions API):
```php
$options['text']['format'] = $this->responseFormatFactory->create($options['output_structure']);
```

Current implementation:
https://github.com/symfony/ai/blob/899ca129323d9d2e68c212b4ca9d5d44f2d96e6e/src/agent/src/StructuredOutput/ResponseFormatFactory.php#L30-L37
What we need (Completions API):
```php
return [
    'type' => 'json_schema',
    'name' => u($responseClass)->afterLast('\\')->toString(),
    'schema' => $this->schemaFactory->buildProperties($responseClass),
    'strict' => true,
];
```
---
The proposed solution is to skip ResponseFormatFactory and leave the decision about the structure to the model, but perhaps there is a better solution.

Before:
https://github.com/symfony/ai/blob/899ca129323d9d2e68c212b4ca9d5d44f2d96e6e/src/agent/src/StructuredOutput/AgentProcessor.php#L68
After:
```php
$input->model->addStructuredOutputToInput($input, $structuredOutputName, $schema, $responseClass);
```

melkamar:
Adding another voice to use the Responses API - it seems like it's the more "feature complete" API to use going forward. OpenAI itself states in [the chat completions docs](https://platform.openai.com/docs/api-reference/chat/create): 

> Starting a new project? We recommend trying [Responses](https://platform.openai.com/docs/api-reference/responses) to take advantage of the latest OpenAI platform features. Compare [Chat Completions with Responses](https://platform.openai.com/docs/guides/responses-vs-chat-completions?api-mode=responses).

**I am using the Azure OpenAI API**.

My use case that is currently blocked is supplying files alongside the prompt. Images work, but when trying to pass through a `Document` like in [the example](https://github.com/symfony/ai/blob/main/examples/openai/pdf-input-binary.php#L28), then that fails with:

```
Invalid Value: 'file'. This model does not support file content types.
```

The payload that is sent to the Azure OpenAI API is this (truncated):

```json
{
  "model": "gpt-4.1",
  "messages": [
    {
      "role": "user",
      "content": [
        {
          "type": "text",
          "text": "What's in the pdf?"
        },
        {
          "type": "file",
          "file" : {
            "filename": "some.pdf",
            "file_data": "data:application/pdf;base64,YWQ="
          }
        }
      ]
    }
  ]
}
```

and the API fails with the message above. The only way I've found to get this to work is to use the Responses API and reference that pre-uploaded PDF.

---

Is there any case for reopening this PR/issue, or should I make a separate one?

PR #152, opened 2025-07-18
[Platform] Leverage Enum everywhere is possible

welcoMattic:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Docs?         | no 
| Issues        | Fix #... 
| License       | MIT

I mostly replaced classes with only public constant with Enums


chr-hertel:
Ha, thanks, I love enums and would love to see them in use more, but there's also a reason why we opt'd out for some cases. but the concept of enums is to strictly limit/validate values to a specific set and we're not the owner of that value set in some cases.
for example with those huggingface inference providers: huggingface might easily add another one - and actually they did add featherless, groq and nscale since i implemented this. so users could still use `'groq'` as string when we only use constants from an interface as some kind of DX helper, but if we limit the interface to an enum, we would need to patch the lib first after huggingface added that provider to enable the user to use it.
that's basically a learning from those GPT versions, and I'd like to apply it wherever it makes sense, because those providers we're building bridges here move really fast on things.

so this logic at least applies to gemini's embeddings types, hugging face providers and tasks - for that Capability thing I get your point - that's something we should do :+1:

OskarStark:
Agree on the capability!

welcoMattic:
Do not merge, I messed up my rebase.

PR #154, opened 2025-07-18
[Platform][Agent] Rework Platform API to invoke and result

### LINKED ISSUE #141: [RFC][Platform] Renaming `Response` to `Result`
Originally the Platform component only focused on wrapping around HTTP calls, and therefore it was quite logical for me to make an interface like `Platform->request($model, $input, $options): Response`, but with integrations like [TransformersPHP](https://github.com/symfony/ai/blob/main/src/platform/src/Bridge/TransformersPHP/Platform.php) or [Bedrock](https://github.com/symfony/ai/blob/main/src/platform/src/Bridge/Bedrock/Platform.php) it became obvious that a) close coupling to HttpClient's ResponseInterface was not ideal and b) also the HTTP-based terminology is not always correct. After introducing the decoupled `ResponsePromise` it is now easier to refactor to a more HTTP-independent approach, and I want to use that timing, to propose a new contract for the Platform:

```php
Platform->execute($model, $input, $options): ResultPromise
```

I feel `execute` is more natural with `$model` and `$input` - and the model execution returns a result. still pretty vague, but sounds more intuitive to me than requesting and model.

What do you think?

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Docs?         | no
| Issues        | Fix #141 
| License       | MIT

We can change the naming later, they said. PhpStorm has great tooling for that, they said. :laughing: :see_no_evil: 

I skipped the ResponseFactory for now - not sure about the Agent vs Platform API and differences here.


OskarStark:
Where is the added line? 😄 
CleanShot 2025-07-18 at 19 37 30@2x

chr-hertel:
> Where is the added line? 😄

Wouldn't it take away the fun of reviewing if I'd tell you right away? :joy:

src/platform/tests/Response/StreamResponseTest.php
```diff
@@ -15,9 +15,9 @@
...
 use PHPUnit\Framework\Attributes\Test;
 use PHPUnit\Framework\TestCase;
-use Symfony\AI\Platform\Response\StreamResponse;
+use Symfony\AI\Platform\Result\StreamResult;
 
-#[CoversClass(StreamResponse::class)]
+#[CoversClass(StreamResult::class)]
 #[Small]
 final class StreamResponseTest extends TestCase
 {
```
OskarStark: class name correct?

chr-hertel: noo, please don't review - it's incomplete

chr-hertel: now :tada:

chr-hertel:
Merging in alignment with Oskar without approvals - not great to review anyways, sorry :grimacing: :innocent:

PR #160, opened 2025-07-18
[Examples] Improve bootstrapping and logging in examples

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | 
| License       | MIT

Brings to things:
* Easier bootstrapping and env var checks
* Enables logging of HTTP calls with `-vv` or tool calls with `-vvv`

Sometimes I'd like to see more details of the execution of an example. This would be possible now:
image

What do you think? does it make the examples too complex to read & write - or helpful?

(only a draft, would roll that out to all examples if feedback is positive)

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -55,7 +55,13 @@ Every example script is a standalone PHP script that can be run from the command
 You can run an example by executing the following command:
 
 ```bash
-php openai/openai/chat.php
+php openai/chat.php
+```
+
+To get more insights into what is happening at runtime, e.g. HTTP and tool calls, you can add `-vv` or `-vvv`:
+
+```bash
+php openai/toolcall-stream.php -vvv
 ```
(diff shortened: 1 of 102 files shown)


src/agent/src/Toolbox/Toolbox.php
chr-hertel, line 1: That should be a separate PR maybe ...

OskarStark: Yes

OskarStark:
I like it very much, it makes the examples less verbose 👍🏻

src/agent/src/Toolbox/Toolbox.php
chr-hertel: done with #162

PR #162, opened 2025-07-19
[Agent][Examples] Remove Toolbox::create(...)

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | 
| License       | MIT

This is just not useful other than in simplified examples - and not even there, see #160

### DIFF
--- a/src/agent/src/Toolbox/Toolbox.php
+++ b/src/agent/src/Toolbox/Toolbox.php
@@ -44,18 +44,13 @@ final class Toolbox implements ToolboxInterface
      * @param iterable<mixed> $tools
      */
     public function __construct(
-        private readonly ToolFactoryInterface $toolFactory,
         iterable $tools,
+        private readonly ToolFactoryInterface $toolFactory = new ReflectionToolFactory(),
         private readonly ToolCallArgumentResolver $argumentResolver = new ToolCallArgumentResolver(),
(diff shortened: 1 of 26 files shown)


src/agent/src/Toolbox/Toolbox.php
```diff
@@ -44,20 +44,15 @@ final class Toolbox implements ToolboxInterface
      * @param iterable<mixed> $tools
      */
     public function __construct(
-        private readonly ToolFactoryInterface $toolFactory,
         iterable $tools,
+        private readonly ToolFactoryInterface $toolFactory = new ReflectionToolFactory(),
```
OskarStark, line 48: This and the other Bc break need a changelog entry

chr-hertel: that changelog thingy breaks my brain - what the perspective there - i mean to what are we comparing while noting the changes?

I understand that changelog _usually_ is basically only about adding lines, but in this case still?
I would just remove this line :joy: 

https://github.com/symfony/ai/blob/24797e3036afde39f35860dc607c3999c8945cec/src/agent/CHANGELOG.md?plain=1#L54

OskarStark: I get your point until we have no release, I am fine with just removing

Issue #77, opened 2025-07-07
Adopt index-based instead of name-based argument config


chr-hertel:
>If I am not mistaken, but I am currently on a phone 📱, we don't use named argument in Symfony, but index based assignments.
>
>cc @Nyholm
>
>Ofc in an extra PR

_Originally posted by @OskarStark in https://github.com/symfony/ai/pull/74#discussion_r2190759057_

stof:
The BC promise of Symfony indeed excludes argument names (except for attribute constructors) for a few reasons:

- when using an interface, argument names are unreliable (PHP does not enforce that implementations use the same argument names, and runtime evaluation of the named arguments is based on the implementation, not based on the type available during static analysis)
- our process for refactoring method signatures (when wanting to remove an argument) is not compatible with being BC with argument names
- when wanting to change what gets injecting in an argument, keeping BC for the name would require keeping the old name forever
- it is in practice almost impossible to keep BC for both positional and named arguments when changing a method signature (which would mean never changing a method signature except by adding new arguments at the end, which would be much more restrictive for maintainers than the current policy)

It would be great for this project to rely on the same BC policy than other Symfony projects to provide clear expectations.

OskarStark:
@yceruto while you are on the bundle config, would you like to take care?

yceruto:
@OskarStark I’d prefer to do it in a separate PR

OskarStark:
cc @valtzu if you want to have a look 👀

chr-hertel:
Fixed with #163 and #165

PR #163, opened 2025-07-19
[AIBundle] Wire & configure services explicitly

valtzu:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | Related to #90
| License       | MIT

Don't autowire or autoconfigure anything in AI Bundle

### DIFF
--- a/src/ai-bundle/config/options.php
+++ b/src/ai-bundle/config/options.php
@@ -11,6 +11,9 @@
 
 namespace Symfony\Component\Config\Definition\Configurator;
 
+use Codewithkyrian\ChromaDB\Client as ChromaDBClient;
+use MongoDB\Client as MongoDBClient;
+use Probots\Pinecone\Client as PineconeClient;
 use Symfony\AI\Platform\PlatformInterface;
 use Symfony\AI\Store\StoreInterface;
 
@@ -154,6 +157,7 @@ return static function (DefinitionConfigurator $configurator): void {
                         ->useAttributeAsKey('name')
                         ->arrayPrototype()
                             ->children()
+                                ->scalarNode('client')->cannotBeEmpty()->defaultValue(ChromaDBClient::class)->end()
                                 ->scalarNode('collection')->isRequired()->end()
                             ->end()
                         ->end()
@@ -163,6 +167,7 @@ return static function (DefinitionConfigurator $configurator): void {
                         ->useAttributeAsKey('name')
                         ->arrayPrototype()
                             ->children()
+                                ->scalarNode('client')->cannotBeEmpty()->defaultValue(MongoDBClient::class)->end()
                                 ->scalarNode('database')->isRequired()->end()
                                 ->scalarNode('collection')->isRequired()->end()
                                 ->scalarNode('index_name')->isRequired()->end()
@@ -176,6 +181,7 @@ return static function (DefinitionConfigurator $configurator): void {
                         ->useAttributeAsKey('name')
                         ->arrayPrototype()
                             ->children()
+                                ->scalarNode('client')->cannotBeEmpty()->defaultValue(PineconeClient::class)->end()
                                 ->scalarNode('namespace')->end()
                                 ->arrayNode('filter')
                                     ->scalarPrototype()->end()
--- a/src/ai-bundle/config/services.php
+++ b/src/ai-bundle/config/services.php
@@ -18,66 +18,114 @@ use Symfony\AI\Agent\Toolbox\AgentProcessor as ToolProcessor;
 use Symfony\AI\Agent\Toolbox\Toolbox;
 use Symfony\AI\Agent\Toolbox\ToolboxInterface;
 use Symfony\AI\Agent\Toolbox\ToolCallArgumentResolver;
+use Symfony\AI\Agent\Toolbox\ToolFactory\AbstractToolFactory;
 use Symfony\AI\Agent\Toolbox\ToolFactory\ReflectionToolFactory;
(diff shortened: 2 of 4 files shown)


src/ai-bundle/config/services.php
```diff
@@ -19,65 +19,89 @@
...
-        ->defaults()
-            ->autowire()
-
+        ->set('ai.platform.contract', Contract::class)
+            ->factory([Contract::class, 'create'])
         // structured output
-        ->set(ResponseFormatFactory::class)
-            ->alias(ResponseFormatFactoryInterface::class, ResponseFormatFactory::class)
-        ->set(StructureOutputProcessor::class)
+        ->set('ai.response_format_factory', ResponseFormatFactory::class)
```
chr-hertel: let's use `ai.agent.` here as prefix as well
```suggestion
        ->set('ai.agent.response_format_factory', ResponseFormatFactory::class)
```

src/ai-bundle/config/services.php
```diff
@@ -19,65 +19,89 @@
...
+            ->args([
+                '$schemaFactory' => service('ai.json_schema_factory'),
+            ])
+        ->set('ai.json_schema_factory', Factory::class)
```
chr-hertel: this would be
```suggestion
        ->set('ai.platform.json_schema_factory', Factory::class)
```

or was there a reason to skip the components here?

src/ai-bundle/src/AIBundle.php
```diff
@@ -158,11 +157,12 @@ private function processPlatformConfig(string $type, array $platform, ContainerB
             $platformId = 'ai.platform.anthropic';
             $definition = (new Definition(Platform::class))
                 ->setFactory(AnthropicPlatformFactory::class.'::create')
-                ->setAutowired(true)
                 ->setLazy(true)
                 ->addTag('proxy', ['interface' => PlatformInterface::class])
                 ->setArguments([
                     '$apiKey' => $platform['api_key'],
+                    '$httpClient' => new Reference('http_client', ContainerInterface::NULL_ON_INVALID_REFERENCE),
+                    '$contract' => new Reference('ai.platform.contract'),
```
chr-hertel: Just injecting the default contract here, will disable the `AnthropicContract` and will cause errors I guess, see

https://github.com/symfony/ai/blob/1d5dabb6356a6fa00015a543f05bc250247fc6a2/src/platform/src/Bridge/Anthropic/PlatformFactory.php#L30-L37

chr-hertel: potentially same for other places

chr-hertel (changes requested):
Thanks for working on this! Will need to give it another read + test, but some things already caught my eye.

src/ai-bundle/src/AIBundle.php
valtzu: Thanks, I'll check

I'd still prefer getting the instance from DI, so I'll add anthropic platform as separate service

valtzu:
Converted from name-based to index-based args to address #77 too

Included in the same PR not to need 2 rounds of testing

src/ai-bundle/config/options.php
```diff
@@ -154,6 +157,7 @@
                         ->useAttributeAsKey('name')
                         ->arrayPrototype()
                             ->children()
+                                ->scalarNode('client_service')->cannotBeEmpty()->defaultValue(ChromaDBClient::class)->end()
```
chr-hertel: don't you think `client` might be enough?

valtzu: Sure, done

chr-hertel:
Wait, does the "Audio Bot" still work for you?

getting this right now:
image

src/ai-bundle/src/AIBundle.php
```diff
@@ -216,10 +218,13 @@ private function processPlatformConfig(string $type, array $platform, ContainerB
...
             $definition = (new Definition(Platform::class))
                 ->setFactory(OpenAIPlatformFactory::class.'::create')
-                ->setAutowired(true)
                 ->setLazy(true)
                 ->addTag('proxy', ['interface' => PlatformInterface::class])
-                ->setArguments(['$apiKey' => $platform['api_key']])
+                ->setArguments([
+                    $platform['api_key'],
+                    new Reference('http_client', ContainerInterface::NULL_ON_INVALID_REFERENCE),
+                    new Reference('ai.platform.contract.default'),
```
chr-hertel: That's causing the error with the audio bot - it needs the Whisper contract extension, that is not part of the default, see here:

https://github.com/symfony/ai/blob/45ea78cd8b62773b4527805e9b7d21d88050e91e/src/platform/src/Bridge/OpenAI/PlatformFactory.php#L48

valtzu: Feels a bit counter-intuitive: I thought OpenAI was meant to be the default contract and now it seems we need a custom contract for it :thinking: 

Anyway, done – the audio demo works now

chr-hertel: The Contract is for the _entire_ Platform - including the model Whisper, but the default is currently only for embeddings and LLM completion.

Whisper doesn't qualify to be in the default Platform contract for me, but still is part of the OpenAI Platform.

chr-hertel (approved):

chr-hertel:
Thank you @valtzu.

valtzu:
Missed a rebase here, main is broken now. :face_with_peeking_eye: 

Fix in #176

PR #165, opened 2025-07-19
[McpBundle] Wire & configure services explicitly

valtzu:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | Related to #90
| License       | MIT

Don't autowire / autoconfigure any services in mcp bundle


valtzu:
Covers McpBundle part of #77 too

demo/config/services.yaml
```diff
@@ -19,10 +19,10 @@ services:
             - '../src/Entity/'
             - '../src/Kernel.php'
 
-    Symfony\AI\McpSdk\Capability\Tool\ToolExecutorInterface:
+    mcp.tool_executor:
```
valtzu, line 22: Not quite sure about this.

I'm not familiar with this MCP bundle at all – is this really how this is supposed to work? Don't we have some autoconfiguration, or explicit config within the app is always needed? :thinking:

chr-hertel: See #168 :)

src/mcp-bundle/config/services.php
```diff
@@ -25,36 +25,51 @@
 
 return static function (ContainerConfigurator $container): void {
     $container->services()
-        ->defaults()
-            ->autowire()
-            ->autoconfigure()
         ->instanceof(NotificationHandlerInterface::class)
             ->tag('mcp.server.notification_handler')
         ->instanceof(RequestHandlerInterface::class)
```
stof: those `instanceof` rules are not necessary, as this file uses explicit tags, and they have no effect outside this file.

Note however that it might make sense to register autoconfiguration rules for those interfaces in the DI extension, for projects implementing those interfaces.

valtzu: Thanks, moved

src/mcp-bundle/config/services.php
```diff
@@ -15,46 +15,55 @@
...
+        ->set('mcp.server.request_handler.tool_call', ToolCallHandler::class)
+            ->args([
+                service('mcp.tool_executor'),
+            ])
+            ->tag('mcp.server.request_handler')
+        ->set('mcp.server.request_handler.tool_list', ToolListHandler::class)
+            ->args([
+                service('mcp.tool_collection'),
+                20,
+            ])
```
chr-hertel, line 47: found out while testing that this is missing the tag:
```suggestion
        ->set('mcp.server.request_handler.tool_list', ToolListHandler::class)
            ->args([
                service('mcp.tool_collection'),
                20,
            ])
            ->tag('mcp.server.request_handler')
```

chr-hertel (approved):

chr-hertel:
Thank you @valtzu.

PR #142, opened 2025-07-17
[Platform] Remove HTTP awareness from Platform

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | 
| License       | MIT

This PR decouples the main Platform implementation from HTTP details, enabling Bedrock and TransformersPHP to drop their own Platform implementation.

I get that dealing with those RawXYZ objects is extra burden, but this is just one refactoring after each other - the goal was to decouple more from HTTP and reduce Platform implementations.

**Key changes are in:**
* HTTP Decoupling:
  * [src/platform/src/ModelClientInterface.php](https://github.com/symfony/ai/pull/142/files#diff-2fa6af72c0c856e895a407798442305cadab6f7f0d0fa0d84f5c26ff6faaccac)
  * [src/platform/src/Platform.php](https://github.com/symfony/ai/pull/142/files#diff-660e8336450572154d3c925521d0b91b8e98f1c6bba0d0ca547e163873549a36)
  * [src/platform/src/Response/ResponsePromise.php](https://github.com/symfony/ai/pull/142/files#diff-ee22066b5a7dfe6e5daccca990feafd609d2afc07412210a9d630e54b057b454)
  * [src/platform/src/ResponseConverterInterface.php](https://github.com/symfony/ai/pull/142/files#diff-2847909fadca63a47d4156f7aec298b9a67992d198922d1f174fa2b961d7abd4)
* Additional Platform Removal
  * [src/platform/src/Bridge/Bedrock/PlatformFactory.php](https://github.com/symfony/ai/pull/142/files#diff-9d90025703ce4f7a3c795b186355435b8b0925de16859c0b4d52ab2cb88d5d51)
  * [src/platform/src/Bridge/TransformersPHP/PlatformFactory.php](https://github.com/symfony/ai/pull/142/files#diff-a6e435b3ff56bbcb596d7f241bfe2926467411d9b489a75573cb705e875c8079)

Everything else is noise but impossible to avoid.
Definitely up for discussing the naming of interfaces and classes involved here. feels a bit odd, but let's move the renaming to follow-up PR, and also please check #141 in that case.

PR #167, opened 2025-07-19
[Platform] Improve error reporting in platform layer

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | 
| License       | MIT

With #142 I accidentally changed the error reporting/handling of the platform layer, since I used `toArray(false)` instead of `toArray()` on the http client's response.

However, maybe it's better to go ahead with that and implement better error handling, since this was annoying to me anyways :speak_no_evil: 

- [ ] Tests
- [ ] Roll out to all bridges

### DIFF
--- a/src/platform/src/Bridge/Replicate/Client.php
+++ b/src/platform/src/Bridge/Replicate/Client.php
@@ -11,6 +11,7 @@
 
 namespace Symfony\AI\Platform\Bridge\Replicate;
 
+use Symfony\AI\Platform\Exception\ResultException;
 use Symfony\Component\Clock\ClockInterface;
 use Symfony\Contracts\HttpClient\HttpClientInterface;
 use Symfony\Contracts\HttpClient\ResponseInterface;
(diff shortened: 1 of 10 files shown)


src/platform/src/Bridge/Replicate/Client.php
```diff
@@ -40,7 +41,11 @@ public function request(string $model, string $endpoint, array $body): ResponseI
...
             'json' => ['input' => $body],
         ]);
-        $data = $response->toArray();
+        $data = $response->toArray(false);
+
+        if (isset($data['detail'])) {
+            throw new ModelException($data['detail'], $data);
+        }
 
         while (!\in_array($data['status'], ['succeeded', 'failed', 'canceled'], true)) {
```
lyrixx, line 50: If the response if 4xx or 5xx, `status` might not exist.

PR #168, opened 2025-07-19
[McpBundle] Compiler pass for tools

TomHart:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | N/A
| License       | MIT

Add a Compiler Pass to find tools and register them in the ToolChain


src/mcp-bundle/CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 CHANGELOG
 =========
 
+0.2
+---
+
+ * Added `\Symfony\AI\McpBundle\DependencyInjection\ContainerBuilder\MCPToolChainCompilerPass`
```
OskarStark: Belongs to 0.1 still

TomHart: Updated

demo/composer.json
chr-hertel, line 1: please revert those changes here - i now it's not that straight forward right now for testing and I have similar changes as a git stash, but let's not bring them into the repo. this part is split into the https://github.com/symfony/ai-demo repo - and that wouldn't be functional with this addition

src/mcp-bundle/src/DependencyInjection/ContainerBuilder/MCPToolChainCompilerPass.php
```diff
@@ -0,0 +1,50 @@
+<?php
+
+declare(strict_types=1);
```
chr-hertel: please remove this

src/mcp-bundle/src/DependencyInjection/ContainerBuilder/MCPToolChainCompilerPass.php
```diff
@@ -0,0 +1,50 @@
...
+use Symfony\AI\McpSdk\Capability\ToolChain;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\Compiler\PriorityTaggedServiceTrait;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Definition;
+
+/**
+ * @author Tom Hart <tom.hart.221@gmail.com>
+ */
+class MCPToolChainCompilerPass implements CompilerPassInterface
```
chr-hertel: we decided for using `Mcp` in the lib - similar to other SDKs
```suggestion
final class McpToolChainCompilerPass implements CompilerPassInterface
```

src/mcp-bundle/src/DependencyInjection/ContainerBuilder/MCPToolChainCompilerPass.php
```diff
@@ -0,0 +1,50 @@
...
+    public function process(ContainerBuilder $container): void
+    {
+        // If the user has already defined this service, don't override it.
+        if ($container->hasDefinition(ToolExecutorInterface::class)) {
+            return;
+        }
+
+        $definition = new Definition(ToolChain::class);
+
+        $taggedServices = $this->findAndSortTaggedServices('mcp_tool', $container);
```
chr-hertel: I'd go with
```suggestion
        $taggedServices = $this->findAndSortTaggedServices('mcp.tool', $container);
```

demo/config/services.yaml
```diff
@@ -19,10 +19,6 @@ services:
...
-    Symfony\AI\McpSdk\Capability\Tool\ToolExecutorInterface:
-        class: Symfony\AI\McpSdk\Capability\ToolChain
-        arguments:
-            - ['@App\MCP\Tools\CurrentTimeTool']
-
-    Symfony\AI\McpSdk\Capability\Tool\CollectionInterface:
-        alias: Symfony\AI\McpSdk\Capability\Tool\ToolExecutorInterface
+    App\MCP\Tools\CurrentTimeTool:
+        tags:
+            - { name: 'mcp_tool' }
```
chr-hertel: let's use `#[Autoconfigure]` on the `CurrentTimeTool` class instead - at least for now - doing autoconfiguration via interface would be the next step maybe

src/mcp-bundle/doc/index.rst
```diff
@@ -63,6 +63,14 @@ Configuration
...
                     url: 'http://localhost:8000/sse' # URL to SSE endpoint of MCP server
 
+To add your tools to the MCP server, add the tag:
+
+.. code-block:: yaml
+
+    App\MCP\Tools\MyCustomTool:
+        tags:
+            - { name: 'mcp_tool' }
+
```
chr-hertel: let's feature the `#[Autoconfigure]` attribute here as well

TomHart:
You're welcome @chr-hertel. I believe I've address everything you mentioned if you'd like to take another look whenever you have time.

src/mcp-bundle/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 CHANGELOG
 =========
 
+
```
OskarStark:
```suggestion
```

src/mcp-bundle/CHANGELOG.md
```diff
@@ -17,4 +18,6 @@ CHANGELOG
  * Add `McpCommand` providing STDIO interface
  * Add bundle configuration for enabling/disabling transports
  * Add cache-based SSE message storage
- * Add service configuration for MCP server setup
\ No newline at end of file
+ * Add service configuration for MCP server setup
+ * Added `\Symfony\AI\McpBundle\DependencyInjection\ContainerBuilder\McpToolChainCompilerPass`
```
OskarStark:
```suggestion
 * Add `Symfony\AI\McpBundle\DependencyInjection\ContainerBuilder\McpToolChainCompilerPass`
```

src/mcp-bundle/doc/index.rst
```diff
@@ -63,6 +63,13 @@ Configuration
                 sse:
                     url: 'http://localhost:8000/sse' # URL to SSE endpoint of MCP server
 
+To add your tools to the MCP server, add the tag:
```
OskarStark:
```suggestion
To add your tools to the MCP server, add the following tag::
```

src/mcp-bundle/doc/index.rst
```diff
@@ -63,6 +63,13 @@ Configuration
...
+
+Example::
```
OskarStark:
```suggestion
```

OskarStark:
Could be a follow up PR, but I would like to have an attribute for this like in the agent

src/mcp-bundle/src/DependencyInjection/ContainerBuilder/McpToolChainCompilerPass.php
```diff
@@ -0,0 +1,50 @@
+<?php
+
+declare(strict_types=1);
```
chr-hertel: Please remove `declare(strict_types=1);`

TomHart:
@chr-hertel my bad on missing the strict types, updated now and rebased into 1 commit

OskarStark (approved):

src/mcp-bundle/doc/index.rst
```diff
@@ -63,6 +63,11 @@ Configuration
                 sse:
                     url: 'http://localhost:8000/sse' # URL to SSE endpoint of MCP server
 
+To add your tools to the MCP server, add the following tag::
+
+    #[AutoconfigureTag('mcp.tool')]
```
stof: this is not the right recommendation IMO:

- `AutoconfigureTag` (like all `Autoconfigure*` attributes) is about defining an autoconfiguration rule applying on all child types (its original intent is to be used on interfaces). Recommending to use it to apply a tag on a service directly is a bad idea, as it can have unintended side effects for non-final classes (as it will also impact autoconfigured services defined with child classes)
- if the goal is to automatically tag *autoconfigured* services implementing ToolExecutorInterface with the `mcp.tool` to make it simpler for projects adding a custom tool, the solution is for the MCP bundle to register an autoconfiguration rule for ToolExecutorInterface, so that projects only need to implement the interface and be done with it.

src/mcp-bundle/src/DependencyInjection/ContainerBuilder/McpToolChainCompilerPass.php
```diff
@@ -0,0 +1,48 @@
...
+    use PriorityTaggedServiceTrait;
+
+    public function process(ContainerBuilder $container): void
+    {
+        // If the user has already defined this service, don't override it.
+        if ($container->hasDefinition(ToolExecutorInterface::class)) {
+            return;
+        }
+
+        $definition = new Definition(ToolChain::class);
```
stof: you could always define a `mcp.tool_chain` service using a tagged iterator, and then defining a `ToolExecutorInterface` alias pointing at it (which projects can still override if they don't want to rely on the tool chain).
There is no need for this compiler pass IMO.

src/mcp-bundle/doc/index.rst
chr-hertel: I was thinking of using the attribute because it's currently hard to decide which interface is the right indicator for the tag here

@Nyholm, it would be the `Symfony\AI\McpSdk\Capability\Tool\IdentifierInterface`, right?

TomHart:
Sorry for the slow response, I've had tonsillitis so took some time to recover. So if I've understood correct (as some of these things are new to me), I'd create a definition of

```yaml
    mcp.tool_chain:
        arguments:
            $executors: !tagged_iterator mcp.tool

    Symfony\AI\McpSdk\Capability\Tool\ToolExecutorInterface: mcp.tool_chain
```

And then apply the `AutoconfigureTag` attribute to the `IdentifierInterface` class, providing @Nyholm confirms that's the correct one to target.

is that correct @chr-hertel @stof?

chr-hertel:
@TomHart no worries at all, hope you're well again :crossed_fingers: 

You are correct here, two things are needed a) registering based on the `IndentifierInterface` and b) injecting the collection into the chain implementation

For a) we can't simply add the `AutoconfigureTag` attribute tho, since that class lives in a different lib than the bundle - the trick here could be, to use the `registerForAutoconfiguration()` method for than in `McpBundle`, see https://symfony.com/doc/current/service_container/tags.html#autoconfiguring-tags

For b) i think you could just move the service definition from the compiler pass into the bundle class as well and use `tagged_iterator('mcp.tool')` for injecting the tools

TomHart:
I've changed it around now. I couldn't get it working with the `instanceof` autoconfigure, it just never picked anything up for some reason. Could you take another look now and see what you think?

Not sure why PHPUnit is failing, could it be re-tried to just see if it's a blip? 

Finally, had to change from `array` to `iterable` for `tagged_iterator` to work, if that's a problem to change I can see if there's any other solutions

chr-hertel:
Looks good to me now, can you please rebase - I just merged #165

TomHart:
Amazing! Done!

chr-hertel:
Thank you @TomHart.

TomHart:
@chr-hertel what's the best way to find the best priority issues to work on?

chr-hertel:
If you're up for something to just on, I'd always point to #16. I think Stores we're quite good progressed lately, but there are quite some Platforms that only are a draft, like OpenRouter or Replicate, but also some missing completely, like Deepseek or Perplexity.

One of my priority topics would be #167 - feel free to have a look at that PR and open PRs against that.

And other than that, the Store component has two extension points that could use some love, especially the `Symfony\AI\Store\Document\LoaderInterface` could use more implementations. Currently it's only a TextFileLoader, but could also be a PdfLoader, an RssLoader, an HtmlLoader, etc ... 

Happy to guide you more if you tell me what's interesting to you :)

PR #169, opened 2025-07-20
Fix the chatbot demo

fabpot:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Docs?         | no 
| Issues        | 
| License       | MIT

Updated the code to the current names.

### DIFF
--- a/demo/src/Audio/Chat.php
+++ b/demo/src/Audio/Chat.php
@@ -17,7 +17,7 @@ use Symfony\AI\Platform\Message\Content\Audio;
 use Symfony\AI\Platform\Message\Message;
 use Symfony\AI\Platform\Message\MessageBag;
 use Symfony\AI\Platform\PlatformInterface;
-use Symfony\AI\Platform\Result\TextResult;
+use Symfony\AI\Platform\Response\TextResponse;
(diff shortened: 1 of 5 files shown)


chr-hertel:
That shouldn't be needed anymore, we merged #154 already

fabpot:
That's because there is a `composer.lock` file in the demo directory. I suggest that we remove it until we have a first release. WDYT?

chr-hertel:
Ah, yes, I feel like that demo doesn't play well right now with the mono repo. removing the composer.lock might help :+1:

chr-hertel:
And maybe switching to Symfony CLI as a next step for the PHP runtime

fabpot:
> And maybe switching to Symfony CLI as a next step for the PHP runtime

I can work on it :)

What about using PostgreSQL for the vector DB? That would allow for a nice and simple way for developers to get started.

chr-hertel:
Postgres would easily integrate with standard DB tooling, yea, i like that - and since we have that bridge now, yes please :)

fabpot:
See #170 for a better fix.

PR #173, opened 2025-07-20
[AIBundle] Fix stream result profiling

### LINKED ISSUE #161: [AIBundle] Streaming responses missing from the profiler
While working on the streaming demo (#118), I discovered streaming AI responses do not show up in the profiler, even though the related HTTP requests / responses do show up there.

Looks like the issue with streaming responses missing in the profiler is due to 2 issues:
1. `DataCollector` does not implement `LateDataCollectorInterface` – this is an easy fix
1. `StreamResponse` holds a `\Generator` and thus, is not serializable

valtzu:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #161
| License       | MIT

Address 2 streaming-related issues:
1. When using `symfony/ai` from within `StreamedResponse` callback (in http controller), no data was collected due to not implementing `LateDataCollectorInterface`
2. An attempt to save `StreamResult` to the profiler storage always failed, because `\Generator`s are not serializable

Both are used in the Turbo Stream Bot demo, and now the profiler works. Though the toolbar still does not auto-update, but that is currently expected behavior.

### DIFF
--- a/src/ai-bundle/src/Profiler/TraceablePlatform.php
+++ b/src/ai-bundle/src/Profiler/TraceablePlatform.php
@@ -14,7 +14,9 @@ namespace Symfony\AI\AIBundle\Profiler;
 use Symfony\AI\Platform\Message\Content\File;
 use Symfony\AI\Platform\Model;
 use Symfony\AI\Platform\PlatformInterface;
+use Symfony\AI\Platform\Result\ResultInterface;
 use Symfony\AI\Platform\Result\ResultPromise;
+use Symfony\AI\Platform\Result\StreamResult;
 
 /**
  * @author Christopher Hertel <mail@christopher-hertel.de>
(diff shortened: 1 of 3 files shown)


src/ai-bundle/src/Profiler/TraceablePlatform.php
```diff
@@ -32,27 +34,50 @@ final class TraceablePlatform implements PlatformInterface
...
+    {
+        return $result = new StreamResult((function () use (&$result, $originalStream) {
+            $this->resultCache[$result] = '';
+            foreach ($originalStream as $chunk) {
+                yield $chunk;
+                if (\is_string($chunk)) {
+                    $this->resultCache[$result] .= $chunk;
+                }
+            }
+        })());
```
chr-hertel, line 81: nice one :clap:

chr-hertel (approved):
Works :+1: 

image

chr-hertel:
Thank you @valtzu.

PR #175, opened 2025-07-20
[AIBundle] Simplify agent-as-tool configuration

### LINKED ISSUE #172: [AIBundle] Simplify agent-as-tool configuration
I'm proposing to simplify agent-as-tool (chain-in-chain) configuration.

Current situation
---

```yaml
  tools:
    - service: ai.agent.my_agent
      name: my_agent
      description: My wonderful agent
      is_agent: true
```

The proposal
---

```yaml
  tools:
    - agent: my_agent # This references the agent name and also serves as default tool name unless name is defined explicitly here
      description: My wonderful agent

### PR DESCRIPTION
**Pull request by valtzu**
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes, kinda
| Docs?         | yes, adjusted
| Issues        | Fix #172
| License       | MIT

Simplify agent as tool setup:
1. Set `agent: name` instead of `is_agent: true, service: ai.agent.name`
2. Use `agent` as fallback for `name`

This is a BC break, but if we wanted to, we could keep `is_agent` around and have the new way only as shortcut.

### CODE REVIEW

File: src/ai-bundle/src/AIBundle.php [hunk 1]
```diff
@@ -319,10 +319,14 @@ private function processAgentConfig(string $name, array $config, ContainerBuilde
 
                 $tools = [];
                 foreach ($config['tools']['services'] as $tool) {
+                    if (isset($tool['agent'])) {
+                        $tool['name'] ??= $tool['agent'];
+                        $tool['service'] ??= \sprintf('ai.agent.%s', $tool['agent']);
```
chr-hertel: `service` can't be set, right?
```suggestion
                        $tool['service'] = \sprintf('ai.agent.%s', $tool['agent']);
```

src/ai-bundle/config/options.php
```diff
@@ -114,18 +114,22 @@
...
-                                            ->booleanNode('is_agent')->defaultFalse()->end()
                                         ->end()
                                         ->beforeNormalization()
                                             ->ifString()
                                             ->then(function (string $v) {
                                                 return ['service' => $v];
                                             })
                                         ->end()
+                                        ->validate()
+                                            ->ifTrue(static fn ($v) => !(empty($v['agent']) xor empty($v['service'])))
```
chr-hertel, line 133: haven't seen that operator for a while :heart_eyes: :+1:

chr-hertel:
needs a rebase

src/ai-bundle/src/AIBundle.php
valtzu: Right :+1: 

My original idea was to allow user-provided agent service too, but in 99% cases it should not be needed, and if it is, then you can just define a service with `class: Symfony\AI\Agent\Toolbox\Tool\Agent`, pass your agent as args and then reference it in the `tools[].service` without defining `tools[].agent` at all.

---

Removed the null-coalescing operator

OskarStark:
Can you please add a test case for the DI config? Thanks

valtzu:
Ok, should be good now

chr-hertel (approved):

chr-hertel:
Thank you @valtzu.

PR #177, opened 2025-07-21
[Agent][AIBundle] Improve tool argument resolving

valtzu:
| Q             | A
| ------------- | ---
| Bug fix?      | yes, kinda
| New feature?  | no
| Docs?         | no
| License       | MIT

Address some issues regarding argument value denormalization:

1. Arrays of objects did not work, even with `serializer` that had `ArrayDenormalizer` in it
1. We did not check for mandatory parameters before
1. Extra arguments from the platform caused an exception – now they are ignored
    * If we still want to throw, that's fine, but it should be a more managed exception instead of `Trying to call getType() on null`
1. Use `symfony/type-info` instead of trying to resolve the type ourselves from reflection


fixtures/Tool/ToolArrayMultidimensional.php
OskarStark, line 1: I think (in a separate PR we should move the fixtures under `src/(agent|*)/tests/Fixtures`

chr-hertel (approved):

chr-hertel:
Good catch, thanks @valtzu.

PR #182, opened 2025-07-22
[Plateform][HuggingFaces] Fix: throw exceptions directly instead of using 'return throw'

### LINKED ISSUE #181: [Plateform][HuggingFace] Incorrect Use of `return throw` in ResultConverter
In ResultConverter.php, the code currently uses return throw to throw exceptions for certain HTTP status codes (503 and 404):

https://github.com/symfony/ai/blob/6c41cd26ea6a2f031aa18dc4d05cb22e581d2134/src/platform/src/Bridge/HuggingFace/ResultConverter.php#L48-L55

This is not valid PHP syntax. Exceptions should be thrown directly using the throw statement, not with return throw.

Suggested Fix: Replace return throw with direct throw statements for all exception cases in ResultConverter.php.

bernard-ng:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #181
| License       | MIT

Replace `return throw` with direct throw statements for all exception cases in ResultConverter

### DIFF
--- a/src/platform/src/Bridge/HuggingFace/ResultConverter.php
+++ b/src/platform/src/Bridge/HuggingFace/ResultConverter.php
@@ -47,11 +47,11 @@ final readonly class ResultConverter implements PlatformResponseConverter
     {
         $httpResponse = $result->getObject();
         if (503 === $httpResponse->getStatusCode()) {
-            return throw new RuntimeException('Service unavailable.');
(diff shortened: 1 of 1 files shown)


chr-hertel (approved):

chr-hertel:
Good catch, thanks @bernard-ng.

PR #184, opened 2025-07-23
[Agent] Add `@throws` tag

VincentLanglet:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | yes
| Issues        | Fix #...
| License       | MIT

Hi @chr-hertel, this lib seems to lack of `@throws` phpdoc.

For static analysis usage (and developer experience) it seems better to tell which methods are throwing exceptions.
I didn't look for every methods but started with two important methods imho.

### DIFF
--- a/src/agent/src/AgentInterface.php
+++ b/src/agent/src/AgentInterface.php
@@ -11,6 +11,7 @@
 
 namespace Symfony\AI\Agent;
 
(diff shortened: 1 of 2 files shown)


OskarStark:
Thank you @VincentLanglet.

PR #186, opened 2025-07-23
Add missing @throws tags to interfaces

OskarStark:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | yes
| Issues        | --
| License       | MIT

Following the pattern from #184, this commit adds missing @throws documentation to interface methods that can throw exceptions. This improves developer experience and helps static analysis tools understand the expected behavior.


nicolas-grekas:
The annotation should be used when the exceptions are part of the API. And then, it should document when they should be expected. Is that the case everywhere?

OskarStark:
> The annotation should be used when the exceptions are part of the API. And then, it should document when they should be expected. Is that the case everywhere?

Thanks, updated

VincentLanglet:
Most people will inject `AgentInterface` into their service rather than `Agent`.
So it's important to keep a `@throws` annotation on interface indeed.

(But obviously it's ok to improve the documentation and be more specific about which exception is thrown and why)

chr-hertel:
I just want to avoid to only add those because PhpStorm inspections are complaining - when we add them, with value please and people should consider them actually 🙏

VincentLanglet:
> I just want to avoid to only add those because PhpStorm inspections are complaining 

More than PhpStorm, there is StaticAnalysis tools like PHPStan and Psalm.
https://phpstan.org/blog/bring-your-exceptions-under-control

A wrong documented interface leads to 
- missing try/catch
- false-positif dead catch

src/mcp-sdk/src/Server/JsonRpcHandler.php
```diff
@@ -54,8 +54,8 @@ public function __construct(
     /**
      * @return iterable<string|null>
      *
-     * @throws ExceptionInterface
-     * @throws \JsonException
+     * @throws ExceptionInterface When a handler throws an exception during message processing
+     * @throws \JsonException     When JSON encoding of the response fails
```
nicolas-grekas, line 58:
```suggestion
     * @throws \JsonException     When JSON decoding of the response fails
```

OskarStark: *encoding* looks correct to me, the JsonException is thrown when encoding the response fails (in the encodeResponse method). The method catches and handles `JsonException` from decoding the input gracefully without re-throwing it, so only encoding failures can propagate up 🤔

VincentLanglet: Btw, I dunno if Symfony should try/catch this exception to throw a Symfony Exception instead...

VincentLanglet (approved):

src/mcp-sdk/src/Server/JsonRpcHandler.php
OskarStark: I would say this is out of scope of this PR and to be discussed

chr-hertel:
> LGTM from afar, a domain expert might review to see if that makes sense as part of the public API of course

meaning we only add them on interfaces/classes designed for user land code to call, right?

So for example the `AgentProcessorInterface` or the `NotificationHandlerInterface` is only to be called by the components themselves and basically internal - so would we add those `throws`  tags there as well?

VincentLanglet:
> So for example the `AgentProcessorInterface` or the `NotificationHandlerInterface` is only to be called by the components themselves and basically internal - so would we add those `throws` tags there as well?

If a non-internal class use an AgentProcessorInterface, it might help to know the exceptions thrown by `AgentProcessorInterface` in order to document the phpdoc of the non-internal class.

Also, if the `AgentProcessorInterface` or the `NotificationHandlerInterface` are supposed to be internal ; shouldn't you add `@internal` on it (?)

chr-hertel (approved):
Let's do it :+1: not that internal since people could implement or instantiate those classes or interfaces - thanks @OskarStark!

PR #188, opened 2025-07-23
[Store][InMemory] Fix cosine similarity sorting

### LINKED ISSUE #187: [Store] Incorrect Similarity Sorting in `InMemoryStore` for Cosine Similarity
### Description

The `InMemoryStore::query()` method currently sorts all similarity strategies using the same logic: ascending order of the calculated "distance". This works well for distance-based strategies like Euclidean or Manhattan, but it results in **incorrect ordering for Cosine Similarity**, which returns higher values for more relevant matches (e.g., 0.9 is *more* similar than 0.3).

### Problem
We set the distance as cosine similarity using:
```php
 'distance' => $strategy($vectorDocument, $vector),
// for cosineStrategy will be 
'distance' => $this->cosineSimilarity($doc, $query)
```

Then the following logic is used across all strategies:

```php
...
usort($currentEmbeddings, fn ($a, $b) => $a['distance'] <=> $b['distance']);
```

For cosine similarity, this is incorrect, as it ranks lower values first (less similar), which is the opposite of what is intended.

### Steps to Reproduce:

1. Create an `InMemoryStore`  instance configured with `self::COSINE_SIMILARITY`
2.  Add several `VectorDocument` instances to the store with varying degrees of similarity to a potential query vector.
3. Perform a `query()` with a target Vector. 
4. Observe the order of the returned VectorDocument objects.

### Expected Behavior:
The VectorDocument instances should be sorted in descending order of their cosine similarity scores (i.e., the most similar documents appear at the beginning of the returned array).

### Actual Behavior:
The VectorDocument instances are sorted in ascending order of their cosine similarity scores (i.e., the least similar documents appear at the beginning of the returned array). This is due to the generic usort comparison `$embedding['distance'] <=> $nextEmbedding['distance']` which implicitly sorts in ascending order.

### Proposed Solution
Normalize cosine similarity as a distance by converting it to a distance with this common conversion:

```php
return 1 - cosineSimilarity(...);
```

This would unify the behavior across all strategies, allowing consistent sorting and interpretation:

- Cosine similarity score (0.0 to 1.0) → distance (0.0 best, 1.0 worst)
- Works with the current usort logic
- Eliminates the need for conditional sorting logic per strategy

RamyHakam:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Docs?         | no 
| Issues        | Fix #187 ... 
| License       | MIT

This PR fixes #187, the issue where cosine similarity values were incorrectly treated as distances in the InMemoryStore::query() method, resulting in reversed result ordering. 

The fix converts cosine similarity to distance using `1-similarity`, which is the common conversion between similarity and distance.

 Ensuring consistent sorting behavior across all strategies. Includes updated unit tests to validate.

### DIFF
--- a/src/store/src/InMemoryStore.php
+++ b/src/store/src/InMemoryStore.php
@@ -20,7 +20,7 @@ use Symfony\AI\Store\Exception\InvalidArgumentException;
  */
 final class InMemoryStore implements VectorStoreInterface
 {
-    public const COSINE_SIMILARITY = 'cosine';
+    public const COSINE_DISTANCE = 'cosine';
     public const ANGULAR_DISTANCE = 'angular';
     public const EUCLIDEAN_DISTANCE = 'euclidean';
     public const MANHATTAN_DISTANCE = 'manhattan';
@@ -32,7 +32,7 @@ final class InMemoryStore implements VectorStoreInterface
     private array $documents = [];
 
     public function __construct(
-        private readonly string $similarity = self::COSINE_SIMILARITY,
+        private readonly string $distance = self::COSINE_DISTANCE,
     ) {
     }
 
@@ -48,13 +48,13 @@ final class InMemoryStore implements VectorStoreInterface
      */
     public function query(Vector $vector, array $options = [], ?float $minScore = null): array
     {
-        $strategy = match ($this->similarity) {
-            self::COSINE_SIMILARITY => $this->cosineSimilarity(...),
+        $strategy = match ($this->distance) {
+            self::COSINE_DISTANCE => $this->cosineDistance(...),
             self::ANGULAR_DISTANCE => $this->angularDistance(...),
             self::EUCLIDEAN_DISTANCE => $this->euclideanDistance(...),
             self::MANHATTAN_DISTANCE => $this->manhattanDistance(...),
             self::CHEBYSHEV_DISTANCE => $this->chebyshevDistance(...),
-            default => throw new InvalidArgumentException(\sprintf('Unsupported similarity strategy "%s"', $this->similarity)),
+            default => throw new InvalidArgumentException(\sprintf('Unsupported distance metric "%s"', $this->distance)),
         };
 
         $currentEmbeddings = array_map(
@@ -78,6 +78,11 @@ final class InMemoryStore implements VectorStoreInterface
             static fn (array $embedding): VectorDocument => $embedding['document'],
             $currentEmbeddings,
         );
+    }
+
+    private function cosineDistance(VectorDocument $embedding, Vector $against): float
+    {
+        return 1 - $this->cosineSimilarity($embedding, $against);
     }
 
     private function cosineSimilarity(VectorDocument $embedding, Vector $against): float
--- a/src/store/tests/InMemoryStoreTest.php
+++ b/src/store/tests/InMemoryStoreTest.php
@@ -23,7 +23,7 @@ use Symfony\Component\Uid\Uuid;
 final class InMemoryStoreTest extends TestCase
 {
     #[Test]
-    public function storeCanSearchUsingCosineSimilarity(): void
+    public function storeCanSearchUsingCosineDistance(): void
     {
         $store = new InMemoryStore();
         $store->add(
@@ -32,7 +32,9 @@ final class InMemoryStoreTest extends TestCase
             new VectorDocument(Uuid::v4(), new Vector([0.3, 0.7, 0.1])),
         );
 
-        self::assertCount(3, $store->query(new Vector([0.0, 0.1, 0.6])));
+        $result = $store->query(new Vector([0.0, 0.1, 0.6]));
+        self::assertCount(3, $result);
+        self::assertSame([0.1, 0.1, 0.5], $result[0]->vector->getData());
 
         $store->add(
             new VectorDocument(Uuid::v4(), new Vector([0.1, 0.1, 0.5])),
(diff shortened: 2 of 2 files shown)


src/store/src/InMemoryStore.php
```diff
@@ -49,7 +49,7 @@ public function add(VectorDocument ...$documents): void
     public function query(Vector $vector, array $options = [], ?float $minScore = null): array
     {
         $strategy = match ($this->similarity) {
-            self::COSINE_SIMILARITY => $this->cosineSimilarity(...),
+            self::COSINE_DISTANCE => $this->cosineDistance(...),
```
OskarStark, line 52: Now we have cosine distance and cosine similarity but only one constant?

src/store/src/InMemoryStore.php
```diff
@@ -80,6 +80,11 @@ public function query(Vector $vector, array $options = [], ?float $minScore = nu
         );
     }
 
+    private function cosineDistance(VectorDocument $embedding, Vector $against): float
+    {
+        return 1 - $this->cosineSimilarity(...);
+    }
+
     private function cosineSimilarity(VectorDocument $embedding, Vector $against): float
```
OskarStark, line 88: I would remove this method and move the body to the cosineDistance method

src/store/src/InMemoryStore.php
```diff
@@ -20,7 +20,7 @@
  */
 final class InMemoryStore implements VectorStoreInterface
 {
-    public const COSINE_SIMILARITY = 'cosine';
+    public const COSINE_DISTANCE = 'cosine';
```
OskarStark, line 23: So we really need to rename this constant? This makes this PR a BC break

src/store/src/InMemoryStore.php
RamyHakam: We still need to calculate `cosineSimilarity()` internally for both `cosineDistance` and `angularDistance` since angular distance is defined as `acos(cosineSimilarity) / π`.

src/store/src/InMemoryStore.php
RamyHakam: I believe we should rename it because the expected result sorting differs between similarity and distance: 
similarity is typically sorted descending (higher is better), while distance is sorted ascending (lower is better).

Also, Since all other metrics use distance,  I think unifying the strategy for consistency and clarity makes sense.

src/store/src/InMemoryStore.php
RamyHakam: Renamed to distanceMetric to reduce confusion and reflect that all strategies use distance, not similarity.

src/store/src/InMemoryStore.php
OskarStark: Got it thanks

src/store/src/InMemoryStore.php
OskarStark: Ok fine then

src/store/src/InMemoryStore.php
```diff
@@ -48,18 +48,18 @@ public function add(VectorDocument ...$documents): void
...
-        $strategy = match ($this->similarity) {
-            self::COSINE_SIMILARITY => $this->cosineSimilarity(...),
+        $distanceMetricStrategy = match ($this->distanceMetric) {
+            self::COSINE_DISTANCE => $this->cosineDistance(...),
             self::ANGULAR_DISTANCE => $this->angularDistance(...),
             self::EUCLIDEAN_DISTANCE => $this->euclideanDistance(...),
             self::MANHATTAN_DISTANCE => $this->manhattanDistance(...),
             self::CHEBYSHEV_DISTANCE => $this->chebyshevDistance(...),
-            default => throw new InvalidArgumentException(\sprintf('Unsupported similarity strategy "%s"', $this->similarity)),
+            default => throw new InvalidArgumentException(\sprintf('Unsupported Distance metric "%s"', $this->distanceMetric)),
```
OskarStark:
```suggestion
            default => throw new InvalidArgumentException(\sprintf('Unsupported distance metric "%s"', $this->distanceMetric)),
```

src/store/src/InMemoryStore.php
```diff
@@ -48,18 +48,18 @@ public function add(VectorDocument ...$documents): void
      */
     public function query(Vector $vector, array $options = [], ?float $minScore = null): array
     {
-        $strategy = match ($this->similarity) {
-            self::COSINE_SIMILARITY => $this->cosineSimilarity(...),
+        $distanceMetricStrategy = match ($this->distanceMetric) {
```
OskarStark:
```suggestion
        $strategy = match ($this->distanceMetric) {
```

no need to rename I guess

src/store/src/InMemoryStore.php
```diff
@@ -32,7 +32,7 @@ final class InMemoryStore implements VectorStoreInterface
     private array $documents = [];
 
     public function __construct(
-        private readonly string $similarity = self::COSINE_SIMILARITY,
+        private readonly string $distanceMetric = self::COSINE_DISTANCE,
```
OskarStark:
```suggestion
        private readonly string $distance = self::COSINE_DISTANCE,
```

chr-hertel:
Makes sense to you @Guikingone?

Guikingone:
Yep, small mistake on my side, nice catch @RamyHakam 👍🏻

chr-hertel (approved):

chr-hertel:
Thank you @RamyHakam.

PR #189, opened 2025-07-23
Fix MariaDB Store query with minScore argument

yoye:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | 
| License       | MIT

Fix SQL query generation with dynamic field name in WHERE clause

Previously, the WHERE clause used a raw format string with a `%1$s` placeholder, but it was not processed by `sprintf`, resulting in the literal `%1$s` being inserted into the SQL. This caused the prepared SQL statement to be incorrect and fail at runtime.

### DIFF
--- a/src/store/src/Bridge/MariaDB/Store.php
+++ b/src/store/src/Bridge/MariaDB/Store.php
@@ -113,7 +113,7 @@ final readonly class Store implements VectorStoreInterface, InitializableStoreIn
                     SQL,
                 $this->vectorFieldName,
                 $this->tableName,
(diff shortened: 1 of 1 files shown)


OskarStark (commented):
Can you please add a testcase to avoid further regression?

OskarStark (approved):

OskarStark:
This test should pass after merge
* https://github.com/symfony/ai/pull/190

OskarStark:
Thank you @yoye.

PR #192, opened 2025-07-23
Replace test method prefix with `#[Test]` attribute

OskarStark:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | --
| License       | MIT

Updated all test files that were using the 'test' prefix convention to use the modern PHPUnit #[Test] attribute instead. This brings consistency across the codebase as the majority of tests were already using the attribute syntax.


valtzu:
Could consider adding php-cs-fixer config

```php
        'php_unit_test_annotation' => ['style' => 'annotation'],
        'php_unit_attributes' => true,
```

PR #193, opened 2025-07-23
[Platform] Add `InMemoryPlatform` and `InMemoryRawResult` for testing

RamyHakam:
| Q             | A
| ------------- | ---
| Bug fix?      |no
| New feature?  | yes 
| Docs?         | yes 
| Issues        | none 
| License       | MIT

This PR adds two test-friendly classes to the Symfony AI Platform component:

###  InMemoryPlatform

The `InMemoryPlatform` is a test-specific implementation of `PlatformInterface` that returns a `TextResult` wrapped in a `ResultPromise`.
 It supports:
- A fixed string response (e.g. `'Fake reply'`)
- A callable response based on the `Model`, `input`, and `options`

This makes it ideal for unit or integration tests, or even functional tests where a full AI stack is not needed.

###  InMemoryRawResult

The `InMemoryRawResult` implements `RawResultInterface` and is used internally by `InMemoryPlatform`. It wraps:

- A structured array (`getData()`)
- A generic object (`getObject()`)

to simulate what real raw results look like in production usage.

### The Problem

While testing new code that uses the AI platform, developers often need to manually create `PlatformInterface` instances. The existing `Platform` and `PlatformFactory` classes are tightly coupled to real AI providers.
 This in-memory implementation simplifies unit and integration testing by allowing developers to inject a dummy platform and return mock results without any external API calls.

### Example

```php
$platform = new InMemoryPlatform('Test reply');
$response = $platform->invoke(new Model('test'), 'Hello');
echo $response->asText(); // "Test reply"

### DIFF
--- a/src/platform/src/InMemoryPlatform.php
+++ b/src/platform/src/InMemoryPlatform.php
@@ -1,0 +1,52 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Platform;
+
+use Symfony\AI\Platform\Result\InMemoryRawResult;
+use Symfony\AI\Platform\Result\ResultPromise;
+use Symfony\AI\Platform\Result\TextResult;
+
+/**
+ * A fake implementation of PlatformInterface that returns fixed or callable responses.
+ *
+ * Useful for unit or integration testing without real API calls.
+ *
+ * @author Ramy Hakam <pencilsoft1@gmail.com>
+ */
+class InMemoryPlatform implements PlatformInterface
+{
+    /**
+     * The mock result can be a string or a callable that returns a string.
+     * If it's a closure, it receives the model, input, and  optionally options as parameters like a real platform call.
+     */
+    public function __construct(private readonly \Closure|string $mockResult)
+    {
+    }
+
+    public function invoke(Model $model, array|string|object $input, array $options = []): ResultPromise
+    {
+        $resultText = $this->mockResult instanceof \Closure
+            ? ($this->mockResult)($model, $input, $options)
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: src/platform/src/InMemoryPlatform.php [hunk 1]
```diff
@@ -0,0 +1,57 @@
...
+ * Useful for unit or integration testing without real API calls.
+ *
+ * @author Ramy Hakam <pencilsoft1@gmail.com>
+ */
+class InMemoryPlatform implements PlatformInterface
+{
+    /**
+     * The mock result can be a string or a callable that returns a string.
+     * If it's a callable, it receives the model, input, and options as parameters like a real platform call.
+     * /**
```
OskarStark:
```suggestion
     *
```

src/platform/src/InMemoryPlatform.php
```diff
@@ -0,0 +1,57 @@
...
+{
+    /**
+     * The mock result can be a string or a callable that returns a string.
+     * If it's a callable, it receives the model, input, and options as parameters like a real platform call.
+     *
+     * @var string|(callable(Model, array<string,mixed>|string|object, array<string,mixed>): string)
+     */
+    private $mockResult;
+
+    public function __construct(callable|string $mockResult = 'Default test response')
```
chr-hertel: let's use constructor property promotion here - and I'm not sure if a default value makes sense here :thinking: 

isn't the usual case that you want to provide something explicitly?

RamyHakam: @chr-hertel  I already tried that, but property promotion does not allow `callable`
PHP will throw `Fatal error: Property cannot have type 'callable'`

And regarding the default value, you are right, it's a valid  point, I will remove it

chr-hertel: how about using `Closure` instead?

RamyHakam: @chr-hertel  Done , Thaks

chr-hertel:
Thank you @RamyHakam.

PR #196, opened 2025-07-24
Add PHP-CS-Fixer rules for PHPUnit attributes

OskarStark:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | --
| License       | MIT

Configure php_unit_test_annotation and php_unit_attributes rules to enforce modern PHPUnit attribute syntax and remove test prefixes from methods that use the #[Test] attribute.

cc @valtzu

### DIFF
--- a/src/agent/tests/Toolbox/MetadataFactory/ChainFactoryTest.php
+++ b/src/agent/tests/Toolbox/MetadataFactory/ChainFactoryTest.php
@@ -47,7 +47,7 @@ final class ChainFactoryTest extends TestCase
     }
 
     #[Test]
-    public function testGetMetadataNotExistingClass(): void
+    public function getMetadataNotExistingClass(): void
     {
         self::expectException(ToolException::class);
         self::expectExceptionMessage('The reference "NoClass" is not a valid tool.');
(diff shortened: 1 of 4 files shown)


src/agent/tests/Toolbox/MetadataFactory/ChainFactoryTest.php
```diff
@@ -47,7 +47,7 @@ protected function setUp(): void
     }
 
     #[Test]
-    public function testGetMetadataNotExistingClass(): void
+    public function getMetadataNotExistingClass(): void
```
nicolas-grekas, line 50: that's strange to me, it blurs the line between test cases and helper methods
we should also consider making the practices on symfony/symfony and symfoyn/ai converge
this looks diverging here to me

derrabus: I agree. Having a `void` method with a `get` name prefix feels espcially weird.

chr-hertel: this is mostly about (not) having the `test*` prefix in the method name, not about the `void`, right?

i don't have a real preference other than being more used to having the prefix over the attribute, but mostly care about streamlining here :)

@OskarStark any hard feelings about going the symfony/symfony way here?

derrabus: > this is mostly about (not) having the `test*` prefix in the method name

Yes.

fabpot: I agree that the new method names are confusing.
I don't think using the Test attribute makes things better.

chr-hertel:
Closed in favor of #214

PR #198, opened 2025-07-24
[Store][PG] Allow to change vector type and index type

lyrixx:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        |
| License       | MIT

Gemini gives me vector of 3072 dimms.
But PG vector can't handle that much dimms. [ref](https://github.com/pgvector/pgvector/issues/461)
So a workarround is to change the vectory type, and so the index.

Example:

```
app=# \d gemini
                   Table "public.gemini"
  Column   |     Type      | Collation | Nullable | Default 
-----------+---------------+-----------+----------+---------
 id        | uuid          |           | not null | 
 metadata  | jsonb         |           |          | 
 embedding | halfvec(3072) |           | not null | 
Indexes:
    "gemini_pkey" PRIMARY KEY, btree (id)
    "gemini_embedding_idx" hnsw (embedding halfvec_cosine_ops)

app=# select count(*) from gemini ;
 count 
-------
   384
(1 row)

```

### DIFF
--- a/src/store/src/Bridge/Postgres/Store.php
+++ b/src/store/src/Bridge/Postgres/Store.php
@@ -120,6 +120,13 @@ final readonly class Store implements VectorStoreInterface, InitializableStoreIn
         return $documents;
     }
 
+    /**
+     * @param array{vector_type?: string, vector_size?: positive-int, index_method?: string, index_opclass?: string} $options
+     *
+     * Good configurations $options are:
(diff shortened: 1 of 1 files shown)


chr-hertel:
Thank you @lyrixx.

PR #200, opened 2025-07-24
[MCP Bundle] (doc) add remote mcp server example

xavierleune:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Docs?         | yes 
| Issues        | N/A 
| License       | MIT

I suggest to add a quick documentation around remote mcp server, as it's actually hard to find a quick how to online. WDYT ?

### DIFF
--- a/src/mcp-bundle/doc/index.rst
+++ b/src/mcp-bundle/doc/index.rst
@@ -63,6 +63,53 @@ Configuration
                 sse:
                     url: 'http://localhost:8000/sse' # URL to SSE endpoint of MCP server
 
+Usage as a remote MCP server
+----------------------------
+
+If your local project runs in docker, it may be easier to run the server as a remote server, thanks to the node package ``mcp-remote``. Node is required.
+
+Configuration for claude desktop:
+
+.. code-block:: json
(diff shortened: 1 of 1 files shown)


chr-hertel:
I agree that more examples help here, but I would rather bring in the SSE/Streamable support here - maybe with an example of the [inspector](https://modelcontextprotocol.io/docs/tools/inspector) - what do you think?

xavierleune:
@chr-hertel thanks, the example for the inspector would be great too ; I think both may be good as the inspector doesn't help to test your mcp in claude desktop. What do you think about covering both ?

xavierleune:
(just in case you missed it: mcp-remote supports streamable even if doc is sparse about it: https://github.com/geelen/mcp-remote/pull/32)

chr-hertel:
Btw, I saw claude desktop working with STDIO over wsl.exe and docker compose exec the other day - so not sure the mentioned problem is an issue anymore

chr-hertel:
Hi @xavierleune any argument for not closing here  for now?

PR #203, opened 2025-07-25
[Agent] Add unit tests for `SimilaritySearch` and `YouTubeTranscriber` tools

OskarStark:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | --
| License       | MIT


src/agent/tests/Toolbox/Tool/YouTubeTranscriberTest.php
```diff
@@ -0,0 +1,52 @@
...
+
+        new YouTubeTranscriber($httpClient);
+    }
+
+    #[Test]
+    public function constructorSucceedsWhenPackageIsInstalled(): void
+    {
+        if (!class_exists(TranscriptListFetcher::class)) {
+            $this->markTestSkipped('mrmysql/youtube-transcript package is not installed');
+        }
```
chr-hertel: how would that happen? get it for extensions, but with the require-dev, that should be safe

OskarStark: 🧐 indeed

PR #209, opened 2025-07-26
[RFC] Add PHPStan rules for forbidden methods and unnecessary named arguments

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | 
| License       | MIT

I wanted to catch some stuff that a) happens to me and b) could be a rule to enforce while working with named arguments if you ask me.

- Add ForbiddenMethodCallsRule to prevent debugging functions like dump(), var_dump(), print_r(), dd() in production code
- Add ForbidUnnecessaryNamedArgumentsRule to detect unnecessary named arguments when all parameters are provided in correct order
- Support function calls, method calls, and constructor calls
- Handle mixed positional and named arguments
- Register both rules in extension.neon for automatic enforcement across all components

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/.phpstan/ForbidNativeExceptionRule.php
+++ b/.phpstan/ForbidNativeExceptionRule.php
@@ -28,6 +28,7 @@ use PHPStan\Rules\RuleErrorBuilder;
  * for better error handling and consistency across the Symfony AI monorepo.
  *
  * @implements Rule<Node>
+ * @author Claude <noreply@anthropic.com>
  */
 final class ForbidNativeExceptionRule implements Rule
 {
--- a/.phpstan/ForbidUnnecessaryNamedArgumentsRule.php
+++ b/.phpstan/ForbidUnnecessaryNamedArgumentsRule.php
@@ -1,0 +1,351 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\PHPStan;
+
+use PhpParser\Node;
+use PhpParser\Node\Arg;
+use PhpParser\Node\Expr\FuncCall;
+use PhpParser\Node\Expr\MethodCall;
+use PhpParser\Node\Expr\StaticCall;
+use PhpParser\Node\Identifier;
+use PHPStan\Analyser\Scope;
+use PHPStan\Reflection\FunctionReflection;
+use PHPStan\Reflection\MethodReflection;
+use PHPStan\Rules\Rule;
+use PHPStan\Rules\RuleError;
+use PHPStan\Rules\RuleErrorBuilder;
+use PHPStan\Type\Type;
+
+/**
+ * PHPStan rule that forbids unnecessary use of named arguments.
+ *
+ * This rule detects when named arguments are used but provide no benefit:
+ * - All required parameters are provided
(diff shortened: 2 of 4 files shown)


.phpstan/ForbidNativeExceptionRule.php
```diff
@@ -28,6 +28,7 @@
  * for better error handling and consistency across the Symfony AI monorepo.
  *
  * @implements Rule<Node>
+ * @author Claude <noreply@anthropic.com>
```
OskarStark, line 31: Should be removed

.phpstan/ForbiddenMethodCallsRule.php
```diff
@@ -0,0 +1,88 @@
...
+/**
+ * PHPStan rule that forbids usage of specific debugging and output functions.
+ *
+ * This rule enforces code quality by preventing debugging functions from being
+ * committed to the codebase, helping maintain clean production code.
+ *
+ * @implements Rule<FuncCall>
+ * @author Claude <noreply@anthropic.com>
+ */
+final class ForbiddenMethodCallsRule implements Rule
```
derrabus, line 31: You could consider using https://github.com/spaze/phpstan-disallowed-calls for this.

.phpstan/ForbidUnnecessaryNamedArgumentsRule.php
```diff
@@ -0,0 +1,351 @@
...
+ *
+ * This rule detects when named arguments are used but provide no benefit:
+ * - All required parameters are provided
+ * - Parameters are in the correct order
+ * - No optional parameters are skipped
+ *
+ * @implements Rule<Node>
+ * @author Claude <noreply@anthropic.com>
+ */
+final class ForbidUnnecessaryNamedArgumentsRule implements Rule
```
OskarStark, line 39: Sounds like a nice one for phpstan official repo 🙂

chr-hertel:
Thanks @OskarStark & @derrabus for having a look! Would you in general agree on that rule about named arguments or is it too opinionated?

.phpstan/ForbidUnnecessaryNamedArgumentsRule.php
chr-hertel: i was rather thinking of PHP CS Fixer - feels a bit too opinionated and more about style than types

derrabus:
> Would you in general agree on that rule about named arguments or is it too opinionated?

What does the rule do? Forbid the usage of named arguments if I could've made the same call with ordered arguments?

It is opinionated all right. Not sure what you're trying to solve. If I have a value object like this:

```php
final readonly class Foo {
    public function __construct(
        public string $a,
        public string $b,
        public string $c,
    ) {}
}
```

Then I wouldn't be allowed to do this?

```php
$foo = new Foo(
    a: 'some value',
    b: 'some value',
    c: 'some value',
);
```

Meh.

chr-hertel:
> Not sure what you're trying to solve.

To be honest, I just don't want to think about it while reviewing code :grimacing: 
(edit: maybe i'm too lazy here :D)

> Then I wouldn't be allowed to do this?
> 
> ```
> $foo = new Foo(
>     a: 'some value',
>     b: 'some value',
>     c: 'some value',
> );
> ```

Yup, and it's not that I care to much, but was thinking that's the easier rule.

But if we don't mind in general, I can just drop thinking about it while reviewing :laughing:

chr-hertel:
no real need, let's close it

PR #211, opened 2025-07-26
[Platform] Add tooling support for ollama to allow agent and toolbox usage

JoshuaBehrens:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes?
| Docs?         | no
| License       | MIT

Running a toolbox agent with ollama was not yet supported. This ensures, that tool messages and responses work with a locally hosted ollama instance. It seems not to be a fix as it was not meant to work and it is not really a mentionable feature as one could expect it to work like the others so I am not really sure to work where to add more content for humans. I tested it with the given new example in `examples/ollama/toolcall.php`.

A current design decision had me struggling a bit. I was not yet able to solve in a good way the fact, that ollama itself cannot claim every model to have tool support. So someone who uses a model needs to know before its usage whether it has model support. This can be queried easily but the capabilities are part of the model, that you need to create before using the platform to query in the name of the model. So this is like hen-egg. If the model itself would not know its capabilities but there is a service, that has model and platform at hand, could evaluate it (and cache it). Is this something we want to establish?

Possible query:
```shell
curl http://localhost:11434/api/chat -d '{
  "model": "llama3:8b-instruct-q6_K",
  "tools": [{}],
  "stream": false
}'
{"error":"llama3:8b-instruct-q6_K does not support tools"}
```

### DIFF
--- a/src/platform/CHANGELOG.md
+++ b/src/platform/CHANGELOG.md
@@ -57,5 +57,6 @@ CHANGELOG
  * Add support for embeddings generation across multiple providers
  * Add response promises for async operations
  * Add InMemoryPlatform and InMemoryRawResult for testing Platform without external Providers calls
+ * Add tool calling support for Ollama platform
 
 
--- a/src/platform/src/Bridge/Ollama/Ollama.php
+++ b/src/platform/src/Bridge/Ollama/Ollama.php
@@ -1,0 +1,74 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Platform\Bridge\Ollama;
+
+use Symfony\AI\Platform\Capability;
+use Symfony\AI\Platform\Model;
+
+/**
+ * @author Joshua Behrens <code@joshua-behrens.de>
+ */
+class Ollama extends Model
+{
+    public const DEEPSEEK_R_1 = 'deepseek-r1';
+    public const GEMMA_3_N = 'gemma3n';
+    public const GEMMA_3 = 'gemma3';
+    public const QWEN_3 = 'qwen3';
+    public const QWEN_2_5_VL = 'qwen2.5vl';
+    public const LLAMA_3_1 = 'llama3.1';
+    public const LLAMA_3_2 = 'llama3.2';
+    public const MISTRAL = 'mistral';
+    public const QWEN_2_5 = 'qwen2.5';
+    public const LLAMA_3 = 'llama3';
+    public const LLAVA = 'llava';
+    public const PHI_3 = 'phi3';
+    public const GEMMA_2 = 'gemma2';
+    public const QWEN_2_5_CODER = 'qwen2.5-coder';
+    public const GEMMA = 'gemma';
+    public const QWEN = 'qwen';
+    public const QWEN_2 = 'qwen2';
+    public const LLAMA_2 = 'llama2';
+
+    private const TOOL_PATTERNS = [
+        '/./' => [
+            Capability::INPUT_MESSAGES,
+            Capability::OUTPUT_TEXT,
+        ],
+        '/^llama\D*3(\D*\d+)/' => [
+            Capability::TOOL_CALLING,
+        ],
+        '/^qwen\d(\.\d)?(-coder)?$/' => [
+            Capability::TOOL_CALLING,
+        ],
+        '/^(deepseek|mistral)/' => [
+            Capability::TOOL_CALLING,
+        ],
+    ];
+
+    /**
+     * @param array<string, mixed> $options
+     */
+    public function __construct(string $name = self::LLAMA_3_2, array $options = [])
+    {
+        $capabilities = [];
+
+        foreach (self::TOOL_PATTERNS as $pattern => $possibleCapabilities) {
+            if (1 === preg_match($pattern, $name)) {
+                foreach ($possibleCapabilities as $capability) {
+                    $capabilities[] = $capability;
+                }
+            }
+        }
+
+        parent::__construct($name, $capabilities, $options);
+    }
+}
--- a/src/platform/tests/Bridge/Ollama/OllamaTest.php
+++ b/src/platform/tests/Bridge/Ollama/OllamaTest.php
@@ -1,0 +1,85 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Platform\Tests\Bridge\Ollama;
+
+use PHPUnit\Framework\Attributes\CoversClass;
+use PHPUnit\Framework\Attributes\DataProvider;
+use PHPUnit\Framework\Attributes\Small;
+use PHPUnit\Framework\TestCase;
+use Symfony\AI\Platform\Bridge\Ollama\Ollama;
+use Symfony\AI\Platform\Capability;
+
+#[CoversClass(Ollama::class)]
+#[Small]
+final class OllamaTest extends TestCase
+{
+    #[DataProvider('provideModelsWithToolCallingCapability')]
+    public function testModelsWithToolCallingCapability(string $modelName)
+    {
+        $model = new Ollama($modelName);
+
+        $this->assertTrue(
+            $model->supports(Capability::TOOL_CALLING),
+            \sprintf('Model "%s" should support tool calling capability', $modelName)
+        );
+    }
+
+    #[DataProvider('provideModelsWithoutToolCallingCapability')]
+    public function testModelsWithoutToolCallingCapability(string $modelName)
+    {
+        $model = new Ollama($modelName);
+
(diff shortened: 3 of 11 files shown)


src/platform/src/Bridge/Ollama/Contract/LlamaContract.php
chr-hertel, line 1: Is this Llama specific? I understood that Ollama tried to have a more agnostic API that works basically with all the models - would be interesting to try that with mistral, qwen or deepseek

src/platform/src/Bridge/Meta/Llama.php
chr-hertel, line 1: Usually we try to couple the capabilities of the model to the name (version).

for tool calling we could go with a "starts with" logic - or in Ollama client we could map the const names to the ollama names, e.g. `new Llama(Llama::V3_2_1B)` => `llama-3.2-1b` becomes `llama3.2:1b` when using Ollama client

src/platform/src/Bridge/Ollama/Contract/LlamaContract.php
JoshuaBehrens: I can not follow up on this. What might be llama specific about it? The issue I had with the assistant message was that the arguments were a JSON encoded string instead of just the arguments as part of the JSON. I doubt that this is model specific.

src/platform/src/Bridge/Meta/Llama.php
JoshuaBehrens: So the logic for rewriting names would be that the prefix can be used to determine the model capabilities? That does not sound like it scales well. Also rewriting names could be confusing in debugging or general usage.

A "starts with" approach based upon regex though could help out here as well. I assume though, that we need more flexibility around ollama when they approach a lot more flexible as well

JoshuaBehrens (commented):
> I general this looks pretty clean and not like your first contribution here - thanks kicking this off - great work!

ty :) likely not my last one. already looking into ollama embeddings

> And rolling out tool calling support to more bridges is def a new feature 👍

anything more you need to have written out for a zero-ver? :D

src/platform/src/Bridge/Ollama/Contract/LlamaContract.php
chr-hertel: Let's call it `OllamaContract` than :+1:

src/platform/src/Bridge/Meta/Llama.php
chr-hertel: In general users should always be able to use the base call `Model` - which is especially helpful with platforms like Ollama

src/platform/CHANGELOG.md
```diff
@@ -57,5 +57,6 @@ CHANGELOG
  * Add support for embeddings generation across multiple providers
  * Add response promises for async operations
  * Add InMemoryPlatform and InMemoryRawResult for testing Platform without external Providers calls
+ * Add Ollama tool support
```
chr-hertel:
```suggestion
 * Add tool calling support for Ollama platform
```

chr-hertel:
Please rebase on main and adopt the changed test style, switching from `#[Test]` attribute to `test` method prefix, see #214.

src/platform/src/Bridge/Ollama/Contract/LlamaContract.php
JoshuaBehrens: Should we rename `\Symfony\AI\Platform\Bridge\Ollama\LlamaModelClient` and `\Symfony\AI\Platform\Bridge\Ollama\LlamaResultConverter` as well?

src/platform/src/Bridge/Meta/Llama.php
JoshuaBehrens: We have checks like `\Symfony\AI\Platform\Bridge\Ollama\LlamaResultConverter::supports` `$model instanceof Llama` in place which does not seem like general use of `Model` is really possible

JoshuaBehrens:
I rebased, changed the test and renamed the class. Anything more :) ?

src/platform/src/Bridge/Ollama/Contract/LlamaContract.php
chr-hertel: Yup, makes a whole lot of sense, true :+1:

src/platform/src/Bridge/Meta/Llama.php
chr-hertel: also true, can you kill that Llama dependency in the `Ollama` bridge?

src/platform/src/Bridge/Meta/Llama.php
```diff
@@ -38,13 +38,17 @@ class Llama extends Model
...
      */
-    public function __construct(string $name = self::V3_1_405B_INSTRUCT, array $options = [])
+    public function __construct(string $name = self::V3_1_405B_INSTRUCT, array $options = [], bool $supportsTools = false)
     {
         $capabilities = [
             Capability::INPUT_MESSAGES,
             Capability::OUTPUT_TEXT,
         ];
 
+        if ($supportsTools) {
```
chr-hertel: what do you think about:
```suggestion
        if (str_starts_with($name, 'llama-3.1') || str_starts_with($name, 'llama-3.2') || str_starts_with($name, 'llama-3.3')) {
```
could be a regex as well, yup

chr-hertel: i guess your regex idea would be nicer :thinking:

src/platform/src/Bridge/Meta/Llama.php
JoshuaBehrens: Claude, Nova and Gemini seem to have a similar issue. I will approach it with a slightly different solution in a different pull request. I assume this class should just not be extendable if it should just act as a struct/poco. Then we can have this in a different discussion :) seems a bit ouf of place to refactor it in this feature

src/platform/src/Bridge/Ollama/Contract/LlamaContract.php
JoshuaBehrens: Done

src/platform/src/Bridge/Meta/Llama.php
JoshuaBehrens: It feels like we do not want to mix Llama with Ollama. When I start to use model names, that are common sense for ollama but not for llama as a provider. I will create a new model class which I want to remove in the follow-up pull request :D

JoshuaBehrens: Have a look at the new Ollama class. I made it a separate commit for review purposes

chr-hertel:
Tested with mistral, deepseek and llama - works :+1:

JoshuaBehrens:
Great I will fix phpstan, rebase and we can get this merged. The new pull request for the model usage can follow after that :)

chr-hertel:
And that unit test for `Symfony\AI\Platform\Bridge\Ollama\Ollama` please :)

src/platform/src/Bridge/Ollama/Ollama.php
```diff
@@ -0,0 +1,72 @@
...
+    public const QWEN_2_5 = 'qwen2.5'; // TOOL
+    public const LLAMA_3 = 'llama3';
+    public const LLAVA = 'llava';
+    public const PHI_3 = 'phi3';
+    public const GEMMA_2 = 'gemma2';
+    public const QWEN_2_5_CODER = 'qwen2.5-coder'; // TOOL
+    public const GEMMA = 'gemma';
+    public const QWEN = 'qwen';
+    public const QWEN_2 = 'qwen2'; // TOOL
+    public const LLAMA_2 = 'llama2';
```
OskarStark: I think we can remove the `// TOOL` comments, once we have a unit test for the regex, using a data provider

JoshuaBehrens:
Sorry for the repetetive pushes. fabpot had some remarks I was not aware of. Especially this is confusing in the config:
grafik
grafik
At one part the suggestion is to have a FQCN on global functions and in another place they don't.

Apart from that I should not have any open tasks on this pull request

OskarStark:
> At one part the suggestion is to have a FQCN on global functions and in another place they don't.

We only use the one which have an optimized usage afterwards, but yes, sometimes its hard to guess :-)

src/platform/tests/Bridge/Ollama/OllamaTest.php
```diff
@@ -0,0 +1,88 @@
...
+use Symfony\AI\Platform\Capability;
+
+#[CoversClass(Ollama::class)]
+#[Small]
+final class OllamaTest extends TestCase
+{
+    /**
+     * @dataProvider provideModelsWithToolCallingCapability
+     */
+    public function testModelsWithToolCallingCapability(string $modelName)
```
chr-hertel: Let's use the attributes please
```suggestion
     #[DataProvider('provideModelsWithToolCallingCapability')
    public function testModelsWithToolCallingCapability(string $modelName)
```

src/platform/tests/Bridge/Ollama/OllamaTest.php
```diff
@@ -0,0 +1,88 @@
...
+
+        $this->assertTrue(
+            $model->supports(Capability::TOOL_CALLING),
+            \sprintf('Model "%s" should support tool calling capability', $modelName)
+        );
+    }
+
+    /**
+     * @dataProvider provideModelsWithoutToolCallingCapability
+     */
```
chr-hertel:
```suggestion
    #[DataProvider('provideModelsWithoutToolCallingCapability')
```

chr-hertel:
Thank you @JoshuaBehrens.

JoshuaBehrens:
Thank you for the good exchange and merge :) next pull requests are in the making

PR #212, opened 2025-07-27
[Demo] Remove --strict to composer validate (using @dev for now on deps)

fabpot:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Docs?         | no 
| Issues        | n/a
| License       | MIT

As we're using unbounded deps for now (`@dev`), we cannot use `--strict` on `composer validate`.

### DIFF
--- a/demo/composer.json
+++ b/demo/composer.json
@@ -103,16 +103,6 @@
             "cache:clear": "symfony-cmd",
             "assets:install %PUBLIC_DIR%": "symfony-cmd",
             "importmap:install": "symfony-cmd"
-        },
-        "pipeline": [
-            "composer validate --strict",
-            "php -l src/**/*.php tests/**/*.php",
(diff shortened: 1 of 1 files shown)


fabpot:
I'm not even sure this `pipeline` entry in `composer.json` is useful. AFAICS, it's not used anywhere and the `php-cs-fixer` line doesn't work either as there is no directory and no PHPCSFixer configuration file. As it's not used in the Github action workflows, maybe we can just remove it? @chr-hertel

chr-hertel:
That's mostly a left over from the demo being a standalone repository - missed to remove it, true

fabpot:
ok, removed.

chr-hertel (approved):

chr-hertel:
Thank you @fabpot.

PR #216, opened 2025-07-27
Simplify choice handling with multiple results

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | 
| License       | MIT

The idea of choices is that a completion result could have multiple results when option `n` larger than `1`. Until now the `Choice` class was rather simple and just assumed text or tool - and therefore in theory other choices where impossible. not sure anyone ever had an issue with this, but this now is more correct with encapsulating multiple results in the `ChoiceResult`. Makes the code even easier and more correct.

Pulled out of #136 to reduce the noise.


src/agent/tests/StructuredOutput/AgentProcessorTest.php
chr-hertel, line 1: this diff might be a good argument to get rid off the covers/uses attributes maybe

src/agent/phpstan.dist.neon
```diff
@@ -15,3 +15,6 @@ parameters:
         -
             identifier: 'symfonyAi.forbidNativeException'
             path: tests/*
+        -
```
OskarStark: ist this already in the main branch?

src/platform/phpstan.dist.neon
```diff
@@ -21,3 +21,6 @@ parameters:
         -
             identifier: missingType.iterableValue
             path: tests/*
+        -
+            message: "#^Method .*::test.*\\(\\) has no return type specified\\.$#"
```
OskarStark: ist this already in the main branch?

OskarStark (approved):

PR #214, opened 2025-07-27
Replace #[Test] attribute by test prefix

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | 
| License       | MIT

Replacing #196 

| Component | `main` | `test-prefix` | |
| ---------- | ------ | -------------- | - |
| Agent |105 tests, 287 assertions | 105 tests, 287 assertions | :heavy_check_mark: |
| AI Bundle | 14 tests, 34 assertions | 14 tests, 34 assertions | :heavy_check_mark: |
| MCP Bundle | no tests | still no tests | :disappointed: |
| MCP SDK | 23 tests, 57 assertions | 23 tests, 57 assertions | :heavy_check_mark: |
| Platform | 381 tests, 801 assertions | 381 tests, 801 assertions | :heavy_check_mark: |
| Store | 90 tests, 411 assertions | 90 tests, 411 assertions | :heavy_check_mark: |

PR #223, opened 2025-07-29
Remove `void` return type from tests

OskarStark:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | --
| License       | MIT

To make Fabbot (and @nicolas-grekas) happy 😃

### DIFF
--- a/demo/tests/Blog/LoaderTest.php
+++ b/demo/tests/Blog/LoaderTest.php
@@ -14,7 +14,6 @@ namespace App\Tests\Blog;
 use App\Blog\FeedLoader;
 use App\Blog\Post;
 use PHPUnit\Framework\Attributes\CoversClass;
-use PHPUnit\Framework\Attributes\Test;
 use PHPUnit\Framework\Attributes\UsesClass;
 use PHPUnit\Framework\TestCase;
 use Symfony\Component\HttpClient\MockHttpClient;
@@ -24,8 +23,7 @@ use Symfony\Component\HttpClient\Response\MockResponse;
 #[UsesClass(Post::class)]
 final class LoaderTest extends TestCase
 {
-    #[Test]
-    public function load(): void
+    public function testLoad()
     {
         $response = MockResponse::fromFile(__DIR__.'/fixtures/blog.rss');
         $client = new MockHttpClient($response);
--- a/demo/tests/Blog/PostTest.php
+++ b/demo/tests/Blog/PostTest.php
@@ -13,15 +13,13 @@ namespace App\Tests\Blog;
 
 use App\Blog\Post;
 use PHPUnit\Framework\Attributes\CoversClass;
-use PHPUnit\Framework\Attributes\Test;
 use PHPUnit\Framework\TestCase;
 use Symfony\Component\Uid\Uuid;
 
 #[CoversClass(Post::class)]
 final class PostTest extends TestCase
 {
-    #[Test]
-    public function postToString(): void
+    public function testPostToString()
     {
         $post = new Post(
             Uuid::v4(),
@@ -43,8 +41,7 @@ final class PostTest extends TestCase
         $this->assertSame($expected, $post->toString());
     }
 
-    #[Test]
-    public function postToArray(): void
+    public function testPostToArray()
     {
         $id = Uuid::v4();
         $post = new Post(
--- a/demo/tests/SmokeTest.php
+++ b/demo/tests/SmokeTest.php
@@ -13,7 +13,6 @@ namespace App\Tests;
 
 use PHPUnit\Framework\Attributes\CoversNothing;
 use PHPUnit\Framework\Attributes\DataProvider;
-use PHPUnit\Framework\Attributes\Test;
 use Symfony\Bundle\FrameworkBundle\Test\WebTestCase;
 use Symfony\UX\LiveComponent\Test\InteractsWithLiveComponents;
 
@@ -22,8 +21,7 @@ final class SmokeTest extends WebTestCase
 {
     use InteractsWithLiveComponents;
 
-    #[Test]
-    public function index(): void
+    public function testIndex()
     {
         $client = static::createClient();
         $client->request('GET', '/');
@@ -33,9 +31,8 @@ final class SmokeTest extends WebTestCase
         self::assertSelectorCount(5, '.card');
     }
 
-    #[Test]
     #[DataProvider('provideChats')]
-    public function chats(string $path, string $expectedHeadline): void
+    public function testChats(string $path, string $expectedHeadline)
     {
         $client = static::createClient();
         $client->request('GET', $path);
--- a/src/agent/tests/Toolbox/MetadataFactory/ChainFactoryTest.php
+++ b/src/agent/tests/Toolbox/MetadataFactory/ChainFactoryTest.php
@@ -45,7 +45,7 @@ final class ChainFactoryTest extends TestCase
         $this->factory = new ChainFactory([$factory1, $factory2]);
     }
 
-    public function testTestGetMetadataNotExistingClass(): void
+    public function testTestGetMetadataNotExistingClass()
     {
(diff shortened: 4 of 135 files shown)


chr-hertel:
I really feel like this is only a symfony/symfony concern 😞

OskarStark:
But the used Fabbot action is now shared across all repos, or shall we make this configurable @nicolas-grekas ?

See
* #221

OskarStark:
@nicolas-grekas any clue regarding?
CleanShot 2025-07-29 at 12 43 31@2x

nicolas-grekas:
This is a general Symfony policy. Adding return types to test cases is just visual debt (TM)

chr-hertel:
ah okay, i thought it's more about avoiding git conflict between different versions - which does not apply here.

is it that you opt in whenever there is a more meaningful return type for tests depending on each other or just avoid that in general? not that i see depending test scenarios a lot, but that's the difference to `void` tests i guess

anyhow no hard feelings - just curious :D

derrabus:
I would not enforce this rule in a brand new code repository, tbh. I understand that this rule has historic reasons in symfony/symfony, but those don't apply here. Let's use types everywhere. The less exceptions to the rule, the better.

OskarStark:
I agree ☝🏻

demo/tests/Blog/LoaderTest.php
```diff
@@ -25,7 +25,7 @@
 final class LoaderTest extends TestCase
 {
     #[Test]
-    public function load(): void
+    public function load()
```
nicolas-grekas:
```suggestion
    public function testLoad()
```

demo/tests/Blog/PostTest.php
```diff
@@ -21,7 +21,7 @@
 final class PostTest extends TestCase
 {
     #[Test]
-    public function postToString(): void
+    public function postToString()
```
nicolas-grekas:
```suggestion
    public function testPostToString()
```

demo/tests/Blog/PostTest.php
```diff
@@ -44,7 +44,7 @@ public function postToString(): void
     }
 
     #[Test]
-    public function postToArray(): void
+    public function postToArray()
```
nicolas-grekas:
```suggestion
    public function testPostToArray()
```

demo/tests/SmokeTest.php
```diff
@@ -23,7 +23,7 @@ final class SmokeTest extends WebTestCase
     use InteractsWithLiveComponents;
 
     #[Test]
-    public function index(): void
+    public function index()
```
nicolas-grekas:
```suggestion
    public function testIndex()
```

demo/tests/SmokeTest.php
```diff
@@ -35,7 +35,7 @@ public function index(): void
 
     #[Test]
     #[DataProvider('provideChats')]
-    public function chats(string $path, string $expectedHeadline): void
+    public function chats(string $path, string $expectedHeadline)
```
nicolas-grekas:
```suggestion
    #[DataProvider('provideChats')]
    public function testChats(string $path, string $expectedHeadline)
```

src/agent/tests/Toolbox/MetadataFactory/ChainFactoryTest.php
```diff
@@ -35,7 +35,7 @@ final class ChainFactoryTest extends TestCase
 {
     private ChainFactory $factory;
 
-    protected function setUp(): void
+    protected function setUp()
```
nicolas-grekas: please fix these, they don't need patching

```suggestion
    protected function setUp(): void
```

src/agent/tests/Toolbox/MetadataFactory/ReflectionFactoryTest.php
```diff
@@ -139,7 +139,7 @@ className: ToolMultiple::class,
         );
     }
 
-    private function assertToolConfiguration(Tool $metadata, string $className, string $name, string $description, string $method, array $parameters): void
+    private function assertToolConfiguration(Tool $metadata, string $className, string $name, string $description, string $method, array $parameters)
```
nicolas-grekas:
```suggestion
    private function assertToolConfiguration(Tool $metadata, string $className, string $name, string $description, string $method, array $parameters): void
```

nicolas-grekas (approved):
The search-n-replace has been a bit too broad ;)

This still makes sense to me. It's not about historical practices, it's about thinking about what one is doing. Each method needs careful thinking at its boundaries, to provide a clear API.
Test case methods are not part of any API. They're entry points. Specifying a return type for them is just more syntax soup IMHO.
A rule like "all methods must have a return type" applies in a context. Ignoring this context (designing APIs) and making the rule a generic one prevents people from thinking about what they're doing.
I wouldn't call the "no-void" policy an exception to the rule, but more like a domain where the rule doesn't apply.

nicolas-grekas:
>   The "/home/runner/work/ai/ai/b/examples" directory does not exist.  

That's because the `.php-cs-fixer.dist.php` needs to account for missing directories, since we check only patched files and the others are not checked-out. Using glob does the job:

```diff
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -43,7 +43,7 @@ return (new PhpCsFixer\Config())
     ->setRiskyAllowed(true)
     ->setFinder(
         (new PhpCsFixer\Finder())
-            ->in([__DIR__.'/demo', __DIR__.'/examples', __DIR__.'/fixtures', __DIR__.'/src'])
+            ->in(__DIR__.'/{demo,examples,fixtures,src}')
             ->append([__FILE__])
```

nicolas-grekas:
For phpstan to be happy, I'd suggest going with this, which matches what fabbot does (we could further restrict to `**/tests/**.php` files):

```diff
 	ignoreErrors:
+		-
+			message: "#^Method .*::test.*\\(\\) has no return type specified\\.$#"
+
```

OskarStark:
Will tackle that topic tomorrow

demo/tests/Blog/LoaderTest.php
chr-hertel: ah, missed those in #214 somehow, thanks

chr-hertel:
Yea, i guess it's rather subjective how you read these methods, but I get your point - and I also don't want to be bothered to define a complex array shape for a `normalize(...): array` method by phpstan if only framework code consumes it - fair.
I guess my hard limit is when it comes to `#[Depends]` scenarios, but i feel like those are rare. not sure why it is different for `setUp()` tho. anyhow, i for sure can live with that - especially if there's automation to remind me about it

chr-hertel (approved):

demo/tests/SmokeTest.php
```diff
@@ -33,9 +32,8 @@ public function index()
         self::assertSelectorCount(5, '.card');
     }
 
-    #[Test]
     #[DataProvider('provideChats')]
-    public function chats(string $path, string $expectedHeadline)
+    public function testsChats(string $path, string $expectedHeadline)
```
chr-hertel:
```suggestion
    public function testChats(string $path, string $expectedHeadline)
```

OskarStark:
@nicolas-grekas if I get it right, we don't have a convenient link to fix CS for the contributors anymore, this is a step back IMHO

nicolas-grekas:
You're right, that's not something supported by GHA.
Instead, I made generated patch copy/pasteable.

nicolas-grekas:
See #226 for exception messages

OskarStark:
> See https://github.com/symfony/ai/pull/226 for exception messages

Lets go with yours first, and then rebase mine on top

nicolas-grekas:
Thank you @OskarStark.

nicolas-grekas:
Thank you @OskarStark.

PR #225, opened 2025-07-29
[MCP SDK] `ToolCallHandler` now returns `structuredContent` if applicable

xavierleune:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no 
| Issues        | Fix #... 
| License       | MIT

As [per documentation](https://modelcontextprotocol.io/specification/2025-06-18/server/tools#structured-content) we should include any structured content in a dedicated field
> Structured content is returned as a JSON object in the structuredContent field of a result.
For backwards compatibility, a tool that returns structured content SHOULD also return the serialized JSON in a TextContent block.

### DIFF
--- a/src/mcp-sdk/src/Capability/Tool/ToolCallResult.php
+++ b/src/mcp-sdk/src/Capability/Tool/ToolCallResult.php
@@ -14,12 +14,12 @@ namespace Symfony\AI\McpSdk\Capability\Tool;
 final readonly class ToolCallResult
 {
     public function __construct(
-        public string $result,
+        public mixed $result,
         /**
          * @var "text"|"image"|"audio"|"resource"|non-empty-string
          */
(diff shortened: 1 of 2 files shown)


JoshuaBehrens (approved):
I have a similar patch in local testing :)

OskarStark:
Please rebase, thanks

xavierleune:
@OskarStark done 👍

chr-hertel (changes requested):
Two things, I think this is good timing to add a test for checking the two options for type text.

Isn't it that with the current approach we end up with something like `"text": "\"foobar\""` for simple, unstructured text values?

OskarStark:
Open to finish this PR @xavierleune ?

chr-hertel:

PR #226, opened 2025-07-30
Fix exception messages

nicolas-grekas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | -
| License       | MIT

As fabbot would suggest (+ minor tweaks)


src/platform/src/Bridge/Mistral/Llm/ResultConverter.php
```diff
@@ -51,13 +51,13 @@ public function convert(RawResultInterface|RawHttpResult $result, array $options
         }
 
         if (200 !== $code = $httpResponse->getStatusCode()) {
-            throw new RuntimeException(\sprintf('Unexpected response code %d: %s', $code, $httpResponse->getContent(false)));
+            throw new RuntimeException(\sprintf('Unexpected response code %d: "%s"', $code, $httpResponse->getContent(false)));
```
xabbuh:
```suggestion
            throw new RuntimeException(\sprintf('Unexpected response code %d: ', $code).$httpResponse->getContent(false));
```

like in the `ResultConverter` for embeddings

PR #233, opened 2025-07-30
[Store] remove `minScore` parameter from `VectorStoreInterface::query`

### LINKED ISSUE #195: [Store][PG] Score is inverted?

DZunke:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #195
| License       | MIT

As discussed within #197 and #195 i have removed the `minScore` argument from the `VectorStoreInterface` as this is semantically not correct for any store. Most stores has never implemented this filter mechanism. It originated from the MongoDB store where it is correct with the internal scoring system but for stores like MariaDB and Postgres the distance calculations are more correct with a `maxScore` option as their results are sorted from `0` as exact match to larger float values for the distance. 

I have changed those two stores and left the MongoDB store with the minScore implementation. 

This is a break in the implementation as the interface has changed.

### DIFF
--- a/src/store/doc/index.rst
+++ b/src/store/doc/index.rst
@@ -85,7 +85,7 @@ This leads to a store implementing two methods::
             // Implementation to add a document to the store
         }
 
-        public function query(Vector $vector, array $options = [], ?float $minScore = null): array
+        public function query(Vector $vector, array $options = []): array
         {
             // Implementation to query the store for documents
             return [];
--- a/src/store/src/Bridge/Azure/SearchStore.php
+++ b/src/store/src/Bridge/Azure/SearchStore.php
@@ -44,7 +44,7 @@ final readonly class SearchStore implements VectorStoreInterface
         ]);
     }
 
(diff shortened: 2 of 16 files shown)


OskarStark (approved):

lyrixx (approved):
Seems good 👍🏼 Thanks

chr-hertel:
Thank you @DZunke.

PR #237, opened 2025-07-31
Simply code-quality matrix

nicolas-grekas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | -
| License       | MIT

For the phpstan job, we don't need any matrix. We only care about what phpstan can report with latest deps.
For the phpunit ones, we should always test with dev dependencies, like we're doing on symfony/symfony.

https://github.com/symfony/ai/pull/237/files?w=1

### DIFF
--- a/.github/workflows/code-quality.yaml
+++ b/.github/workflows/code-quality.yaml
@@ -21,50 +21,38 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
-                php-version: [ '8.2', '8.3', '8.4']
-                dependency-version: ['']
-                symfony-version: ['']
-                minimum-stability: ['stable']
-                include:
-                    # lowest deps
-                    -   php-version: '8.2'
-                        dependency-version: 'lowest'
-                    # LTS version of Symfony
-                    -   php-version: '8.2'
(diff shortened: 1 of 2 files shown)


.github/workflows/code-quality.yaml
```diff
@@ -21,50 +21,38 @@ jobs:
...
 
-            -   name: Setup PHP
-                uses: shivammathur/setup-php@v2
-                with:
-                    php-version: '8.2'
-                    tools: flex
+            - name: Setup PHP
+              uses: shivammathur/setup-php@v2
+              with:
+                  php-version: '8.2'
```
Jibbarth:
```suggestion
                  php-version: '${{ matrix.php-version }}'
```
? 
Or maybe we can get rid of the following part ?

```
matrix:
    php-version: [ '8.4' ]
```

nicolas-grekas: indeed, fixed!
we can keep the matrix I think

chr-hertel:
Thank you @nicolas-grekas.

PR #239, opened 2025-07-31
[Chat] Meilisearch message store

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no 
| Issues        | None
| License       | MIT

Hi 👋🏻 

Here's a small POC to store message bags outside of memory (for long term agents), the implementation is built around MS but the main interfaces are defined (Redis, Doctrine, etc, any bridge can use it) for any bridge (and heavily inspired by the `Store` package).

Let me know if the idea is interesting, if not, feel free to close the PR 🙂


OskarStark:
Wouldn't it be more related to Chat and the MessageStore?

Guikingone:
Ah yes, I wasn't sure about the interface to use, I'll check the store one.

OskarStark (approved):
PR title and body would need an update please

Guikingone:
Yes, I need to finish the test in the same time, I'll push the updated version once it's done 🙂

examples/misc/persistent-chat_meilisearch.php
chr-hertel, line 1: let's stick to dash instead of underscore please:
filename should be `examples/misc/persistent-chat-meilisearch.php`

chr-hertel: and it makes sense to rename the other one now, maybe `examples/misc/persistent-chat-memory.php`

examples/misc/persistent-chat_meilisearch.php
```diff
@@ -0,0 +1,44 @@
...
+$store->initialize();
+
+$chat = new Chat($agent, $store);
+
+$messages = new MessageBag(
+    Message::forSystem('You are a helpful assistant. You only answer with short sentences.'),
+);
+
+$chat->initiate($messages);
+$chat->submit(Message::ofUser('My name is Christopher.'));
```
chr-hertel, line 35:
```suggestion
$chat->submit(Message::ofUser('My name is Guillaume.'));
```

:laughing:

chr-hertel: (feel free to omit this one - it's friday and my brain is already half way gone)

Guikingone:
@chr-hertel I agree, the `Store` component should contains everything related to storing messages, vectors and more, this way, we can easily "update" the API without forcing the release of `Agent`, plus, we can easily share the API thanks to interfaces.

Maybe we can move the classes/interfaces after this PR to prevent any bloating of it?

chr-hertel:
> Maybe we can move the classes/interfaces after this PR to prevent any bloating of it?

Agreed :+1:

OskarStark:
What about a dedicated MessageStore component? or too much?

Guikingone:
@OskarStark Well, here's my humble opinion (I don't know how this package was initially designed and what are the "rules" that it must follow to comply with Symfony approach): 

The `Store` package is enough, its description along with the documentation must be updated to inform that it "store" more than just vectors and is not "only designed" for RAG, this approach would allow to keep the current packages and add new features (like graph databases and more) without compromising the overall developer experience (too much packages tend to complicate the developer experience, you're using X namespaces, you don't know which one do what, etc)

Maybe we could find a new "way" to split the `Store` package (new sub namespaces and more, let's say `Bridge` then two namespaces `VectorStore` then `MessageStore`, simple examples to gave the idea) to keep it coherent and "open" for future modifications but IMO, it's enough for the "storage" part of the AI initiative.

chr-hertel:
One thing I wonder is about having multiple conversations happening in parallel. Let's say per user. How would that work with this implementation? separate index?

Guikingone:
Actually, yes, multiple index.

Another solution might be to use a "session" per agent, the session is initiated by the agent (constructor call, maybe an UUID?) then sent to the message store to "lock" messages per agent.

I don't see a big technical bottleneck but IMO, this refactoring must be brought via another PR (I can work on it if you want).

examples/misc/persistent-chat-meilisearch.php
```diff
@@ -0,0 +1,44 @@
...
+$agent = new Agent($platform, $llm, logger: logger());
+$store = new MessageStore(
+    http_client(),
+    env('MEILISEARCH_HOST'),
+    env('MEILISEARCH_API_KEY'),
+    'chat',
+);
+$store->initialize();
+
+$chat = new Chat($agent, $store);
```
OskarStark: What about letting the Chat call initialize internally if it is some kind of initalizeable store?

Just thinking out loud

src/chat/src/Bridge/Meilisearch/MessageStore.php
```diff
@@ -0,0 +1,177 @@
...
+/**
+ * @author Guillaume Loulier <personal@guillaumeloulier.fr>
+ */
+final readonly class MessageStore implements InitializableMessageStoreInterface, MessageStoreInterface
+{
+    public function __construct(
+        private HttpClientInterface $httpClient,
+        private string $endpointUrl,
+        #[\SensitiveParameter] private string $apiKey,
+        private string $indexName,
```
OskarStark: Does it make sense to provide a default here?

src/chat/src/Bridge/Meilisearch/MessageStore.php
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    /**
+     * @param array<string, mixed>|list<array<string, mixed>> $payload
+     *
+     * @return array<string, mixed>
+     */
+    private function request(string $method, string $endpoint, array $payload = []): array
+    {
+        $url = \sprintf('%s/%s', $this->endpointUrl, $endpoint);
```
OskarStark: Can be inlined

src/chat/tests/Bridge/Meilisearch/MessageStoreTest.php
```diff
@@ -0,0 +1,310 @@
...
+                'http_code' => 400,
+            ]),
+        ], 'http://localhost:7700');
+
+        $store = new MessageStore(
+            $httpClient,
+            'http://localhost:7700',
+            'test',
+            'test',
+        );
```
OskarStark: One line please

examples/misc/persistent-chat-meilisearch.php
Guikingone: I'm thinking the same way, the actual API is not built for initializeable store, that's one of the issue I'm facing with #254

chr-hertel:
Hi @Guikingone, i guess this one got lost - still want to wrap this up?

Guikingone:
Hi @chr-hertel, yes, of course, shouldn't we wait until #675 is merged? This way we can directly move the class into the correct directory 🤔

chr-hertel:
Needs a rebase now

OskarStark:
Needs a rebase again

examples/misc/persistent-chat-meilisearch.php
```diff
@@ -9,22 +9,28 @@
...
 use Symfony\AI\Platform\Message\Message;
 use Symfony\AI\Platform\Message\MessageBag;
-use Symfony\Component\Cache\Adapter\ArrayAdapter;
 
 require_once dirname(__DIR__).'/bootstrap.php';
 
 $platform = PlatformFactory::create(env('OPENAI_API_KEY'), http_client());
+$llm = new Gpt(Gpt::GPT_4O_MINI);
+
+$agent = new Agent($platform, $llm, logger: logger());
```
OskarStark:
```suggestion

$agent = new Agent($platform, 'gpt-40-mini', logger: logger());
```

examples/misc/persistent-chat-meilisearch.php
```diff
@@ -9,22 +9,28 @@
...
+$store = new MessageStore(
+    http_client(),
+    env('MEILISEARCH_HOST'),
+    env('MEILISEARCH_API_KEY'),
+    'chat',
```
OskarStark:
```suggestion
    indexName: 'chat',
```

examples/misc/persistent-chat-meilisearch.php
OskarStark:
```suggestion
echo $message->getContent().\PHP_EOL;
```
? I think we made the property private

Guikingone:
@OskarStark Yes, I'm facing a small issue during tests, might need a day or two to fix it and ensure that everything is working as expecting, I'll take a look at your suggestions ASAP.

OskarStark:
No worries

src/chat/src/Bridge/Meilisearch/MessageStore.php
```diff
@@ -0,0 +1,180 @@
...
+/**
+ * @author Guillaume Loulier <personal@guillaumeloulier.fr>
+ */
+final readonly class MessageStore implements ManagedStoreInterface, MessageStoreInterface
+{
+    public function __construct(
+        private HttpClientInterface $httpClient,
+        private string $endpointUrl,
+        #[\SensitiveParameter] private string $apiKey,
+        private string $indexName,
```
OskarStark: Shall we provide a default here?

docs/components/chat.rst
```diff
@@ -31,6 +31,20 @@ with a ``Symfony\AI\Agent\AgentInterface`` and a ``Symfony\AI\Chat\MessageStoreI
 
     $chat->submit(Message::ofUser('Hello'));
 
+You can find more advanced usage in combination with an Agent using the store for long-term context:
+
+* `Long-term context with Cache`_
+* `Long-term context with HttpFoundation session`_
+* `Long-term context with InMemory`_
```
chr-hertel: i feel like "long-term" is not really accurate when writing about InMemory (basically done with the process terminating) and session (also not thaat persistent).

Guikingone: Agree, I'll fix it 👍🏻

examples/chat/persistent-chat-meilisearch.php
chr-hertel, line 1: doesn't work for me:
image

Guikingone: Strange, I'll take a look 🤔

Guikingone: Found the issue, fixed 🙂

chr-hertel (approved):
Works - let's get this merged 
image

chr-hertel:
Thank you @Guikingone.

Issue #158, opened 2025-07-18
[Agent] SimilaritySearch in Azure Search has no `id` in hits


apfelbox:
Hi,

so we are currently trying out the new bundles and they work really great so far! Nice job!

However, we found an issue and are unsure on how to proceed.

As context: we use Azure-based ChatGPT with Azure Search-based embeddings (RAG).

So we fetch the chunks for the request from Azure Search, however it currently breaks: the structure of the `value`s in the result look like this:

```php
[
	"@search.score" => "...",
	"chunk_id" => "...",
	"parent_id" => "...",
	"chunk" => "...",
	"title" => null,
	"text_vector" => "...",
]
```

However, this currently breaks, as the SearchStore tries to access `id` in this array: 
https://github.com/symfony/ai/blob/main/src/store/src/Bridge/Azure/SearchStore.php#L92

And we don't have an `id`.

So my question: is that something we need to configure on our side or is the assumption in the code currently wrong, that there always is an id?

Comments:
- Neither `chunk_id` nor `parent_id` are in any way Uuid-compatible
- When we change the line to `Uuid::v4()`, it appears to work

Thanks for any pointers and for the great bundles

chr-hertel:
Hey, it's been a while since I played around with that store. How's the data getting into that store? with built-in indexer of Azure?

I think we have two options here:
1. force users to have an ID as UUID in their schema (doesn't sound ideal)
2. make ID column configurable, like mapping `chunk_id` to it + open up the type declaration
3. drop the ID on our end completely and rely on users handling that themselves with the metadata

@OskarStark were you using Azure AI Search as well? what's your take here

OskarStark:
My take was, that I entered the data via our store, I guess that's not the case here, am I right @apfelbox ?

In fact this is something we already discussed (with no result) with @alcaeus: "How can we retrieve data with a different indexed format aka. how can we make project objects indexable for our stores."

@apfelbox do you have an idea in mind how to fix this? We could change the contract, we could make it configurable via options etc.

apfelbox:
I am actually not sure on how we index the data right now (as we are currently experimenting with everything quite a lot), however I am pretty sure that we just put a set of markdown file into an Azure Storage bucket and push it through the indexer.
The indexer then chunks it, calculates necessary overlap, etc.. and that's what we get in the search endpoint.

[...]

Issue #18, opened 2025-06-07
[Store] Support user land objects for embedding instead of TextDocument


chr-hertel:
Typically applications already have domain objects at hand and it would increase DX to simply use that for Embedder instead of our document classes, see https://github.com/symfony/ai/blob/main/examples/store/mongodb-similarity-search.php

A solution should be able to also cover Doctrine ORM and ODM.

paulinevos:
Would like to work on this

PR #244, opened 2025-08-01
[Store] Add support for ClickHouse

lyrixx:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        |
| License       | MIT

We already have ClickHouse in our infrastructure, So I'm benchmarkin it.
And, let's share my code 🎉

As you can see, I let some extension points (for the metadata). AFAIK, it's not possible to
store **structured / indexed** JSON in CH. We have to store it as `String` instead. 
We could instead use better typed structure, but Symfony cannot know the type, obviously.
So with string, if we want to filter the results, we need to use `[JSONExtractString](https://clickhouse.com/docs/sql-reference/functions/json-functions#jsonextractstring)`.
And It's slow!

So with the current extensions points, I could write the following code:

```php
class ClickHouseStore extends SymfonyClickHouseStore
{
    public function initialize(array $options = []): void
    {
        $sql = <<<'SQL'
            CREATE TABLE IF NOT EXISTS {{ table }} (
                id UUID,
                metadata String,
                embedding Array(Float32),
                crawlId UUID,
            ) ENGINE = MergeTree()
            ORDER BY id
        SQL;

        $this->execute('POST', $sql);
    }

    protected function formatVectorDocument(VectorDocument $document): array
    {
        $formatted = parent::formatVectorDocument($document);

        $formatted['crawlId'] = $document->metadata['crawlId'];

        return $formatted;
    }
}
```

And then:

```
$documents = $clickHouseStore->query(
    $vector,
    [
        'where' => 'crawlId = {crawlId:String} AND id != {currentId:UUID} AND score < 0.1',
        'params' => [
            'crawlId' => $crawlPoId,
            'currentId' => $row['id'],
        ],
    ],
);
```

---

If the PR could be accepted, I'll write tests


chr-hertel:
the need for extension is not really ClickHouse specific, is it?
I think we need an extension point somewhere here anyways. a low level mapping/hydration thingy.
with #18 we have a similar topic i'd say - and with #158 we also have an issue of making too many assumption. totally open for ideas. :)

lyrixx:
> the need for extension is not really ClickHouse specific, is it?

Indeed 👍🏼 The current StoreInterface is a bit limited, but I eventually managed to do what I need to POC something.

> I think we need an extension point somewhere here anyways. a low level mapping/hydration thingy.

Yes! 

---

~~Should I work on tests?~~ **edit**: done ✅

src/store/src/Bridge/ClickHouse/Store.php
```diff
@@ -0,0 +1,182 @@
...
+                'default_format' => 'JSON',
+            ],
+        ];
+
+        foreach ($params as $key => $value) {
+            $options['query']['param_'.$key] = $value;
+        }
+
+        return $this
+            ->httpClient
```
OskarStark:
```suggestion
        return $this->httpClient
```

src/store/tests/Bridge/ClickHouse/StoreTest.php
```diff
@@ -0,0 +1,289 @@
...
+            }))
+            ->willReturn($response);
+
+        $store->add($document);
+    }
+
+    public function testAddMultipleDocuments()
+    {
+        $httpClient = $this->createMock(HttpClientInterface::class);
+        $response = $this->createMock(ResponseInterface::class);
```
OskarStark: Let's use MockHttpClient?

OskarStark:
Please dd the bundle configuration for this new store, thanks

lyrixx:
@OskarStark Thanks for the review. I addressed your comments

OskarStark:
Lets merge after a rebase, thanks!

src/ai-bundle/src/AiBundle.php
```diff
@@ -481,6 +483,32 @@ private function processStoreConfig(string $type, array $stores, ContainerBuilde
...
+                    $httpClient = new Reference($store['http_client']);
+                } else {
+                    $httpClient = new Definition(HttpClientInterface::class);
+                    $httpClient
+                        ->setFactory([HttpClient::class, 'createForBaseUri'])
+                        ->setArguments([$store['dsn']])
+                    ;
+                }
+
+                $definition = new Definition(ChromaDbStore::class);
```
OskarStark: ChromaDb?

lyrixx: Good catch. Thanks

OskarStark (approved):

OskarStark:
Thank you @lyrixx.

PR #248, opened 2025-08-02
[Demo] Render markdown in the stream example

### LINKED ISSUE #201: [DEMO]  Turbo stream bot demo and markdown to html
Hi,

In the turbo stream bot demo, the rendering is really impressive.

The display is done piece by piece.

Is it possible to keep the display piece by piece and interpret the markdown in html ?

Thanks.

valtzu:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | Fix #201
| License       | MIT

Previously markdown was only converted to HTML on full page reload, now we render the whole partial message server side while streaming. This is a lot less efficient than only sending the new parts of the stream, but much easier to implement _(still no custom JS needed)_ – and I don't think we should make the demos too complex.

### DIFF
--- a/demo/src/Stream/Chat.php
+++ b/demo/src/Stream/Chat.php
@@ -57,8 +57,7 @@ final class Chat
 
         $response = '';
         foreach ($stream as $chunk) {
-            yield $chunk;
-            $response .= $chunk;
+            yield $response .= $chunk;
         }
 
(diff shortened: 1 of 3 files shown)


chr-hertel (approved):

chr-hertel:
Thank you @valtzu.

PR #252, opened 2025-08-03
[Platform] Enhance in memory platform to support all `ResultInterface` types

RamyHakam:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Docs?         | yes 
| Issues        | none 
| License       | MIT

Enhances the existing `InMemoryPlatform` to support all' ResultInterface' types, making it easier to test all result types.
Previously limited to text responses, it now supports vectors, images, binary data, choices, and any custom result types.


src/platform/src/InMemoryPlatform.php
```diff
@@ -26,17 +27,31 @@ class InMemoryPlatform implements PlatformInterface
...
+                    options: $options
+                );
+            }
+
+            $resultText = (string) $result;
+        } else {
+            $resultText = $this->mockResult;
+        }
 
         $textResult = new TextResult($resultText);
```
chr-hertel: Make sense, yes, but would like to see the code less complex please.
For example this could also work, right?

```php
    public function invoke(Model $model, array|string|object $input, array $options = []): ResultPromise
    {
        $result = is_string($this->mockResult) ? $this->mockResult : ($this->mockResult)($model, $input, $options);

        if ($result instanceof ResultInterface) {
            return $this->createPromise($result, $options);
        }

        return $this->createPromise(new TextResult($result), $options);
    }

    private function createPromise(ResultInterface $result, array $options): ResultPromise
    {
        $rawResult = $result->getRawResult() ?? new InMemoryRawResult(
            ['text' => $result->getContent()],
            (object) ['text' => $result->getContent()],
        );

        return new ResultPromise(static fn () => $result, $rawResult, $options);
    }
```

chr-hertel (changes requested):

src/platform/src/InMemoryPlatform.php
RamyHakam: You are right, that is simpler and less complex

chr-hertel (approved):

chr-hertel:
Thank you @RamyHakam.

PR #253, opened 2025-08-03
[AiBundle] Register fault tolerant default toolbox

valtzu:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix https://github.com/php-llm/llm-chain-bundle/issues/102
| License       | MIT

Fault tolerant toolbox was never registered if you used `tools: true` instead of defining tools explicitly.


OskarStark (approved):

OskarStark:
Good catch, thanks @valtzu.

PR #257, opened 2025-08-04
[Store] Add PSR-6 support for vector store

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | None
| License       | MIT

Hi 👋🏻 

This PR aims to introduce a `CacheStore` based on PSR-6, it's heavily inspired by the `CacheStore` from `MessageStore` layer, this store introduce a refactoring of the `InMemoryStore` to use common logic for both stores (as we can't rely on the `CacheItemPool` implementation).

Thanks for the feedback.


src/store/src/VectorStoreTrait.php
```diff
@@ -0,0 +1,139 @@
...
+namespace Symfony\AI\Store;
+
+use Symfony\AI\Platform\Vector\Vector;
+use Symfony\AI\Store\Document\VectorDocument;
+use Symfony\AI\Store\Exception\InvalidArgumentException;
+
+/**
+ * @author Guillaume Loulier <personal@guillaumeloulier.fr>
+ */
+trait VectorStoreTrait
```
chr-hertel: does that have to be a trait? at first glance it looks like this could just be a standalone service class injected to both stores, or do I miss something here?

chr-hertel:
That is really an interesting idea, since the PSR 6 cache has quite some implementations. I wonder tho if we would still recommend to use that only in testing or development. I would assume that it doesn't scale well with a lot of documents since the search needs to happen in PHP runtime with all documents loaded, right?

src/store/src/VectorStoreTrait.php
Guikingone: Well, I'm open to debate on this one, IMO, the trait is "easier" as it doesn't require any service injection, if we move the logic to a service, the problem is: How do we call it?

And the naming is another debate 😅

Guikingone:
Not sure about the performances side, IMO, that's not a big problem (we already use the Cache layer from Sf for heavy operations, that's not a big deal), If you fear that it can impact applications, we can recommend to use it only in dev/test envs, it's just a documentation concern 🤔

src/store/src/VectorStoreTrait.php
chr-hertel: i would argue the other way around - a separate class would be the easier choice as there is little to no cohesion but a functional dependency. and all the usual arguments would count - including testability.

let's go with a optional dependency like:
```php
    public function __construct(
        private readonly string $distanceCalculator = new DistanceCalculator(),
    ) {
    }
```

since the options are not part of that standalone contract, you could design the method quite explicit - even with an enum for the strategy

Guikingone: Ah, I see, yes, didn't though about the optional dependency, this could open the possibility to extend it later with a custom interface, I like this approach 👍🏻

Guikingone: Improved with a service / enum.

src/store/doc/index.rst
```diff
@@ -40,12 +40,15 @@ You can find more advanced usage in combination with an Agent using the store fo
...
 * `Similarity Search with Meilisearch (RAG)`_
 * `Similarity Search with memory storage (RAG)`_
+* `Similarity Search with PSR-6 Cache (RAG)`_
 * `Similarity Search with MongoDB (RAG)`_
 * `Similarity Search with Pinecone (RAG)`_
 * `Similarity Search with Qdrant (RAG)`_
 * `Similarity Search with SurrealDB (RAG)`_
 * `Similarity Search with Neo4j (RAG)`_
 
+PS: Both `InMemory` and `PSR-6 cache` vector store are designed to be primarily used in tests / development environments.
```
OskarStark:
```suggestion
PS: Both `InMemory` and `PSR-6 cache` vector stores are designed to be primarily used in tests / development environments.
```
?

src/store/doc/index.rst
```diff
@@ -60,6 +63,7 @@ Supported Stores
 * `Qdrant`_
 * `SurrealDB`_
 * `Neo4j`_
+* `PSR-6 Cache`_
```
OskarStark: While on it, can you please sort the stores alphabetically? Thanks

src/store/src/DistanceCalculator.php
```diff
@@ -0,0 +1,134 @@
...
+
+use Symfony\AI\Platform\Vector\Vector;
+use Symfony\AI\Store\Document\VectorDocument;
+
+/**
+ * @author Guillaume Loulier <personal@guillaumeloulier.fr>
+ */
+final readonly class DistanceCalculator
+{
+    public function __construct(private DistanceStrategyEnum $strategy = DistanceStrategyEnum::COSINE_DISTANCE)
```
OskarStark:
```suggestion
    public function __construct(
    private DistanceStrategyEnum $strategy = DistanceStrategyEnum::COSINE_DISTANCE
    )
```

chr-hertel: better even
```suggestion
    public function __construct(
        private DistanceStrategyEnum $strategy = DistanceStrategyEnum::COSINE_DISTANCE,
    )
```

src/store/src/DistanceStrategyEnum.php
```diff
@@ -0,0 +1,24 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Store;
+
+/**
+ * @author Guillaume Loulier <personal@guillaumeloulier.fr>
+ */
+enum DistanceStrategyEnum: string
```
chr-hertel: Doesn't need the suffix if you ask me :)

```suggestion
enum DistanceStrategy: string
```

src/store/src/DistanceCalculator.php
```diff
@@ -0,0 +1,134 @@
...
+
+    /**
+     * @param VectorDocument[] $documents
+     * @param array{
+     *     maxItems?: positive-int
+     * }|array{} $options If maxItems is provided, only the top N results will be returned
+     *
+     * @return VectorDocument[]
+     */
+    public function doCalculate(array $documents, Vector $vector, array $options = []): array
```
chr-hertel: can you please change the vague `$options` to a concrete set of arguments, and we can skip the `do` now :)
```suggestion
    public function calculate(array $documents, Vector $vector, int $maxItems): array
```

chr-hertel: or the $maxItems would be nullable i guess

src/store/src/VectorStoreTrait.php
chr-hertel: Looks better now, thanks 👍

src/store/doc/index.rst
```diff
@@ -41,10 +41,13 @@ You can find more advanced usage in combination with an Agent using the store fo
...
 * `Similarity Search with memory storage (RAG)`_
 * `Similarity Search with MongoDB (RAG)`_
+* `Similarity Search with Neo4j (RAG)`_
 * `Similarity Search with Pinecone (RAG)`_
+* `Similarity Search with PSR-6 Cache (RAG)`_
 * `Similarity Search with Qdrant (RAG)`_
 * `Similarity Search with SurrealDB (RAG)`_
-* `Similarity Search with Neo4j (RAG)`_
+
+PS: Both `InMemory` and `PSR-6 cache` vector stores are designed to be primarily used in tests / development environments.
```
GromNaN: You can add details. The environment doesn't matter, the data volume is the problem.

```suggestion
PS: Both `InMemory` and `PSR-6 cache` vector stores will load all the data into the memory of the PHP process. They can be used only the amount of data fits in the PHP memory limit, typically for testing.
```

chr-hertel (approved):
Thanks @Guikingone - especially for going the extra mile of refactoring/extracting the calculation - def worth it tho 👍

src/store/src/DistanceCalculator.php
```diff
@@ -0,0 +1,133 @@
...
+        );
+
+        usort(
+            $currentEmbeddings,
+            static fn (array $embedding, array $nextEmbedding): int => $embedding['distance'] <=> $nextEmbedding['distance'],
+        );
+
+        if (null !== $maxItems && $maxItems < \count($currentEmbeddings)) {
+            $currentEmbeddings = \array_slice($currentEmbeddings, 0, $maxItems);
+        }
```
junaidbinfarooq, line 58:
```suggestion
    public function calculate(array $documents, Vector $vector, ?int $maxItems = 0): array
    {
        $strategy = match ($this->strategy) {
            DistanceStrategy::COSINE_DISTANCE => $this->cosineDistance(...),
            DistanceStrategy::ANGULAR_DISTANCE => $this->angularDistance(...),
            DistanceStrategy::EUCLIDEAN_DISTANCE => $this->euclideanDistance(...),
            DistanceStrategy::MANHATTAN_DISTANCE => $this->manhattanDistance(...),
            DistanceStrategy::CHEBYSHEV_DISTANCE => $this->chebyshevDistance(...),
        };

        if ($maxItems < \count($documents)) {
            $documents = \array_slice($documents, 0, $maxItems);
        }

        $currentEmbeddings = array_map(
            static fn (VectorDocument $vectorDocument): array => [
                'distance' => $strategy($vectorDocument, $vector),
                'document' => $vectorDocument,
            ],
            $documents,
        );

        usort(
            $currentEmbeddings,
            static fn (array $embedding, array $nextEmbedding): int => $embedding['distance'] <=> $nextEmbedding['distance'],
        );
```
How about this?
Instead of always looping over and calculating the distances for all the existing documents, we only do so for the required number of documents if a value is provided for the `$maxItems` argument.

Guikingone: Well, the idea is not to limit the items that we iterate over but rather limiting the returned items once the distance is calculated.

By moving the condition before the mapping/sorting phase, you take the risk of finding items that are not relevant/with a lower score in the final array.

Guikingone:
The configuration in the bundle package has been added (better to do it now than in another PR 😄 )

src/store/src/CacheStore.php
```diff
@@ -0,0 +1,73 @@
...
+final readonly class CacheStore implements VectorStoreInterface
+{
+    public function __construct(
+        private CacheItemPoolInterface $cache,
+        private DistanceCalculator $distanceCalculator = new DistanceCalculator(),
+        private string $cacheKey = '_vectors',
+    ) {
+        if (!interface_exists(CacheItemPoolInterface::class)) {
+            throw new RuntimeException('For using the CacheStore as vector store, a PSR-6 cache implementation is required. Try running "composer require symfony/cache" or another PSR-6 compatible cache.');
+        }
```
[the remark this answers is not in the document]
OskarStark: I think you are wrong @copilot

src/store/doc/index.rst
```diff
@@ -39,12 +39,16 @@ You can find more advanced usage in combination with an Agent using the store fo
...
 * `Similarity Search with MongoDB (RAG)`_
 * `Similarity Search with Neo4j (RAG)`_
 * `Similarity Search with Pinecone (RAG)`_
+* `Similarity Search with PSR-6 Cache (RAG)`_
 * `Similarity Search with Qdrant (RAG)`_
 * `Similarity Search with SurrealDB (RAG)`_
-* `Similarity Search with memory storage (RAG)`_
+
+PS: Both `InMemory` and `PSR-6 cache` vector stores will load all the data into the memory of the PHP process.
+They can be used only the amount of data fits in the PHP memory limit, typically for testing.
```
OskarStark:
```suggestion
.. note::

    Both `InMemory` and `PSR-6 cache` vector stores will load all the data into the
    memory of the PHP process. They can be used only the amount of data fits in the
    PHP memory limit, typically for testing.
```

src/store/src/CacheStore.php
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    public function add(VectorDocument ...$documents): void
+    {
+        $cacheItem = $this->cache->getItem($this->cacheKey);
```
OskarStark: Can w use something like:
```php
// The callable will only be executed on a cache miss.
$value = $pool->get('my_cache_key', function (ItemInterface $item): string {
    $item->expiresAfter(3600);

    // ... do some HTTP request or heavy computations
    $computedValue = 'foobar';

    return $computedValue;
});
```

src/store/doc/index.rst
chr-hertel: looks like this didn't happen

src/store/doc/index.rst
```diff
@@ -97,12 +102,13 @@ This leads to a store implementing two methods::
 .. _`Retrieval Augmented Generation`: https://de.wikipedia.org/wiki/Retrieval-Augmented_Generation
 .. _`Similarity Search with MariaDB (RAG)`: https://github.com/symfony/ai/blob/main/examples/rag/mariadb.php
 .. _`Similarity Search with MongoDB (RAG)`: https://github.com/symfony/ai/blob/main/examples/rag/mongodb.php
-.. _`Similarity Search with Pinecone (RAG)`: https://github.com/symfony/ai/blob/main/examples/rag/pinecone.php
+.. _`Similarity Search with PSR-6 Cache (RAG)`: https://github.com/symfony/ai/blob/main/examples/rag/cache.php
```
chr-hertel: Needs to be moved down after #264

Guikingone: Ah yes, totally forgot about this PR, sorry 😅

src/store/doc/index.rst
Guikingone: Ah, forgot to commit the modification, sorry

Guikingone: Ok, this time is the good one 😅

OskarStark:
Thank you @Guikingone.

PR #263, opened 2025-08-05
[Agent] ElevenLabs TTS tool

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | None
| License       | MIT

Hi 👋🏻 

This PR aims to introduce a new native tool to allows the TTS pipeline from the agent, the HTTP call is performed against ElevenLabs API (only for TTS currently).

Once received, the file is stored using the `$path` property, the unit test is performed against the fixture file (not sure about this one but it works).

### DIFF
--- a/src/agent/src/Toolbox/Tool/ElevenLabs.php
+++ b/src/agent/src/Toolbox/Tool/ElevenLabs.php
@@ -19,6 +19,8 @@ use Symfony\Contracts\HttpClient\HttpClientInterface;
 
 /**
  * @author Guillaume Loulier <personal@guillaumeloulier.fr>
+ *
+ * @see https://elevenlabs.io/
  */
 #[AsTool('text_to_speech', description: 'Convert text to speech / voice')]
 final readonly class ElevenLabs


src/agent/src/Toolbox/Tool/ElevenLabs.php
```diff
@@ -0,0 +1,55 @@
...
+            ],
+            'json' => [
+                'text' => $text,
+                'model_id' => $this->model,
+            ],
+        ]);
+
+        $file = \sprintf('%s/%s.mp3', $this->path, uniqid());
+
+        $filesystem = new Filesystem();
```
OskarStark, line 64: as we have symfony/filesystem in `require_dev` only, lets check if its available and throw a helpful exception if not

src/agent/src/Toolbox/Tool/ElevenLabs.php
```diff
@@ -0,0 +1,55 @@
...
+namespace Symfony\AI\Agent\Toolbox\Tool;
+
+use Symfony\AI\Agent\Toolbox\Attribute\AsTool;
+use Symfony\Component\Filesystem\Filesystem;
+use Symfony\Contracts\HttpClient\HttpClientInterface;
+
+/**
+ * @author Guillaume Loulier <personal@guillaumeloulier.fr>
+ */
+#[AsTool('eleven_labs', description: 'convert text to speech / voice')]
```
OskarStark: `eleven_labs` sounds weird, is it really helpful for the LLM to know when to use this tool?

OskarStark:
```suggestion
#[AsTool('eleven_labs', description: 'Convert text to speech / voice')]
```

Guikingone: Totally agree, locally, it works but I'm not convinced about the name, maybe just `text_to_speech`?

src/agent/src/Toolbox/Tool/ElevenLabs.php
```diff
@@ -0,0 +1,55 @@
...
+    public function __construct(
+        private HttpClientInterface $httpClient,
+        #[\SensitiveParameter] private string $apiKey,
+        private string $path,
+        private string $model,
+        private string $voice,
+    ) {
+    }
+
+    public function __invoke(string $text): array
```
OskarStark, line 50: > Method Symfony\AI\Agent\Toolbox\Tool\ElevenLabs::__invoke() return type has no value type specified in iterable type array.

Guikingone: Yes, currently working on it 😅

src/agent/src/Toolbox/Tool/ElevenLabs.php
```diff
@@ -0,0 +1,55 @@
...
+    {
+        $response = $this->httpClient->request('POST', \sprintf('https://api.elevenlabs.io/v1/text-to-speech/%s?output_format=mp3_44100_128', $this->voice), [
+            'headers' => [
+                'xi-api-key' => $this->apiKey,
+            ],
+            'json' => [
+                'text' => $text,
+                'model_id' => $this->model,
+            ],
+        ]);
```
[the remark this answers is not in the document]
OskarStark: No, thanks

src/agent/src/Toolbox/Tool/ElevenLabs.php
```diff
@@ -0,0 +1,55 @@
...
+
+        $file = \sprintf('%s/%s.mp3', $this->path, uniqid());
```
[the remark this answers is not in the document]
OskarStark: no thanks

src/agent/src/Toolbox/Tool/ElevenLabs.php
Guikingone: Done 🙂

src/agent/src/Toolbox/Tool/ElevenLabs.php
Guikingone: Changed to `text_to_speech`.

src/agent/composer.json
```diff
@@ -41,7 +41,9 @@
         "symfony/css-selector": "^6.4 || ^7.1",
         "symfony/dom-crawler": "^6.4 || ^7.1",
         "symfony/event-dispatcher": "^6.4 || ^7.1",
-        "symfony/http-foundation": "^6.4 || ^7.1"
+        "symfony/filesystem": "^7.3",
```
OskarStark: Also filesystem 6.4 ?

Guikingone: Nice catch, fixed

chr-hertel (changes requested):
ElevenLabs :partying_face: 

Great stuff - thanks for having a look at that platform! So there are multiple things in here - and we should split it:

1.
instead of focussing on the tool, let's bring in a Platform bridge first please. basically a `Symfony\AI\Platform\Bridge\ElevenLabs` and text input, binary result

2.
you're twisting my mind here, but you're right - we should out of the box enable agents to use platforms as a tool as well - not only agents.
and maybe even 3, for example in the demo project there is some kind of pre processing with whisper (speech2text) and this could be also post processing text2speech - this should be standardized somehow ...

a lot of stuff in those 165 lines of code :laughing: 

are you up for starting with the platform bridge?

Guikingone:
> are you up for starting with the platform bridge?

Yes, of course.

I'll take a look at the pre-processing part 👍🏻

OskarStark:
open for finishing this PR @Guikingone ?

Guikingone:
Absolutely but probably not in the current "way of doing thing", we'll probably need a new PR to introduce "voice providers" to handle audio in response of a message, the current API is not "flexible enough" to handle it properly, I'm working on the provider since a few days, the PR should be out this week 🙂

chr-hertel:
closed for now in alignment with @Guikingone - TTS & media support stay relevant tho :+1:

PR #267, opened 2025-08-06
[AIBundle] Add Ollama support

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Docs?         | yes
| Issues        | None
| License       | MIT

Hi 👋🏻 

Noticed that `Ollama` cannot be used as a platform in the bundle, here's the fix along with tests.


src/ai-bundle/src/AiBundle.php
```diff
@@ -214,28 +215,28 @@ private function processPlatformConfig(string $type, array $platform, ContainerB
...
                 ->addTag('proxy', ['interface' => PlatformInterface::class])
                 ->setArguments([
-                    $platform['api_key'],
+                    $platform['host_url'],
                     new Reference('http_client', ContainerInterface::NULL_ON_INVALID_REFERENCE),
-                    new Reference('ai.platform.contract.openai'),
+                    new Reference('ai.platform.contract.default'),
                 ])
-                ->addTag('ai.platform');
+                ->addTag('symfony_ai.platform');
```
OskarStark: why does this diff looks so weird?

src/ai-bundle/config/services.php
```diff
@@ -44,6 +45,8 @@
             ->factory([AnthropicContract::class, 'create'])
         ->set('ai.platform.contract.google', Contract::class)
             ->factory([GeminiContract::class, 'create'])
+        ->set('ai.platform.contract.ollama', Contract::class)
+        ->factory([OllamaContract::class, 'create'])
```
OskarStark:
```suggestion
            ->factory([OllamaContract::class, 'create'])
```

src/ai-bundle/src/AiBundle.php
Guikingone: Because I moved the different conditions by name order, don't know why Github is displaying them like that 🙁

Guikingone: Capture d’écran 2025-08-06 à 11 03 54

OskarStark: > Because I moved the different conditions by name order, don't know why Github is displaying them like that 🙁

Can you revert and we do this in an extra PR?

Guikingone: Done, should be better 🙂

OskarStark:
Thank you Guillaume.

PR #273, opened 2025-08-06
[AI Bundle] [AIBundle] add initialize method call during container compilation

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | #260 
| License       | MIT

Hi 👋🏻 

Last one for today, the method `initialize` must be called during the container compilation (not all stores implement the interface so we can't define a "routine" for it) if we want to prevent any errors at runtime.

### DIFF
--- a/src/ai-bundle/src/AiBundle.php
+++ b/src/ai-bundle/src/AiBundle.php
@@ -507,6 +507,7 @@ final class AiBundle extends AbstractBundle
                 $definition
                     ->addTag('ai.store')
                     ->setArguments($arguments);
+                $definition->addMethodCall('initialize');
 
                 $container->setDefinition('ai.store.'.$type.'.'.$name, $definition);
             }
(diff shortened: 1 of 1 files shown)


Guikingone:
Ci fails due to zip retrieval, don't know, seems random 😄

chr-hertel:
Wait, this feels like doing `doctrine:schema:update` in the container :thinking: 

I don't think we should do that and rather think how we want to do that kind of store management.
Similar to the messenger setup and doctrine schema management, I see that as separate tool.

One key aspect for me here is: the library itself does not rely on that interface/method, and only offers that as convenience. btw, should we have a `tearDown` or `destroy` method as well?! i think so

Guikingone:
Agree on the "messenger-like" approach, maybe a `ai:setup-store` command?

Agree also on the extra methods, could be useful to prevent any errors during the initialization command.

chr-hertel:
Let's go with `ai:store:setup $name` - could think of some `ai:store:drop $name` and `ai:store:index $name` as well ... :thinking:

chr-hertel:
Created this issue as follow up #285

We can close here, right?

Guikingone:
Yes

PR #274, opened 2025-08-06
[Platform][Anthropic] Allow beta feature flags to be passed into platform invocations

TomLisankie:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | Yes
| Docs?         | Yes
| Issues        | See below
| License       | MIT

Anthropic [supports](https://docs.anthropic.com/en/api/beta-headers) passing flags for beta features via beta headers in requests to its API. This PR allows for passing in the desired beta features via the `$options` parameter in the `invoke` function like so:

```php
$platform->invoke($model, $messageBag, [
            'temperature' => 1.0,
            'max_tokens' => 20000,
            // can now include beta features like this:
            'beta_features' => [
                'code-execution-2025-05-22',
                'mcp-client-2025-04-04'
            ]
        ]);
```

If the beta features option is set and contains at least one element, the beta header is constructed inside the model client and then the option is removed from `$options` before the request is built and sent to Anthropic.

If we want to support beta feature flags for other providers in the future, we should perhaps add another parameter to the `invoke` function for platforms and the `request` function for model clients in order to not pollute `$options` (after all, `$options` is meant to hold _model_ options not _platform_ options).

### DIFF
--- a/src/platform/src/Bridge/Anthropic/ModelClient.php
+++ b/src/platform/src/Bridge/Anthropic/ModelClient.php
@@ -39,15 +39,26 @@ final readonly class ModelClient implements ModelClientInterface
 
     public function request(Model $model, array|string $payload, array $options = []): RawHttpResult
     {
+        $headers = [
+            'x-api-key' => $this->apiKey,
+            'anthropic-version' => $this->version,
+        ];
+
         if (isset($options['tools'])) {
             $options['tool_choice'] = ['type' => 'auto'];
         }
 
+        if (
+            isset($options['beta_features'])
+            && \is_array($options['beta_features'])
+            && !empty($options['beta_features'])
+        ) {
+            $headers['anthropic-beta'] = implode(',', $options['beta_features']);
+            unset($options['beta_features']);
+        }
+
         return new RawHttpResult($this->httpClient->request('POST', 'https://api.anthropic.com/v1/messages', [
-            'headers' => [
-                'x-api-key' => $this->apiKey,
-                'anthropic-version' => $this->version,
-            ],
+            'headers' => $headers,
             'json' => array_merge($options, $payload),
         ]));
     }
--- a/src/platform/tests/Bridge/Anthropic/ModelClientTest.php
+++ b/src/platform/tests/Bridge/Anthropic/ModelClientTest.php
@@ -1,0 +1,114 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Platform\Bridge\Anthropic\Tests;
+
+use PHPUnit\Framework\Attributes\CoversClass;
+use PHPUnit\Framework\TestCase;
+use Symfony\AI\Platform\Bridge\Anthropic\Claude;
+use Symfony\AI\Platform\Bridge\Anthropic\ModelClient;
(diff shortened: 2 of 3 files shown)


OskarStark:
I am fine using the `$options` array here.

Can you please add a test case? Thanks

TomLisankie:
> Can you please add a test case? Thanks

Sure thing.

TomLisankie:
@OskarStark Done.

src/platform/doc/index.rst
```diff
@@ -65,7 +65,7 @@ start for vendor-specific models and their capabilities, see ``Symfony\AI\Platfo
 supports a specific feature, like ``Capability::INPUT_AUDIO`` or ``Capability::OUTPUT_IMAGE``.
 
 **Options** are additional parameters that can be passed to the model, like ``temperature`` or ``max_tokens``, and are
-usually defined by the specific models and their documentation.
+usually defined by the specific models and their documentation. Flags for beta features are also passed in here, handled, and then removed before being sent to the provider (beta feature flags are currently only supported for the `Anthropic` platform). A more robust approach for handling provider feature flags may be implemented in the future.
```
chr-hertel: Please remove this again, I don't think we should give those model specific examples in this rather generic section.

src/platform/src/Bridge/Anthropic/ModelClient.php
```diff
@@ -39,15 +39,22 @@ public function supports(Model $model): bool
...
+        $headers = [
+            'x-api-key' => $this->apiKey,
+            'anthropic-version' => $this->version,
+        ];
+
         if (isset($options['tools'])) {
             $options['tool_choice'] = ['type' => 'auto'];
         }
 
+        if (isset($options['beta_features']) && !empty($options['beta_features'])) {
```
chr-hertel: what do you think about also checking for array here?

src/platform/tests/Bridge/Anthropic/ModelClientTest.php
```diff
@@ -0,0 +1,113 @@
...
+    private MockHttpClient $httpClient;
+    private ModelClient $modelClient;
+    private Claude $model;
+
+    protected function setUp(): void
+    {
+        $this->model = new Claude();
+    }
+
+    private function parseHeaders(array $headers): array
```
chr-hertel: please move `private` methods below `public` methods

i wonder why php-cs-fixer doesn't complain :thinking:

chr-hertel (changes requested):
Good catch, those specific things slip easily through, thanks!

minor comments from my end

src/platform/tests/Bridge/Anthropic/ModelClientTest.php
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    public function testAnthropicBetaHeaderIsNotSetWhenBetaFeaturesIsNotProvided(): void
+    {
+        $this->httpClient = new MockHttpClient(function ($method, $url, $options) {
+            $headers = $this->parseHeaders($options['headers']);
+
+            $this->assertArrayNotHasKey('anthropic-beta', $headers);
+
+            return new MockResponse('{"success": true}');
```
OskarStark: Please use JsonMockResponse

TomLisankie:
Will address these issues on Monday.

TomLisankie:
It's done now. My apologies for the delay, I had some higher priority work to attend to.

chr-hertel (approved):
> It's done now. My apologies for the delay, I had some higher priority work to attend to.

No need to apologize - we all have a day job to make the living :D
Looks good to me know, thanks!

chr-hertel:
Thank you @TomLisankie.

PR #281, opened 2025-08-07
[AIBundle] Improve injection aliases

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | None
| License       | MIT

Hi 👋🏻 

This PR aims to add "injection aliases", those allows the user to use the following code while defining X agents: 

```php
public function __construct(
    private AgentInterface $fooAgent,
) {
}
```

Why the `Agent` suffix? Because we don't know, maybe one day, a service will be called `foo` and it will trigger an error, it seems better to define a suffix and ensure that this bundle is the only one to use it.

### DIFF
--- a/src/ai-bundle/tests/DependencyInjection/AiBundleTest.php
+++ b/src/ai-bundle/tests/DependencyInjection/AiBundleTest.php
@@ -30,6 +30,21 @@ class AiBundleTest extends TestCase
         $this->buildContainer($this->getFullConfig());
     }
 
+    public function testInjectionAgentAliasIsRegistered()
+    {
+        $container = $this->buildContainer([
+            'ai' => [
+                'agent' => [
+                    'my_agent' => [
+                        'model' => ['class' => 'Symfony\AI\Platform\Bridge\OpenAi\Gpt'],
+                    ],
+                ],
+            ],
+        ]);
+
+        $this->assertTrue($container->hasAlias('Symfony\AI\Agent\AgentInterface $myAgentAgent'));
+    }
+
     #[TestWith([true], 'enabled')]
     #[TestWith([false], 'disabled')]
     public function testFaultTolerantAgentSpecificToolbox(bool $enabled)
--- a/src/ai-bundle/src/AiBundle.php
+++ b/src/ai-bundle/src/AiBundle.php
@@ -27,13 +27,13 @@ use Symfony\AI\AiBundle\Profiler\TraceableToolbox;
 use Symfony\AI\AiBundle\Security\Attribute\IsGrantedTool;
 use Symfony\AI\Platform\Bridge\Anthropic\PlatformFactory as AnthropicPlatformFactory;
 use Symfony\AI\Platform\Bridge\Azure\OpenAi\PlatformFactory as AzureOpenAiPlatformFactory;
+use Symfony\AI\Platform\Bridge\Cerebras\PlatformFactory as CerebrasPlatformFactory;
 use Symfony\AI\Platform\Bridge\Gemini\PlatformFactory as GeminiPlatformFactory;
 use Symfony\AI\Platform\Bridge\LmStudio\PlatformFactory as LmStudioPlatformFactory;
 use Symfony\AI\Platform\Bridge\Mistral\PlatformFactory as MistralPlatformFactory;
 use Symfony\AI\Platform\Bridge\Ollama\PlatformFactory as OllamaPlatformFactory;
 use Symfony\AI\Platform\Bridge\OpenAi\PlatformFactory as OpenAiPlatformFactory;
 use Symfony\AI\Platform\Bridge\OpenRouter\PlatformFactory as OpenRouterPlatformFactory;
-use Symfony\AI\Platform\Bridge\Cerebras\PlatformFactory as CerebrasPlatformFactory;
 use Symfony\AI\Platform\Model;
 use Symfony\AI\Platform\ModelClientInterface;
 use Symfony\AI\Platform\Platform;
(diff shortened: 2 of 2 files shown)


src/ai-bundle/tests/DependencyInjection/AiBundleTest.php
```diff
@@ -30,6 +30,21 @@ public function testExtensionLoadDoesNotThrow()
...
+            'ai' => [
+                'agent' => [
+                    'my_agent' => [
+                        'model' => ['class' => 'Symfony\AI\Platform\Bridge\OpenAi\Gpt'],
+                    ],
+                ],
+            ],
+        ]);
+
+        $this->assertTrue($container->hasAlias('.Symfony\AI\Agent\AgentInterface $my_agentAgent'));
```
OskarStark: Shouldn't it be $myAgentAgent?

Guikingone: Well, I'm not sure, while adding the call, the container export the service definition as the following: 

```xml
<service id="Symfony\AI\Agent\AgentInterface $fooAgent" alias="ai.agent.foo"/>
```

I bet Symfony is exporting the service along with the FQCN, don't know why to be fair but looks like the "default behavior" 🤔

OskarStark: Your name is `my_agent`, so IMHO it should be `myAgent` + `Agent` suffix, to be more clear, I would expect `$fooBarAgent` when I name it `foo_bar` and not `$foo_barAgent`. How does Monolog is converting the names for example or HttpClient?

Guikingone: Well, regarding the method: 

```php
/**
 * Registers an autowiring alias that only binds to a specific argument name.
 *
 * The argument name is derived from $name if provided (from $id otherwise)
 * using camel case: "foo.bar" or "foo_bar" creates an alias bound to
 * "$fooBar"-named arguments with $type as type-hint. Such arguments will
 * receive the service $id when autowiring is used.
 */
public function registerAliasForArgument(string $id, string $type, ?string $name = null): Alias
{
    $parsedName = (new Target($name ??= $id))->getParsedName();

    if (!preg_match('/^[a-zA-Z_\x7f-\xff]/', $parsedName)) {
        if ($id !== $name) {
            $id = \sprintf(' for service "%s"', $id);
        }

        throw new InvalidArgumentException(\sprintf('Invalid argument name "%s"'.$id.': the first character must be a letter.', $name));
    }

    if ($parsedName !== $name) {
        $this->setAlias('.'.$type.' $'.$name, $type.' $'.$parsedName);
    }

    return $this->setAlias($type.' $'.$parsedName, $id);
}
```

It look like Sf is not doing the job properly, here's the method usage for HttpClient's: 

```php
$container->registerAliasForArgument($name, HttpClientInterface::class);

if ($hasPsr18) {
    $container->setDefinition('psr18.'.$name, new ChildDefinition('psr18.http_client'))
        ->replaceArgument(0, new Reference($name));

    $container->registerAliasForArgument('psr18.'.$name, ClientInterface::class, $name);
}

if ($hasHttplug) {
    $container->setDefinition('httplug.'.$name, new ChildDefinition('httplug.http_client'))
        ->replaceArgument(0, new Reference($name));

    $container->registerAliasForArgument('httplug.'.$name, HttpAsyncClient::class, $name);
}
```

BTW, If I try to define a custom agent in my application called `translation`, the resolved argument is `translationAgent`.

Guikingone: Regarding Monolog, there's nothing more in the extension: 

```php
protected function createLogger(string $channel, string $loggerId, ContainerBuilder $container)
{
    if (!in_array($channel, $this->channels)) {
        $logger = new ChildDefinition('monolog.logger_prototype');
        $logger->replaceArgument(0, $channel);
        $container->setDefinition($loggerId, $logger);
        $this->channels[] = $channel;
    }

    $parameterName = $channel . 'Logger';

    $container->registerAliasForArgument($loggerId, LoggerInterface::class, $parameterName);
}
```

Guikingone: Ok, found the issue.

Looks like if you set `foo_toto`, Symfony resolve it as `fooToto`, the catch is that if the `$name` is not the same as the parsed name: 

```php
if ($parsedName !== $name) {
    $this->setAlias('.'.$type.' $'.$name, $type.' $'.$parsedName);
}
```

In this case, the `$name` is used (so `foo_toto`) rather than `fooToto` (the parsed one) 🤔 

EDIT:

One solution might be to call the `Target` class while defining the alias: 

```php
$container->registerAliasForArgument('ai.agent.'.$name, AgentInterface::class, (new Target($name.'Agent'))->getParsedName());
```

OskarStark: Thanks for digging, I am fine either way, but would like to end with `$myAgentAgent` for `my_agent` 😄 

Maybe something which can be unified? cc @nicolas-grekas

Guikingone: Ok, just pushed the fix with the `Target` call, don't see any other way to do it as the replacement is defined inside the class, IMO, that's not a big deal to use the attribute but it depends on the rules Symfony applies here so if it's doesn't fit, maybe we can c/c the replacement approach in the `AiBundle` class.

Any way, this approach allows to define an agent called `foo_bar` and expect an argument `fooBarAgent` in services methods.

chr-hertel (approved):
Tested it - works! :+1:

chr-hertel:
Thank you @Guikingone.

PR #282, opened 2025-08-07
[OpenAI] Add GPT 5 Models

DZunke:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | 
| License       | MIT

The new GPT Models are out, so i have added them to the platform with default configuration for image support and structured output. 

https://platform.openai.com/docs/models/gpt-5
https://platform.openai.com/docs/models/gpt-5-mini
https://platform.openai.com/docs/models/gpt-5-nano
https://platform.openai.com/docs/models/gpt-5-chat-latest


chr-hertel (changes requested):
on the pulse of things :D

when bringing in `gpt-5-chat` here, you need to make the tool calling capability dynamic as well - it's not supported with that model.

chr-hertel:
Thank you @DZunke.

PR #284, opened 2025-08-07
[Platform] Add Cerebras

junaidbinfarooq:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | #16 
| License       | MIT

This PR adds support for the [Cerebras](https://inference-docs.cerebras.ai/) platform

### DIFF
--- a/examples/cerebras/chat.php
+++ b/examples/cerebras/chat.php
@@ -1,0 +1,29 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use Symfony\AI\Agent\Agent;
+use Symfony\AI\Platform\Bridge\Cerebras\Model;
+use Symfony\AI\Platform\Bridge\Cerebras\PlatformFactory;
+use Symfony\AI\Platform\Message\Message;
+use Symfony\AI\Platform\Message\MessageBag;
+
+require_once dirname(__DIR__).'/bootstrap.php';
+
+$platform = PlatformFactory::create(env('CEREBRAS_API_KEY'), http_client());
+
+$agent = new Agent($platform, new Model(), logger: logger());
+$messages = new MessageBag(
+    Message::forSystem('You are a helpful assistant.'),
+    Message::ofUser('How is the weather in Tokyo today?'),
+);
+$result = $agent->call($messages);
+
+echo $result->getContent().\PHP_EOL;
--- a/src/platform/doc/index.rst
+++ b/src/platform/doc/index.rst
@@ -346,6 +346,8 @@ This allows fast and isolated testing of AI-powered features without relying on 
 
 * `Parallel GPT Calls`_
 * `Parallel Embeddings Calls`_
+* `Cerebras Chat`_
+* `Cerebras Streaming`_
 
 .. note::
 
@@ -392,3 +394,5 @@ This allows fast and isolated testing of AI-powered features without relying on 
 .. _`Parallel Embeddings Calls`: https://github.com/symfony/ai/blob/main/examples/misc/parallel-embeddings.php
 .. _`LM Studio`: https://lmstudio.ai/
 .. _`LM Studio Catalog`: https://lmstudio.ai/models
+.. _`Cerebras Chat`: https://github.com/symfony/ai/blob/main/examples/cerebras/chat.php
+.. _`Cerebras Streaming`: https://github.com/symfony/ai/blob/main/examples/cerebras/stream.php
--- a/src/platform/src/Bridge/Cerebras/Model.php
+++ b/src/platform/src/Bridge/Cerebras/Model.php
@@ -1,0 +1,39 @@
+<?php
+
+namespace Symfony\AI\Platform\Bridge\Cerebras;
+
+use Symfony\AI\Platform\Capability;
+use \Symfony\AI\Platform\Model as BaseModel;
+
+/**
+ * @author Junaid Farooq <ulislam.junaid125@gmail.com>
+ */
+final class Model extends BaseModel
+{
+    public const LLAMA_4_SCOUT_17B_16E_INSTRUCT = 'llama-4-scout-17b-16e-instruct';
+    public const LLAMA3_1_8B = 'llama3.1-8b';
+    public const LLAMA_3_3_70B = 'llama-3.3-70b';
+    public const LLAMA_4_MAVERICK_17B_128E_INSTRUCT = 'llama-4-maverick-17b-128e-instruct';
+    public const QWEN_3_32B = 'qwen-3-32b';
+    public const QWEN_3_235B_A22B_INSTRUCT_2507 = 'qwen-3-235b-a22b-instruct-2507';
+    public const QWEN_3_235B_A22B_THINKING_2507 = 'qwen-3-235b-a22b-thinking-2507';
+    public const QWEN_3_CODER_480B = 'qwen-3-coder-480b';
+    public const GPT_OSS_120B = 'gpt-oss-120b';
+
+    public const CAPABILITIES = [
+        Capability::INPUT_MESSAGES,
+        Capability::OUTPUT_TEXT,
+        Capability::OUTPUT_STREAMING,
+    ];
+
+    /**
+     * @see https://inference-docs.cerebras.ai/api-reference/chat-completions for details like options
+     */
+    public function __construct(
+        string $name = self::LLAMA3_1_8B,
+        array $capabilities = self::CAPABILITIES,
+        array $options = [],
+    ) {
+        parent::__construct($name, $capabilities, $options);
+    }
+}
--- a/src/platform/src/Bridge/Cerebras/ModelClient.php
+++ b/src/platform/src/Bridge/Cerebras/ModelClient.php
@@ -1,0 +1,64 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Platform\Bridge\Cerebras;
+
+use Symfony\AI\Platform\Exception\InvalidArgumentException;
+use Symfony\AI\Platform\Model as BaseModel;
+use Symfony\AI\Platform\ModelClientInterface;
+use Symfony\AI\Platform\Result\RawHttpResult;
+use Symfony\Component\HttpClient\EventSourceHttpClient;
+use Symfony\Contracts\HttpClient\HttpClientInterface;
+
+/**
+ * @author Junaid Farooq <ulislam.junaid125@gmail.com>
+ */
+final readonly class ModelClient implements ModelClientInterface
+{
+    private EventSourceHttpClient $httpClient;
+
+    public function __construct(
+        HttpClientInterface $httpClient,
+        #[\SensitiveParameter] private string $apiKey,
+    ) {
+        if ('' === $apiKey) {
+            throw new InvalidArgumentException('The API key must not be empty.');
+        }
+
+        if (!str_starts_with($apiKey, 'csk-')) {
+            throw new InvalidArgumentException('The API key must start with "csk-".');
+        }
+
+        $this->httpClient = $httpClient instanceof EventSourceHttpClient ? $httpClient : new EventSourceHttpClient($httpClient);
+    }
+
+    public function supports(BaseModel $model): bool
+    {
+        return $model instanceof Model;
+    }
+
+    public function request(BaseModel $model, array|string $payload, array $options = []): RawHttpResult
+    {
+        return new RawHttpResult(
+            $this->httpClient->request(
+                'POST', 'https://api.cerebras.ai/v1/chat/completions',
+                [
+                    'headers' => [
+                        'Content-Type' => 'application/json',
+                        'Authorization' => sprintf('Bearer %s', $this->apiKey),
+                    ],
+                    'json' => \is_array($payload) ? array_merge($payload, $options) : $payload,
+                ]
+            )
+        );
+    }
+}
+
--- a/src/platform/src/Bridge/Cerebras/PlatformFactory.php
+++ b/src/platform/src/Bridge/Cerebras/PlatformFactory.php
@@ -1,0 +1,34 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Platform\Bridge\Cerebras;
+
+use Symfony\AI\Platform\Platform;
+use Symfony\Component\HttpClient\EventSourceHttpClient;
+use Symfony\Contracts\HttpClient\HttpClientInterface;
+
+/**
+ * @author Junaid Farooq <ulislam.junaid125@gmail.com>
+ */
+final readonly class PlatformFactory
+{
+    public static function create(
+        #[\SensitiveParameter] string $apiKey,
+        ?HttpClientInterface $httpClient = null,
+    ): Platform {
+        $httpClient = $httpClient instanceof EventSourceHttpClient ? $httpClient : new EventSourceHttpClient($httpClient);
+
+        return new Platform(
+            [new ModelClient($httpClient, $apiKey)],
+            [new ResultConverter()],
+        );
+    }
+}
--- a/src/platform/tests/Bridge/Cerebras/ModelClientTest.php
+++ b/src/platform/tests/Bridge/Cerebras/ModelClientTest.php
@@ -1,0 +1,95 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Platform\Tests\Bridge\Cerebras;
+
+use PHPUnit\Framework\Attributes\CoversClass;
+use PHPUnit\Framework\Attributes\Small;
+use PHPUnit\Framework\Attributes\TestWith;
+use PHPUnit\Framework\Attributes\UsesClass;
+use PHPUnit\Framework\TestCase;
+use Symfony\AI\Platform\Bridge\Cerebras\Model;
+use Symfony\AI\Platform\Bridge\Cerebras\ModelClient;
+use Symfony\AI\Platform\Exception\InvalidArgumentException;
+use Symfony\Component\HttpClient\MockHttpClient;
+use Symfony\Component\HttpClient\Response\JsonMockResponse;
+
+/**
+ * @author Junaid Farooq <ulislam.junaid125@gmail.com>
+ */
+#[CoversClass(ModelClient::class)]
+#[UsesClass(Model::class)]
+#[Small]
+class ModelClientTest extends TestCase
+{
+    public function testItDoesNotAllowAnEmptyKey()
+    {
+        $this->expectException(InvalidArgumentException::class);
+        $this->expectExceptionMessage('The API key must not be empty.');
+
+        new ModelClient(new MockHttpClient(), '');
+    }
+
+    #[TestWith(['api-key-without-prefix'])]
+    #[TestWith(['pk-api-key'])]
+    #[TestWith(['SK-api-key'])]
+    #[TestWith(['skapikey'])]
+    #[TestWith(['sk api-key'])]
+    #[TestWith(['sk'])]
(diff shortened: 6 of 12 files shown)


examples/cerebras/chat.php
```diff
@@ -0,0 +1,35 @@
...
+    Message::ofUser('How is the weather in Tokyo today?'),
+);
+$result = $agent->call($messages, [
+    'max_completion_tokens' => 50,
+    'temperature' => 0.2,
+]);
+
+echo $result->getContent().\PHP_EOL;
+
+echo "\nChat finished successfully.\n";
```
chr-hertel: This is not needed - at least it would be the first example. Let's be consistent and remove it please.

src/platform/src/Bridge/Cerebras/ApiKey.php
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+namespace Symfony\AI\Platform\Bridge\Cerebras;
+
+use Symfony\AI\Platform\Exception\InvalidArgumentException;
+
+/**
+ * @author Junaid Farooq <ulislam.junaid125@gmail.com>
+ */
+final readonly class ApiKey implements \Stringable
```
chr-hertel: the extra class is not needed, please treat the api key as simple scalar string

src/platform/src/Bridge/Cerebras/ModelType.php
```diff
@@ -0,0 +1,46 @@
...
+ */
+
+namespace Symfony\AI\Platform\Bridge\Cerebras;
+
+use Symfony\AI\Platform\Model;
+
+/**
+ * @author Junaid Farooq <ulislam.junaid125@gmail.com>
+ */
+enum ModelType: string
```
chr-hertel: Please use string constants directly in the model.
this approach limits the usage to only models known by this library and not newer ones.

examples/cerebras/chat.php
junaidbinfarooq: Ok np. Removed.

src/platform/src/Bridge/Cerebras/ApiKey.php
junaidbinfarooq: May I know the reason for not adding such a class? Thought it better encapsulates the validation logic.
I removed it, though.

src/platform/src/Bridge/Cerebras/ModelType.php
junaidbinfarooq: Cool.

examples/cerebras/chat.php
```diff
@@ -0,0 +1,35 @@
...
+
+use Symfony\AI\Agent\Agent;
+use Symfony\AI\Platform\Bridge\Cerebras\ApiKey;
+use Symfony\AI\Platform\Bridge\Cerebras\Model;
+use Symfony\AI\Platform\Bridge\Cerebras\PlatformFactory;
+use Symfony\AI\Platform\Message\Message;
+use Symfony\AI\Platform\Message\MessageBag;
+
+require_once dirname(__DIR__).'/bootstrap.php';
+
```
OskarStark, line 19: please check for the env var like in other examples, thanks

examples/cerebras/chat.php
```diff
@@ -0,0 +1,35 @@
...
+$platform = PlatformFactory::create(new ApiKey(env('CEREBRAS_API_KEY')), http_client());
+
+$agent = new Agent($platform, new Model(), logger: logger());
+$messages = new MessageBag(
+    Message::forSystem('You are a helpful assistant.'),
+    Message::ofUser('How is the weather in Tokyo today?'),
+);
+$result = $agent->call($messages, [
+    'max_completion_tokens' => 50,
+    'temperature' => 0.2,
```
OskarStark: is this needed? Lets keep the example ASAP (as slim as possible) 😄

src/agent/doc/index.rst
```diff
@@ -558,4 +558,4 @@ useful when certain interactions shouldn't be influenced by the memory context::
 .. _`Structured Output with PHP class`: https://github.com/symfony/ai/blob/main/examples/openai/structured-output-math.php
 .. _`Structured Output with array`: https://github.com/symfony/ai/blob/main/examples/openai/structured-output-clock.php
 .. _`Chat with static memory`: https://github.com/symfony/ai/blob/main/examples/memory/static.php
-.. _`Chat with embedding search memory`: https://github.com/symfony/ai/blob/main/memory/mariadb.php
+.. _`Chat with embedding search memory`: https://github.com/symfony/ai/blob/main/examples/memory/embedding-search.php
```
OskarStark: This change looks wrong to me

src/platform/doc/index.rst
```diff
@@ -89,6 +89,7 @@ usually defined by the specific models and their documentation.
     * `LM Studio Catalog`_ and `HuggingFace`_ Models  with `LM Studio`_ as Platform.
     * All models provided by `HuggingFace`_ can be listed with a command in the examples folder,
       and also filtered, e.g. ``php examples/huggingface/_model-listing.php --provider=hf-inference --task=object-detection``
+    * Various models Models provided by `Cerebras`_ Platform.
```
OskarStark: You need do add a link for Cerebras at the end of the file

src/platform/src/Bridge/Cerebras/Model.php
```diff
@@ -0,0 +1,39 @@
...
+{
+    public const LLAMA_4_SCOUT_17B_16E_INSTRUCT = 'llama-4-scout-17b-16e-instruct';
+    public const LLAMA3_1_8B = 'llama3.1-8b';
+    public const LLAMA_3_3_70B = 'llama-3.3-70b';
+    public const LLAMA_4_MAVERICK_17B_128e_INSTRUCT = 'llama-4-maverick-17b-128e-instruct (preview)';
+    public const QWEN_3_32B = 'qwen-3-32b';
+    public const QWEN_3_235B_a22B_INSTRUCT_2507 = 'qwen-3-235b-a22b-instruct-2507 (preview)';
+    public const QWEN_3_235B_a22B_thinking_2507 = 'qwen-3-235b-a22b-thinking-2507 (preview)';
+    public const QWEN_3_CODER_480B = 'qwen-3-coder-480b (preview)';
+    public const GPT_OSS_120B = 'gpt-oss-120b (preview)';
```
OskarStark: Please uppercase for all constants

src/platform/src/Bridge/Cerebras/Model.php
```diff
@@ -0,0 +1,39 @@
...
+        Capability::OUTPUT_STREAMING,
+    ];
+
+    /**
+     * @see https://inference-docs.cerebras.ai/api-reference/chat-completions for details like options
+     */
+    public function __construct(
+        string $name = self::LLAMA3_1_8B,
+        array $capabilities = self::CAPABILITIES,
+        array $options = ['temperature' => 1.0, 'max_completion_tokens' => 1000],
```
OskarStark: do we need to set this or is it the default from the API anyway? In this case I am in favor of removing it here.

src/platform/src/Bridge/Cerebras/ModelClient.php
```diff
@@ -0,0 +1,65 @@
...
+ * @author Junaid Farooq <ulislam.junaid125@gmail.com>
+ */
+final readonly class ModelClient implements ModelClientInterface
+{
+    private EventSourceHttpClient $httpClient;
+
+    public function __construct(
+        HttpClientInterface $httpClient,
+        #[\SensitiveParameter]
+        private string $apiKey,
```
OskarStark: on one line please

src/platform/src/Bridge/Cerebras/ModelClient.php
```diff
@@ -0,0 +1,65 @@
...
+
+        $this->httpClient = $httpClient instanceof EventSourceHttpClient ? $httpClient : new EventSourceHttpClient($httpClient);
+    }
+
+    public function supports(BaseModel $model): bool
+    {
+        return $model instanceof Model;
+    }
+
+    public function request(BaseModel $model, array|string $payload, array $options = []): RawHttpResult
```
OskarStark, line 48: why the `array|string` for the payload here? I would prefer `array` only

src/platform/src/Bridge/Cerebras/ModelType.php
OskarStark, line 1: Can be removed, we have constants now

examples/cerebras/chat.php
junaidbinfarooq: Not really :smile: 
Just added the options to demonstrate them being overridden.
Removed now to keep it simple.

src/platform/src/Bridge/Cerebras/PlatformFactory.php
```diff
@@ -0,0 +1,35 @@
...
+use Symfony\Contracts\HttpClient\HttpClientInterface;
+
+/**
+ * @author Junaid Farooq <ulislam.junaid125@gmail.com>
+ */
+final readonly class PlatformFactory
+{
+    public static function create(
+        #[\SensitiveParameter]
+        string $apiKey,
```
OskarStark:
```suggestion
        #[\SensitiveParameter] string $apiKey,
```

src/platform/tests/Bridge/Cerebras/ModelClientTest.php
```diff
@@ -0,0 +1,104 @@
...
+/**
+ * @author Junaid Farooq <ulislam.junaid125@gmail.com>
+ */
+#[CoversClass(ModelClient::class)]
+#[UsesClass(Model::class)]
+#[Small]
+class ModelClientTest extends TestCase
+{
+    #[Test]
+    public function it_does_not_allow_an_empty_key(): void
```
OskarStark: No `: void` in test methods and also rename methods to:
```suggestion
    public function testItDoesNotAllowAnEmptyKey()
```

examples/cerebras/chat.php
junaidbinfarooq: Could you clarify?
Many other scripts follow the same approach as far as I can see.

src/platform/src/Bridge/Cerebras/ApiKey.php
chr-hertel: sure, I understand that those classes promote the API key as part of the libraries business objects, and I agree that checking for an empty string is valid. but that's it.

I know that there is a second validation rule, but that is also a bit questionable when you ask me: why should we on library side have an opinion on the format of the api key cerebras might accept - we can't really validate and should spend the effort rather on API error code conversion instead. Just think of the change scenario here: Cerebras decides to change the token generation - happened before with OpenAI for example - a user gets a new token but can't use it with the lib because we thought it might only be `csk` - that's brittle to me.

So an extra class for checking if a string is empty is not needed in my opinion and would be quite some extra work if we'd do that all the time instead of scalar values.

junaidbinfarooq:
> Hi @junaidbinfarooq - welcome to this repository - great to meet you!
> 
> I think contributing a new platform is a great way to get started - maybe looking at the other implementations that are already around helps as well.
> 
> I left some comments, and will continue reviewing when the pipeline is green - let me know if you have some questions or need more guidance!

Thank you @chr-hertel

src/platform/src/Bridge/Cerebras/ModelClient.php
junaidbinfarooq: Just followed the signature declared in the interface.

src/platform/src/Bridge/Cerebras/ModelType.php
junaidbinfarooq: It has already been removed.

src/platform/tests/Bridge/Cerebras/ModelClientTest.php
junaidbinfarooq: Ah! I had in mind that casing might come up during the review.
Btw, is there anything wrong with having _void_ as the return type here? I see it is removed at other places too.

junaidbinfarooq: Phpstan is also complaining about missing return types :smile:

examples/bootstrap.php
```diff
@@ -17,7 +17,7 @@
 use Symfony\Component\HttpClient\HttpClient;
 use Symfony\Contracts\HttpClient\HttpClientInterface;
 
-require_once __DIR__.'/vendor/autoload.php';
+require_once __DIR__.'/../src/platform/vendor/autoload.php';
```
chr-hertel: please revert this. you can execute the examples with you local changes like this:

```
cd examples
composer install
../link
```

this creates symlinks to your local versions of the packages in the vendor folder of the examples

examples/cerebras/chat.php
chr-hertel: yes, that's not needed anymore, but was moved into the `env(...)` function

src/agent/doc/index.rst
chr-hertel: partially right, i'd say
```suggestion
.. _`Chat with embedding search memory`: https://github.com/symfony/ai/blob/main/examples/memory/mariadb.php
```

src/platform/src/Bridge/Cerebras/Model.php
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * @author Junaid Farooq <ulislam.junaid125@gmail.com>
+ */
+final class Model extends BaseModel
+{
+    public const LLAMA_4_SCOUT_17B_16E_INSTRUCT = 'llama-4-scout-17b-16e-instruct';
+    public const LLAMA3_1_8B = 'llama3.1-8b';
+    public const LLAMA_3_3_70B = 'llama-3.3-70b';
+    public const LLAMA_4_MAVERICK_17B_128e_INSTRUCT = 'llama-4-maverick-17b-128e-instruct (preview)';
```
chr-hertel: are you sure about the `(preview)` part in the string?
```suggestion
    public const LLAMA_4_MAVERICK_17B_128e_INSTRUCT = 'llama-4-maverick-17b-128e-instruct';
```

chr-hertel: same for the others

src/platform/src/Bridge/Cerebras/Model.php
```diff
@@ -0,0 +1,39 @@
...
+    public const GPT_OSS_120B = 'gpt-oss-120b (preview)';
+
+    public const CAPABILITIES = [
+        Capability::INPUT_MESSAGES,
+        Capability::OUTPUT_TEXT,
+        Capability::OUTPUT_STREAMING,
+    ];
+
+    /**
+     * @see https://inference-docs.cerebras.ai/api-reference/chat-completions for details like options
```
chr-hertel, line 30: i like that service link :+1:

src/platform/src/Bridge/Cerebras/ApiKey.php
junaidbinfarooq: @chr-hertel 
Thank you for the explanation.
Makes sense.

src/agent/doc/index.rst
junaidbinfarooq: I had actually added a few links for the new bridge there, but removed them later on, and this remained.
The link seemed broken, but I didn't look deeply into it.
Removed for now.

src/platform/doc/index.rst
junaidbinfarooq: Added a couple of links at the end.

src/platform/src/Bridge/Cerebras/Model.php
junaidbinfarooq: Oh! tried uppercasing them all, but didn't realize some are still lowercased.

src/platform/src/Bridge/Cerebras/Model.php
junaidbinfarooq: Couldn't find default values for these options.
Removed now, though.

examples/bootstrap.php
junaidbinfarooq: Oh! Committed by mistake.
I tried linking the vendor folder to the local changes, but it didn't work for some reason.
I kept getting `SomeClass` not found exception, so I did this as a temporary measure to run the examples.

src/agent/doc/index.rst
junaidbinfarooq: I reverted the change for now and will fix the link in another PR, as other such stuff may be broken, so they are better fixed together if it is the case.

src/platform/src/Bridge/Cerebras/Model.php
junaidbinfarooq: Thanks.

src/platform/src/Bridge/Cerebras/Model.php
junaidbinfarooq: Oh! Forgot to remove the suffixes.
Done now.

src/platform/tests/Bridge/Cerebras/ModelClientTest.php
junaidbinfarooq: @OskarStark 
Could you let me know what to do about phpstan as it breaks the CI due to missing return types in the test methods?

junaidbinfarooq: nvm. Fixed it now.
I see `platform/phpstan.dist.neon` ignores such errors for methods wth prefixes set to _test_, which wasn't so in my case as I had used the `Test` attribute from _phpunit_ library instead of _test_ prefix.

src/platform/src/Bridge/Cerebras/ModelType.php
chr-hertel: It is not in use anymore, but still in the branch

src/platform/src/Bridge/Cerebras/Model.php
junaidbinfarooq: Done now.

src/platform/src/Bridge/Cerebras/ModelType.php
junaidbinfarooq: Removed it now.

examples/cerebras/chat.php
OskarStark: Right, sorry

src/platform/src/Bridge/Cerebras/ModelClient.php
OskarStark: Right, sorry

src/platform/tests/Bridge/Cerebras/ModelClientTest.php
```diff
@@ -0,0 +1,99 @@
...
+    {
+        $client = new ModelClient(new MockHttpClient(), 'csk-1234567890abcdef');
+
+        self::assertTrue($client->supports(new Model(Model::GPT_OSS_120B)));
+    }
+
+    public function testItSuccessfullyInvokesTheModel()
+    {
+        $httpClient = new MockHttpClient(
+            new MockResponse(
```
OskarStark: Please use `JsonMockResponse`

src/platform/tests/Bridge/Cerebras/ModelClientTest.php
```diff
@@ -0,0 +1,99 @@
...
+        $data = $result->getData();
+        $info = $result->getObject()->getInfo();
+
+        self::assertNotEmpty($data);
+        self::assertNotEmpty($info);
+        self::assertSame('POST', $info['http_method']);
+        self::assertSame('https://api.cerebras.ai/v1/chat/completions', $info['url']);
+        self::assertSame(
+            '{"model":"llama-3.3-70b","input":{"messages":[{"role":"user","content":"Hello, world!"}]},"temperature":0.5}',
+            json_encode($data, \JSON_THROW_ON_ERROR),
```
OskarStark: we can assert on array, no need to json_encode the result, right?

src/platform/tests/Bridge/Cerebras/ModelClientTest.php
junaidbinfarooq: Done.

src/platform/tests/Bridge/Cerebras/ModelClientTest.php
junaidbinfarooq: Indeed, done now.
It was at some point an array, IIRC.
For some reason, I changed it back to JSON.

junaidbinfarooq:
> Please add AiBundle integration if needed

Done.

src/platform/src/Bridge/Cerebras/ModelClient.php
junaidbinfarooq: Done already.

OskarStark (approved):
LGTM

junaidbinfarooq:
@chr-hertel 
WDYT?

src/agent/doc/index.rst
junaidbinfarooq: Fixed [here](https://github.com/symfony/ai/pull/291).

chr-hertel (approved):
Good for a first step - thanks for that @junaidbinfarooq!
Needs a follow up about tool calling capabilities of some pre-defined models.

chr-hertel:
Thank you @junaidbinfarooq.

junaidbinfarooq:
> Good for a first step - thanks for that @junaidbinfarooq! Needs a follow up about tool calling capabilities of some pre-defined models.

I considered it but then it seemed the tool calling capability isn't very reliable:
image

PR #240, opened 2025-07-31
Draft: Streamable Http

xavierleune:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Docs?         | Todo 
| Issues        | Fix #166 #5 
| License       | MIT

This pull request is currently a draft / work in progress (before getting away from keyboard) to enable support of Streamable Http. There is still a lot to do, but it's kinda working. Quick demo with a tool call that sends a bunch of notif before sending the result: [youtube](https://youtu.be/ffzV194-DPA).

Key points:
- I don't use the `Server` for streamable http, it's quite impractical as it hides too much the internal logic
- JsonRpcHandler now handle one message at a time, as batching has been dropped
- I think we should separate message encoding / decoding and message handling, it might improve the workflow
- The `*Handler` Api needs a rework to work well with streams, it's currently only a quick poc in `ToolCallHandler` and it's kinda ugly

Todo:
- [x] Support for [Session Management](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#session-management)
- [x] Basic support for [Resumability and Redelivery](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#resumability-and-redelivery)
- [x] Dropped support for [Json-Rpc batching](https://modelcontextprotocol.io/specification/2025-06-18/changelog#major-changes)
- [ ] Better mapping between Json RPC schema and `Symfony\AI\McpSdk\Message` (with separation between Server sent objects and Client sent objects)
- [ ] Tests

Technical stuff:
- [ ] Param for session duration
- [ ] Cleaner support for notifications sent by the client
- [ ] Simplify session management (drop left-over `SessionIdentifierResolver` and so)
- [ ] CS / Copyright / ...

PR #290, opened 2025-08-08
[MCP SDK] Handle closed connection in `StreamTransport` to avoid stuck worker

### LINKED ISSUE #283: [MCP SDK] StreamTransport connection not properly detected as closed, causing infinite worker loops
Hi!

The `StreamTransport` implementation in the MCP SDK seems to have an issue where client disconnections are not properly detected, causing PHP workers to potentially run indefinitely and exhausting FPM worker limits.

The issue seems to comes from how PHP's `connection_aborted()` function works with TCP connections. In the current implementation:

```php
// StreamTransport.php
public function isConnected(): bool
{
    return 0 === connection_aborted();
}
```

```php
// Server.php  
while ($transport->isConnected()) {
    // ... processing loop
    usleep(1000);
}
```

The core issue: `connection_aborted()` returns `0` (connection alive) until PHP attempts to flush data to the client and detects the disconnection. With Server-Sent Events, if no events are being sent, PHP never attempts to write to the connection and therefore may never discovers that the client has disconnected (there are multiple StackOverflow issues discussing about this).

### Steps to Reproduce

1. Clone the [demo repository](https://github.com/Gregory-Gerard/symfony-ai-demo-infinite-sse-reproduction) with SSE enabled (vendor directory included with modified `symfony/mcp-sdk` for logging demonstration)
2. Start server: `symfony serve`
3. Connect with: `curl 'http://localhost:8000/_mcp/sse'`
4. Terminate the curl connection (Ctrl/Cmd+C)
5. Server continues logging "Transport is still connected" indefinitely

The demo repository includes a modified `vendor/symfony/mcp-sdk/src/Server.php` with added logging on line 33 to demonstrate the infinite loop behavior when connections are not properly detected as closed. We can also observe FPM workers not closing if needed (even with `pm.max_spare_servers = 1`.

### Potentiel solutions

With both solution we should make `StreamTransport` aware of last ping time, so we may need to remove the `readonly` on this class. Is it okay?

#### Implement MCP Ping

The [MCP specification includes a ping utility](https://modelcontextprotocol.io/specification/2025-03-26/basic/utilities/ping), that seems specifically designed for connection health checks:

```php
// In StreamTransport, add periodic ping functionality
private function sendPing(): void
{
    $ping = json_encode([
        'jsonrpc' => '2.0',
        'id' => uniqid(),
        'method' => 'ping'
    ]);
    
    $this->flushEvent('message', $ping);
}

public function isConnected(): bool
{
    // Send ping periodically (e.g., every 30 seconds)
    if ($this->shouldSendPing()) {
        $this->sendPing();
    }
    
    return 0 === connection_aborted();
}
```

When the client is disconnected, the ping packet won't be acknowledged, and PHP will detect the broken connection.

#### Simple periodic heartbeat characters

Send a minimal heartbeat character periodically (similar to what Figma's MCP server does):

```php
public function isConnected(): bool
{
    if ($this->shouldSendHeartbeat()) {
        $this->flushEvent('heartbeat', ' '); // Send minimal data
    }
    
    return 0 === connection_aborted();
}
```

Example from Figma's MCP server, every 30 seconds it send a `:`:

Image

---

I'm happy to contribute a PR implementing the preferred solution, or any other better solution. Please let me know which approach aligns best.

Gregory-Gerard:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| Docs?         | no
| Issues        | Fix #283
| License       | MIT

WIP: I'll update this description with full details as I progress 👍

### DIFF
--- a/src/mcp-sdk/src/Server.php
+++ b/src/mcp-sdk/src/Server.php
@@ -13,21 +13,41 @@ namespace Symfony\AI\McpSdk;
 
 use Psr\Log\LoggerInterface;
 use Psr\Log\NullLogger;
+use Symfony\AI\McpSdk\Exception\TransportNotConnectedException;
+use Symfony\AI\McpSdk\Message\Error;
+use Symfony\AI\McpSdk\Message\Request;
+use Symfony\AI\McpSdk\Message\Response;
 use Symfony\AI\McpSdk\Server\JsonRpcHandler;
+use Symfony\AI\McpSdk\Server\KeepAliveSessionInterface;
+use Symfony\AI\McpSdk\Server\PendingResponse;
+use Symfony\AI\McpSdk\Server\PendingResponseBag;
 use Symfony\AI\McpSdk\Server\TransportInterface;
+use Symfony\Component\Clock\ClockInterface;
+use Symfony\Component\Uid\Uuid;
 
-final readonly class Server
+final class Server
 {
+    private ?TransportInterface $transport = null;
+
+    private PendingResponseBag $pendingResponses;
+
     public function __construct(
         private JsonRpcHandler $jsonRpcHandler,
+        private ClockInterface $clock,
+        private ?KeepAliveSessionInterface $keepAliveSession = null,
         private LoggerInterface $logger = new NullLogger(),
     ) {
+        $this->pendingResponses = new PendingResponseBag($clock, new \DateInterval('PT30S'));
     }
 
     public function connect(TransportInterface $transport): void
     {
+        $this->transport = $transport;
+
         $transport->initialize();
         $this->logger->info('Transport initialized');
+
+        $this->keepAliveSession?->start();
 
         while ($transport->isConnected()) {
             foreach ($transport->receive() as $message) {
@@ -38,6 +58,15 @@ final readonly class Server
                 try {
                     foreach ($this->jsonRpcHandler->process($message) as $response) {
                         if (null === $response) {
+                            continue;
+                        }
+
+                        if ($response instanceof Response || $response instanceof Error) {
+                            if ($this->pendingResponses->resolve($response)) {
+                                continue;
+                            }
+
+                            $this->logger->warning(\sprintf('No handler found for response id "%s".', $response->id), ['response' => $response]);
                             continue;
                         }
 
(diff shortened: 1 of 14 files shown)


OskarStark:
Hmm one is called runtime, the other one transport, having a runtime interface on a transport feels weird 😐 any ideas?

Gregory-Gerard:
Yes it feels weird to me too. But I don't have other ideas for now for testing `StreamTransport` and mocking `connection_aborted` for example. 

I could override `connection_aborted` (I found an example in Slim framework) in `StreamTransport` in my test, but I don't really like this idea too.

If someone have guidance, I'll happily modify this!

Gregory-Gerard:
@OskarStark would you be okay if I try the namespaced function override approach just for `connection_aborted` via a test-only polyfill? We can keep flush/write as-is and assert via PHP’s output buffering. 

I couldn’t find other example testing this; even HttpFoundation’s `EventStreamResponse` doesn’t check specifically `connection_aborted`. 

This would avoids the `Runtime` dependency and any constructor changes. If you’re fine with it, I’ll add a dev-only polyfill that shadows `Symfony\AI\McpSdk\Server\Transport\Sse\connection_aborted()` in tests.

OskarStark:
I would say give it a try, why not

Gregory-Gerard:
Re @OskarStark! I implemented the namespaced override only for `connection_aborted`. `StreamTransport` keeps native calls; tests assert via output capture; and no new `RuntimeInterface`. If you’re good with this, I’ll proceed with the concrete SSE ping/heartbeat fix.

src/mcp-sdk/tests/Server/Transport/Sse/StreamTransportTest.php
```diff
@@ -0,0 +1,130 @@
...
+        $this->assertInstanceOf(\Generator::class, $generator);
+        $this->assertSame('hello', $generator->current());
+        $generator->next();
+        $this->assertFalse($generator->valid());
+    }
+
+    public function testSendEmitsMessageEventAndFlushes()
+    {
+        $store = $this->createMock(StoreInterface::class);
+        $id = Uuid::v7();
```
OskarStark: Let's inline one time variables

OskarStark:
Yes why not 🙌🙂

src/mcp-sdk/tests/Server/Transport/Sse/StreamTransportTest.php
Gregory-Gerard: Done!

Gregory-Gerard:
Hi @OskarStark @chr-hertel (friendly tag because you were tagged in the issue originally)!

Quick update:

- Still WIP. Next up: make ping interval and request timeouts configurable per MCP spec (https://modelcontextprotocol.io/specification/2025-06-18/basic/utilities/ping). I may introduce a small immutable ServerConfig injected into `Server` to avoid constructor bloat.
- Ping (or rather the timeout) is non-strict for now because some clients don't reply explicitly to pings. I'll add a "strict" mode (default off) so we can enforce spec behavior without breaking common clients (spec: https://modelcontextprotocol.io/specification/2025-06-18/basic/lifecycle#timeouts, e.g. issue with a strict timeout https://github.com/punkpeye/fastmcp/issues/38).
- Major changes so far:
  - Added `PendingResponse`/`PendingResponseBag` to track outgoing requests, resolve inbound responses/errors, and GC on timeout. This currently requires `JsonRpcHandler` to yield `Response|Error`. Longer-term I'd prefer splitting JSON encode/decode from processing handlers as proposed in PR #240, but seems out of scope here, but happy to adjust later if you prefer.
  - Added `KeepAliveSessionInterface` + `KeepAliveSession` to automate periodic pings.
  - `Server::sendRequest(...)` with callback and timeout handling is in; next I'll add per-request timeout overrides and emit a `cancel` notification to the client on timeout (which implies adding `sendNotification`), still per the spec.
  - Injected a `ClockInterface` to make testing easier and remove a direct dependency on system clock with `usleep`
- If everything is okay, I will of course update the `mcp-bundle` with new configuration, docs, etc.
- Also, last thing but not least, we may need to hardcode the `ping` capability in `RequestHandlers` (because most implementation does), and client/server **SHOULD** always respond to a `ping`, any ideas about this?

WDYT about my directions? Happy to have feedback before polishing thing!

src/mcp-sdk/src/Server.php
```diff
@@ -13,22 +13,42 @@
...
-final readonly class Server
+final class Server
 {
+    private ?TransportInterface $transport = null;
+
+    private PendingResponseBag $pendingResponses;
+
     public function __construct(
         private JsonRpcHandler $jsonRpcHandler,
+        private ClockInterface $clock,
```
Gregory-Gerard, line 36: Thinking about adding a default `Clock` like in Symfony Messenger, wdyt?

chr-hertel:

PR #292, opened 2025-08-08
[Platform] Add ElevenLabs as platform

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | #263 
| License       | MIT

Hi 👋🏻 

As discussed in #263, `ElevenLabs` can be more than just a tool for agents, the idea of a platform was discussed so here it is.

PS: Both TTS and STT examples are working "as expected" (the TTS one generate a file as we can't run the file).


examples/.env
```diff
@@ -43,6 +43,10 @@ HUGGINGFACE_KEY=
 # For using OpenRouter
 OPENROUTER_KEY=
 
+# For using ElevenLabs
+ELEVEN_LABS_URL=https://api.elevenlabs.io/v1
```
chr-hertel, line 47: is there another URL that would be usable?
we don't need to promote something to configuration level if the value always is the same

src/ai-bundle/doc/index.rst
```diff
@@ -85,6 +89,12 @@ Configuration
                 tools: # If undefined, all tools are injected into the agent, use "tools: false" to disable tools.
                     - 'Symfony\AI\Agent\Toolbox\Tool\Wikipedia'
                 fault_tolerant_toolbox: false # Disables fault tolerant toolbox, default is true
+            audio:
+                platform: 'ai.platform.eleven_labs'
+                model:
+                    class: 'Symfony\AI\Platform\Bridge\ElevenLabs'
+                    name: !php/const Symfony\AI\Platform\Bridge\ElevenLabs\TEXT_TO_SPEECH
```
chr-hertel: this looks wrong, doesn't it?

src/platform/src/Bridge/ElevenLabs/ElevenLabs.php
```diff
@@ -0,0 +1,31 @@
...
+
+use Symfony\AI\Platform\Model;
+
+/**
+ * @author Guillaume Loulier <personal@guillaumeloulier.fr>
+ */
+final class ElevenLabs extends Model
+{
+    public const TEXT_TO_SPEECH = 'text-to-speech';
+    public const SPEECH_TO_TEXT = 'speech-to-text';
```
chr-hertel: this is more like a task or capability, if i get it right, that's the list of models they provide:
* Eleven v3 (alpha)
* Eleven Multilingual v2
* Eleven Flash v2.5
* Eleven Turbo v2.5
* Eleven Turbo v2
* Eleven Flash v2
* Eleven English v1
* Eleven Multilingual v1

src/platform/src/Bridge/ElevenLabs/ElevenLabsClient.php
```diff
@@ -0,0 +1,104 @@
...
+
+/**
+ * @author Guillaume Loulier <personal@guillaumeloulier.fr>
+ */
+final readonly class ElevenLabsClient implements ModelClientInterface
+{
+    public function __construct(
+        private HttpClientInterface $httpClient,
+        private string $hostUrl,
+        private string $apiKey,
```
chr-hertel: needs a
```suggestion
        #[\SensitiveParameter] private string $apiKey,
```

src/platform/src/Bridge/ElevenLabs/ElevenLabsResultConverter.php
```diff
@@ -0,0 +1,70 @@
...
+    {
+        $payload = $result->getObject()->getContent();
+
+        $path = \sprintf('%s/%s.mp3', $this->outputPath, uniqid());
+
+        $filesystem = new Filesystem();
+        $filesystem->dumpFile($path, $payload);
+
+        return new AudioResult($path, (new MimeTypes())->guessMimeType($path));
+    }
```
chr-hertel: let's not mix in file system as a concern here, but just return a binary object with `BinaryResult` and don't we know the mime type of the response? at least you chose `.mp3` as file type already

src/platform/src/Contract/Normalizer/Message/Content/AudioNormalizer.php
```diff
@@ -45,6 +45,7 @@ public function normalize(mixed $data, ?string $format = null, array $context =
             'type' => 'input_audio',
             'input_audio' => [
                 'data' => $data->asBase64(),
+                'path' => $data->asPath(),
```
chr-hertel: does that have impact to other audio examples, e.g. `examples/openai/audio-input.php`? maybe we need to isolate that in an elevenlabs specific contract

chr-hertel (changes requested):
Was a bit in a hurry but left some comments - definitely looking forward to play around with it - thanks for working on it! :pray:

examples/.env
Guikingone: Agree

src/ai-bundle/doc/index.rst
Guikingone: Yes, the problem is between my computer and my chair 😄

src/platform/src/Contract/Normalizer/Message/Content/AudioNormalizer.php
Guikingone: Well, open to debate on this one, the `$path` attribut is always defined (I agree on the fact that it can be null) thanks to `File` so IMHO, using the method is not a problem (when correctly typed), returning an array with a value "null" in certain cases is not a problem as long as the API is typed to do so.

src/platform/src/Bridge/ElevenLabs/ElevenLabsResultConverter.php
Guikingone: Fixed, nice catch 👍🏻

examples/.env
chr-hertel: Then remove it please :pray:

chr-hertel: I'm confused - is there another host possible?

src/platform/composer.json
```diff
@@ -40,11 +40,13 @@
...
         "phpstan/phpstan-symfony": "^2.0.6",
         "phpunit/phpunit": "^11.5",
+        "symfony/ai-agent": "@dev",
         "symfony/console": "^6.4 || ^7.1",
         "symfony/dotenv": "^6.4 || ^7.1",
-        "symfony/ai-agent": "@dev",
         "symfony/event-dispatcher": "^6.4 || ^7.1",
+        "symfony/filesystem": "^7.3",
         "symfony/finder": "^6.4 || ^7.1",
+        "symfony/mime": "^7.3",
```
chr-hertel: we don't need this anymore, right?

src/platform/src/Contract/Normalizer/Message/Content/AudioNormalizer.php
chr-hertel: It breaks the example `openai/audio-input.php` - please revert and bring in another Normalizer specifically for ElevenLabs

examples/.gitignore
```diff
@@ -2,3 +2,4 @@
 .transformers-cache
 composer.lock
 vendor
+tmp
```
OskarStark, line 5: instead of this, what about using `sys_get_temp_dir` as default?

examples/elevenlabs/speech-to-text.php
```diff
@@ -0,0 +1,28 @@
...
+
+$platform = PlatformFactory::create(
+    apiKey: env('ELEVEN_LABS_API_KEY'),
+    outputPath: __DIR__.'/tmp',
+    httpClient: http_client()
+);
+$model = new ElevenLabs(ElevenLabs::SCRIBE_V1);
+$file = Audio::fromFile(dirname(__DIR__, 2).'/fixtures/audio.mp3');
+
+$result = $platform->invoke($model, $file);
```
OskarStark:
```suggestion

$result = $platform->invoke($model, Audio::fromFile(dirname(__DIR__, 2).'/fixtures/audio.mp3'));
```

src/ai-bundle/config/options.php
```diff
@@ -40,6 +40,13 @@
                             ->end()
                         ->end()
                     ->end()
+                    ->arrayNode('eleven_labs')
+                        ->children()
+                            ->scalarNode('host')->end()
+                            ->scalarNode('api_key')->isRequired()->end()
+                            ->scalarNode('output_path')->isRequired()->end()
```
OskarStark: what about having a sensible default here using kernel.project_dir/var ?

examples/.gitignore
Guikingone: Well, thing is, this directory is only used for examples, if we want the developer who's using the examples to access the file, the `sys_get_temp_dir` is probably not the best fit 🤔

src/ai-bundle/config/options.php
Guikingone: What about `%kernel.project_dir%/var/eleven_labs`?

examples/elevenlabs/text-to-speech.php
```diff
@@ -0,0 +1,29 @@
...
+    outputPath: dirname(__DIR__).'/tmp',
+    httpClient: http_client(),
+);
+$model = new ElevenLabs(options: [
+    'voice' => 'Dslrhjl3ZpzrctukrQSN', // Brad (https://elevenlabs.io/app/voice-library?voiceId=Dslrhjl3ZpzrctukrQSN)
+]);
+
+$result = $platform->invoke($model, new Text('Hello world'));
+
+echo $result->asBase64().\PHP_EOL;
```
chr-hertel: what do you think about enabling this?
```bash
php elevenlabs/text-to-speech.php | mpg123 -
```
that works for me to easily run the example, but requires this:
```suggestion
echo $result->asBinary();
```

(didn't figure out a way to play it from php directly)

Guikingone: Fixed, nice catch 🙂

Guikingone:
BTW, this PR does not include _all_ the features exposed by EL, I will probably bring more of them once its merged, the STT and TTS are enough for now 😅

src/platform/src/Bridge/ElevenLabs/ElevenLabsResultConverter.php
```diff
@@ -0,0 +1,65 @@
...
+
+    public function convert(RawResultInterface $result, array $options = []): ResultInterface
+    {
+        /** @var ResponseInterface $response */
+        $response = $result->getObject();
+
+        return match (true) {
+            str_contains($response->getInfo('url'), 'speech-to-text') => new TextResult($result->getData()['text']),
+            str_contains($response->getInfo('url'), 'text-to-speech') => $this->doConvertTextToSpeech($result),
```
chr-hertel: let's skip persisting the file please:

```suggestion
            str_contains($response->getInfo('url'), 'text-to-speech') => new BinaryResult($result->getObject()->getContent(), 'audio/mpeg'),
```

my reasoning:
* we return the binary data and developer can do with it what they want
* from `ResultConverterInterface` point of view, writing on the file system is an unexpected side effect
* if someone would use a file system abstraction we would mess with that or would need to take it into account

chr-hertel (changes requested):
only getting started myself to play around with elevenlabs, but we're almost good to merge as a start.

chr-hertel:
Thank you @Guikingone.

PR #293, opened 2025-08-08
[AIBundle] Cache store configuration improved

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | yes
| Issues        | #257 
| License       | MIT

Hi 👋🏻 

The configuration wasn't complete enough to handle custom strategy / key for the cache store, here's the fix, tests are validated (both in the project and in a custom project too).


src/ai-bundle/src/AiBundle.php
```diff
@@ -494,8 +496,19 @@ private function processStoreConfig(string $type, array $stores, ContainerBuilde
             foreach ($stores as $name => $store) {
                 $arguments = [
                     new Reference($store['service']),
+                    new Definition(DistanceCalculator::class),
+                    $store['key'] ?? '_vectors',
```
OskarStark: Shouldn't this key default come from the configuration or not been set if the default is already set via constructor?

Guikingone: Hum, yes, by default, it's set to `_vectors` so the usage of the key is optional by design, I'll take a look to remove it.

src/ai-bundle/doc/index.rst
```diff
@@ -91,6 +91,14 @@ Configuration
...
                 default:
                     collection: 'my_collection'
+            cache:
+                ollama:
+                    cache:
+                        service: 'cache.app'
+                        cache_key: 'ollama'
+                        strategy: 'chebyshev'
+                    memory:
+                        strategy: 'chebyshev'
```
chr-hertel: why is there an `ollama` level in-between - isn't it like this?
```suggestion
            cache:
                cache:
                    service: 'cache.app'
                    cache_key: 'ollama'
                    strategy: 'chebyshev'
                memory:
                    strategy: 'chebyshev'
```

Guikingone: Nope, stores accept arrays, the doc is wrongly built but that's my fault, I'm pushing the fix.

Guikingone: Improved

src/ai-bundle/doc/index.rst
```diff
@@ -91,6 +91,15 @@ Configuration
...
                     collection: 'my_collection'
+            cache:
+                cache:
+                    research:
+                        service: 'cache.app'
+                        cache_key: 'research'
+                        strategy: 'chebyshev'
+                memory:
+                    ollama:
+                        strategy: 'chebyshev'
```
chr-hertel: if i get it right, this should be:
```suggestion
            cache:
                research:
                    service: 'cache.app'
                    cache_key: 'research'
                    strategy: 'chebyshev'
            memory:
                in_memory:
                    strategy: 'chebyshev'
```

Guikingone: Of course, I lost myself in the spaces 😄

OskarStark (approved):

src/store/src/CacheStore.php
```diff
@@ -32,6 +32,10 @@ public function __construct(
         if (!interface_exists(CacheItemPoolInterface::class)) {
             throw new RuntimeException('For using the CacheStore as vector store, a PSR-6 cache implementation is required. Try running "composer require symfony/cache" or another PSR-6 compatible cache.');
         }
+
+        if (!interface_exists(CacheInterface::class)) {
+            throw new RuntimeException('For using the CacheStore as vector store, a symfony/contracts cache implementation is required. Try running "composer require symfony/cache" or another symfony/contracts compatible cache.');
+        }
```
chr-hertel: Why is that check needed as well?

Guikingone: Because we're using `$this->cache->get($this->cacheKey, static fn (): array => []);` and the method is not present in the PSR interface.

chr-hertel: it just looks unintuitive to me to require both - can we limit to one of the interfaces?

Guikingone: We can limit to `symfony/cache`, I personaly don't see any interest of using the PSR interface without a wrapper like Sf / other.

src/store/src/CacheStore.php
chr-hertel, line 1: Let's remove the `CacheItemPoolInterface` then here completely :+1:

chr-hertel:
Needs a rebase after #318

src/store/src/CacheStore.php
Guikingone: Hum nope, it's required for the static analysis, without the `CacheItemPoolInterface`, PHPStan isn't capable of resolving the methods from CIP.

Guikingone: The only way to allows it to resolve it is by adding: 

```php
private CacheInterface&CacheItemPoolInterface $cache,
```

chr-hertel:
Thank you @Guikingone.

Issue #272, opened 2025-08-06
[Platform] Support Google Vertext AI as a platform


aszenz:
Allow using gemini models from google cloud, gemini models are accessible via Google Vertex AI which is a similar kind of service to AWS bedrock

https://cloud.google.com/vertex-ai

We could use the google cloud php ai sdk to implement it as it supports credential chain discovery, similar to async bedrock runtime client, see:

https://github.com/googleapis/google-cloud-php-ai-platform

Example rawPredict request:

```php
$predictionServiceClient = new Google\Cloud\AIPlatform\V1\Client\PredictionServiceClient();
$modelUrl = \sprintf(
    'projects/%s/locations/%s/%s',
    $this->googleCloudProject,
    self::GOOGLE_CLOUD_LOCATION,
    $modelId,
);
$response = $predictionServiceClient->rawPredict(
    request: RawPredictRequest::build(
        endpoint: $modelUrl,
        httpBody: new HttpBody(
            [
                'data' => \json_encode($payload)
            ]
        )
    )
);
```

See preview relevant issue: #146

chr-hertel:
Sounds good, PR welcome :)

junaidbinfarooq:
Now that https://github.com/symfony/ai/pull/297 is merged, I guess this can be closed.
WDYT @chr-hertel @OskarStark?

Issue #146, opened 2025-07-17
Normalize Gemini VS Google


lyrixx:
ATM, the code base use `Gemini` or `Google` in class name, namespace, or configuration key.

It could be confusing I guess. 
Anyway it's not predictable.

I think I would be better to stick to only one name (maybe Gemini).

WDYT?

chr-hertel:
Man that's chaotic - but I have a hard time to get google's terminology straight here.

In [Google's AI Studio](https://aistudio.google.com/) you can get an API Key, what they call `GEMINI_API_KEY` in their examples, and they also call it Gemini API. But on the other hand Gemini is only one model family of Google, and there is Gemma or text embedding models as well. And Vertex AI. Oof ... 😆

chr-hertel:
Okay, maybe we should have two bridges:
* Vertex -> rather the Google Cloud product for hosting all kinds of models
* Gemini -> access to Gemini API with that subset of models

@valtzu, you've been playing around with Google/Gemini mostly, right? What makes sense from your point of view?

valtzu:
I think it's the simplest for us if we keep the bridges separate like you @chr-hertel suggested – because they are completely separate in Google's docs too and even though [Gemini is available via Vertex](https://cloud.google.com/vertex-ai/generative-ai/docs/start/quickstart?usertype=apikey#rest), the request schema seemed slightly different _(unless it's just outdated docs, getting plenty of 404 there 🙂)_.

I would prefer to have both (`Google` & `Gemini`/`Vertex`) in the fully-qualified class name and even configuration key, however it does not seem the current convention, f.e. `Bedrock` does not mention `Aws`/`Amazon` in the name.

I see @OskarStark already pushed a commit to change the current `Google` to `Gemini`, it makes sense to me, at least as long as we don't have Vertex implemented.

chr-hertel:
Meaning for the current bridge to be renamed from `Google` to `Gemini` 👍

developkariyer:
Since Google does not provide an official client for PHP, following the Python path might be more compatible to what Google aims - with the assumption that they will give priority to updating their own Python libraries to follow API changes.

With this in mind, current status is, there are two distinct libraries, which have completely different usage methodology:

google.genai: Used with API keys. Not all models are available in this API.
vertexai: Provides more models. Used with oauth2 authentication and I guess requires some google tools installed on server, at least it is what I did, in order not to fight with the documentation and do whatever google says to make things work.

[...]

PR #297, opened 2025-08-10
[Platform] Add support for Google vertex AI

junaidbinfarooq:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | #16 #272 
| License       | MIT

# Changes proposed:
- The PR aims to add support for Google's Vertex AI
- Support is added for text generation, embeddings, server tools, etc
- I created a separate directory inside the **Bridge** directory for the said bridge. Moving forward, we could either create a new namespace named **Google** and add two namespaces inside it, one each for **Gemini** and **VertexAi**, or we could simply keep **VertexAi** and remove the other, as this api already supports **Gemini** and more. Either way, #146 would be addressed.
- I exempted the use of api key from the implementation so far as **VertexAi** supports two auth mechanisms, Api keys and ADC, ie, Application Default Credentials. The former is recommended for testing while the latter is recommended for production, IIRC. However, it does need to install gCloud on the server and locally too (if verifying the behavior without an api key). The Google auth library is also needed for automatic credential discovery.
- More information can be found at https://cloud.google.com/vertex-ai/generative-ai/docs


junaidbinfarooq:
@chr-hertel @OskarStark 
While I am working on this, your feedback could prove useful and perhaps also steer the development toward a better direction.

src/platform/composer.json
```diff
@@ -22,6 +22,7 @@
     "require": {
         "php": ">=8.2",
         "ext-fileinfo": "*",
+        "google/auth": "^1.47",
```
chr-hertel: Please move it to `require-dev`  and add a check for the class existence to the implementation, like in the Bedrock one for example:
https://github.com/symfony/ai/blob/11ef98f74e281f549f250c39e659838b8cf16142/src/platform/src/Bridge/Bedrock/PlatformFactory.php#L37-L39

src/platform/src/Bridge/VertexAi/PlatformFactory.php
```diff
@@ -0,0 +1,49 @@
...
+final readonly class PlatformFactory
+{
+    public static function create(
+        string $location,
+        string $projectId,
+        ?HttpClientInterface $httpClient = null,
+        ?Contract $contract = null,
+    ): Platform {
+        if (!class_exists(ApplicationDefaultCredentials::class)) {
+            throw new RuntimeException('For using the Bedrock platform, the google/auth package is required. Try running "composer require google/auth".');
```
aszenz:
```suggestion
            throw new RuntimeException('For using the Vertex AI platform, google/auth package is required for authentication via application default credentials. Try running "composer require google/auth".');
```

faizanakram99 (approved):
lgtm, ship it

src/platform/src/Bridge/VertexAi/PlatformFactory.php
junaidbinfarooq: I did it already.
Pushing the change in a few moments.

junaidbinfarooq:
> lgtm, ship it

Docs and AI bundle integration need to be added first.

src/platform/src/Bridge/VertexAi/Contract/AssistantMessageNormalizer.php
```diff
@@ -0,0 +1,67 @@
...
+     *
+     * @return array{
+     *     array{
+     *         text: string,
+     *         functionCall?: array{
+     *             name: string,
+     *             args?: array<int|string, mixed>
+     *         }
+     *      }
+     *     }
```
OskarStark: The indention looks wrong

examples/vertexai/bootstrap.php
```diff
@@ -0,0 +1,31 @@
...
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use Google\Auth\ApplicationDefaultCredentials;
```
chr-hertel, line 12: this needs a new require in the `examples/composer.json` file - cannot run the examples locally at this point, sorry

chr-hertel (changes requested):

src/ai-bundle/doc/index.rst
```diff
@@ -52,6 +52,9 @@ Configuration
                     api_version: '%env(AZURE_GPT_VERSION)%'
             gemini:
                 api_key: '%env(GEMINI_API_KEY)%'
+            vertexAi:
```
chr-hertel: would be 
```suggestion
            vertex_ai:
```

src/platform/doc/vertexai-server-tools.rst
chr-hertel, line 1: Instead of a similar doc to the gemini one, please combine the two files a la Server Tools - supported by VertexAI and Gemini API or sth

src/ai-bundle/doc/index.rst
```diff
@@ -34,7 +34,7 @@ Configuration
                     class: 'Symfony\AI\Platform\Bridge\OpenAi\Gpt'
                     name: !php/const Symfony\AI\Platform\Bridge\OpenAi\Gpt::GPT_4O_MINI
 
-**Advanced Example with Anthropic, Azure, Gemini and multiple agents**
+**Advanced Example with Anthropic, Azure, Gemini, VertexAi and multiple agents**
```
chr-hertel: the official product name is "Vertex AI" i guess
```suggestion
**Advanced Example with Anthropic, Azure, Gemini, Vertex AI and multiple agents**
```

src/platform/doc/index.rst
```diff
@@ -73,13 +73,15 @@ usually defined by the specific models and their documentation.
     * `OpenAI's GPT`_ with `OpenAI`_ and `Azure`_ as Platform
     * `Anthropic's Claude`_ with `Anthropic`_ and `AWS Bedrock`_ as Platform
     * `Meta's Llama`_ with `Azure`_, `Ollama`_, `Replicate`_ and `AWS Bedrock`_ as Platform
-    * `Gemini`_ with `Google`_ and `OpenRouter`_ as Platform
+    * `Gemini`_ with `Google`_, `VertexAi`_ and `OpenRouter`_ as Platform
```
chr-hertel:
```suggestion
    * `Gemini`_ with `Google`_, `Vertex AI`_ and `OpenRouter`_ as Platform
```

chr-hertel:
Thanks @junaidbinfarooq for your work on this - I think we're getting somewhere.
Please see my comments and review once again yourself with those in mind - not sure I catched everything.

In general it great to have that bridge, I still wonder tho about the amount of duplicated code with the Gemini bridge and we would need to support API Keys as well ... but both topics are fine as follow up

src/platform/src/Bridge/VertexAi/Contract/AssistantMessageNormalizer.php
junaidbinfarooq: In the PHP doc, the IDE doesn't apparently help with the indentation in this case.
I tried to correct the indentation now.

examples/vertexai/bootstrap.php
junaidbinfarooq: Not a biggie.
I did think about it, but somehow it skipped my mind, probably because the examples worked for me as I included the autoload file in another way.

src/ai-bundle/doc/index.rst
junaidbinfarooq: Wouldn't it require a change inside `AiBundle.php` too? While integrating the bridge into the AI bundle, I looked for other bridges that had multiple words in them and found that they were added as a single word. And I followed suit.
I did, however, change it to _vertexai_.
Given your input on it, however, I'd be happy to change it to _vertex_ai_.

src/platform/doc/vertexai-server-tools.rst
junaidbinfarooq: I did this on purpose so that when we later make a decision about whether we need both the bridges or only Vertex AI, then we could make such a refactor.

junaidbinfarooq:
> Thanks @junaidbinfarooq for your work on this - I think we're getting somewhere. Please see my comments and review once again yourself with those in mind - not sure I catched everything.
> 
> In general it great to have that bridge, I still wonder tho about the amount of duplicated code with the Gemini bridge and we would need to support API Keys as well ... but both topics are fine as follow up

@chr-hertel 
You're welcome :smile:
The duplication is expected as I did not touch the existing Gemini bridge code.
This is what I also pointed out in the PR description about possibly having Vertex AI only as we move forward.
If this is done, there would be no duplication anywhere as far as Google bridges are concerned.
And yes, I do believe this is better done in a follow-up PR after a proper discussion on the topic.

src/platform/doc/vertexai-server-tools.rst
chr-hertel: For me removing the support for Gemini API as inference platform is not on the table - only if google would sunset the product.

However, i don't have a good solution on mind in general for this situation - I guess having a dependency from vertex bridge to gemini bridge when you want to use gemini with vertex would be the easiest as a start i guess.

junaidbinfarooq: @chr-hertel 
That might not help the case. Accessing Gemini via Vertex AI is still different from accessing Gemini directly, as far as I could see. There are some subtle differences in how requests and receive data, so merging them or making one dependent on the other may result in unexpected complexity.
We could instead keep both as sub-namespaces under a namespace named _Google_.
On another thought, we recently integrated Google Gemini AI into our software to evaluate it, but there is a possibility that we might migrate to accessing it via Vertex AI

chr-hertel:
@junaidbinfarooq are all examples that you added working for you?

i need to apply those changes, any clue why?
image

junaidbinfarooq:
> @junaidbinfarooq are all examples that you added working for you?
> 
> i need to apply those changes, any clue why? image

@chr-hertel 
I just rechecked the vertex ai api and also ran the examples again to verify they are still working as expected. And it is apparent that they are working:
image
Besides, I didn't use versions in model IDs as that is not recommended, IIRC.

chr-hertel:
Hmm, super weird :thinking: 

And where is that `$location-aiplatform.googleapi.com/...` coming from?
Docs say it's just `aiplatform` as subdomain:
image
See https://console.cloud.google.com/vertex-ai/publishers/google/model-garden/gemini-2.5-flash-lite

junaidbinfarooq:
@chr-hertel 
It is actually weird considering I can see both the URLs in the documentation, one that you pointed out and another that I used. For instance:
1. If you head over to [Api quickstart](https://cloud.google.com/vertex-ai/generative-ai/docs/start/quickstart?usertype=adc#make-first-request), you will see the URL that is used in the code. `API_ENDPOINT` basically refers to `{REGION}-aiplatform.googleapis.com`. I used the latter because it is easier for library users to set the location instead of the entire base URL.
2. In the api reference, which is what I mostly used, admittedly (or surprisingly), the page for the `generateContent` endpoint doesn't show the URL to access the api, but the one for Embeddings does. And it is exactly what I used. Please go to [this](https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/text-embeddings-api#syntax) for more details.
3. There are also other pages that have variations of the same URL.

P.S. I made a change now to remove the location from the start of the request URL for both text and embedding generation after verifying it works. It matches the URL that you referenced, but the location is still there in the middle.

chr-hertel:
and also with that endpoint change the examples `vertexai/token-metadata.php` and `vertexai/toolcall.php` raise errors for me:
```
PHP Fatal error:  Uncaught Symfony\AI\Platform\Exception\RuntimeException: Error from Gemini API: "Publisher Model `projects/XXXXXXXXXXX/locations/global/publishers/google/models/gemini-2.0-flash-lite` not found." in /path/to/repo/src/platform/src/Bridge/VertexAi/Gemini/ResultConverter.php:48
Stack trace:
#0 /path/to/repo/src/platform/src/Result/ResultPromise.php(48): Symfony\AI\Platform\Bridge\VertexAi\Gemini\ResultConverter->convert()
#1 /path/to/repo/src/platform/src/Result/ResultPromise.php(37): Symfony\AI\Platform\Result\ResultPromise->await()
#2 /path/to/repo/src/agent/src/Agent.php(82): Symfony\AI\Platform\Result\ResultPromise->getResult()
#3 /path/to/repo/examples/vertexai/token-metadata.php(28): Symfony\AI\Agent\Agent->call()
#4 {main}
  thrown in /path/to/repo/src/platform/src/Bridge/VertexAi/Gemini/ResultConverter.php on line 48
```
---
Btw, code-wise we're good here, thanks for that! I just want to be able to run all examples locally to make sure all integrations are still working. And I'm not super deep into google cloud. is there something special that you configured in vertex? global region? any models or endpoints in action here?

junaidbinfarooq:
> and also with that endpoint change the examples `vertexai/token-metadata.php` and `vertexai/toolcall.php` raise errors for me:
> 
> ```
> PHP Fatal error:  Uncaught Symfony\AI\Platform\Exception\RuntimeException: Error from Gemini API: "Publisher Model `projects/XXXXXXXXXXX/locations/global/publishers/google/models/gemini-2.0-flash-lite` not found." in /path/to/repo/src/platform/src/Bridge/VertexAi/Gemini/ResultConverter.php:48
> Stack trace:
> #0 /path/to/repo/src/platform/src/Result/ResultPromise.php(48): Symfony\AI\Platform\Bridge\VertexAi\Gemini\ResultConverter->convert()
> #1 /path/to/repo/src/platform/src/Result/ResultPromise.php(37): Symfony\AI\Platform\Result\ResultPromise->await()
> #2 /path/to/repo/src/agent/src/Agent.php(82): Symfony\AI\Platform\Result\ResultPromise->getResult()
> #3 /path/to/repo/examples/vertexai/token-metadata.php(28): Symfony\AI\Agent\Agent->call()
> #4 {main}
>   thrown in /path/to/repo/src/platform/src/Bridge/VertexAi/Gemini/ResultConverter.php on line 48
> ```
> 
> Btw, code-wise we're good here, thanks for that! I just want to be able to run all examples locally to make sure all integrations are still working. And I'm not super deep into google cloud. is there something special that you configured in vertex? global region? any models or endpoints in action here?

@chr-hertel 
All the examples work for me as they did earlier:
image

## Changes I made:
### Vertex AI:
1. Followed the instructions as mentioned [here](https://cloud.google.com/vertex-ai/generative-ai/docs/start/gcp-auth)
2. Set the env var `GOOGLE_CLOUD_PROJECT` to the ID of the newly created project on GCloud, created specifically for this purpose (though an existing project should work well too with the right set of permissions/access, I think), and set the env var `GOOGLE_CLOUD_LOCATION` to _us-central1_.

### Locally:
1. Initialized the gcloud CLI as mentioned in the aforementioned linked page.
2. Have been using `platform/vendor/autoload.php` directly inside the `examples/bootstrap.php` file right from the start.

junaidbinfarooq:
@chr-hertel 
Could you let me know if you were able to run the examples or if something still needs to be done here?
cc @OskarStark

chr-hertel:
Code-wise all looks good to me - mostly depends on me to find the issue of testing. Not gonna happen this week tho, sorry.

src/platform/src/Bridge/VertexAi/TokenOutputProcessor.php
```diff
@@ -0,0 +1,87 @@
...
+     * @throws ServerExceptionInterface
+     * @throws RedirectionExceptionInterface
+     * @throws DecodingExceptionInterface
+     * @throws ClientExceptionInterface
+     */
+    public function processOutput(Output $output): void
+    {
+        $metadata = $output->result->getMetadata();
+
+        if ($output->result instanceof StreamResult) {
```
VincentLanglet, line 43: Cf https://github.com/symfony/ai/pull/347#discussion_r2297300546 I'm dunno if StreamResult should be handled here...

Maybe it's better to have the same behavior everywhere `// Streams have to be handled manually as the tokens are part of the streamed chunks` and to open an issue first to discuss about it ?

junaidbinfarooq: I left a comment there.

junaidbinfarooq:
> Code-wise all looks good to me - mostly depends on me to find the issue of testing. Not gonna happen this week tho, sorry.

@chr-hertel 
Any update on this?

junaidbinfarooq:
> Code-wise all looks good to me - mostly depends on me to find the issue of testing. Not gonna happen this week tho, sorry.

@chr-hertel 
Could you let me know if you managed to give it a shot?

OskarStark:
@chr-hertel is currently AFK for some days.

junaidbinfarooq:
> @chr-hertel is currently AFK for some days.

Oh ok.
Thanks for the update.

chr-hertel:
one thing i learned lately, that it fails due to the region - that was a good pointer of yours - so you could reproduce with switching to `global` - not sure how to deal with that tho - add a hint in docs?

junaidbinfarooq:
> one thing i learned lately, that it fails due to the region - that was a good pointer of yours - so you could reproduce with switching to `global` - not sure how to deal with that tho - add a hint in docs?

@chr-hertel 
Sure, why not!
I added a dedicated doc for Vertex AI that has this information, besides other information.

src/platform/doc/vertexai.rst
chr-hertel, line 1: :+1:

chr-hertel:
Thank you @junaidbinfarooq.

PR #301, opened 2025-08-11
Use child model classes as factories instead

### LINKED ISSUE #114: Custom URL ability or generic bridges?

JoshuaBehrens:
This commit bricks several instanceof checks. Other changes needs to be done as well. See the other pull requests:

* https://github.com/symfony/ai/pull/299
* https://github.com/symfony/ai/pull/300

| Q             | A
| ------------- | ---
| Bug fix?      | "yes"
| New feature?  | no
| Docs?         | no
| Issues        | Fix #114 
| License       | MIT

In other issues and pull requests the model class should be used as a simple struct, only holding scalar values and should not have any business logic effect by inheritance. This has been the case for embeddings, completions and other API calls like Whisper audio, and Dall-e images. Right now this makes it difficult to share different instances of the model class with various providers, although a lot of providers share the same models, and therefore likely the same capabilities. This eventually will reduce the amount of code you have to write to establish new providers that replicate existing API's with a well-known subset of models or add better support for custom models like OpenAI fine-tuned models or ollamas self-built models.

This is pull request is about to suggest a way to reflector the usage of the model classes to only use the model class as a struct without using inheritance. I stick to the previous classes to simplify code changes and the current effects of different types of models being grouped in a class. The class hierarchy has been broken up, so the previously sub classes of the model class are no longer extending the model class, and the model of the class has been made final to prevent changes like this in the future. Right now, it will break instanceof checks in several classes and can only be merged if the linked requests merged. It is not part of the others, as the approach how to solve this particular issue can be discussed separately.

### DIFF
--- a/src/platform/src/Bridge/OpenAi/Gpt.php
+++ b/src/platform/src/Bridge/OpenAi/Gpt.php
@@ -18,7 +18,7 @@ use Symfony\AI\Platform\Model;
  * @author Christopher Hertel <mail@christopher-hertel.de>
  * @author Oskar Stark <oskarstark@googlemail.com>
  */
-class Gpt extends Model
+final class Gpt
 {
     public const GPT_35_TURBO = 'gpt-3.5-turbo';
     public const GPT_35_TURBO_INSTRUCT = 'gpt-3.5-turbo-instruct';
@@ -62,10 +62,10 @@ class Gpt extends Model
     /**
      * @param array<mixed> $options The default options for the model usage
      */
-    public function __construct(
+    public static function create(
         string $name = self::GPT_4O,
         array $options = ['temperature' => 1.0],
-    ) {
+    ): Model {
         $capabilities = [
             Capability::INPUT_MESSAGES,
             Capability::OUTPUT_TEXT,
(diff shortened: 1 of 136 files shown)


src/platform/src/Bridge/OpenAi/Gpt.php
```diff
@@ -85,6 +85,6 @@ public function __construct(
             $capabilities[] = Capability::OUTPUT_STRUCTURED;
         }
 
-        parent::__construct($name, $capabilities, $options);
+        return new Model($name, $capabilities, $options);
```
VincentLanglet, line 88: When tracing/logging my ai usage I'm currently use the two following tags:
- The model, given by `$model->getName();`
- The provider, given by `new \ReflectionClass($model)->getShortName()`

This PR is breaking the way I "compute" the provider of the following model ; any suggestion for such need ?

OskarStark:
you may want to have a look at
* #401

chr-hertel:
Moving in a different direction here, see #640 - thanks for the proposal tho - let's see where we end up with :D

JoshuaBehrens:
Thank you for the link to the alternative :) I will have a look whether it helps me out on the overall topic

PR #303, opened 2025-08-11
[Platform][Anthropic] Fix TypeError in AssistantMessageNormalizer when toolCalls is null

### LINKED ISSUE #302: AssistantMessageNormalizer throws TypeError when toolCalls is null
The AssistantMessageNormalizer::normalize() method in vendor/symfony/ai-platform/src/Bridge/Anthropic/Contract/AssistantMessageNormalizer.php throws a TypeError when $data->toolCalls is null.

Error message
array_map(): Argument #2 ($array) must be of type array, null given

Root Cause
The current code uses $data->toolCalls ?? [] as a fallback, but this doesn't prevent the issue because array_map() is called regardless of whether toolCalls is null or not. The code assumes toolCalls will always be an array when it may be null.

Current Code (Problematic)
php'content' => array_map(static function (ToolCall $toolCall) {
    return [
        'type' => 'tool_use',
        'id' => $toolCall->id,
        'name' => $toolCall->name,
        'input' => [] !== $toolCall->arguments ? $toolCall->arguments : new \stdClass(),
    ];
}, $data->toolCalls ?? []),

Muneer-Shafi:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #302
| License       | MIT

## Summary

Fixes a `TypeError` in `AssistantMessageNormalizer::normalize()` that occurs when `$data->toolCalls` is null in the Anthropic bridge.

## Problem

The current implementation calls `array_map()` on potentially null `toolCalls`:

```php
'content' => array_map(static function (ToolCall $toolCall) {
    // ...
}, $data->toolCalls ?? []),

### DIFF
--- a/src/platform/src/Bridge/Anthropic/Contract/AssistantMessageNormalizer.php
+++ b/src/platform/src/Bridge/Anthropic/Contract/AssistantMessageNormalizer.php
@@ -26,22 +26,12 @@ final class AssistantMessageNormalizer extends ModelContractNormalizer implement
 {
     use NormalizerAwareTrait;
 
-    protected function supportedDataClass(): string
-    {
-        return AssistantMessage::class;
-    }
-
-    protected function supportsModel(Model $model): bool
-    {
-        return $model instanceof Claude;
-    }
-
     /**
      * @param AssistantMessage $data
      *
      * @return array{
      *     role: 'assistant',
-     *     content: list<array{
+     *     content: string|list<array{
      *         type: 'tool_use',
      *         id: string,
      *         name: string,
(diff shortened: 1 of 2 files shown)


Muneer-Shafi:
> Hi @Muneer-Shafi and welcome to the repository! :) Thanks for working on that fix - can you please bring in a test case to prove error and fix?

@chr-hertel  i have updated test for Claude AssistantMessageNormalizerTest

chr-hertel:
Thank you @Muneer-Shafi.

PR #304, opened 2025-08-11
[Platform]: Ollama structured output support

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | Kind of
| New feature?  | no
| Docs?         | yes
| Issues        | MIT
| License       | MIT

Hi 👋🏻 

This PR aims to introduce the support for structured output via Ollama, the feature is fully functional since a few versions ans by default, the client does not map the "format" to the HTTP request, here's the fix/feature.


src/platform/tests/Bridge/Ollama/OllamaClientTest.php
```diff
@@ -0,0 +1,90 @@
...
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Bridge\Ollama;
```
chr-hertel: should be
```suggestion
namespace Symfony\AI\Platform\Tests\Bridge\Ollama;
```

src/platform/src/Bridge/Ollama/OllamaClient.php
chr-hertel, line 1: What's the reason for changing the methods order? Is it because one is public? I think it shouldn't be public - could you fix that please while you're at it? :pray:

src/platform/src/Bridge/Ollama/Ollama.php
```diff
@@ -45,6 +45,7 @@ class Ollama extends Model
         '/./' => [
             Capability::INPUT_MESSAGES,
             Capability::OUTPUT_TEXT,
+            Capability::OUTPUT_STRUCTURED,
```
chr-hertel, line 48: is it valid just to assume all models support that?

Guikingone: AFAIK, yes, every model exposed by Ollama supports it: 

- https://ollama.com/blog/structured-outputs
- https://github.com/ollama/ollama/blob/main/docs/api.md#request-structured-outputs

That's why Ollama is not "exposing" the capability, looks like the engine is doing the formatting phase itself when tools are supported individually by each models.

chr-hertel: okay, great :+1:

chr-hertel (approved):

chr-hertel:
Thank you @Guikingone.

OskarStark:
Nice, thanks 🙏

PR #307, opened 2025-08-12
[Examples] add AWS credentials environment validator

sadiqk2:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Docs?         | no 
| Issues        | 
| License       | MIT

- Created `EnvValidator` to centralize AWS credentials environment variable validation
- Updated `examples/bedrock/` to use the new validator


sadiqk2:
@chr-hertel  Can you review this ?

OskarStark (commented):
We usually haven env() shortcut in other examples, I would prefer to use this

sadiqk2:
> We usually haven env() shortcut in other examples, I would prefer to use this

@OskarStark you mean to env() instead of global var thingy

chr-hertel (changes requested):
The difference with these examples is that `env(...)` is usually used to access the `$_SERVER` and throw an error if not set. In our Bedrock bridge we don't have to do that since the aws lib we're using is fetching the env vars itself.

Again, like with #313, I'm not convinced to have an extra class here. I see the point that this looks a bit odd right now, but unless we can come up with a better idea than an extra class, i'd rather keep it like that.

Issue #208, opened 2025-07-25
[Bundle] Activate TokenUsage output processors in the bundle


Renrhaf:
At the moment the bundle does not activate the output processors related to token usage.
It might be usefull to have them enabled when they exists so that we can retrieve token usage data in the output metadata.

chr-hertel:
makes sense, i think it should be configurable tho. maybe default true, but you might not want to have it in metadata.

junaidbinfarooq:
@chr-hertel 
I tried addressing this one [#311](https://github.com/symfony/ai/pull/311).
Could you have a look at the changes and let me know what you think?

OskarStark:
If it is not too complicated I would love to see a PR fixing this bug and only use #311 for the new proposal

junaidbinfarooq:
@OskarStark 
It is not necessarily a bug IMO, rather something that hasn't been implemented so far.

OskarStark:
Ah got it, thanks

PR #311, opened 2025-08-13
[Platform] Standardise token usage

junaidbinfarooq:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | #208
| License       | MIT

# Changes proposed:

The PR aims to standardize the token usage extraction and its availability in the AI bundle. Essentially, the following is done:
- Makes token usage explicit by adding a dedicated dto for token information
- Populates the token usage dto inside different token output processors and then adds it to the metadata object

### DIFF
--- a/examples/openai/token-metadata.php
+++ b/examples/openai/token-metadata.php
@@ -15,6 +15,7 @@ use Symfony\AI\Platform\Bridge\OpenAi\PlatformFactory;
 use Symfony\AI\Platform\Bridge\OpenAi\TokenOutputProcessor;
 use Symfony\AI\Platform\Message\Message;
 use Symfony\AI\Platform\Message\MessageBag;
+use Symfony\AI\Platform\Result\Metadata\TokenUsage;
 
 require_once dirname(__DIR__).'/bootstrap.php';
 
@@ -33,8 +34,11 @@ $result = $agent->call($messages, [
 ]);
 
 $metadata = $result->getMetadata();
+$tokenUsage = $metadata->get('token_usage');
 
-echo 'Utilized Tokens: '.$metadata['total_tokens'].\PHP_EOL;
-echo '-- Prompt Tokens: '.$metadata['prompt_tokens'].\PHP_EOL;
-echo '-- Completion Tokens: '.$metadata['completion_tokens'].\PHP_EOL;
-echo 'Remaining Tokens: '.$metadata['remaining_tokens'].\PHP_EOL;
+assert($tokenUsage instanceof TokenUsage);
+
+echo 'Utilized Tokens: '.$tokenUsage->totalTokens.\PHP_EOL;
+echo '-- Prompt Tokens: '.$tokenUsage->promptTokens.\PHP_EOL;
+echo '-- Completion Tokens: '.$tokenUsage->completionTokens.\PHP_EOL;
+echo 'Remaining Tokens: '.$tokenUsage->remainingTokens.\PHP_EOL;
--- a/src/platform/src/Result/Metadata/TokenUsage.php
+++ b/src/platform/src/Result/Metadata/TokenUsage.php
@@ -1,0 +1,56 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Platform\Result\Metadata;
+
+/**
+ * @author Junaid Farooq <ulislam.junaid125@gmail.com>
+ */
+final class TokenUsage implements \JsonSerializable
+{
+    public function __construct(
+        public ?int $promptTokens = null,
+        public ?int $completionTokens = null,
+        public ?int $thinkingTokens = null,
+        public ?int $cachedTokens = null,
+        public ?int $remainingTokens = null,
+        public ?int $remainingTokensMinute = null,
+        public ?int $remainingTokensMonth = null,
+        public ?int $totalTokens = null,
+    ) {
+    }
+
+    /**
+     * @return array{
+     *      prompt_tokens: ?int,
+     *      completion_tokens: ?int,
+     *      thinking_tokens: ?int,
+     *      cached_tokens: ?int,
+     *      remaining_tokens: ?int,
+     *      remaining_tokens_minute: ?int,
+     *      remaining_tokens_month: ?int,
+     *      total_tokens: ?int,
+     *  }
+     */
+    public function jsonSerialize(): array
+    {
+        return [
+            'prompt_tokens' => $this->promptTokens,
+            'completion_tokens' => $this->completionTokens,
+            'thinking_tokens' => $this->thinkingTokens,
+            'cached_tokens' => $this->cachedTokens,
+            'remaining_tokens' => $this->remainingTokens,
+            'remaining_tokens_minute' => $this->remainingTokensMinute,
+            'remaining_tokens_month' => $this->remainingTokensMonth,
+            'total_tokens' => $this->totalTokens,
+        ];
+    }
+}
--- a/src/platform/tests/Bridge/OpenAi/TokenOutputProcessorTest.php
+++ b/src/platform/tests/Bridge/OpenAi/TokenOutputProcessorTest.php
@@ -20,6 +20,7 @@ use Symfony\AI\Platform\Bridge\OpenAi\TokenOutputProcessor;
 use Symfony\AI\Platform\Message\MessageBagInterface;
 use Symfony\AI\Platform\Model;
 use Symfony\AI\Platform\Result\Metadata\Metadata;
+use Symfony\AI\Platform\Result\Metadata\TokenUsage;
 use Symfony\AI\Platform\Result\RawHttpResult;
 use Symfony\AI\Platform\Result\ResultInterface;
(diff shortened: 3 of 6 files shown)


examples/openai/token-metadata.php
```diff
@@ -23,18 +24,26 @@
...
     Message::forSystem('You are a pirate and you write funny.'),
     Message::ofUser('What is the Symfony framework?'),
 );
+
 $result = $agent->call($messages, [
     'max_tokens' => 500, // specific options just for this call
 ]);
 
-$metadata = $result->getMetadata();
+if (null === $tokenUsage = $result->getTokenUsage()) {
```
OskarStark: Can this happen?

src/platform/src/Result/TokenUsage/TokenUsage.php
```diff
@@ -0,0 +1,53 @@
...
+
+namespace Symfony\AI\Platform\Result\TokenUsage;
+
+/**
+ * @author Junaid Farooq <ulislam.junaid125@gmail.com>
+ */
+final class TokenUsage implements \JsonSerializable
+{
+    public function __construct(
+        public ?int $promptTokens = null,
```
OskarStark:
```suggestion
        public ?int $prompt = null,
```
Maybe remove all the Token suffix here? They feel superfluous

examples/openai/token-metadata.php
junaidbinfarooq: Yes, if you mean `null === $tokenUsage`, when _token_usage_ is marked false in the config, as far as I can see.
Edit: Here it is unnecessary.

src/platform/src/Result/TokenUsage/TokenUsage.php
junaidbinfarooq: No problem.
I would not have kept the suffix if the object were simply named something like `Tokens`, but I thought `TokenUsage` is a better name. Or perhaps the DTO should be named as `Tokens`.

OskarStark: To be discussed with @chr-hertel

junaidbinfarooq: I did remove the suffixes BTW.

chr-hertel (changes requested):
Sorry, but we need to slim this down a bit - in general smaller PRs, that build on top of each other are favorable to me. so let's break it down like you did:

> Makes token usage explicit
I like having the value object, we should definitely have that :+1: 
But I'm not in favor of having the `getTokenUsage()` directly on the result since the result is a high lever abstraction and token usage is quite specific to remote inference of GenAI models - not all models have that. let's keep it in the metadata please.

> Adds support to automatically register token usage extractors so that the token usage information is available in the result when a model is invoked
That's a bit misleading since token usage was already sitting on an extension point => `OutputProcessor` and your design decision to use that extension point to add another extension point doesn't really resonate with me. why is the new extension point needed? it feels like we're just adding them because we can.

> Makes the token usage auto-registration configurable
For me this is too early, i don't think that a use-case, do you? how often do we see the case for someone to bring their own token usage extractor.

---

all in all, let's start here please with the TokenUsage value object in the metadata class - that's an easy step we see the same picture :+1:

src/platform/src/Result/TokenUsage/TokenUsage.php
OskarStark: Thanks, not sure now that this was the right decision 😅

junaidbinfarooq: `TokenUsage` for the class name and _tokens_ as the suffixes for the property names sounds more logical to me.
Or conversely, `Tokens` for the class name and no suffixes for the property names would also be a good option.

OskarStark: TokenUsage and tokens as suffix sound best to me

junaidbinfarooq:
@chr-hertel 
> Makes token usage explicit
I like having the value object, we should definitely have that 👍
But I'm not in favor of having the getTokenUsage() directly on the result since the result is a high lever abstraction and token usage is quite specific to remote inference of GenAI models - not all models have that. let's keep it in the metadata please.

Hmm, I initially thought about keeping it in metadata, but having it inside the Result object also made sense since an API should send us this information. I, so far, have seen such information in the responses of whatever APIs I tried, but perhaps you are right, and other models out there do not provide such information.

> Adds support to automatically register token usage extractors so that the token usage information is available in the result when a model is invoked
That's a bit misleading since token usage was already sitting on an extension point => OutputProcessor and your design decision to use that extension point to add another extension point doesn't really resonate with me. why is the new extension point needed? it feels like we're just adding them because we can.

Nope, the idea, instead, was to introduce an extension point specific to token usage extraction and exposure. This would also simplify for bridges to their own implentations of the extractor interface if needed.

> Makes the token usage auto-registration configurable
For me this is too early, i don't think that a use-case, do you? how often do we see the case for someone to bring their own token usage extractor.

So far, in this repository, I have found three such extractors (currently TokenOutputProcessor classes), including the new Google Vertex AI bridge, and all three of them perform the extraction differently, albeit not entirely. As you also rightly pointed out above, token usage is something specific to GenAI models, and not all of them have it. Not all of them will expose the data in a similar fashion, either.

src/platform/src/Result/TokenUsage/TokenUsage.php
junaidbinfarooq: :man_facepalming: :smile:

junaidbinfarooq:
@OskarStark @chr-hertel 
I have reverted the previous changes and only kept the token usage information. Also, edited the PR description.
Please review the changes.

src/platform/src/Result/Metadata/TokenUsage.php
```diff
@@ -0,0 +1,54 @@
...
+        public ?int $cachedTokens = null,
+        public ?int $remainingTokens = null,
+        public ?int $remainingTokensMinute = null,
+        public ?int $remainingTokensMonth = null,
+        public ?int $totalTokens = null,
+    ) {
+    }
+
+    /**
+     * @return array<string, int|null>
```
OskarStark: We can describe the array better as we know all keys

src/platform/tests/Bridge/OpenAi/TokenOutputProcessorTest.php
```diff
@@ -70,8 +72,11 @@ public function testItAddsRemainingTokensToMetadata()
         $processor->processOutput($output);
 
         $metadata = $output->result->getMetadata();
+        $tokenUsage = $metadata->get('token_usage');
```
OskarStark, line 75: What about having a dedicated setTokenUsage/getTokenUsage on the metadata object?

src/platform/src/Result/Metadata/TokenUsage.php
junaidbinfarooq: Yes, done.

src/platform/tests/Bridge/OpenAi/TokenOutputProcessorTest.php
junaidbinfarooq: Not sure about this.
Metadata object already has a lot of methods, and each of them is generic in nature as far as I can see, and doesn't concern any specific model characteristic.

src/platform/src/Result/Metadata/TokenUsage.php
```diff
@@ -0,0 +1,66 @@
...
+        ];
+    }
+
+    /**
+     * @return TokenUsageArray
+     */
+    public function jsonSerialize(): array
+    {
+        return $this->toArray();
+    }
```
chr-hertel: where do we `json_encode` the token usage?

chr-hertel: ah i guess when forwarding to frontend clients maybe, fair

src/platform/src/Result/Metadata/TokenUsage.php
```diff
@@ -0,0 +1,66 @@
...
+            'total_tokens' => $this->totalTokens,
+        ];
+    }
+
+    /**
+     * @return TokenUsageArray
+     */
+    public function jsonSerialize(): array
+    {
+        return $this->toArray();
```
chr-hertel: you can move it down and get rid of the `toArray()`

```suggestion
        return [
            'prompt_tokens' => $this->promptTokens,
            'completion_tokens' => $this->completionTokens,
            'thinking_tokens' => $this->thinkingTokens,
            'cached_tokens' => $this->cachedTokens,
            'remaining_tokens' => $this->remainingTokens,
            'remaining_tokens_minute' => $this->remainingTokensMinute,
            'remaining_tokens_month' => $this->remainingTokensMonth,
            'total_tokens' => $this->totalTokens,
        ];
```

junaidbinfarooq: Yeah, wasn't sure about `toArray` either.
Removed it and moved php-doc to `jsonSerialize` method now.

junaidbinfarooq:
> almost done i'd say - thanks already!

I am working on autoconfiguring token usage processors inside the AI bundle and based that branch on this one.
A PR will arrive after this one is finished.

junaidbinfarooq:
> Good to merge from my end - @OskarStark?

@OskarStark 
Could you have a look?

chr-hertel:
Thank you @junaidbinfarooq.

Issue #285, opened 2025-08-07
[Store] Add lifecycle management for stores


chr-hertel:
Currently there is the `InitializableStoreInterface` which was originally introduced with MongoDB, but was adopted quite well while rolling out more stores.

This interface is not used by the library itself, but its examples and could provide another feature for managing the lifecycle of a store. Similar to doctrine schema management or messenger setup.

To level this up, we would need to add the counterpart method, e.g. `drop` or `tearDown` or `destroy` or `killItWithFire`, which could be combined with corresponding commands like `ai:store:initialize` and `ai:store:destory`.

We should aling:
* Do we want to have that feature?
  * If not, that brings in the question about dropping the interface
* How should we name the methods - and is it only the two of them?

I'm def in favor of having that in the lib + commands.

chr-hertel:
What do you think about `ManagedStoreInterface` with `setup` and `drop`?

we only need the functionality if the store component should manage the store, not only use it.

cc @Guikingone & @OskarStark

OskarStark:
Love it 👍🏻

Guikingone:
Agree on the name and methods, makes more sense than `DroppableStoreInterface` 😅

PR #314, opened 2025-08-14
[Store] Add Milvus

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | None
| License       | MIT

Hi 👋🏻 

This PR aim to add the support for `Milvus` in the `Store` component, it also introduce a `DroppableStoreInterface` that allows to remove the current database (because it's boring to remove the volumes and drop the services from Docker 😄 ).

Thanks for the feedback.


src/store/src/DroppableStoreInterface.php
```diff
@@ -0,0 +1,20 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Store;
+
+/**
+ * @author Guillaume Loulier <personal@guillaumeloulier.fr>
+ */
+interface DroppableStoreInterface
```
OskarStark, line 17: Nice idea, but not a big fan of the name of the interface 🤔

Guikingone: Me neither to be honest 😅

chr-hertel: let's move that discussion to another PR - i see this more in combination with #285 and extending/renaming the `InitializableStoreInterface`

examples/compose.yaml
chr-hertel, line 1: oof, do we really need `etcd` and `minio` as well? seems a bit heavy to me, tbh :D

chr-hertel:
Needs a rebase after #317

examples/compose.yaml
Guikingone: Sadly, if we want the Docker container, yes, without them, Milvus cannot be started properly (it stored credentials in Minio and more), It's a shame, I know 😄

chr-hertel (changes requested):
Alright, I played a little bit around with it and also running Milvus on cloud.zilliz.com.

On thing, that made it harder for me is the database - this is not actually needed. With the cloud service i ran into 403, but worked when commenting it out - and also locally it works without creating the database first.

I think the database should at least be optional - maybe if database is not null or something?

Guikingone:
Ah yes, tested the cloud version and it comes with an existing database, hum, one solution might be to add a `forceDatabaseCreation` (name open to debate 😄 ) option in the `initialize` method, this way, if the option is set to `true`, we create the database, otherwise, we "expect" that it exist 🤔

Guikingone:
Small update here @chr-hertel @OskarStark, I removed the `DroppableStoreInterface` following the discussion in #285, it's better to do a "big" PR to updated than introducing something that will be removed for clarity later.

examples/rag/milvus.php
```diff
@@ -0,0 +1,72 @@
...
+use Symfony\AI\Store\Document\Vectorizer;
+use Symfony\AI\Store\Indexer;
+use Symfony\Component\Uid\Uuid;
+
+require_once dirname(__DIR__).'/bootstrap.php';
+
+// initialize the store
+$store = new Store(
+    httpClient: http_client(),
+    endpointUrl: env('MILVUS_HOST'),
```
OskarStark, line 34: In other stores we name it Base URL, the env is called HOST, looks weird/complex

Guikingone: In most stores, it's called `X_HOST` when it comes to the env variable, do you want me to rename the env variables?

OskarStark: Alright then let's keep it like this for now. Can be tackled separately if needed

Guikingone:
I just rebased the branch against `main`, @OskarStark, I will work on renaming the env variable in a new PR 🙂

chr-hertel:
Thank you @Guikingone.

PR #318, opened 2025-08-14
[Store] Introduce local bridge for in-memory and cache based stores

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | 
| License       | MIT

Didn't catch that before - I'd rather keep the root clean and i don't think everyone needs this. "local" was just my best idea at this point - input/feedback welcome!

Sits on top of #317


OskarStark (approved):
Makes sense

PR #324, opened 2025-08-16
[Platform] Add stream for ElevenLabs TTS

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | None
| License       | MIT

Hi 👋🏻 

This PR aim to introduce the support for streaming TTS.

### DIFF
--- a/examples/elevenlabs/text-to-speech-as-stream.php
+++ b/examples/elevenlabs/text-to-speech-as-stream.php
@@ -1,0 +1,35 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use Symfony\AI\Platform\Bridge\ElevenLabs\ElevenLabs;
+use Symfony\AI\Platform\Bridge\ElevenLabs\PlatformFactory;
+use Symfony\AI\Platform\Message\Content\Text;
+
+require_once dirname(__DIR__).'/bootstrap.php';
+
+$platform = PlatformFactory::create(
+    apiKey: env('ELEVEN_LABS_API_KEY'),
+    httpClient: http_client(),
+);
+$model = new ElevenLabs(options: [
+    'voice' => 'Dslrhjl3ZpzrctukrQSN', // Brad (https://elevenlabs.io/app/voice-library?voiceId=Dslrhjl3ZpzrctukrQSN)
+    'stream' => true,
+]);
+
+$result = $platform->invoke($model, new Text('The first move is what sets everything in motion.'));
+
+$content = '';
+
+foreach ($result->asStream() as $chunk) {
+    echo $chunk;
+}
+
+echo \PHP_EOL;
--- a/src/platform/src/Bridge/ElevenLabs/ElevenLabsClient.php
+++ b/src/platform/src/Bridge/ElevenLabs/ElevenLabsClient.php
@@ -42,7 +42,7 @@ final readonly class ElevenLabsClient implements ModelClientInterface
         }
 
         if (\in_array($model->getName(), [ElevenLabs::SCRIBE_V1, ElevenLabs::SCRIBE_V1_EXPERIMENTAL], true)) {
-            return $this->doSpeechToTextRequest($model, $payload, $options);
+            return $this->doSpeechToTextRequest($model, $payload);
         }
 
         $capabilities = $this->retrieveCapabilities($model);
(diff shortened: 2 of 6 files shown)


OskarStark:
Let's use GitHub draft instead of the prefix in the title please

Guikingone:
Not needed anymore, I added the `Draft` prefix to prevent any reviews before pushing my refactoring 😅

src/platform/src/Bridge/ElevenLabs/ElevenLabsClient.php
```diff
@@ -86,6 +85,21 @@ private function doTextToSpeechRequest(Model $model, array|string $payload, arra
...
+                'headers' => [
+                    'xi-api-key' => $this->apiKey,
+                ],
+                'json' => [
+                    'text' => $payload['text'],
+                    'model_id' => $model->getName(),
+                ],
+            ]);
+
+            return new RawHttpResult($this->httpClient->stream($streamSource));
```
chr-hertel: a bit different than the other streaming implementations we have, but quite an interesting idea to return the `ResponseStreamInterface` instead of the `ResponseInterface`.

is this blocking or still enables concurrency of requests?

src/platform/src/Bridge/ElevenLabs/ElevenLabsClient.php
```diff
@@ -86,6 +85,21 @@ private function doTextToSpeechRequest(Model $model, array|string $payload, arra
...
+        if ($stream) {
+            $streamSource = $this->httpClient->request('POST', \sprintf('%s/text-to-speech/%s/stream', $this->hostUrl, $voice), [
+                'headers' => [
+                    'xi-api-key' => $this->apiKey,
+                ],
+                'json' => [
+                    'text' => $payload['text'],
+                    'model_id' => $model->getName(),
+                ],
+            ]);
```
chr-hertel: to much of duplication here - the difference is only the extra `/stream` at the endpoint?

src/platform/src/Bridge/ElevenLabs/ElevenLabsResultConverter.php
```diff
@@ -32,10 +34,23 @@ public function supports(Model $model): bool
...
+                        continue;
+                    }
+
+                    if ('' === $chunk->getContent()) {
+                        continue;
+                    }
+
+                    yield $chunk;
+                }
+            })()),
```
chr-hertel: can you please move that to a private method and we keep the match lean?

src/platform/src/Result/RawHttpResult.php
chr-hertel, line 1: looking at this, i'm rather in favor of still returning the `ResponseInterface` :thinking: 

what's the benefit of directly returning the `ResponseStreamInterface`?

Guikingone: Mostly static analysis and the fact that `stream` from HttpClient is typed with this interface, `ResponseStreamInterface` does not extend `ResponseInterface` so we're locked if we don't allow the stream one.

src/platform/src/Bridge/ElevenLabs/ElevenLabsClient.php
chr-hertel: this isn't solved from my point of view: streaming a response makes it blocking, right?

Guikingone: I'm not sure about the question but AFAIK, until we call `getContent()`, the request is asynchronous by design isn't it? In the converter, we return the `chunk` rather than the content 🤔 

The `getContent()` call is not performed until we call the example file (or loop through the chunks) 🤔

examples/elevenlabs/text-to-speech-as-stream.php
chr-hertel, line 1: how do you test that it works?

i used `php elevenlabs/text-to-speech.php | mpg123 -` for the other example, but with that there is no benefit in streaming

src/platform/src/Bridge/ElevenLabs/ElevenLabsClient.php
```diff
@@ -86,16 +85,25 @@ private function doTextToSpeechRequest(Model $model, array|string $payload, arra
...
             ],
             'json' => [
                 'text' => $payload['text'],
                 'model_id' => $model->getName(),
             ],
-        ]));
+        ]);
+
+        return $stream
+            ? new RawHttpResult($this->httpClient->stream($response))
```
chr-hertel: in my understanding this call of `stream` method break the concurrency feature and should be shifted to the `ElevenLabsResultConverter`

examples/elevenlabs/text-to-speech-as-stream.php
Guikingone: I pushed a fix with a modification of the `PlatformFactory` 🙂

chr-hertel:
Thank you @Guikingone.

PR #327, opened 2025-08-18
[Platform] Fix fatal error with string payloads

### LINKED ISSUE #326: [Platform] Fatal Error: array_merge() called with string payload in ModelClient implementations
## Description:
The ModelClientInterface defines `$payload` as array|string but many implementations use `array_merge($payload, $options)`  without type checking, causing fatal errors when a string payload is passed.

## Affected Files
Multiple ModelClient implementations fail to handle string payloads properly, 

## Current Behavior
Interface allows: `request(Model $model, array|string $payload, array $options = [])`
Many implementations call: `array_merge($payload, $options)`
Result: Fatal error when $payload is string: array_merge(): Expected parameter 1 to be an array, string given

## Expected Behavior
All implementations should handle both array and string payloads consistently.
Inconsistent Implementation
Some implementations handle this correctly:

Albert/EmbeddingsModelClient.php:45
Albert/GptModelClient.php:50
Cerebras/ModelClient.php:58

These use: \is_array($payload) ? array_merge($payload, $options) : $payload
Or assigning the payload to` ['input => $payload]` before merge 
But many others don't perform this check.

RamyHakam:
…lementations

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  no 
| Docs?         | no 
| Issues        | Fix #326 
| License       | MIT

Issue
The `ModelClientInterface` allows array|string $payload, but all implementations use `array_merge($payload, $options)` without type checking, causing fatal errors when string payloads are passed.

Fix
Remove unnecessary string support from the interface since all implementations expect arrays:

### DIFF
--- a/src/platform/src/Bridge/Albert/EmbeddingsModelClient.php
+++ b/src/platform/src/Bridge/Albert/EmbeddingsModelClient.php
@@ -38,11 +38,11 @@ final readonly class EmbeddingsModelClient implements ModelClientInterface
         return $model instanceof Embeddings;
     }
 
-    public function request(Model $model, array|string $payload, array $options = []): RawResultInterface
+    public function request(Model $model, array $payload, array $options = []): RawResultInterface
     {
         return new RawHttpResult($this->httpClient->request('POST', \sprintf('%s/embeddings', $this->baseUrl), [
             'auth_bearer' => $this->apiKey,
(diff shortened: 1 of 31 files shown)


chr-hertel:
This breaks some huggingface examples, and embeddings - for example for openai.

Let's discuss in #326 first

PR #328, opened 2025-08-19
[Store] Add ManagedStoreInterface

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | #285 #314 
| License       | MIT

Hi 👋🏻 

This PR aim to introduce the `ManagedStoreInterface` as discussed in #285, `Store` is a critical part of this library so I'm not gonna pretend that I know how each store currently work but here's the refactoring for most of them, tests must be improved but here's the first "draft".

This PR does not include the related commands.

### DIFF
--- a/examples/rag/mongodb.php
+++ b/examples/rag/mongodb.php
@@ -54,7 +54,7 @@ $indexer = new Indexer($vectorizer, $store, logger());
 $indexer->index($documents);
 
 // initialize the index
-$store->initialize();
+$store->setup();
 
 $model = new Gpt(Gpt::GPT_4O_MINI);
 
--- a/src/store/CHANGELOG.md
+++ b/src/store/CHANGELOG.md
@@ -6,7 +6,7 @@ CHANGELOG
 
  * Add core store interfaces:
    - `StoreInterface` for storing and querying document stores
-   - `InitializableStoreInterface` for stores requiring initialization
+   - `ManagedStoreInterface` for stores, which can be initialized or dropped
  * Add document types:
    - `TextDocument` for raw text with UUID, content, and metadata
    - `VectorDocument` for vectorized documents with embeddings and similarity scores
--- a/src/store/src/Bridge/Postgres/Store.php
+++ b/src/store/src/Bridge/Postgres/Store.php
@@ -17,7 +17,7 @@ use Symfony\AI\Platform\Vector\VectorInterface;
 use Symfony\AI\Store\Document\Metadata;
 use Symfony\AI\Store\Document\VectorDocument;
 use Symfony\AI\Store\Exception\InvalidArgumentException;
-use Symfony\AI\Store\InitializableStoreInterface;
+use Symfony\AI\Store\ManagedStoreInterface;
 use Symfony\AI\Store\StoreInterface;
 use Symfony\Component\Uid\Uuid;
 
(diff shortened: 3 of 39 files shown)


src/store/src/Bridge/Postgres/Store.php
```diff
@@ -38,6 +38,49 @@ public function __construct(
     ) {
     }
 
+    /**
+     * @param array{vector_type?: string, vector_size?: positive-int, index_method?: string, index_opclass?: string} $options
+     *
+     * Good configurations $options are:
```
OskarStark:
```suggestion
     * Good configuration $options are:
```

src/store/tests/Bridge/Local/CacheStoreTest.php
```diff
@@ -27,6 +27,33 @@
 #[UsesClass(Vector::class)]
 final class CacheStoreTest extends TestCase
 {
+    public function testStoreCanInitialize()
```
OskarStark:
```suggestion
    public function testStoreCanSetup()
```

Same for the others

OskarStark (commented):
No docs to adjust or changelog?

Guikingone:
Yes, need to work on the changelog and tests, won't update the changelog until the tests are fixed 🙂

Guikingone:
Yes, need to work on the changelog and tests, won't update the changelog until the tests are fixed 🙂

Regarding the documentation, the current state is that `InitializableStoreInterface` is the interface for initialization, should we list the change to `ManagedStoreInterface` or just change the name in the file? 🤔

OskarStark:
Just change the class name in the changelog as we haven't released yet

OskarStark:
Can you create an issue for creating the commands? Thanks

examples/.env
```diff
@@ -84,29 +84,29 @@ ALBERT_API_URL=
 MARIADB_URI=pdo-mysql://root@127.0.0.1:3309/my_database
 
 # Meilisearch
-MEILISEARCH_HOST=http://127.0.0.1:7700
```
OskarStark: Hmm while seeing that much changes to the ENVs, let's keep HOST

src/store/CHANGELOG.md
```diff
@@ -6,7 +6,7 @@ CHANGELOG
 
  * Add core store interfaces:
    - `StoreInterface` for storing and querying document stores
-   - `InitializableStoreInterface` for stores requiring initialization
+   - `ManagedStoreInterface` for stores requiring initialization
```
OskarStark:
```suggestion
   - `ManagedStoreInterface` for stores, which can be initialized or dropped
```
?

OskarStark (approved):

examples/rag/mongodb.php
```diff
@@ -38,6 +38,9 @@
     vectorFieldName: 'vector',
 );
 
+// initialize the index
+$store->setup();
```
OskarStark: Please keep the order, in MongoDB we cannot initialize the index without documents.

Same for the other stores, let's keep the diff as small as possible

Guikingone: Ah yes, I moved them during tests 🙁

OskarStark (approved):

chr-hertel:
oh wow, nice one :muscle: 

needs a rebase tho after #288

chr-hertel (approved):
Thanks for that @Guikingone!

Didn't test it - and honestly wouldn't know how at this point. The PHPUnit tests won't be sufficient going forward.
For the other integrations we currently use the examples and the runner for doing integration tests - when we bring in the management commands we should also think about some integration test layer here with all the store bridges.

chr-hertel:
Thank you @Guikingone.

PR #329, opened 2025-08-19
[AI Bundle] Activate token usage output processors in ai bundle

### LINKED ISSUE #208: [Bundle] Activate TokenUsage output processors in the bundle

junaidbinfarooq:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | Fix #208 Continues #311
| License       | MIT

# Changes proposed:
- Adds support to automatically configure the token usage processors so that they are available to end users (devs) when needed
- Makes the availability of token usage configurable and disabled by default


examples/mistral/token-metadata.php
```diff
@@ -0,0 +1,45 @@
...
+
+$metadata = $result->getMetadata();
+$tokenUsage = $metadata->get('token_usage');
+
+assert($tokenUsage instanceof TokenUsage);
+
+echo 'Utilized Tokens: '.$tokenUsage->totalTokens.\PHP_EOL;
+echo '-- Prompt Tokens: '.$tokenUsage->promptTokens.\PHP_EOL;
+echo '-- Completion Tokens: '.$tokenUsage->completionTokens.\PHP_EOL;
+echo 'Remaining Tokens: '.$tokenUsage->remainingTokens.\PHP_EOL;
```
OskarStark: We are using this in other we examples, what about ab helper method: print_token_usage($metadata) for the examples and print all values which are not null?

src/ai-bundle/tests/DependencyInjection/Compiler/RegisterTokenUsageProcessorPassTest.php
```diff
@@ -0,0 +1,106 @@
...
+ * @author Junaid Farooq <ulislam.junaid125@gmail.com>
+ */
+#[CoversClass(RegisterTokenUsageProcessorPass::class)]
+#[UsesClass(ContainerBuilder::class)]
+#[UsesClass(Definition::class)]
+class RegisterTokenUsageProcessorPassTest extends TestCase
+{
+    public function testProcessRegistersTokenUsageProcessors()
+    {
+        // Arrange
```
OskarStark: Please remove such comments

junaidbinfarooq: I added the comments on purpose as that is a general recommendation on how to structure your tests, while there are blank lines inside different sections within the tests.

junaidbinfarooq:
> Off topic, but should also be then integrated in the profiler 🙌🤓

Can that be done separately?

examples/mistral/token-metadata.php
junaidbinfarooq: Done.

OskarStark:
Yes should be a follow up PR, maybe create an issue to not forget?

examples/mistral/token-metadata.php
OskarStark: Thanks

examples/bootstrap.php
```diff
@@ -52,3 +54,15 @@ function logger(): LoggerInterface
...
+function print_token_usage(Metadata $metadata): void
+{
+    $tokenUsage = $metadata->get('token_usage');
+
+    assert($tokenUsage instanceof TokenUsage);
+
+    echo 'Utilized Tokens: '.$tokenUsage->totalTokens.\PHP_EOL;
+    echo '-- Prompt Tokens: '.$tokenUsage->promptTokens.\PHP_EOL;
+    echo '-- Completion Tokens: '.$tokenUsage->completionTokens.\PHP_EOL;
+    echo 'Remaining Tokens: '.$tokenUsage->remainingTokens.\PHP_EOL;
```
OskarStark: Let's print all tokens here which are not null as we have more properties in the TokenUsage object

junaidbinfarooq: Done.

src/ai-bundle/config/options.php
```diff
@@ -96,6 +96,7 @@
                                 ->end()
                             ->end()
                         ->end()
+                        ->booleanNode('token_usage')->defaultFalse()->end()
```
```suggestion
                        ->booleanNode('track_token_usage')->defaultFalse()->end()
```
?

junaidbinfarooq: Ah! It was initially _track_token_usage_, but I later changed it to _token_usage_.
Changed it to _track_token_usage_ now.

OskarStark (approved):

junaidbinfarooq:
@chr-hertel 
Could you have a look?

src/ai-bundle/config/options.php
```diff
@@ -96,6 +96,7 @@
                                 ->end()
                             ->end()
                         ->end()
+                        ->booleanNode('track_token_usage')->defaultFalse()->end()
```
OskarStark: please add `->info()` node and add something to the README about this new option, thanks

junaidbinfarooq: Done.
Also added some info in `ai-bundle/doc/index.rst`.

OskarStark (approved):

examples/bootstrap.php
chr-hertel, line 1: :+1:

examples/mistral/token-metadata.php
chr-hertel, line 1: :heart_eyes:

src/ai-bundle/config/options.php
```diff
@@ -96,6 +96,10 @@
                                 ->end()
                             ->end()
                         ->end()
+                        ->booleanNode('track_token_usage')
+                            ->defaultFalse()
```
chr-hertel: the overhead here is super small, let's go with `defaultTrue` don't you think?

junaidbinfarooq:
> Please remove the indirection via the attribute here. there is no benefit, but only adds complexity to maintain.
> 
> We know in the bridge exactly for which platform the output processor is intended, could just register them as service (where does that currently happen?) and check for their existence while building the agent, like you did in `AiBundle` class.
> 
> And please enable and test it with the demo - does it already work for you?

Do you mean we should register such classes as services individually manually?
I tried to register them dynamically, and this is why the attribute exists.

> And please enable and test it with the demo - does it already work for you?

Integration with the demo app is already done, and I was awaiting this PR's merge so that I could follow it up with another PR that would contain these changes. @OskarStark earlier suggested trying to make the respective changes in the profiler, and this is why I didn't commit/push the changes that pertain to the demo app to keep the surface area of the changes small.
I tried testing the demo app with **Ollama**. Though everything worked fine, the token usage was obviously empty since the said bridge doesn't have any output processors for token usage yet, which I intended to add as well.

junaidbinfarooq:
> > Do you mean we should register such classes as services individually manually?
> 
> Yes please - we don't need to decouple via attribute. That also makes it more brittle. More moving parts, more places to break :D let's keep it simple and register the services directly when the option is `true` programmatic in the `AiBundle` class
> 
> > I tried testing the demo app with Ollama. Though everything worked fine, the token usage was obviously empty since the said bridge doesn't have any output processors for token usage yet, which I intended to add as well.
> 
> didn't work for me with openai, but not sure i got it right. let's see after the change 👍

@chr-hertel 
I made the following changes in the latest commit:
- Removed the attribute to tag token usage output processors
- Registered the token usage output processors as services manually inside the AI bundle
- Added token usage information to the documentation

Would it make sense to follow this PR up with another one to add the token usage stuff for demo and probably profiler as well, to keep this PR small?

VincentLanglet:
> Would it make sense to follow this PR up with another one to add the token usage stuff for demo and probably profiler as well, to keep this PR small?

Multiple small PR would be great to unlock other TokenOutProcessor PR too :)

src/ai-bundle/src/AiBundle.php
```diff
@@ -462,6 +464,25 @@ private function processAgentConfig(string $name, array $config, ContainerBuilde
...
+
+            if (!$container->hasDefinition('ai.platform.token_usage_processor.mistral')) {
+                $container->setDefinition(
+                    'ai.platform.token_usage_processor.mistral',
+                    new Definition(MistralTokenOutputProcessor::class),
+                );
+            }
+
+            $outputProcessors[] = new Reference('ai.platform.token_usage_processor.openai');
+            $outputProcessors[] = new Reference('ai.platform.token_usage_processor.mistral');
```
VincentLanglet: This is wrong.

All the processors will be executed and the last one will override every others with an early return like
```
if (!\array_key_exists('usage', $content)) {
            $metadata->add('token_usage', $tokenUsage);

            return;
        }
```

We have to add the right token usage processor based on the agent model/platform.

junaidbinfarooq:
> > > Do you mean we should register such classes as services individually manually?
> > 
> > 
> > Yes please - we don't need to decouple via attribute. That also makes it more brittle. More moving parts, more places to break :D let's keep it simple and register the services directly when the option is `true` programmatic in the `AiBundle` class
> > > I tried testing the demo app with Ollama. Though everything worked fine, the token usage was obviously empty since the said bridge doesn't have any output processors for token usage yet, which I intended to add as well.
> > 
> > 
> > didn't work for me with openai, but not sure i got it right. let's see after the change 👍
> 
> @chr-hertel I made the following changes in the latest commit:
> 
> * Removed the attribute to tag token usage output processors
> * Registered the token usage output processors as services manually inside the AI bundle
> * Added token usage information to the documentation
> 
> Would it make sense to follow this PR up with another one to add the token usage stuff for demo and probably profiler as well, to keep this PR small?

@chr-hertel 
Could you have a look at the latest changes and let me know what you think?

src/ai-bundle/src/AiBundle.php
chr-hertel: Yes, I agree, you already had some logic, that might help here based on
```php
$platform = u($config['platform'])->after('ai.platform.')->toString();
```
(not sure it works like that, but that's what you had :D)

src/ai-bundle/config/options.php
```diff
@@ -87,6 +88,19 @@
...
+                            ->children()
+                                ->booleanNode('enabled')
+                                    ->isRequired()
+                                    ->defaultTrue()
+                                    ->info('Enable tracking of token usage for this agent')
+                                ->end()
+                                ->scalarNode('processor')
+                                    ->info('Token usage processor to use e.g., OpenAiTokenOutputProcessor::class.')
+                                    ->defaultValue(OpenAiTokenOutputProcessor::class)
+                                ->end()
```
VincentLanglet: Now, the configuration feel really odd.

If I enable the tokenUsage, I would expect the right TokenOutputProcessor to be used. There is no reason to use OpenAiTokenOutputProcessor if I'm not using OpenAI.
Also, I have to pass the right outputprocessor. I don't need a bool `enabled`, it could be just be a processor config options and I pass the right one directly or null if I don't want one.
But then if we start adding `processor` options, instead of a specific one for the tokenUsage we should just add one generic which resolve https://github.com/symfony/ai/issues/346

The first implementation was the nearest of the right one IMHO, with something like
```
if ($config['track_token_usage'] ?? true) {
            $platform = u($config['platform'])->after('ai.platform.')->toString();

            if ($container->has('ai.platform.token_usage_processor.'.$platform)) {
                $outputProcessors[] = new Reference('ai.platform.token_usage_processor.'.$platform);
            }
        }
```
And in `processPlatformConfig` or another method you should add the definition of the existing token_usage_processor.

What was the issue ?

junaidbinfarooq: Didn't feel too good to me either. The original implementation used an attribute to tag the token usage processors. Since I removed the attribute in another commit, the said processors were registered manually.
Then you rightly pointed out one processor may override another, since both the existing token output processors were added as output processors.

> Also, I have to pass the right outputprocessor. I don't need a bool enabled, it could be just be a processor config options and I pass the right one directly or null if I don't want one.

Having only the processor for this config again seems a bit and the combination of an _enabled_ option plus the _processor_ feels a natural choice. But then providing a default for the processor option may or may not be the right thing to do. For now, I'll remove the default.

VincentLanglet: > Didn't feel too good to me either. The original implementation used an attribute to tag the token usage processors. Since I removed the attribute in another commit, the said processors were registered manually.
Then you rightly pointed out one processor may override another, since both the existing token output processors were added as output processors.

There is nothing wrong to register the processor manually, you just need to register the right one.

I just tell you that 
```
$outputProcessors[] = new Reference('ai.platform.token_usage_processor.openai');
$outputProcessors[] = new Reference('ai.platform.token_usage_processor.mistral');
```
was wrong.

But
```
if ($config['track_token_usage'] ?? true) {
     $platform = u($config['platform'])->after('ai.platform.')->toString();

     if ($container->has('ai.platform.token_usage_processor.'.$platform)) {
          $outputProcessors[] = new Reference('ai.platform.token_usage_processor.'.$platform);
     }
}
```
wasn't.

You just need somewhere else to write
```
$container->setDefinition('ai.platform.token_usage_processor.mistral', MistralTokenOutputProcessor);
$container->setDefinition('ai.platform.token_usage_processor.openai', OpenAiTokenOutputProcessor);
```

junaidbinfarooq: I know what was fine and what wasn't.
Also, adding the token processors to the output processors array and including their service definitions in the container are also to be done here, within the agentic config registration, and do not need to be moved elsewhere.
The only thing that needs thought is whether to get the right platform from the existing config somehow or to add a new option that lets developers explicitly specify the processor. The current state of the config lets us do the latter.

chr-hertel: @junaidbinfarooq with #352 it would be possible to easily register custom processors, and with
```yaml
                    enabled: true
                    processor: 'Symfony\AI\Platform\Bridge\OpenAi\TokenOutputProcessor'
```
the configuration indeed is a bit bloated and not what i intended, maybe i wasn't clear enough on that.

didn't get the issue with your first solution here:
```
if ($config['track_token_usage'] ?? true) {
     $platform = u($config['platform'])->after('ai.platform.')->toString();

     if ($container->has('ai.platform.token_usage_processor.'.$platform)) {
          $outputProcessors[] = new Reference('ai.platform.token_usage_processor.'.$platform);
     }
}
```
---

goal for me would be to have a slim config `track_token_usage: true` and if an agent uses a platform where we have the token usage processor implemented, and therefore registered as a service, we attach it to the array of used output processors.
doesn't that make sense?

chr-hertel (changes requested):
Please reduce the needed configuration to
```yaml
track_token_usage: false
```

See https://github.com/symfony/ai/pull/329#discussion_r2309864896

And then we're good to merge after me testing it out :) :+1:

src/ai-bundle/doc/index.rst
```diff
@@ -65,6 +65,9 @@ Configuration
             rag:
                 platform: 'ai.platform.azure.gpt_deployment'
                 structured_output: false # Disables support for "output_structure" option, default is true
+                track_token_usage:
+                    enabled: true # Enable tracking of token usage for the agent, default is true
+                    processor: 'Symfony\AI\Platform\Bridge\OpenAi\TokenOutputProcessor' # Token usage processor to use
```
VincentLanglet: need to be updated

src/platform/src/Result/Metadata/TokenUsage/TokenUsage.php
```diff
@@ -9,7 +9,7 @@
  * file that was distributed with this source code.
  */
 
-namespace Symfony\AI\Platform\Result\Metadata;
+namespace Symfony\AI\Platform\Result\Metadata\TokenUsage;
```
VincentLanglet: Is it needed to introduce a new folder inside Metadata (when there is only one class so far in it) ?

src/ai-bundle/doc/index.rst
junaidbinfarooq: Oh! skipped my eyes.
Updated.
Thanks.

src/platform/src/Result/Metadata/TokenUsage/TokenUsage.php
junaidbinfarooq: Absolutely not.
There were more than one class related to each other and were hence grouped together in another PR.
That is not the case anymore, so I removed the dir and moved this class to the `Metadata` dir now.

src/ai-bundle/doc/index.rst
```diff
@@ -230,6 +231,82 @@ The attribute ``IsGrantedTool`` can be added on class- or method-level - even mu
...
+* **Remaining tokens**: Number of remaining tokens in rate limits (when available)
+* **Cached tokens**: Number of cached tokens used (when available)
+* **Thinking tokens**: Number of reasoning tokens used (for models that support reasoning)
+
+The token usage information can be accessed from the result metadata::
+
+    use Symfony\AI\Agent\AgentInterface;
+    use Symfony\AI\Platform\Message\Message;
+    use Symfony\AI\Platform\Message\MessageBag;
+    use Symfony\AI\Platform\Result\Metadata\TokenUsage\TokenUsage;
```
OskarStark, line 255: Correct namespace? 

I am currently on a phone and cannot check

OskarStark:
Looks quite good to me

examples/bootstrap.php
```diff
@@ -11,6 +11,8 @@
 
 use Psr\Log\LoggerAwareInterface;
 use Psr\Log\LoggerInterface;
+use Symfony\AI\Platform\Metadata\TokenUsage;
+use Symfony\AI\Platform\Result\Metadata\Metadata;
```
chr-hertel:
```suggestion
use Symfony\AI\Platform\Metadata\Metadata;
use Symfony\AI\Platform\Metadata\TokenUsage;
```

src/ai-bundle/doc/index.rst
chr-hertel, line 255: No
```suggestion
    use Symfony\AI\Platform\Metadata\TokenUsage;
```

chr-hertel:
Thank you @junaidbinfarooq.

chr-hertel:
Fixed some namespaces and docs on merging + vertexai support

PR #331, opened 2025-08-19
[MCP SDK] Example update to use proper logging

### LINKED ISSUE #306: [MCP SDK] Logging
Hi.
Example for running MCP Server uses 
```php
$logger = new SymfonyConsole\Logger\ConsoleLogger($output);
```
as logger. 

This won't work with STDIO transport 100%. 
Logs go to STDERR with proper verbosity.
Logs written not in JSON, but as string `[info] .....` which breaks tools like `npx @modelcontextprotocol/inspector` for example, and you won't see any logs there.

This can be changed, for example app can create own JSON logger (something like [this](https://github.com/ineersa/mcp-sql-server/blob/main/src/Logger/JsonConsoleLogger.php)), then tools start to see logs.

Image

Writing own logger is not best solution, and example is not really working out of the box.

Maybe example can be changed to utilize Monolog logger with STDERR stream handler and JSON formatter?

Also this will work only on debugging stage with `npx @modelcontextprotocol/inspector` and won't work with other tools. For example LM Studio or PHPStorm when they see output in STDERR they stop using tool since they think that it's actual error.

If need user still needs logging on server side, I guess logging to file would be best choice, maybe you could add some examples and explanations?

There is [Logging Capability](https://modelcontextprotocol.io/specification/2025-06-18/server/utilities/logging) in MCP protocol, which can be utilized to keep logs in 1 place (client in this case), but it's not implemented now.

ineersa:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #306
| License       | MIT

Added `monolog` as `dev` dependency.
Swapped `ConsoleLogger` to Monolog -> STDERR with JsonFormatter.
Added commentary with explanation

### DIFF
--- a/src/mcp-sdk/composer.json
+++ b/src/mcp-sdk/composer.json
@@ -22,7 +22,8 @@
         "phpstan/phpstan": "^2.1",
         "phpunit/phpunit": "^11.5",
         "symfony/console": "^6.4 || ^7.0",
-        "psr/cache": "^3.0"
+        "psr/cache": "^3.0",
+        "monolog/monolog": "^3.9"
     },
     "suggest": {
         "symfony/console": "To use SymfonyConsoleTransport for STDIO",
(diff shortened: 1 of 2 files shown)


chr-hertel:

PR #333, opened 2025-08-20
 Fix exceptions to use `sprintf` for string formatting

OskarStark:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | --
| License       | MIT

Replace string concatenation and interpolation with sprintf in all exception messages for consistency and better formatting.


chr-hertel:
I'm afraid it needs patches in the tests as well

OskarStark:
It does, will have a look soon

Issue #330, opened 2025-08-19
[Store] Add lifecycle management commands


Guikingone:
Hi 👋🏻 

As discussed in #285, the `Store` component must be able to `setup` and `drop` stores while using dedicated commands.

As the interface `ManagedStoreInterface` is introduced via #328, a new PR must be created to add the commands.

chr-hertel:
This should also enable us to do integration tests for those aspects of the bridges.

+ documentation

chr-hertel:
@Guikingone do you consider this done?

Guikingone:
Yes 👍

PR #335, opened 2025-08-20
[Store]: ManagedStoreInterface commands added

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | #330 
| License       | MIT

Hi 👋🏻 

As discussed in #330, here's the commands required to setup/drop the stores, by default, I moved the commands in the `Store` component but wasn't sure about it.


src/ai-bundle/config/console.php
```diff
@@ -0,0 +1,32 @@
...
+namespace Symfony\Component\DependencyInjection\Loader\Configurator;
+
+use Symfony\AI\Store\Command\DropStoreCommand;
+use Symfony\AI\Store\Command\SetupStoreCommand;
+
+return static function (ContainerConfigurator $container): void {
+    $container->services()
+        ->set('console.command.ai.setup_store', SetupStoreCommand::class)
+            ->args([
+                service('messenger.receiver_locator'),
```
OskarStark: receiver?

Guikingone: Yeah, duplicated code 😅

src/store/src/Command/DropStoreCommand.php
```diff
@@ -0,0 +1,96 @@
...
+            )
+        ;
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $io = new SymfonyStyle($input, $output);
+
+        $storeNames = $this->storeNames;
+        // do we want to set up only one store?
```
OskarStark: I think that's the main use case, if needed you can also run the command multiple times.

Guikingone: Wasn't sure about this one, if we only support one store, good for me, I'll update the commands.

src/ai-bundle/config/console.php
```diff
@@ -0,0 +1,32 @@
...
+
+use Symfony\AI\Store\Command\DropStoreCommand;
+use Symfony\AI\Store\Command\SetupStoreCommand;
+
+return static function (ContainerConfigurator $container): void {
+    $container->services()
+        ->set('console.command.ai.setup_store', SetupStoreCommand::class)
+            ->args([
+                service('ai.store_locator'),
+                [], // Receiver names
```
OskarStark:
```suggestion
                [], // Store names
```

src/ai-bundle/config/console.php
```diff
@@ -0,0 +1,32 @@
...
+            ])
+            ->tag('console.command')
+        ->set('console.command.ai.drop_store', DropStoreCommand::class)
+            ->args([
+                service('ai.store_locator'),
+                [], // Receiver names
```
OskarStark:
```suggestion
                [], // Store names
```

src/ai-bundle/src/AiBundle.php
```diff
@@ -115,11 +116,18 @@ public function loadExtension(array $config, ContainerConfigurator $container, C
...
-            $builder->setAlias(StoreInterface::class, reset($stores));
+
+        $references = [];
+        foreach ($stores as $storeName) {
+            $references[$storeName] = new Reference($storeName);
         }
 
+        $builder->getDefinition('ai.store_locator')->replaceArgument(0, $references);
+        $builder->getDefinition('console.command.ai.setup_store')->replaceArgument(0, $stores);
+        $builder->getDefinition('console.command.ai.drop_store')->replaceArgument(0, $stores);
```
OskarStark: I would only enable all this if we have at least one configured store, WDYT?

src/store/src/Command/DropStoreCommand.php
```diff
@@ -0,0 +1,96 @@
...
+        if ($input->mustSuggestArgumentValuesFor('store')) {
+            $suggestions->suggestValues($this->storeNames);
+        }
+    }
+
+    protected function configure(): void
+    {
+        $this
+            ->addArgument('store', InputArgument::OPTIONAL, 'Name of the store to drop')
+            ->setHelp(<<<EOF
```
OskarStark, line 54: AFAIK we can now add the help to the `#[AsCommand]` attribute

src/store/src/DroppableStoreInterface.php
OskarStark, line 1: Removed it in https://github.com/symfony/ai/pull/336, please rebase

src/ai-bundle/src/AiBundle.php
Guikingone: Agree

src/store/src/Command/DropStoreCommand.php
Guikingone: Depending on the version, the `help` attribute is not always supported, should we stick to the method or just force the Sf version?

OskarStark: Ah ok then let's keep it 👍🏻😊

src/ai-bundle/src/AiBundle.php
```diff
@@ -115,9 +116,18 @@ public function loadExtension(array $config, ContainerConfigurator $container, C
         foreach ($config['store'] ?? [] as $type => $store) {
             $this->processStoreConfig($type, $store, $builder);
         }
+
         $stores = array_keys($builder->findTaggedServiceIds('ai.store'));
-        if (1 === \count($stores)) {
-            $builder->setAlias(StoreInterface::class, reset($stores));
```
chr-hertel, line 140: unexpected to see this being removed - was it intentional?

src/ai-bundle/src/AiBundle.php
```diff
@@ -507,6 +517,7 @@ private function processStoreConfig(string $type, array $stores, ContainerBuilde
                     ->setArguments($arguments);
 
                 $container->setDefinition('ai.store.'.$type.'.'.$name, $definition);
+                $container->registerAliasForArgument('ai.store.'.$name, StoreInterface::class, (new Target($name.'Store'))->getParsedName());
```
chr-hertel, line 630: ah okay, that was unexpected as well :laughing:

chr-hertel:
> I moved the commands in the Store component but wasn't sure about it.

no no, that the right decision from my point of view :+1:

chr-hertel:
Alright, let's talk about functional testing here - that's important to me with that amount of bridges we can't go with manual.

My first idea was to bring in simple store examples, that would setup, add, query and drop. but that would be super similar to the rag ones - well without the agent tho.
so a) we adopt be commands in those examples, b) we bring in new one with only those for operations, and c) $yourIdea :man_shrugging: :laughing:

src/ai-bundle/src/AiBundle.php
Guikingone: Well, yes, I'm tired of using `#[Autowire]` attribute when injection stores so, here's the solution 😄

Guikingone: I can move this part to another PR if it's an issue here, that's not a big deal.

Guikingone:
What about improving the rag examples by adding the `drop` calls when it's possible? 

Maybe just before returning the content to the user?

Don't know if adding examples with commands brings any benefits to the user or we're talking about single-file commands that we can call in the existing examples 🤔

OskarStark:
I would keep the examples, but what about using setup and store in CI to see if they are working? Could be a follow up PR imho

OskarStark:
Docs are missing

chr-hertel:
If we could shift it to GitHub actions, that would be fine as well - and it's true, that there is little benefit to users with examples only setup-add-query-drop, true.

We already skipped it in #328 - so we should bring it in now.

src/ai-bundle/src/AiBundle.php
chr-hertel: nonoo, all good. is using `StoreInterface $store` still supported if only one is present than?

Guikingone: As long as you define a store, yes, the alias is defined for each of them so you can inject store even if there's only one of them defined.

src/ai-bundle/src/AiBundle.php
Guikingone: Yes, as the aliases are defined for each store, I'm not sure if we need this alias anymore 🤔

Guikingone:
I was wondering, shouldn't we use the `demo` in actions to show how to use the commands? 

We can build a container then make the commands accessible but we already have the `demo` project with configuration, maybe we can improve it and use it as a "functional" test in CI? 🤔

.github/workflows/integration-tests.yaml
```diff
@@ -0,0 +1,126 @@
...
+            chromadb:
+                image: chromadb/chroma
+                ports:
+                    - '8001:8000'
+
+        env:
+            SYMFONY_REQUIRE: ${{ matrix.symfony-version || '>=6.4' }}
+
+        steps:
+            - uses: actions/checkout@v4
```
OskarStark:
```suggestion
            - uses: actions/checkout@v5
```

examples/compose.yaml
```diff
@@ -56,7 +56,18 @@ services:
...
+      environment:
+          - IS_PERSISTENT=TRUE
+          - PERSIST_DIRECTORY=/chroma/chroma # this is the default path, change it as needed
+          - ANONYMIZED_TELEMETRY=FALSE
+      volumes:
+          - chroma_vlm:/chroma/chroma
+      ports:
+          - '8001:8000'
+
+    # Milvus services
```
OskarStark:
```suggestion
   # Milvus services
```

chr-hertel:
> I was wondering, shouldn't we use the `demo` in actions to show how to use the commands?
> 
> We can build a container then make the commands accessible but we already have the `demo` project with configuration, maybe we can improve it and use it as a "functional" test in CI? 🤔

The demo currently has only one store - and I like that simplicity there. But I get your point.

So I see two options, test the functional integration only on service level with a PHPUnit testcase neglecting the command level, or bring in a small example script with bootstrapping a console application and registering the commands there. i'm not sure we need a separate container image but go with the setup-php base. but maybe i'm missing something here.

Guikingone:
Let's add a small console + commands then 👍🏻

src/ai-bundle/src/AiBundle.php
chr-hertel: please revert this, see #355 

It should still be possible to just use `StoreInterface $store` if only store is configured - which is basically quite common, i'd say

Guikingone: BTW, regarding #355, as I brought the issue, do you want me to work on the fix?

Guikingone: Fix submitted in #358.

examples/commands/stores.php
```diff
@@ -0,0 +1,124 @@
...
+        '_commands',
+    ),
+    'milvus' => static fn (): MilvusStore => new MilvusStore(
+        http_client(),
+        env('MILVUS_HOST'),
+        env('MILVUS_API_KEY'),
+        env('MILVUS_DATABASE'),
+        '_commands',
+    ),
+    'mongodb' => static fn (): MongoDbStore => new MongoDbStore(
```
OskarStark, line 65: Not sure this is working with the local version. I think it's an atlas only feature.

cc @GromNaN

examples/compose.yaml
OskarStark, line 1: Can you fix the indention in a separate PR to keep it clean and the diff small?

examples/composer.json
OskarStark, line 1: Can we fix this separately?

examples/rag/neo4j.php
OskarStark, line 1: Those ones can be fixed separately

src/store/src/Bridge/ClickHouse/Store.php
```diff
@@ -25,12 +25,12 @@
 /**
  * @author Grégoire Pineau <lyrixx@lyrixx.info>
  */
-class Store implements ManagedStoreInterface, StoreInterface
+final readonly class Store implements ManagedStoreInterface, StoreInterface
```
OskarStark: Are the other stores final? I would not make it final, as Symfony is in general more open

Guikingone: Yes, all of them are `final` except this one.

examples/commands/stores.php
Guikingone: Nice catch, I'm removing it.

GromNaN: Vector Search Indexes are available in MongoDB Atlas, Enterprise and Atlas Local. You can run it locally with the [`mongodb/mongodb-atlas-local` docker image](https://www.mongodb.com/docs/atlas/cli/current/atlas-cli-deploy-docker/).

Guikingone: So in theory, we could test vectors locally? 🤔

GromNaN: Yes, that's how I'm testing this feature in the Laravel-MongoDB package. 

  https://github.com/mongodb/laravel-mongodb/blob/c483b99a69ea9916d17719b86e7e5ea2f8b44c8a/.github/workflows/build-ci-atlas.yml#L45

OskarStark: Many thanks 🙏

src/store/src/Bridge/ClickHouse/Store.php
OskarStark: Fixed in 
* https://github.com/symfony/ai/pull/366

.github/workflows/integration-tests.yaml
```diff
@@ -0,0 +1,144 @@
...
+                    POSTGRES_USER: postgres
+                    POSTGRES_PASSWORD: postgres
+                ports:
+                    - '5432:5432'
+
+        env:
+            SYMFONY_REQUIRE: ${{ matrix.symfony-version || '>=6.4' }}
+
+        steps:
+            - uses: actions/checkout@v4
```
OskarStark:
```suggestion
            - uses: actions/checkout@v5
```

examples/composer.json
```diff
@@ -8,16 +8,18 @@
         "async-aws/bedrock-runtime": "^1.1",
         "codewithkyrian/transformers": "^0.6.1",
         "doctrine/dbal": "^3.3|^4.0",
+        "mongodb/mongodb": "^2.1",
         "mrmysql/youtube-transcript": "^0.0.5",
         "php-http/discovery": "^1.20",
-        "probots-io/pinecone-php": "^1.1",
+        "probots-io/pinecone-php": "^1.0",
```
OskarStark: why the downgrade?

examples/compose.yaml
OskarStark: I streamlined the compose.yaml file in
* #379

examples/composer.json
```diff
@@ -8,16 +8,18 @@
...
         "psr/http-factory-implementation": "*",
         "symfony/ai-agent": "@dev",
         "symfony/ai-platform": "@dev",
         "symfony/ai-store": "@dev",
         "symfony/cache": "^6.4|^7.0",
         "symfony/console": "^6.4|^7.0",
         "symfony/css-selector": "^6.4|^7.0",
+        "symfony/dependency-injection": "^7.3",
```
OskarStark: 6.4 missing

examples/composer.json
Guikingone: Because it's already `^1.0` in the `Store`package: 

```json
"require-dev": {
    "codewithkyrian/chromadb-php": "^0.2.1 || ^0.3 || ^0.4",
    "doctrine/dbal": "^3.3 || ^4.0",
    "mongodb/mongodb": "^1.21 || ^2.0",
    "phpstan/phpstan": "^2.0",
    "phpunit/phpunit": "^11.5",
    "probots-io/pinecone-php": "^1.0",
    "symfony/cache": "^6.4 || ^7.1",
    "symfony/console": "^6.4 || ^7.1",
    "symfony/dependency-injection": "^6.4 || ^7.1"
},
```

Without this downgrade, we can't test the `Pinecone` store for each PHP version.

OskarStark: Can you provide a PR with exactly this explanation? I would like to stay focused on the topic here. I also streamlined most of the compose.yaml changes, so you should be able to rebase

chr-hertel: I know it's my fault that this is growing, but we can see the value already - really nice! :+1: 

I would also propose to split these changes into separate PRs each:
* New example `examples/rag/clickhouse.php` 
* Bug fix in `src/store/src/Bridge/Local/CacheStore.php`
* Bug fix in `src/store/src/Bridge/SurrealDb/Store.php`
* The env defaults in `examples/.env`

src/ai-bundle/src/AiBundle.php
chr-hertel: but we're not running in the same issue here like #355?

examples/composer.json
Guikingone: I'll open the PRs today 🙂

src/ai-bundle/src/AiBundle.php
Guikingone: I added the same logic as for agents: 

```php
# ...

$stores = array_keys($builder->findTaggedServiceIds('ai.store'));

if (1 === \count($stores)) {
    $builder->setAlias(StoreInterface::class, reset($stores));
}
```

src/ai-bundle/config/console.php
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+namespace Symfony\Component\DependencyInjection\Loader\Configurator;
+
+use Symfony\AI\Store\Command\DropStoreCommand;
+use Symfony\AI\Store\Command\SetupStoreCommand;
+
+return static function (ContainerConfigurator $container): void {
+    $container->services()
+        ->set('console.command.ai.setup_store', SetupStoreCommand::class)
```
OskarStark:
```suggestion
        ->set('ai.command.setup_store', SetupStoreCommand::class)
```
?

src/ai-bundle/config/console.php
```diff
@@ -0,0 +1,32 @@
...
+            ->args([
+                service('ai.store_locator'),
+                [], // Store names
+            ])
+            ->tag('console.command')
+        ->set('console.command.ai.drop_store', DropStoreCommand::class)
```
OskarStark:
```suggestion
        ->set('ai.command.drop_store', DropStoreCommand::class)
```

src/ai-bundle/src/AiBundle.php
```diff
@@ -94,6 +94,7 @@ public function configure(DefinitionConfigurator $definition): void
     public function loadExtension(array $config, ContainerConfigurator $container, ContainerBuilder $builder): void
     {
         $container->import('../config/services.php');
+        $container->import('../config/console.php');
```
OskarStark: this should only be registered once we have one store, right?

src/store/src/Command/SetupStoreCommand.php
```diff
@@ -0,0 +1,91 @@
...
+        if (!$this->storeLocator->has($storeName)) {
+            throw new RuntimeException(\sprintf('The "%s" store does not exist.', $storeName));
+        }
+
+        $store = $this->storeLocator->get($storeName);
+        if (!$store instanceof ManagedStoreInterface) {
+            $io->note(\sprintf('The "%s" store does not support setup.', $storeName));
+
+            return Command::FAILURE;
+        }
```
OskarStark: Should we move this to the `initialize` method?

src/store/src/Command/DropStoreCommand.php
```diff
@@ -0,0 +1,91 @@
...
+        if (!$this->storeLocator->has($storeName)) {
+            throw new RuntimeException(\sprintf('The "%s" store does not exist.', $storeName));
+        }
+
+        $store = $this->storeLocator->get($storeName);
+        if (!$store instanceof ManagedStoreInterface) {
+            $io->note(\sprintf('The "%s" store does not support be dropped.', $storeName));
+
+            return Command::FAILURE;
+        }
```
OskarStark: Should we move this to the `initialize` method?

src/ai-bundle/src/AiBundle.php
Guikingone: Not sure, we could maybe have commands in it that are not related to `stores`.

Maybe we could split the file into sub-files that only contains commands for specific components?

OskarStark: lets inline the command definition the in the services.php file and not register them when no stores. no need for a separate file

src/store/src/Command/DropStoreCommand.php
```diff
@@ -0,0 +1,96 @@
...
+    protected function initialize(InputInterface $input, OutputInterface $output)
+    {
+        $storeName = $input->getArgument('store');
+        if (!$this->storeLocator->has($storeName)) {
+            throw new RuntimeException(\sprintf('The "%s" store does not exist.', $storeName));
+        }
+
+        $store = $this->storeLocator->get($storeName);
+        if (!$store instanceof ManagedStoreInterface) {
+            throw new RuntimeException(\sprintf('The "%s" store does not support be dropped.', $storeName));
```
OskarStark:
```suggestion
            throw new RuntimeException(\sprintf('The "%s" store does not support to be dropped.', $storeName));
```

src/store/src/Command/SetupStoreCommand.php
```diff
@@ -0,0 +1,96 @@
...
+
+        $storeName = $input->getArgument('store');
+
+        $store = $this->storeLocator->get($storeName);
+
+        try {
+            $store->setup();
+            $io->success(\sprintf('The "%s" store was set up successfully.', $storeName));
+        } catch (\Exception $e) {
+            throw new RuntimeException(\sprintf('An error occurred while setting up the "%s" store: ', $storeName).$e->getMessage(), 0, $e);
```
OskarStark:
```suggestion
            throw new RuntimeException(\sprintf('An error occurred while setting up the "%s" store: ', $storeName).$e->getMessage(), previous: $e);
```

src/store/src/Command/DropStoreCommand.php
```diff
@@ -0,0 +1,96 @@
...
+
+        $storeName = $input->getArgument('store');
+
+        $store = $this->storeLocator->get($storeName);
+
+        try {
+            $store->drop();
+            $io->success(\sprintf('The "%s" store was dropped successfully.', $storeName));
+        } catch (\Exception $e) {
+            throw new RuntimeException(\sprintf('An error occurred while dropping the "%s" store: ', $storeName).$e->getMessage(), 0, $e);
```
OskarStark:
```suggestion
            throw new RuntimeException(\sprintf('An error occurred while dropping the "%s" store: ', $storeName).$e->getMessage(), previous: $e);
```

src/store/doc/index.rst
```diff
@@ -77,6 +77,28 @@ Supported Stores
...
+
+.. code-block:: yaml
+
+    # config/packages/ai.yaml
+    ai:
+        # ...
+    store:
+        chroma_db:
+            symfonycon:
+                collection: 'symfony_blog'
```
OskarStark: indention looks wrong, `store` must be under `ai`

src/store/src/Command/SetupStoreCommand.php
```diff
@@ -0,0 +1,96 @@
...
+    {
+        if ($input->mustSuggestArgumentValuesFor('store')) {
+            $suggestions->suggestValues($this->storeNames);
+        }
+    }
+
+    protected function configure(): void
+    {
+        $this
+            ->addArgument('store', InputArgument::OPTIONAL, 'Name of the store to setup')
```
chr-hertel: why is it `InputArgument::OPTIONAL`? should be required, no?

chr-hertel: same with the other command

OskarStark:
Needs a rebase after 
* https://github.com/symfony/ai/pull/395

src/store/src/Command/DropStoreCommand.php
```diff
@@ -0,0 +1,96 @@
...
+    {
+        if ($input->mustSuggestArgumentValuesFor('store')) {
+            $suggestions->suggestValues($this->storeNames);
+        }
+    }
+
+    protected function configure(): void
+    {
+        $this
+            ->addArgument('store', InputArgument::REQUIRED, 'Name of the store to drop')
```
OskarStark, line 52: I think its common to add a `--force` option?

src/ai-bundle/config/services.php
```diff
@@ -145,5 +148,24 @@
...
+            ->tag('console.command')
+        ->set('ai.command.drop_store', DropStoreCommand::class)
+            ->args([
+                service('ai.store_locator'),
+                [], // Store names
+            ])
+            ->tag('console.command')
+
+        // Discovery
+        ->set('ai.store_locator', ServiceLocator::class)
```
OskarStark: do we need this service or can we do the same like I do for the agents?

Guikingone: To be honest? Everything is possible at this stage 😄 

My PR is based on the implementation for Messenger, it can easily be updated to use your approach, I didn't look at it until you asked 😅

Guikingone: I moved the commands to your approach, easier to handle in the bundle, nice catch 👍🏻

OskarStark (approved):

OskarStark:
Needs a rebase after 
* #459

Guikingone:
Rebase done 👍🏻

OskarStark:

PR #339, opened 2025-08-21
[Platform][Ollama] Support streaming output

### LINKED ISSUE #338: [Platform] Support streaming output for Ollama
Currently, if you set `stream` to `true` when setting Ollama options it will throw an error.

example:
```
$result = $agent->call(
    $messages, [
        'stream' => true,
    ]
);
```
results in:
```
PHP Fatal error:  Uncaught Symfony\Component\HttpClient\Exception\JsonException: Syntax error for "http://localhost:11434/api/chat". in /.../vendor/symfony/http-client/Response/CommonResponseTrait.php:90
Stack trace:
#0 /.../vendor/symfony/ai-platform/src/Result/RawHttpResult.php(28): Symfony\Component\HttpClient\Response\AsyncResponse->toArray()
#1 /.../vendor/symfony/ai-platform/src/Bridge/Ollama/OllamaResultConverter.php(47): Symfony\AI\Platform\Result\RawHttpResult->getData()
#2 /.../vendor/symfony/ai-platform/src/Result/ResultPromise.php(48): Symfony\AI\Platform\Bridge\Ollama\OllamaResultConverter->convert()
#3 /.../vendor/symfony/ai-platform/src/Result/ResultPromise.php(37): Symfony\AI\Platform\Result\ResultPromise->await()
#4 /.../vendor/symfony/ai-agent/src/Agent.php(82): Symfony\AI\Platform\Result\ResultPromise->getResult()
#5 /.../src/app.php(26): Symfony\AI\Agent\Agent->call()
#6 {main}
  thrown in /.../vendor/symfony/http-client/Response/CommonResponseTrait.php on line 90
```

logicalor:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| Docs?         | yes/no
| Issues        | Fix #338
| License       | MIT

Both a bug fix and a feature?  Support streaming output from Ollama, preventing an error being thrown if streaming output is sent with the chat request.

### DIFF
--- a/src/platform/src/Bridge/Ollama/OllamaMessageChunk.php
+++ b/src/platform/src/Bridge/Ollama/OllamaMessageChunk.php
@@ -1,0 +1,45 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Platform\Bridge\Ollama;
+
+/**
+ * @author Shaun Johnston <shaun@snj.au>
+ */
+final readonly class OllamaMessageChunk
+{
+    /**
+     * @param array<string, mixed> $message
+     */
+    public function __construct(
+        public readonly string $model,
+        public readonly \DateTimeImmutable $created_at,
+        public readonly array $message,
+        public readonly bool $done,
+    ) {
+    }
+
+    public function __toString(): string
+    {
+        // Return the assistant's message content if available
+        return $this->message['content'] ?? '';
+    }
+
+    public function getContent(): ?string
+    {
+        return $this->message['content'] ?? null;
+    }
+
+    public function getRole(): ?string
+    {
+        return $this->message['role'] ?? null;
+    }
+}
--- a/src/platform/src/Bridge/Ollama/OllamaResultConverter.php
+++ b/src/platform/src/Bridge/Ollama/OllamaResultConverter.php
@@ -15,12 +15,17 @@ use Symfony\AI\Platform\Exception\RuntimeException;
 use Symfony\AI\Platform\Model;
 use Symfony\AI\Platform\Result\RawResultInterface;
 use Symfony\AI\Platform\Result\ResultInterface;
+use Symfony\AI\Platform\Result\StreamResult;
 use Symfony\AI\Platform\Result\TextResult;
 use Symfony\AI\Platform\Result\ToolCall;
 use Symfony\AI\Platform\Result\ToolCallResult;
 use Symfony\AI\Platform\Result\VectorResult;
 use Symfony\AI\Platform\ResultConverterInterface;
 use Symfony\AI\Platform\Vector\Vector;
+use Symfony\Component\HttpClient\Chunk\FirstChunk;
+use Symfony\Component\HttpClient\Chunk\LastChunk;
+use Symfony\Component\HttpClient\EventSourceHttpClient;
+use Symfony\Contracts\HttpClient\ResponseInterface;
 
 /**
  * @author Christopher Hertel <mail@christopher-hertel.de>
@@ -34,6 +39,10 @@ final readonly class OllamaResultConverter implements ResultConverterInterface
 
     public function convert(RawResultInterface $result, array $options = []): ResultInterface
     {
+        if ($options['stream'] ?? false) {
+            return new StreamResult($this->convertStream($result->getObject()));
+        }
+
         $data = $result->getData();
 
         return \array_key_exists('embeddings', $data)
@@ -82,5 +91,27 @@ final readonly class OllamaResultConverter implements ResultConverterInterface
                 $data['embeddings'],
             ),
         );
+    }
+
+    private function convertStream(ResponseInterface $result): \Generator
+    {
+        foreach ((new EventSourceHttpClient())->stream($result) as $chunk) {
+            if ($chunk instanceof FirstChunk || $chunk instanceof LastChunk) {
+                continue;
+            }
+
+            try {
+                $data = json_decode($chunk->getContent(), true, 512, \JSON_THROW_ON_ERROR);
+            } catch (\JsonException $e) {
+                throw new RuntimeException('Failed to decode JSON: '.$e->getMessage());
+            }
+
+            yield new OllamaMessageChunk(
+                $data['model'],
+                new \DateTimeImmutable($data['created_at']),
+                $data['message'],
+                $data['done'],
+            );
+        }
     }
 }
--- a/examples/ollama/stream.php
+++ b/examples/ollama/stream.php
@@ -1,0 +1,36 @@
+<?php
+
+/*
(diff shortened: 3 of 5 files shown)


OskarStark:
Can you please add an example in examples/ directory? Thanks

logicalor:
> Can you please add an example in examples/ directory? Thanks

@OskarStark updated with an example

OskarStark:
Thanks

It looks like your committer email is not associated with your GitHub account

logicalor:
> Thanks
> 
> It looks like your committer email is not associated with your GitHub account

@OskarStark thanks, hadn't noticed that. Added.

src/platform/src/Bridge/Ollama/OllamaMessageChunk.php
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Platform\Bridge\Ollama;
+
+class OllamaMessageChunk
```
Guikingone: Could be `final` and `readonly` maybe?

I'm not sure if the format used by Ollama is shared by OpenAI and others 🤔

logicalor: @Guikingone good point re: `final` and `readonly`.  I will update.

Ollama does provide an OpenAI-compatible endpoint at `/v1/chat/completions` (https://ollama.com/blog/openai-compatibility) meaning you would be able to point an OpenAI platform at it and it should work.

This lib is using the native Ollama HTTP API (https://github.com/ollama/ollama/blob/main/docs/api.md).  The chunk format from streaming mode is different to OpenAI's chunk format (https://platform.openai.com/docs/api-reference/chat_streaming/streaming)

src/platform/src/Bridge/Ollama/OllamaMessageChunk.php
```diff
@@ -0,0 +1,65 @@
...
+     */
+    public static function fromArray(array $data): self
+    {
+        return new self(
+            $data['model'] ?? '',
+            new \DateTimeImmutable($data['created_at'] ?? ''),
+            $data['message'] ?? [],
+            $data['done'] ?? false
+        );
+    }
```
Guikingone: Personal opinion here, if we only use this method once, I don't see the benefits, we could move the `new self...` to the `fromJsonString` method.

src/platform/src/Bridge/Ollama/OllamaResultConverter.php
```diff
@@ -83,4 +92,18 @@ public function doConvertEmbeddings(array $data): ResultInterface
...
+        foreach ((new EventSourceHttpClient())->stream($result) as $chunk) {
+            if ($chunk instanceof FirstChunk || $chunk instanceof LastChunk) {
+                continue;
+            }
+
+            $msg = OllamaMessageChunk::fromJsonString($chunk->getContent());
+            if ($msg) {
+                yield $msg;
+            }
+        }
```
Guikingone, line 115: Question for @chr-hertel (not relevant for the review), isn't the same situation that we're facing in #324?

src/platform/src/Bridge/Ollama/OllamaMessageChunk.php
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Platform\Bridge\Ollama;
+
+final readonly class OllamaMessageChunk
```
Guikingone, line 17: BTW, do not forget to add your informations here as author.

src/platform/src/Bridge/Ollama/OllamaMessageChunk.php
logicalor: @Guikingone thanks, I agree - I've removed the method and put the logic into `fromJsonString`.

src/platform/src/Bridge/Ollama/OllamaMessageChunk.php
logicalor: Updated 🙏

src/platform/src/Bridge/Ollama/OllamaResultConverter.php
chr-hertel: It's not since this is only working on the response in the ResultConverter, which is only called when finally fetching the content. In #324 the `stream` is called already in the client

examples/ollama/chat-ollama-streaming.php
```diff
@@ -0,0 +1,35 @@
...
+    Message::ofUser('Tina has one brother and one sister. How many sisters do Tina\'s siblings have?'),
+);
+
+// Stream the response
+$result = $agent->call($messages, ['stream' => true]);
+
+// Emit each chunk as it is received
+foreach ($result as $chunk) {
+    echo $chunk->getContent();
+}
```
chr-hertel:
```suggestion
}
echo \PHP_EOL;
```

examples/ollama/chat-ollama-streaming.php
chr-hertel, line 1: Let's call the file only `stream.php` - it is more consistent with the other stream examples, and `ollama` is basically the folder already

chr-hertel:
which ollama version do you use?
image

src/platform/src/Bridge/Ollama/OllamaMessageChunk.php
```diff
@@ -0,0 +1,60 @@
...
+        // Return the assistant's message content if available
+        return $this->message['content'] ?? '';
+    }
+
+    public static function fromJsonString(string $json): ?self
+    {
+        $data = json_decode($json, true);
+        if (!\is_array($data)) {
+            return null;
+        }
```
chr-hertel: please move this to the result converter - it's their job to convert to data. fine with having the value object, but let's keep it simple than :+1:

chr-hertel (changes requested):
Thanks @logicalor for reporting and working on this - we need this :)

Made some comments, but i think all in all we're close to merging it :+1:

src/platform/src/Bridge/Ollama/OllamaMessageChunk.php
logicalor: @chr-hertel I have removed the fromJsonString method and put the logic into `OllamaResultConverter::convertStream` instead.

logicalor:
> which ollama version do you use?

@chr-hertel I was running `0.10.1` but just upgraded to `0.11.7`.

examples/ollama/chat-ollama-streaming.php
logicalor: Updated 👍

chr-hertel (approved):
Works & looks good :+1:

chr-hertel:
Thank you @logicalor.

logicalor:
> Thank you @logicalor.

🙏

PR #340, opened 2025-08-21
[Store][ChromaDB] Add query filtering

valx76:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| License       | MIT

Very simple update allowing to add ChromaDB filters using the `$options` array parameter (actually unused for ChromaDB).
Before, we needed to create kind of a _fake_ decorator as I did [here](https://github.com/valx76/SymfonyAI-POC/blob/master/src/AI/StoreDecorator.php) (I didn't find any other way)..

I noticed the `query` method doesn't have any test yet so I imagine it'll be done in the future, which is why there is no test update in this PR.

### DIFF
--- a/src/store/src/Bridge/ChromaDb/Store.php
+++ b/src/store/src/Bridge/ChromaDb/Store.php
@@ -52,12 +52,17 @@ final readonly class Store implements StoreInterface
         $collection->add($ids, $vectors, $metadata, $originalDocuments);
     }
 
+    /**
+     * @param array{where?: array<string, string>, whereDocument?: array<string, mixed>} $options
+     */
     public function query(Vector $vector, array $options = []): array
     {
         $collection = $this->client->getOrCreateCollection($this->collectionName);
         $queryResponse = $collection->query(
             queryEmbeddings: [$vector->getData()],
             nResults: 4,
+            where: $options['where'] ?? null,
+            whereDocument: $options['whereDocument'] ?? null,
         );
 
         $documents = [];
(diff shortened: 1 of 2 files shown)


OskarStark:
Would love to see some tests for this store in this PR, thanks.

Should we define the array shape for $options here, which keys are allowed?

src/store/src/Bridge/ChromaDb/Store.php
```diff
@@ -52,12 +52,19 @@ public function add(VectorDocument ...$documents): void
         $collection->add($ids, $vectors, $metadata, $originalDocuments);
     }
 
+    /**
+     * @param array{'where'?: array<string, string>, 'whereDocument'?: array<string, mixed>} $options
```
OskarStark:
```suggestion
     * @param array{where?: array<string, string>, whereDocument?: array<string, mixed>} $options
```

src/store/src/Bridge/ChromaDb/Store.php
```diff
@@ -52,12 +52,19 @@ public function add(VectorDocument ...$documents): void
...
+     *
+     * @return VectorDocument[]
```
OskarStark:
```suggestion
```
Should be already on the interface, right?

valx76: You are right!

valx76:
Will try the tests now :)

OskarStark (approved):

OskarStark:
Thank you @valx76.

OskarStark:
> Will try the tests now :)

Thanks, feel free to provide a new PR then if you find some time. 👍

valx76:
> > Will try the tests now :)
> 
> Thanks, feel free to provide a new PR then if you find some time. 👍

Will do! 😄

PR #345, opened 2025-08-24
[Agent][Platform] Add `Metadata` to `MessageBag`

### LINKED ISSUE #344: Add metadata to MessageBag
Hi @chr-hertel @OskarStark,

Every time an agent is called I'd like to log the call.

For this, I can either
- decorated all the Agents with a CompilerPass doing the log.
- or implements an Input/Output processor to do it.
 
In the log I plan to write data like the model/platform use but also some extra data like an id/name/version of the prompt.

I feel odd passing such message-related information in `options` field and think the messageBag should be aware of such data because if I call three time an agent with the same message bag, the same id/name/version should be sent to every time.

Do you agree with something like a `MetadataAwareMessageBagInterface` implemented by MessageBag ?
Do you see another solution ?

I know I can implements my own MessageBag (`implements MessageBagInterface`) but I thought I wouldn't the only use with such need, and since `MessageBag` is marked with `@final`, it's not recommended to extends it.

VincentLanglet:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | Fix #344
| License       | MIT


chr-hertel (approved):

chr-hertel (changes requested):
The `MessageBagTest` should be extended, please.

chr-hertel (approved):

chr-hertel:
Thank you @VincentLanglet.

PR #347, opened 2025-08-24
Add TokenOutputProcessor for Gemini

VincentLanglet:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes>
| Docs?         | no 
| Issues        | Fix #... 
| License       | MIT

According to https://ai.google.dev/api/generate-content?hl=en#UsageMetadata


OskarStark (commented):
Can you please add an example in examples/ folder? Thanks

src/platform/tests/Bridge/Gemini/TokenOutputProcessorTest.php
```diff
@@ -0,0 +1,162 @@
...
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Bridge\Gemini;
```
OskarStark: Looks wrong

chr-hertel:
Should wait for #329

VincentLanglet:
> Should wait for #329

Yeah, I will rebase after ;)

src/platform/src/Bridge/Gemini/TokenOutputProcessor.php
```diff
@@ -0,0 +1,56 @@
...
+use Symfony\Contracts\HttpClient\ResponseInterface;
+
+final class TokenOutputProcessor implements OutputProcessorInterface
+{
+    public function processOutput(Output $output): void
+    {
+        if ($output->result instanceof StreamResult) {
+            // Streams have to be handled manually as the tokens are part of the streamed chunks
+            return;
+        }
```
junaidbinfarooq, line 27: @VincentLanglet
From my experience of working with the **Vertex AI**, the response does include _token usage_ information even when using streaming. Please check if it is available when you directly access the **Gemini AI** API.
For reference, have a look at the `TokenOutputProcessor` class in the **Vertex AI** PR.
You could probably even compose this one from that service.

VincentLanglet: I looked at your PR and I dunno if it should be this way.

All existing TokenOutputProcessor are doing this way currently and the message `// Streams have to be handled manually as the tokens are part of the streamed chunks` does not seems to say the usage cannot be found, but it's inside the stream and (maybe ?) it might not be recommended to iterate on the whole stream just to look for this data...
For instance some Generators can only be read once, so it might be annoying to consume it just for the token usage and loosing all the content...

So I think it more a general issue (about how all the TokenOutputProcessor should handle StreamResult and not just mine) and I would recommend to do it in another issue/PR. WDYT @junaidbinfarooq ?

junaidbinfarooq: @VincentLanglet 
I look at it in a different way, which is as follows:
- A contributor added a token usage processor to the OpenAI bridge initially, IIRC, but for some reason did not choose to process the stream result for token usage, or perhaps it was not available there (not all bridges are supposed to work in the same way).
- This was followed by another contributor who added a similar service to the Mistral bridge and followed the lead from the previous implementation.
- Yet again, another implementation was added by you here, and you followed suit.
- My implementation obviously preceded yours, but I chose to process the token usage even when streaming the results as it was available there.

junaidbinfarooq: > So I think it more a general issue (about how all the TokenOutputProcessor should handle StreamResult and not just mine) and I would recommend to do it in another issue/PR.

I do not see it as appropriate to leave this unimplemented, only to be followed up by another PR to do it without knowing the reason(s) behind it being unimplemented in other PRs.
As far as the existing implementations are concerned, yes, it can be done in a separate PR IMO, as those PRs are already done and not still open, while yours is.
But this is better left to the maintainers who have the final say.

VincentLanglet: > A contributor added a token usage processor to the OpenAI bridge initially, IIRC, but for some reason did not choose to process the stream result for token usage, or perhaps it was not available there (not all bridges are supposed to work in the same way).

Contributors are @Renrhaf and @DZunke ; they might want to explain themself about it rather making suppositions.

> but I chose to process the token usage even when streaming the results as it was available there.

And yet you totally ignored the issues I exposed that some Generators could only be read once, so it might be annoying to consume it just for the token usage and loosing all the content...

> But this is better left to the maintainers who have the final say.

Sure, and anyway this won't be merged before https://github.com/symfony/ai/pull/329 so it should be the focus

chr-hertel: Fine by me to go incremental here and skip token usage for streamed responses at the first shot.

VincentLanglet:
> Should wait for #329

Since #329 seems currently stuck/unfinished, I wonder if this PR couldn't be review/merged first and then https://github.com/symfony/ai/pull/329 will rebased it.

This way I'll be able to use my Processor @chr-hertel :)

chr-hertel:
Ready for rebase :+1:

examples/gemini/token-metadata.php
```diff
@@ -0,0 +1,39 @@
...
+$result = $agent->call($messages);
+
+$metadata = $result->getMetadata();
+$tokenUsage = $metadata->get('token_usage');
+
+assert($tokenUsage instanceof TokenUsage);
+
+echo 'Utilized Tokens: '.$tokenUsage->totalTokens.\PHP_EOL;
+echo '-- Prompt Tokens: '.$tokenUsage->promptTokens.\PHP_EOL;
+echo '-- Completion Tokens: '.$tokenUsage->completionTokens.\PHP_EOL;
```
OskarStark: after rebase, use
```suggestion
print_token_usage($result->getMetadata());
```

src/platform/src/Result/Metadata/TokenUsage.php
```diff
@@ -46,6 +48,7 @@ public function jsonSerialize(): array
             'prompt_tokens' => $this->promptTokens,
             'completion_tokens' => $this->completionTokens,
             'thinking_tokens' => $this->thinkingTokens,
+            'tool_tokens' => $this->toolTokens,
```
OskarStark: `tool_tokens` should be added to `examples/bootstrap.php` `print_token_usage` function

chr-hertel (changes requested):
Also needs the bundle registration now :)

OskarStark:
> Also needs the bundle registration now :)

Indeed, I missed that

OskarStark (approved):

VincentLanglet:
Ready @chr-hertel

OskarStark:
Thank you @VincentLanglet.

PR #348, opened 2025-08-24
[Agent] Expose Agent model

VincentLanglet:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no 
| Issues        | Fix #... 
| License       | MIT

Similar to https://github.com/symfony/ai/issues/344 @OskarStark 
Do you think it could be ok to expose the Agent model ?

It would be useful to access data like model name for logging purpose (without the need for reflection class).

But I'm dunno if it should be in the AgentInterface, your call.

### DIFF
--- a/src/agent/src/Agent.php
+++ b/src/agent/src/Agent.php
@@ -54,6 +54,11 @@ final readonly class Agent implements AgentInterface
         $this->outputProcessors = $this->initializeProcessors($outputProcessors, OutputProcessorInterface::class);
     }
 
+    public function getModel(): Model
+    {
+        return $this->model;
+    }
+
     /**
      * @param array<string, mixed> $options
      *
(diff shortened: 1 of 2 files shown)


chr-hertel:
My heart always feels a bit sad when exposing stuff, but i think your use-case is valid :+1: 

Regarding the interface I'm a bit torn but currently in favor of adding it as well. :thinking:

VincentLanglet:
> Regarding the interface I'm a bit torn but currently in favor of adding it as well. 🤔

I was hesitating about it because:
- Symfony codebase itself currently never use AgentInterface::getModel
- Some people might implements an `AgentInterface` with multiples models (based on the option ?)

Do you confirm I should update the interface in this PR @chr-hertel or you prefer another opinion ?

chr-hertel (approved):
Let's keep it like this - no change in the interface for now

OskarStark:
Thank you @VincentLanglet.

PR #350, opened 2025-08-24
 Remove MessageBagInterface

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | 
| License       | MIT

The MessageBag was final for some time and the interface was needed to bring in more capabilities that users want to have on their end. The interface was quite heavy tho and changes would also be quite impactful - not saying adding a method to a class that gets extended is fun, but still less annoying from maintenance point of view - i hope :grimacing:


OskarStark (approved):

OskarStark:
Thank you @chr-hertel.

Issue #346, opened 2025-08-24
[AI Bundle] Provide an easy way to add input/output processor to an agent via config


VincentLanglet:
Looking at https://github.com/symfony/ai/blob/main/src/ai-bundle/config/options.php
and https://github.com/symfony/ai/blob/main/src/ai-bundle/src/AiBundle.php

I don't see an easy way to add an input/output processor to a platform or an agent.
They currently only added with custom config like
- structured_output
- system_prompt
- ...

So it's not easy to add a custom input/output processor. Am I wrong ?

VincentLanglet:
I'm open to do the PR but I'd like your opinion on how the config should be modified @OskarStark @chr-hertel :)

chr-hertel:
I think two ways would be great:

1. adding it via config on the agent - especially if there are third party services that you want as a input/output processor
2. adding it on your own code with an attribute `#[AsOutputProcessor(agent: 'xyz')` 

makes sense?

OskarStark:
Is this done?

VincentLanglet:
https://github.com/symfony/ai/pull/352 is not merged yet

chr-hertel:
Implemented with #352

PR #352, opened 2025-08-25
[AI Bundle][Agent] Add `#[AsInputProcessor]` and `#[AsOutputProcessor]`

VincentLanglet:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | TODO
| Issues        | Fix #... 
| License       | MIT

Reference https://github.com/symfony/ai/issues/346#issuecomment-3218405736

cc @chr-hertel (@OskarStark) I'd like your opinion on this
- I added a priority param, I feel like it could be useful to handle the order of the processor
- I'd like your opinion about what should be the default priority of the existing processor (agent_processor, structured_output_processor, system_prompt)

### DIFF
--- a/src/agent/src/Attribute/AsInputProcessor.php
+++ b/src/agent/src/Attribute/AsInputProcessor.php
@@ -1,0 +1,26 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Agent\Attribute;
+
+#[\Attribute(\Attribute::TARGET_CLASS | \Attribute::IS_REPEATABLE)]
+final readonly class AsInputProcessor
+{
+    /**
+     * @param string|null $agent the service id of the agent which will use this processor,
+     *                           null to register this processor for all existing agents
+     */
+    public function __construct(
+        public ?string $agent = null,
+        public int $priority = 0,
+    ) {
+    }
+}
--- a/src/agent/src/Attribute/AsOutputProcessor.php
+++ b/src/agent/src/Attribute/AsOutputProcessor.php
@@ -1,0 +1,26 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Agent\Attribute;
+
+#[\Attribute(\Attribute::TARGET_CLASS | \Attribute::IS_REPEATABLE)]
+final readonly class AsOutputProcessor
+{
+    /**
+     * @param string|null $agent the service id of the agent which will use this processor,
+     *                           null to register this processor for all existing agents
+     */
+    public function __construct(
+        public ?string $agent = null,
+        public int $priority = 0,
+    ) {
+    }
+}
--- a/src/ai-bundle/config/services.php
+++ b/src/ai-bundle/config/services.php
@@ -68,8 +68,6 @@ return static function (ContainerConfigurator $container): void {
                 service('ai.agent.response_format_factory'),
                 service('serializer'),
             ])
-            ->tag('ai.agent.input_processor')
-            ->tag('ai.agent.output_processor')
 
         // tools
         ->set('ai.toolbox.abstract', Toolbox::class)
@@ -111,8 +109,6 @@ return static function (ContainerConfigurator $container): void {
             ])
         ->set('ai.tool.agent_processor', ToolProcessor::class)
             ->parent('ai.tool.agent_processor.abstract')
-            ->tag('ai.agent.input_processor')
-            ->tag('ai.agent.output_processor')
             ->arg('index_0', service('ai.toolbox'))
         ->set('ai.security.is_granted_attribute_listener', IsGrantedToolAttributeListener::class)
             ->args([
--- a/src/ai-bundle/doc/index.rst
+++ b/src/ai-bundle/doc/index.rst
@@ -150,6 +150,34 @@ Use the `Agent` service to leverage models and tools::
         }
     }
 
+**Register Processors**
+
+By default, all services implementing the ``InputProcessorInterface`` or the
+``OutputProcessorInterface`` interfaces are automatically applied to every ``Agent``.
+
+This behavior can be overridden/configured with the ``#[AsInputProcessor]`` and
+the ``#[AsOutputProcessor]`` attributes::
+
+    use Symfony\AI\Agent\Input;
+    use Symfony\AI\Agent\InputProcessorInterface;
+    use Symfony\AI\Agent\Output;
+    use Symfony\AI\Agent\OutputProcessorInterface;
+
+    #[AsInputProcessor(priority: 99)] // This applies to every agent
+    #[AsOutputProcessor(agent: 'my_agent_id')] // The output processor will only be registered for 'my_agent_id'
+    final readonly class MyService implements InputProcessorInterface, OutputProcessorInterface
+    {
+        public function processInput(Input $input): void
+        {
+            // ...
+        }
+
+        public function processOutput(Output $output): void
+        {
+            // ...
+        }
+    }
+
 **Register Tools**
 
 To use existing tools, you can register them as a service:
--- a/src/ai-bundle/src/AiBundle.php
+++ b/src/ai-bundle/src/AiBundle.php
@@ -14,6 +14,8 @@ namespace Symfony\AI\AiBundle;
 use Google\Auth\ApplicationDefaultCredentials;
 use Symfony\AI\Agent\Agent;
 use Symfony\AI\Agent\AgentInterface;
+use Symfony\AI\Agent\Attribute\AsInputProcessor;
+use Symfony\AI\Agent\Attribute\AsOutputProcessor;
 use Symfony\AI\Agent\InputProcessor\SystemPromptInputProcessor;
 use Symfony\AI\Agent\InputProcessorInterface;
 use Symfony\AI\Agent\OutputProcessorInterface;
@@ -22,6 +24,7 @@ use Symfony\AI\Agent\Toolbox\FaultTolerantToolbox;
 use Symfony\AI\Agent\Toolbox\Tool\Agent as AgentTool;
 use Symfony\AI\Agent\Toolbox\ToolFactory\ChainFactory;
 use Symfony\AI\Agent\Toolbox\ToolFactory\MemoryToolFactory;
+use Symfony\AI\AiBundle\DependencyInjection\ProcessorCompilerPass;
 use Symfony\AI\AiBundle\Exception\InvalidArgumentException;
 use Symfony\AI\AiBundle\Profiler\TraceablePlatform;
 use Symfony\AI\AiBundle\Profiler\TraceableToolbox;
@@ -83,6 +86,13 @@ use function Symfony\Component\String\u;
  */
 final class AiBundle extends AbstractBundle
 {
+    public function build(ContainerBuilder $container)
+    {
+        parent::build($container);
+
+        $container->addCompilerPass(new ProcessorCompilerPass());
+    }
+
     public function configure(DefinitionConfigurator $definition): void
     {
         $definition->import('../config/options.php');
@@ -143,10 +153,25 @@ final class AiBundle extends AbstractBundle
             ]);
         });
 
+        $builder->registerAttributeForAutoconfiguration(AsInputProcessor::class, static function (ChildDefinition $definition, AsInputProcessor $attribute): void {
+            $definition->addTag('ai.agent.input_processor', [
+                'agent' => $attribute->agent,
+                'priority' => $attribute->priority,
+            ]);
+        });
+
+        $builder->registerAttributeForAutoconfiguration(AsOutputProcessor::class, static function (ChildDefinition $definition, AsOutputProcessor $attribute): void {
+            $definition->addTag('ai.agent.output_processor', [
+                'agent' => $attribute->agent,
+                'priority' => $attribute->priority,
+            ]);
+        });
+
         $builder->registerForAutoconfiguration(InputProcessorInterface::class)
-            ->addTag('ai.agent.input_processor');
+            ->addTag('ai.agent.input_processor', ['tagged_by' => 'interface']);
         $builder->registerForAutoconfiguration(OutputProcessorInterface::class)
-            ->addTag('ai.agent.output_processor');
+            ->addTag('ai.agent.output_processor', ['tagged_by' => 'interface']);
+
         $builder->registerForAutoconfiguration(ModelClientInterface::class)
             ->addTag('ai.platform.model_client');
         $builder->registerForAutoconfiguration(ResultConverterInterface::class)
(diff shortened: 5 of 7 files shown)


src/ai-bundle/src/AiBundle.php
```diff
@@ -135,6 +149,20 @@ public function loadExtension(array $config, ContainerConfigurator $container, C
             ]);
         });
 
+        $builder->registerAttributeForAutoconfiguration(AsInputProcessor::class, static function (ChildDefinition $definition, AsInputProcessor $attribute): void {
+            $definition->addTag('ai.input_processor', [
```
VincentLanglet: There is also
```
        $builder->registerForAutoconfiguration(InputProcessorInterface::class)
            ->addTag('ai.agent.input_processor');
        $builder->registerForAutoconfiguration(OutputProcessorInterface::class)
            ->addTag('ai.agent.output_processor');
```
This feel like duplicated tags... but the purpose is different.

Should i remove the autoconfiguration for InputProcessorInterface/OutputProcessorInterface (which is currently not used anywhere in the symfony code) to use `ai.agent.*` tag ?

VincentLanglet:
I think it's ready for a first review @chr-hertel @OskarStark to get your opinion on this :)

src/ai-bundle/src/AiBundle.php
junaidbinfarooq: The input and output processors already implement from `InputProcessorInterface` and `OutputProcessorInterface` interfaces and are automatically registered for autoconfiguration at lines 138-141.
Would there be an extra case(s) that are not already covered by this, for which we need the new attributes? Without a genuine use case, the new attributes are seemingly adding indirection to the autoregistration process IMO?

junaidbinfarooq: I added a similar attribute in one of my recent PRs, as there was a need to register a particular subset of services in addition to them implementing the `OutputProcessorInterface` interface and being autoregistered here. But I was still asked by @chr-hertel to remove the new attribute and instead register those services manually here.

VincentLanglet: > Would there be an extra case(s) that are not already covered by this, for which we need the new attributes? Without a genuine use case, the new attributes are seemingly adding indirection to the autoregistration process IMO?

Sure:
- I invite you to read https://github.com/symfony/ai/issues/346#issuecomment-3218405736 first
- You can see that AsInputProcessor reference an agent and a priority while InputProcessorInterface not. You'll implements the interface but then you can add `AsInputProcessor(agent: 'a'), AsInputProcessor(agent: 'b') ...` to apply the processor on those agent.

chr-hertel: @junaidbinfarooq The difference here to the `AsTokenProcessor` attribute before is that we don't need to be that specific but can have a broader extension point => Input/OutputProcessor, that's why we ended up with #346 

@VincentLanglet i see the conflict here, you're right, but we can solve it, i guess
1. if someone implements the `InputProcessorInterface` this gets registered for all agents
2. if someone implements the interface but configures it with `AsInputProcessor(agent: 'blog')` that would only be registered for that agent

and i think we can use the same tag here `ai.agent.input_processor`

(same for the output processor of course)

VincentLanglet: > 1. if someone implements the `InputProcessorInterface` this gets registered for all agents
> 2. if someone implements the interface but configures it with `AsInputProcessor(agent: 'blog')` that would only be registered for that agent
> 
> and i think we can use the same tag here `ai.agent.input_processor`
> 
> (same for the output processor of course)

When `AsInputProcessor` is used as attribute, it still required to implement the InputProcessorInterface ; so the service will ends with the two tags...

Cf the Agent code
```
array_map(fn (InputProcessorInterface $processor) => $processor->processInput($input), $this->inputProcessors);
```

Also I don't think it's a good idea to auto-register class implementing InputProcessorInterface for all agents. 
(For instance it will imply that all the TokenUsageProcessor of Symfony will be automatically executed for all the agents, without extra modification)

I'm kinda tempted to just remove the code
```
$builder->registerForAutoconfiguration(InputProcessorInterface::class)
            ->addTag('ai.agent.input_processor');
        $builder->registerForAutoconfiguration(OutputProcessorInterface::class)
            ->addTag('ai.agent.output_processor');
```
because those registration is not used anywhere...

src/ai-bundle/src/AiBundle.php
```diff
@@ -135,10 +147,20 @@ public function loadExtension(array $config, ContainerConfigurator $container, C
             ]);
         });
 
-        $builder->registerForAutoconfiguration(InputProcessorInterface::class)
-            ->addTag('ai.agent.input_processor');
-        $builder->registerForAutoconfiguration(OutputProcessorInterface::class)
```
VincentLanglet: I removed the default tag to not auto-configure processor on every agent by default.

People will need to manually add those tag with config or with the freshly new attributes.

src/agent/src/Attribute/AsOutputProcessor.php
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+namespace Symfony\AI\Agent\Attribute;
+
+#[\Attribute(\Attribute::TARGET_CLASS | \Attribute::IS_REPEATABLE)]
+final readonly class AsOutputProcessor
+{
+    /**
+     * @param string|null $agent The service id of the agent which will use this processor.
+     *                           Use NULL in order to register this processor on all the existing agents.
```
OskarStark:
```suggestion
     * @param string|null $agent The service id of the agent which will use this processor, null to register this processor for all existing agents.
```

src/agent/src/Attribute/AsInputProcessor.php
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+namespace Symfony\AI\Agent\Attribute;
+
+#[\Attribute(\Attribute::TARGET_CLASS | \Attribute::IS_REPEATABLE)]
+final readonly class AsInputProcessor
+{
+    /**
+     * @param string|null $agent The service id of the agent which will use this processor.
+     *                           Use NULL in order to register this processor on all the existing agents.
```
OskarStark:
```suggestion
     * @param string|null $agent The service id of the agent which will use this processor, null to register this processor for all existing agents.
```

src/agent/src/Attribute/AsInputProcessor.php
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+namespace Symfony\AI\Agent\Attribute;
+
+#[\Attribute(\Attribute::TARGET_CLASS | \Attribute::IS_REPEATABLE)]
+final readonly class AsInputProcessor
+{
+    /**
+     * @param string|null $agent the service id of the agent which will use this processor, null to register this processor for all existing agents
+     */
```
OskarStark: Sorry, I missed the linbreak after `this processor,` same for the other, thanks!

src/ai-bundle/src/AiBundle.php
```diff
@@ -109,7 +117,11 @@ public function loadExtension(array $config, ContainerConfigurator $container, C
         }
 
         foreach ($config['agent'] as $agentName => $agent) {
-            $this->processAgentConfig($agentName, $agent, $builder);
+            $this->processAgentConfig(
+                $agentName,
+                $agent,
+                $builder,
+            );
```
OskarStark: lets revert this line break

src/agent/src/Attribute/AsInputProcessor.php
VincentLanglet: No problem, udpated

VincentLanglet:
@chr-hertel I think the PR is ready for a new review

chr-hertel (changes requested):
Please make sure it is not needed to use the attribute if an `Input` or `OutputProcessor` is meant to catch all agents. With implementing the interface the intent of the developer is clear and the feature of `#[AsOutputProcessor]` is then to limit it to one (or multiple) agents instead of all.

VincentLanglet:
> Please make sure it is not needed to use the attribute if an `Input` or `OutputProcessor` is meant to catch all agents. With implementing the interface the intent of the developer is clear and the feature of `#[AsOutputProcessor]` is then to limit it to one (or multiple) agents instead of all.

@chr-hertel I don't think it
- is easily possible
- should be recommended

As an exemple there are the following intput/outputprocessor
https://github.com/symfony/ai/blob/4f83f33b1e622568ef3db5b35cbf13f1023c3447/src/ai-bundle/src/AiBundle.php#L443-L526

If implementing the interface automatically register the processor it means that (for example):
- Using `$config['structured_output'] = false` on all your agent will let the processor without attribute and then will automatically register it to EVERY agent instead.
- As soon as you use `$config['structured_output'] = true` on ONE agent, it will remove it from every other agents.

So it may require some extra reflection about how to disable a processor from the auto-discovery logic you want.

Also, in an incremental implementation of this feature we could prefer
- A PR to allow people manually register processor with the new attributes/tags
- A PR to automatically add the attribute/tag to the class implementing the interface in some conditions.

They don't have to be in the same PR.

chr-hertel:
No, don't get me wrong, I don't think it is useful to register _all_ Input/OutputProcessor implementations based on the interface - especially the ones that we provide out of the box, like the toolbox or structured output - that should only be based on the bundle config.

I'm talking about the library user land. If someone implements their own processor - default is all agents, limited with attribute

VincentLanglet:
> I'm talking about the library user land. If someone implements their own processor - default is all agents, limited with attribute

Then, how I make a difference (in the config) between an InputProcessor/OutputProcessor from symfony/ai VS from the user lib ? Because if I use 
```
$builder->registerForAutoconfiguration(InputProcessorInterface::class)
```
it will impact both.

Also, the feature is currently doable with just `#[AsInputProcessor]` since I consider the attribute with "null" agent means every agents.

chr-hertel:
> Then, how I make a difference (in the config) between an InputProcessor/OutputProcessor from symfony/ai VS from the user lib ? Because if I use
> 
> ```
> $builder->registerForAutoconfiguration(InputProcessorInterface::class)
> ```
> 
> it will impact both.

I thought, that with your changes in `src/ai-bundle/config/services.php` that should be fine. Autoconfiguration is not enabled for lib services in the bundle - and maybe it makes the bundle config even easier to add on tag per agent using the structured output or toolbox. but if that doesn't work since i'm overlooking something, i can take a deeper look later on code-level.

VincentLanglet:
> I thought, that with your changes in `src/ai-bundle/config/services.php` that should be fine. Autoconfiguration is not enabled for lib services in the bundle - and maybe it makes the bundle config even easier to add on tag per agent using the structured output or toolbox. but if that doesn't work since i'm overlooking something, i can take a deeper look later on code-level.

I tried on my project and
```
dd($container->findTaggedServiceIds('ai.agent.input_processor'));
```
returns
```
array:2 [
  "ai.agent.structured_output_processor" => array:1 [
    0 => []
  ]
  "ai.tool.agent_processor" => array:1 [
    0 => []
  ]
]
```
so indeed, after my PR it will returns nothing by default ; I'll update the PR.

src/ai-bundle/config/services.php
```diff
@@ -64,8 +64,6 @@
                 service('ai.agent.response_format_factory'),
                 service('serializer'),
             ])
-            ->tag('ai.agent.input_processor')
-            ->tag('ai.agent.output_processor')
```
VincentLanglet, line 72: I have to remove the tag to not autoregister those.

src/ai-bundle/src/AiBundle.php
```diff
@@ -440,42 +462,46 @@ private function processAgentConfig(string $name, array $config, ContainerBuilde
...
                 if ($config['fault_tolerant_toolbox'] && !$container->hasDefinition('ai.fault_tolerant_toolbox')) {
                     $container->setDefinition('ai.fault_tolerant_toolbox', new Definition(FaultTolerantToolbox::class))
                         ->setArguments([new Reference('.inner')])
                         ->setDecoratedService('ai.toolbox');
                 }
 
-                $inputProcessors[] = new Reference('ai.tool.agent_processor');
-                $outputProcessors[] = new Reference('ai.tool.agent_processor');
+                $container->getDefinition('ai.tool.agent_processor')
+                    ->addTag('ai.agent.input_processor', ['agent' => $name, 'priority' => -10])
```
VincentLanglet, line 529: The main discussion might be about the priority of the registered processor @chr-hertel.

By default custom processor are registered with the 0 priority meaning they are applied BEFORE all the processors from Symfony.
If we want the opposite, I need to use a positive priority here.

VincentLanglet:
Should be re-ready then @chr-hertel :)

chr-hertel:
What I did to test this with the demo application:

I added two output processors, as they are easy to test:
```php
namespace App;

use Symfony\AI\Agent\Output;
use Symfony\AI\Agent\OutputProcessorInterface;
use Symfony\AI\Platform\Result\TextResult;

final readonly class StrRevProcessor implements OutputProcessorInterface
{
    public function processOutput(Output $output): void
    {
        $result = $output->result;

        if (!$result instanceof TextResult) {
            return;
        }

        $output->result = new TextResult(strrev($result->getContent()));
    }
}
```

and

```php
namespace App;

use Symfony\AI\Agent\Attribute\AsOutputProcessor;
use Symfony\AI\Agent\Output;
use Symfony\AI\Agent\OutputProcessorInterface;
use Symfony\AI\Platform\Result\TextResult;

#[AsOutputProcessor(agent: 'ai.agent.wikipedia')]
final readonly class UpperProcessor implements OutputProcessorInterface
{
    public function processOutput(Output $output): void
    {
        $result = $output->result;

        if (!$result instanceof TextResult) {
            return;
        }

        $output->result = new TextResult(strtoupper($result->getContent()));
    }
}
```

But the `UpperProcessor` was not limited to the wikipedia agent, but also registered to all others, based on the interface.

image

---

I agree that it is a bit tricky to fix this, but with your PR all essential parts were already there and this would be my solution to this problem here: d3118940f21a7ea74f045d58535ee73464b6a6c6

src/ai-bundle/doc/index.rst
```diff
@@ -150,6 +150,34 @@ Use the `Agent` service to leverage models and tools::
         }
     }
 
+**Register Processors**
+
+By default, all services implementing the `InputProcessorInterface` or the
+`OutputProcessorInterface` interfaces are automatically applied to every `Agent`.
+
+This behavior can be overridden/configured with the `#[AsInputProcessor]` and
+the `#[AsOutputProcessor]` attributes::
```
OskarStark:
```suggestion
By default, all services implementing the ``InputProcessorInterface`` or the
``OutputProcessorInterface`` interfaces are automatically applied to every ``Agent``.

This behavior can be overridden/configured with the ``#[AsInputProcessor]`` and
the ``#[AsOutputProcessor]`` attributes::
```

src/ai-bundle/doc/index.rst
```diff
@@ -150,6 +150,34 @@ Use the `Agent` service to leverage models and tools::
...
+
+    use Symfony\AI\Agent\Input;
+    use Symfony\AI\Agent\InputProcessorInterface;
+    use Symfony\AI\Agent\Output;
+    use Symfony\AI\Agent\OutputProcessorInterface;
+
+    #[AsInputProcessor(priority: 99)] // This apply to every agents
```
OskarStark:
```suggestion
    #[AsInputProcessor(priority: 99)] // This applies to every agent
```

src/ai-bundle/doc/index.rst
```diff
@@ -150,6 +150,34 @@ Use the `Agent` service to leverage models and tools::
...
+    #[AsOutputProcessor(agent: 'my_agent_id')] // The output processor will only be registered for this specific agent
```
OskarStark:
```suggestion
    #[AsOutputProcessor(agent: 'my_agent_id')] // The output processor will only be registered for 'my_agent_id'
```

VincentLanglet:
@chr-hertel I think https://github.com/symfony/ai/pull/352/commits/4d2cf87e9ff5e8971dd007b08acda20d4f7d33b0 might be enough for your issue

src/ai-bundle/src/AiBundle.php
OskarStark: @chr-hertel please advise, thanks

chr-hertel (approved):
Works for my example now as well :+1:

chr-hertel:
Thank you @VincentLanglet.

PR #357, opened 2025-08-27
[Platform][OpenAI] Add file input normalizer

glengemann:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | Fix #... 
| License       | MIT

Hi there!

Currently, if you try to add a file into the message input, the following error is thrown: 

```php
// PHP Fatal error:  Uncaught Symfony\Component\Serializer\Exception\NotNormalizableValueException: Could not normalize object of type "Symfony\AI\Platform\Message\Content\File", no supporting normalizer found.

Message::ofUser(
    'What is this document about?',
    File::fromFile(dirname(__DIR__, 2).'/fixtures/document.pdf'),
)
```

This PR adds support for including files as inputs in the OpenAI completions API, as described in the official [documentation](https://platform.openai.com/docs/guides/pdf-files?api-mode=chat&lang=javascript#base64-encoded-files). The implementation follows the same approach already used for Gemini and Anthropic.

### DIFF
--- a/src/platform/src/Bridge/OpenAi/Contract/FileNormalizer.php
+++ b/src/platform/src/Bridge/OpenAi/Contract/FileNormalizer.php
@@ -1,0 +1,49 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Platform\Bridge\OpenAi\Contract;
+
+use Symfony\AI\Platform\Bridge\OpenAi\Gpt;
+use Symfony\AI\Platform\Contract\Normalizer\ModelContractNormalizer;
+use Symfony\AI\Platform\Message\Content\File;
+use Symfony\AI\Platform\Model;
+
+/**
+ * @author Guillermo Lengemann <guillermo.lengemann@gmail.com>
+ */
+class FileNormalizer extends ModelContractNormalizer
+{
+    /**
+     * @param File $data
+     *
+     * @return array{type: 'file', file: array{filename: string, file_data: 'base64'}}
+     */
+    public function normalize(mixed $data, ?string $format = null, array $context = []): array
+    {
+        return [
+            'type' => 'file',
+            'file' => [
+                'filename' => $data->getFilename(),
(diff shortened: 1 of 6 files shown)


src/platform/src/Bridge/OpenAi/Contract/DocumentNormalizer.php
```diff
@@ -0,0 +1,49 @@
...
+
+use Symfony\AI\Platform\Bridge\OpenAi\Gpt;
+use Symfony\AI\Platform\Contract\Normalizer\ModelContractNormalizer;
+use Symfony\AI\Platform\Message\Content\File;
+use Symfony\AI\Platform\Model;
+
+/**
+ * @author Guillermo Lengemann <guillermo.lengemann@gmail.com>
+ */
+class DocumentNormalizer extends ModelContractNormalizer
```
chr-hertel: shouldn't it be `FileNormalizer` than? feels inconsistent with `'type' => 'file'` and `File::class` in the `supportedDataClass`

```suggestion
class FileNormalizer extends ModelContractNormalizer
```

chr-hertel:
One thing that got me thinking is that we use the OpenAI contract mostly as default - meaning that the normalizers are mostly in namespace `Symfony\AI\Platform\Contract\Normalizer` - but I tested with Ollama and Cerebras, that are basically using that contract, but got errors - so i guess it's fine to only have it for OpenAI instead of default.

chr-hertel (changes requested):
Thanks for this, would love to merge that - one tiny thing noted about the normalizer name.

chr-hertel:
Thank you @glengemann.

src/platform/src/Bridge/OpenAi/Contract/FileNormalizer.php
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * @author Guillermo Lengemann <guillermo.lengemann@gmail.com>
+ */
+class FileNormalizer extends ModelContractNormalizer
+{
+    /**
+     * @param File $data
+     *
+     * @return array{type: 'file', file: array{filename: string, file_data: 'base64'}}
```
OskarStark, line 27: Is "base_64" as string really correct?

chr-hertel: No, it's not => #384

PR #359, opened 2025-08-27
Voyage embedding update

natewiebe13:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| License       | MIT

Added additional options for working with Voyage embeddings. I did also bump the default model used, as the newer one is cheaper and higher quality (ref: https://blog.voyageai.com/2025/05/20/voyage-3-5/)


src/platform/src/Bridge/Voyage/ModelClient.php
```diff
@@ -34,11 +34,16 @@ public function supports(Model $model): bool
 
     public function request(Model $model, object|string|array $payload, array $options = []): RawHttpResult
     {
+        $modelOptions = $model->getOptions();
```
chr-hertel: Please use the `$options` parameter of the method instead, it is a result of the model's options and the `invoke` method and allows users to overwrite once defined options from request to request, see `Platform::invoke`:

https://github.com/symfony/ai/blob/e7bec238e4f695da16ce00843ad0d7d7100d27b7/src/platform/src/Platform.php#L47-L50

chr-hertel (changes requested):
Thanks for this, small change request but close to merge :+1:

src/platform/src/Bridge/Voyage/ModelClient.php
```diff
@@ -34,11 +34,16 @@ public function supports(Model $model): bool
...
+
         return new RawHttpResult($this->httpClient->request('POST', 'https://api.voyageai.com/v1/embeddings', [
             'auth_bearer' => $this->apiKey,
             'json' => [
                 'model' => $model->getName(),
                 'input' => $payload,
+                'input_type' => $modelOptions['input_type'] ?? null,
+                'truncation' => $modelOptions['truncation'] ?? true,
+                'output_dimension' => $modelOptions['dimensions'] ?? null,
             ],
```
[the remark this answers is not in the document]
OskarStark: In this case the api respond with a proper error

src/platform/src/Bridge/Voyage/Voyage.php
```diff
@@ -19,17 +19,34 @@
...
+    public const CODE_3 = 'voyage-code-3';
+    /** Fixed 1536 dimensions */
     public const CODE_2 = 'voyage-code-2';
 
+    public const INPUT_TYPE_DOCUMENT = 'document';
+    public const INPUT_TYPE_QUERY = 'query';
+
     /**
-     * @param array<string, mixed> $options
+     * @param array{dimensions?: int, input_type?: self::INPUT_TYPE_*, truncation?: bool} $options
```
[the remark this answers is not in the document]
OskarStark: No it's fine

src/platform/src/Bridge/Voyage/Voyage.php
```diff
@@ -19,17 +19,34 @@
  */
 class Voyage extends Model
 {
+    /** Supported dimensions: 2048, 1024, 512, or 256 */
```
GromNaN: It's useful to have this information here. Could we validate the `dimensions` option in the constructor and indicate the dimensions supported in case of error?

chr-hertel: I'm usually a bit reluctant to bring in those validations because we don't really own that and don't rely on that ourselves, but it is only with Voyage.

Change scenario could be that they open up or change the support, the lib user want to leverage it, but is limited by our design - and we would need to patch and release to enable them.

We should check tho what happens if the user submits wrong options to the API, and if the error message gets back to the developer - that one of our weak spots still ...

chr-hertel: @natewiebe13 by any chance, did you check that already?

src/platform/src/Bridge/Voyage/Voyage.php
```diff
@@ -19,17 +19,34 @@
...
 
+    public const INPUT_TYPE_DOCUMENT = 'document';
+    public const INPUT_TYPE_QUERY = 'query';
+
     /**
-     * @param array<string, mixed> $options
+     * @param array{dimensions?: int, input_type?: self::INPUT_TYPE_*, truncation?: bool} $options
      */
-    public function __construct(string $name = self::V3, array $options = [])
+    public function __construct(string $name = self::V3_5_LITE, array $options = [])
```
OskarStark: lets not change the default version

OskarStark:
Thank you, I made some changes while merging.

src/platform/src/Bridge/Voyage/Voyage.php
natewiebe13: @chr-hertel a `\Symfony\AI\Platform\Exception\RuntimeException` is thrown with `Response does not contain embedding data.`

natewiebe13:
> Thank you, I made some changes while merging.

@OskarStark the one thing I'll mention is that the comments regarding dimensions was copied from `\Symfony\AI\Platform\Bridge\Gemini\Embeddings`. From a consistency standpoint, it might be a good idea to either remove them from that class or come up with a better way of including this info. It is nice to have so that way you don't always have to jump into docs to see what's available.

src/platform/src/Bridge/Voyage/Voyage.php
natewiebe13: I wonder if it might be best to not specify a default. People are notorious for using the defaults because it's easy. Keeping an outdated, more expensive model as a default likely isn't the best long-term. If not, what would the process be for updating the default? If the hesitation is going to a "lite" version, what about V3_5 as the new default?

OskarStark:
Let's remove them there as well I would say. We are not able to keep track of this, and if they may change the number of dimensions in the future. Is this the raw exception from the API? I think not and there is an open PR/issue to improve error logging from the platforms, so we should tackle this with better logging

natewiebe13:
@OskarStark that's the content from the exception being thrown (`vendor/symfony/ai-platform/src/Bridge/Voyage/ResultConverter.php:39`).

The raw response json is:
```json
{"detail":"Value '1234' supplied for argument 'output_dimension' is not valid -- accepted values for 'voyage-3.5-lite' are [256, 512, 1024, 2048]."}
```

OskarStark:
Thats super helpful, so lets propagate this raw error from the vendor somehow

PR #366, opened 2025-08-29
[Store][ClickHouse] Make store `final readonly`

OskarStark:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | --
| License       | MIT

### DIFF
--- a/src/store/src/Bridge/ClickHouse/Store.php
+++ b/src/store/src/Bridge/ClickHouse/Store.php
@@ -25,12 +25,12 @@ use Symfony\Contracts\HttpClient\ResponseInterface;
 /**
  * @author Grégoire Pineau <lyrixx@lyrixx.info>
  */
-class Store implements ManagedStoreInterface, StoreInterface
+final readonly class Store implements ManagedStoreInterface, StoreInterface
 {
     public function __construct(
-        private readonly HttpClientInterface $httpClient,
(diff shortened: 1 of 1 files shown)


lyrixx:
Hello, I thinks it's a bad idea to mark it as final.

For exemple, I override the store, to override some parts:

```php
class ClickHouseStore extends SymfonyClickHouseStore
{
    public function initialize(array $options = []): void
    {
        $sql = <<<'SQL'
            CREATE TABLE IF NOT EXISTS {{ table }} (
                id UUID,
                metadata String,
                embedding Array(Float32),
                crawlId UUID,
                -- INDEX embedding_idx embedding TYPE vector_similarity('hnsw', 'cosineDistance', 3072)
            ) ENGINE = MergeTree()
            PRIMARY KEY crawlId
        SQL;

        $this->execute('POST', $sql);
    }

    protected function formatVectorDocument(VectorDocument $document): array
    {
        $formatted = parent::formatVectorDocument($document);

        $formatted['crawlId'] = $document->metadata['crawlId'];

        return $formatted;
    }
}
```

Composition here is not easy, and seems overkill.

Could we consider to revert this PR?

OskarStark:
Yes, let's open them up all please, thanks

PR #369, opened 2025-08-29
[AI Bundle][Platform] Fix Anthropic platform definition

### LINKED ISSUE #363: Invalid constructor argument 3 for service "ai.platform.anthropic": argument 2 must be defined before in dev-main
Bug Report

  Environment:
  - symfony/ai-bundle: dev-main (installed from git on 2025-08-25)
  - Symfony: 7.3
  - PHP: 8.3
  - Platform: Docker with FrankenPHP

  Issue:
  When configuring Anthropic agents in ai.yaml, the application fails to clear cache and throws a service definition error.

  Error Message:
  In DefinitionErrorExceptionPass.php line 48:
  Invalid constructor argument 3 for service "ai.platform.anthropic": argument 2 must be defined before. Check your service definition.

  Configuration Used:
```yaml
  ai:
      platform:
          anthropic:
              api_key: '%env(ANTHROPIC_API_KEY)%'

      agent:
          blog_generator:
              platform: 'ai.platform.anthropic'
              model:
                  class: 'Symfony\AI\Platform\Bridge\Anthropic\Claude'
                  name: !php/const Symfony\AI\Platform\Bridge\Anthropic\Claude::SONNET_35
              system_prompt: "Tu es un rédacteur web expert."
```

  Steps to Reproduce:
  1. Install symfony/ai-bundle from dev-main
  2. Configure Anthropic platform with valid API key
  3. Configure at least one agent using Anthropic Claude models
  4. Run symfony console cache:clear

  Expected Behavior:
  Cache should clear successfully and Anthropic agents should be available.

  Actual Behavior:
  Cache clear fails with service definition error related to constructor arguments.

  Additional Context:
  - Mistral configuration works perfectly with the same setup
  - The error suggests an issue with the service definition in the Anthropic bridge
  - This prevents using Anthropic agents entirely
  - Switching to Mistral platform works as expected

  Workaround:
  Currently using Mistral as alternative, but Anthropic support is needed for production requirements.

  Service Definition Debug:
  The error seems to indicate that the ai.platform.anthropic service has incorrect constructor argument ordering or missing dependencies.

  Would appreciate guidance on:
  1. Is this a known issue with dev-main?
  2. Are there specific configuration requirements for Anthropic not documented?
  3. Is there a fix or workaround available?

VincentLanglet:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no 
| Issues        | Fix #363
| License       | MIT

### DIFF
--- a/src/platform/src/Bridge/Anthropic/ModelClient.php
+++ b/src/platform/src/Bridge/Anthropic/ModelClient.php
@@ -27,7 +27,6 @@ final readonly class ModelClient implements ModelClientInterface
     public function __construct(
         HttpClientInterface $httpClient,
         #[\SensitiveParameter] private string $apiKey,
-        private string $version = '2023-06-01',
     ) {
         $this->httpClient = $httpClient instanceof EventSourceHttpClient ? $httpClient : new EventSourceHttpClient($httpClient);
     }
@@ -41,7 +40,7 @@ final readonly class ModelClient implements ModelClientInterface
     {
         $headers = [
             'x-api-key' => $this->apiKey,
-            'anthropic-version' => $this->version,
+            'anthropic-version' => '2023-06-01',
         ];
 
         if (isset($options['tools'])) {
--- a/src/platform/src/Bridge/Anthropic/PlatformFactory.php
+++ b/src/platform/src/Bridge/Anthropic/PlatformFactory.php
@@ -25,14 +25,13 @@ final readonly class PlatformFactory
     public static function create(
         #[\SensitiveParameter]
         string $apiKey,
-        string $version = '2023-06-01',
         ?HttpClientInterface $httpClient = null,
         ?Contract $contract = null,
     ): Platform {
         $httpClient = $httpClient instanceof EventSourceHttpClient ? $httpClient : new EventSourceHttpClient($httpClient);
 
         return new Platform(
-            [new ModelClient($httpClient, $apiKey, $version)],
+            [new ModelClient($httpClient, $apiKey)],
             [new ResultConverter()],
             $contract ?? AnthropicContract::create(),
         );
--- a/src/ai-bundle/src/AiBundle.php
+++ b/src/ai-bundle/src/AiBundle.php
@@ -174,15 +174,11 @@ final class AiBundle extends AbstractBundle
                 ->setLazy(true)
                 ->addTag('proxy', ['interface' => PlatformInterface::class])
                 ->setArguments([
-                    0 => $platform['api_key'],
-                    2 => new Reference('http_client', ContainerInterface::NULL_ON_INVALID_REFERENCE),
-                    3 => new Reference('ai.platform.contract.anthropic'),
+                    $platform['api_key'],
+                    new Reference('http_client', ContainerInterface::NULL_ON_INVALID_REFERENCE),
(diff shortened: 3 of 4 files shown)


VincentLanglet:
@Krystopher88 could you try this and confirm it solves your issue ?

src/platform/src/Bridge/Anthropic/ModelClient.php
```diff
@@ -27,7 +27,7 @@
     public function __construct(
         HttpClientInterface $httpClient,
         #[\SensitiveParameter] private string $apiKey,
-        private string $version = '2023-06-01',
+        private ?string $version = null,
```
OskarStark: Why this change?

src/platform/src/Bridge/Anthropic/PlatformFactory.php
```diff
@@ -25,7 +25,7 @@
     public static function create(
         #[\SensitiveParameter]
         string $apiKey,
-        string $version = '2023-06-01',
+        ?string $version = null,
```
xabbuh: What about moving this parameter at the end instead?

src/platform/src/Bridge/Anthropic/ModelClient.php
VincentLanglet: Platform is doing 
```
new ModelClient($httpClient, $apiKey, $version)
```

Which means that, either
- It has to know the default value, so `$version = '2023-06-01',` is duplicated
- It requires the logic 
```
null === $version ? new ModelClient($httpClient, $apiKey) : new ModelClient($httpClient, $apiKey, $version);
```

Also, since I needed to pass the platform version in the AiBundle config, I didn't want to pass 
```
$platform['version'] ?? '2023-06-01',
```
with the version hard coded a third time.

src/platform/src/Bridge/Anthropic/PlatformFactory.php
VincentLanglet: Might be a solution.

But still I'm unsure it a nice pattern to duplicate the default value `'2023-06-01'` between the factory and the Model, no ?

chr-hertel (changes requested):
Thanks for jumping on that @VincentLanglet!
I'm usually not a huge fan of nullable arguments, and it's true that this is a bit weird in handling and duplicated here.

One thing I noticed when looking at the Anthropics docs again: there are only two version strings at this point `2023-01-01` and the `2023-06-01` we're currently using. and guess what, only one is supported by our implementation here anyways. with `2023-01-01` you get the error "anthropic-version: "2023-01-01" not allowed for this endpoint" error.

So, if you're up for it, let's just clean that up, remove all arguments, the bundle config option and hard code the version in the header: `'anthropic-version' => '2023-06-01'` - easier and we don't have to debate the nullable and where not to duplicate stuff :laughing:

OskarStark:
Please fix PHPStan, thanks

OskarStark:
Thanks for fixing this bug Vincent.

chr-hertel:
@OskarStark saw my comment? any thoughts on that?

OskarStark:
You mean regarding the defaults? Mixed feelings, as if there is a new version, you can only use it with a new release of the lib, like for the models

PR #373, opened 2025-08-30
[AI Bundle][OpenAI] Register normalizer contract

glengemann:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #... 
| License       | MIT

Christopher, I think we need to register the normalizer contract so that it’s available when using the Symfony bundle.


chr-hertel (commented):
Guillermo, i think you are right :)

chr-hertel:
Thank you @glengemann.

PR #387, opened 2025-08-31
[Platform] Fix VertexAI server tools

### LINKED ISSUE #386: [Platform] VertexAI server tools not working at all
Server tools in recently-introduced VertexAI platform are not working at all.

1. The server-tools.php example does not use any server tools, instead, it fetches the url client-side.
2. Tools are provided to VertexAI using wrong structure – server tools should not be inside `functionDeclarations`
3. Providing server tools to VertexAI within `symfony/ai` unnecessarily differs from providing `server_tools` to Gemini platform
4. It is difficult to pass in an empty object from yaml bundle config – that's why Gemini platform converts  `true` in `server_tools` to an empty object

I will try to address all these

valtzu:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | yes
| Issues        | Fix #386
| License       | MIT

1. Align "Gemini via VertexAI" server tools configuration with Gemini platform
    - This makes it easier to migrate from Gemini -> Vertex and vice-versa 
3. Fix the server tools example to actually use server tools
4. Fix how server tools are provided to VertexAI

Worth noting that unlike Gemini API, it seems VertexAI does not support client-side tools when using server-side tools, the api is returning `Multiple tools are supported only when they are all search tools`. This is why the `Clock` & `Toolbox` was removed from the example.

### DIFF
--- a/examples/vertexai/server-tools.php
+++ b/examples/vertexai/server-tools.php
@@ -10,9 +10,6 @@
  */
 
 use Symfony\AI\Agent\Agent;
-use Symfony\AI\Agent\Toolbox\AgentProcessor;
-use Symfony\AI\Agent\Toolbox\Tool\Clock;
-use Symfony\AI\Agent\Toolbox\Toolbox;
 use Symfony\AI\Platform\Bridge\VertexAi\Gemini\Model;
 use Symfony\AI\Platform\Bridge\VertexAi\PlatformFactory;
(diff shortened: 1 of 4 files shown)


chr-hertel (approved):

chr-hertel:
Thank you @valtzu.

PR #392, opened 2025-09-01
[Examples] Allow probots-io/pinecone-php 1.0

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | None
| License       | MIT

Hi 👋🏻 

As discussed in #335, the `Pinecone` dependency for examples must be downgraded if we want to being able to use it during functional tests (mostly commands), this also synchronize the dependency with the one defined in `Store`.


OskarStark:
Thank you @Guikingone.

PR #395, opened 2025-09-01
[AI Bundle] Add chat console command for interactive AI agent conversations

OskarStark:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | --
| License       | MIT

Implement a new console command `bin/console ai:chat` that allows users to interactively chat with different AI agents (blog, stream, youtube, wikipedia, audio). The command supports custom system prompts and maintains conversation history.

### Demo

CleanShot 2025-09-03 at 07 57 09@2x
CleanShot 2025-09-03 at 07 56 53@2x

### DIFF
--- a/src/ai-bundle/config/services.php
+++ b/src/ai-bundle/config/services.php
@@ -21,6 +21,7 @@ use Symfony\AI\Agent\Toolbox\ToolCallArgumentResolver;
 use Symfony\AI\Agent\Toolbox\ToolFactory\AbstractToolFactory;
 use Symfony\AI\Agent\Toolbox\ToolFactory\ReflectionToolFactory;
 use Symfony\AI\Agent\Toolbox\ToolResultConverter;
+use Symfony\AI\AiBundle\Command\ChatCommand;
 use Symfony\AI\AiBundle\Profiler\DataCollector;
 use Symfony\AI\AiBundle\Profiler\TraceableToolbox;
 use Symfony\AI\AiBundle\Security\EventListener\IsGrantedToolAttributeListener;
@@ -137,5 +138,12 @@ return static function (ContainerConfigurator $container): void {
         ->set('ai.platform.token_usage_processor.gemini', GeminiTokenOutputProcessor::class)
         ->set('ai.platform.token_usage_processor.openai', OpenAiTokenOutputProcessor::class)
         ->set('ai.platform.token_usage_processor.vertexai', VertexAiTokenOutputProcessor::class)
+
+        // commands
+        ->set('ai.command.chat', ChatCommand::class)
+            ->args([
+                tagged_locator('ai.agent', indexAttribute: 'name'),
+            ])
+            ->tag('console.command')
     ;
 };
--- a/src/ai-bundle/src/AiBundle.php
+++ b/src/ai-bundle/src/AiBundle.php
@@ -457,7 +457,7 @@ final class AiBundle extends AbstractBundle
 
         // AGENT
         $agentDefinition = (new Definition(Agent::class))
-            ->addTag('ai.agent')
+            ->addTag('ai.agent', ['name' => $name])
             ->setArgument(0, new Reference($config['platform']))
             ->setArgument(1, new Reference('ai.agent.'.$name.'.model'));
 
--- a/src/ai-bundle/src/Command/ChatCommand.php
+++ b/src/ai-bundle/src/Command/ChatCommand.php
@@ -1,0 +1,213 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\AiBundle\Command;
+
+use Symfony\AI\Agent\AgentInterface;
+use Symfony\AI\AiBundle\Exception\InvalidArgumentException;
+use Symfony\AI\Platform\Message\Message;
+use Symfony\AI\Platform\Message\MessageBag;
+use Symfony\AI\Platform\Result\TextResult;
+use Symfony\Component\Console\Attribute\AsCommand;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Completion\CompletionInput;
+use Symfony\Component\Console\Completion\CompletionSuggestions;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Question\ChoiceQuestion;
+use Symfony\Component\Console\Style\SymfonyStyle;
+use Symfony\Component\DependencyInjection\ServiceLocator;
+
+/**
+ * @author Oskar Stark <oskarstark@gmail.com>
+ */
+#[AsCommand(
+    name: 'ai:chat',
+    description: 'Chat with an agent',
+)]
+final class ChatCommand extends Command
+{
+    private AgentInterface $agent;
+    private string $agentName;
+
+    /**
+     * @param ServiceLocator<AgentInterface> $agents
+     */
+    public function __construct(
+        private readonly ServiceLocator $agents,
+    ) {
+        parent::__construct();
+    }
+
+    public function complete(CompletionInput $input, CompletionSuggestions $suggestions): void
+    {
+        if ($input->mustSuggestArgumentValuesFor('agent')) {
+            $suggestions->suggestValues($this->getAvailableAgentNames());
+        }
+    }
+
+    protected function configure(): void
+    {
+        $this
+            ->addArgument('agent', InputArgument::OPTIONAL, 'The name of the agent to chat with')
+            ->setHelp(
+                <<<'HELP'
+                The <info>%command.name%</info> command allows you to chat with different agents.
+
+                Usage:
+                  <info>%command.full_name% [<agent_name>]</info>
+
+                Examples:
+                  <info>%command.full_name% wikipedia</info>
+
+                If no agent is specified, you'll be prompted to select one interactively.
+                
+                The chat session is interactive. Type your messages and press Enter to send.
+                Type 'exit' or 'quit' to end the conversation.
+                HELP
+            );
+    }
+
+    protected function interact(InputInterface $input, OutputInterface $output): void
+    {
+        // Skip interaction in non-interactive mode
+        if (!$input->isInteractive()) {
+            return;
+        }
+
+        $agentArg = $input->getArgument('agent');
+
+        // If agent is already provided and valid, nothing to do
+        if ($agentArg) {
+            return;
+        }
+
+        $availableAgents = $this->getAvailableAgentNames();
+
+        if (0 === \count($availableAgents)) {
+            throw new InvalidArgumentException('No agents are configured.');
+        }
+
+        $question = new ChoiceQuestion(
+            'Please select an agent to chat with:',
+            $availableAgents,
+            0
+        );
+        $question->setErrorMessage('Agent %s is invalid.');
+
+        /** @var \Symfony\Component\Console\Helper\QuestionHelper $helper */
+        $helper = $this->getHelper('question');
+        $selectedAgent = $helper->ask($input, $output, $question);
+
+        $input->setArgument('agent', $selectedAgent);
+    }
+
+    protected function initialize(InputInterface $input, OutputInterface $output): void
+    {
+        // Initialization will be done in execute() after interact() has run
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        // Initialize agent (moved from initialize() to execute() so it runs after interact())
+        $availableAgents = array_keys($this->agents->getProvidedServices());
+
+        if (0 === \count($availableAgents)) {
+            throw new InvalidArgumentException('No agents are configured.');
+        }
+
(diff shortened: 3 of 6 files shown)


OskarStark:
Ready for review

src/ai-bundle/src/Command/ChatCommand.php
```diff
@@ -0,0 +1,189 @@
...
+use Symfony\Component\Console\Question\ChoiceQuestion;
+use Symfony\Component\Console\Style\SymfonyStyle;
+use Symfony\Component\DependencyInjection\Attribute\AutowireLocator;
+use Symfony\Component\DependencyInjection\ServiceLocator;
+
+/**
+ * @author Oskar Stark <oskarstark@gmail.com>
+ */
+#[AsCommand(
+    name: 'chat',
```
yceruto: maybe the name could include a namespace, for example `ai:chat`?

src/ai-bundle/src/Command/ChatCommand.php
```diff
@@ -0,0 +1,189 @@
...
+        #[AutowireLocator('ai.agent', indexAttribute: 'key')]
+        private readonly ServiceLocator $agents,
+    ) {
+        parent::__construct();
+    }
+
+    protected function configure(): void
+    {
+        $this
+            ->addArgument('agent', InputArgument::OPTIONAL, 'The name of the agent to chat with')
```
GromNaN, line 61: You can add a callback for the bash completion.

src/ai-bundle/src/Command/ChatCommand.php
```diff
@@ -0,0 +1,189 @@
...
+            throw new InvalidArgumentException('No agents are configured.');
+        }
+
+        $agentArg = $input->getArgument('agent');
+        $this->agentName = \is_string($agentArg) ? $agentArg : '';
+
+        // If no agent specified or invalid agent, prompt for selection
+        if (!$this->agentName || !isset($agentMap[$this->agentName])) {
+            if ($this->agentName && !isset($agentMap[$this->agentName])) {
+                $io->warning(\sprintf('Agent "%s" not found.', $this->agentName));
```
GromNaN: The error message should list available agents.

src/ai-bundle/src/Command/ChatCommand.php
```diff
@@ -0,0 +1,189 @@
...
+            $question = new ChoiceQuestion(
+                'Please select an agent to chat with:',
+                array_keys($availableAgents),
+                0
+            );
+            $question->setErrorMessage('Agent %s is invalid.');
+
+            /** @var \Symfony\Component\Console\Helper\QuestionHelper $helper */
+            $helper = $this->getHelper('question');
+            $selectedAgent = $helper->ask($input, $output, $question);
```
GromNaN: The best practice is to move the interactive question in the [interact method](https://github.com/symfony/symfony/blob/7278aa035af69b83d398b7220669ecb61be5a639/src/Symfony/Component/Console/Command/Command.php#L242-L244). And make the `agent` argument required.

src/ai-bundle/src/Command/ChatCommand.php
```diff
@@ -0,0 +1,189 @@
...
+                    $io->writeln('');
+                    $io->writeln($result->getContent());
+                    $io->newLine();
+
+                    $messages->add(Message::ofAssistant($result->getContent()));
+                } else {
+                    $io->error('Unexpected response type from agent');
+                }
+            } catch (\Exception $e) {
+                $io->error(\sprintf('Error: %s', $e->getMessage()));
```
GromNaN, line 193: In verbose mode, it is useful to have the exception backtrace.

src/ai-bundle/src/Command/ChatCommand.php
```diff
@@ -0,0 +1,189 @@
...
+        $io->info('Type your message and press Enter. Type "exit" or "quit" to end the conversation.');
+        $io->newLine();
+
+        $messages = new MessageBag();
+        $systemPromptDisplayed = false;
+
+        while (true) {
+            $userInput = $io->ask('You');
+
+            if (!\is_string($userInput)) {
```
GromNaN: Should you skip in case of blank line also?

GromNaN (commented):
Nice feature!
I think there are small details we can improve.

src/ai-bundle/src/Command/ChatCommand.php
OskarStark: ofc

src/ai-bundle/src/Command/ChatCommand.php
OskarStark: done

GromNaN (approved):

OskarStark:
Ready to merge from my side

src/ai-bundle/src/AiBundle.php
```diff
@@ -95,6 +96,11 @@ public function loadExtension(array $config, ContainerConfigurator $container, C
     {
         $container->import('../config/services.php');
 
+        // COMMANDS
+        $builder->register('ai.command.chat', ChatCommand::class)
+            ->addTag('console.command')
+            ->setArgument(0, new Reference('ai.locator.agents'));
```
GromNaN: The command is a service, Why is it defined here and not in `service.php` like any other service?

The service definition would use `tagged_locator` instead of the `ai.locator.agents` definition.

I don't understand why you need to manually populate the `ai.locator.agents` from `$builder->findTaggedServiceIds('ai.agent')`.

OskarStark: Tried different things, but you are right. Fixed

chr-hertel (approved):
Great for testing :+1:

chr-hertel:
oh, i wonder, did you try to use streaming? otherwise maybe a follow up

src/ai-bundle/src/Command/ChatCommand.php
```diff
@@ -0,0 +1,251 @@
...
+            throw new InvalidArgumentException('No agents are configured.');
+        }
+
+        $agentArg = $input->getArgument('agent');
+        $this->agentName = \is_string($agentArg) ? $agentArg : '';
+
+        // Validate the agent
+        if (!$this->agentName) {
+            throw new InvalidArgumentException(\sprintf('Agent name is required. Available agents: "%s"', implode(', ', array_keys($availableAgents))));
+        }
```
chr-hertel: if you check this here, we never end up with the "Please select an agent to chat with" choice ...

image

chr-hertel (changes requested):
found that one bug about the `ChoiceQuestion`

src/ai-bundle/src/Command/ChatCommand.php
OskarStark: Will check

OskarStark:
> oh, i wonder, did you try to use streaming? otherwise maybe a follow up

No not yet, lets follow up

src/ai-bundle/src/AiBundle.php
```diff
@@ -457,7 +457,7 @@ private function processAgentConfig(string $name, array $config, ContainerBuilde
 
         // AGENT
         $agentDefinition = (new Definition(Agent::class))
-            ->addTag('ai.agent')
+            ->addTag('ai.agent', ['key' => 'ai.agent.'.$name])
```
GromNaN: You can name the key `name` and just pass the unprefixed name. That would prevent having to deal with the prefixed and unprefixed names in the command: the agent name argument cannot be prefixed with `ai.agent.`.

```suggestion
            ->addTag('ai.agent', ['name' => $name])
```

src/ai-bundle/src/Command/ChatCommand.php
```diff
@@ -0,0 +1,253 @@
...
+final class ChatCommand extends Command
+{
+    private AgentInterface $agent;
+    private string $agentName;
+
+    /**
+     * @param ServiceLocator<AgentInterface> $agents
+     */
+    public function __construct(
+        #[TaggedLocator('ai.agent', indexAttribute: 'key')]
```
GromNaN: This attribute is read by autowiring compiler pass. It's not necessary here unless you enable autowiring on this service definition.

src/ai-bundle/src/Command/ChatCommand.php
OskarStark: Should be good now?

src/ai-bundle/src/Command/ChatCommand.php
```diff
@@ -0,0 +1,253 @@
...
+                $name = substr($serviceId, \strlen('ai.agent.'));
+            } else {
+                $name = $serviceId;
+            }
+            $names[] = $name;
+            // Also include the full service ID for completion
+            $names[] = $serviceId;
+        }
+
+        return array_unique($names);
```
GromNaN: If the agent are indexed by their short name, this function becomes a lot simpler. You cannot have duplicates in array keys.

```suggestion
        return array_keys($this->agents->getProvidedServices());
```

src/ai-bundle/src/Command/ChatCommand.php
```diff
@@ -0,0 +1,253 @@
...
+        if (0 === \count($shortNames)) {
+            throw new InvalidArgumentException('No agents are configured.');
+        }
+
+        // Interactive selection
+        $question = new ChoiceQuestion(
+            'Please select an agent to chat with:',
+            $shortNames,
+            0
+        );
```
GromNaN: Also simpler with only short agent names.

```suggestion
        $availableAgents = $this->getAvailableAgentNames();

        if (0 === \count($availableAgents)) {
            throw new InvalidArgumentException('No agents are configured.');
        }

        // Interactive selection
        $question = new ChoiceQuestion(
            'Please select an agent to chat with:',
            $availableAgents,
            0
        );
```

src/ai-bundle/config/services.php
```diff
@@ -137,5 +138,12 @@
         ->set('ai.platform.token_usage_processor.gemini', GeminiTokenOutputProcessor::class)
         ->set('ai.platform.token_usage_processor.openai', OpenAiTokenOutputProcessor::class)
         ->set('ai.platform.token_usage_processor.vertexai', VertexAiTokenOutputProcessor::class)
+
+        // commands
+        ->set('ai.command.chat', ChatCommand::class)
+            ->args([
+                tagged_locator('ai.agent', indexAttribute: 'key'),
```
GromNaN:
```suggestion
                tagged_locator('ai.agent', indexAttribute: 'name'),
```

src/ai-bundle/src/Command/ChatCommand.php
chr-hertel: u sure? :D

src/ai-bundle/src/Command/ChatCommand.php
OskarStark: Indeed, to much project development 😄

OskarStark:
I agree with @GromNaN that we can simplify this a lot if we only allow `my_agent` instead of `ai.agent.my_agent` **or** `my_agent`

GromNaN (approved):

OskarStark:
It works when I provide a correct one, or a wrong one. But it does not work when I provide none... investigating.

OskarStark:
Works now!

chr-hertel:
Common, i feel like you're trolling me ... :joy: 

image

I'm going to bed now - sad :D

OskarStark:
CleanShot 2025-09-03 at 07 56 53@2x
CleanShot 2025-09-03 at 07 57 09@2x

OskarStark:
Got a 👍 from @chr-hertel via WhatsApp 😄

src/ai-bundle/src/Command/ChatCommand.php
```diff
@@ -0,0 +1,213 @@
...
+                The chat session is interactive. Type your messages and press Enter to send.
+                Type 'exit' or 'quit' to end the conversation.
+                HELP
+            );
+    }
+
+    protected function interact(InputInterface $input, OutputInterface $output): void
+    {
+        // Skip interaction in non-interactive mode
+        if (!$input->isInteractive()) {
```
stof, line 83: useless check. The parent class already skips calling the `interact` method entirely.

src/ai-bundle/src/Command/ChatCommand.php
```diff
@@ -0,0 +1,213 @@
...
+    {
+        if ($input->mustSuggestArgumentValuesFor('agent')) {
+            $suggestions->suggestValues($this->getAvailableAgentNames());
+        }
+    }
+
+    protected function configure(): void
+    {
+        $this
+            ->addArgument('agent', InputArgument::OPTIONAL, 'The name of the agent to chat with')
```
stof, line 61: The argument should be marked as required, as you use the `interact` method to fill it anyway. this provides better documentation (and standard error reporting)

src/ai-bundle/src/Command/ChatCommand.php
```diff
@@ -0,0 +1,213 @@
...
+    {
+        // Initialize agent (moved from initialize() to execute() so it runs after interact())
+        $availableAgents = array_keys($this->agents->getProvidedServices());
+
+        if (0 === \count($availableAgents)) {
+            throw new InvalidArgumentException('No agents are configured.');
+        }
+
+        $agentArg = $input->getArgument('agent');
+        $this->agentName = \is_string($agentArg) ? $agentArg : '';
```
stof, line 129: Why using instance properties (making the service stateful) when they are only used locally ? You should use local variables instead.

OskarStark:
Thanks for your feedback @stof, makes sense

PR #398, opened 2025-09-01
[Store] Add Weaviate

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | None
| License       | MIT

Hi 👋🏻 

This PR aims to introduce the support for `Weaviate` as a store.


src/ai-bundle/config/options.php
```diff
@@ -348,6 +348,17 @@
                             ->end()
                         ->end()
                     ->end()
+                    ->arrayNode('weaviate')
+                    ->normalizeKeys(false)
```
OskarStark: Curious, why do we need this? Inline all others are using it too, but is it needed?

OskarStark: Created issue https://github.com/symfony/ai/issues/404

OskarStark:
```suggestion
```
not needed

src/ai-bundle/config/options.php
```diff
@@ -337,6 +337,16 @@
                             ->end()
                         ->end()
                     ->end()
+                    ->arrayNode('weaviate')
+                    ->useAttributeAsKey('name')
```
```suggestion
                        ->useAttributeAsKey('name')
```

chr-hertel:
Thank you @Guikingone.

PR #408, opened 2025-09-02
[Platform][Gemini][Embeddings] Remove comments for dimensions

OskarStark:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | Follows #359 
| License       | MIT

cc @natewiebe13

### DIFF
--- a/src/platform/src/Bridge/Gemini/Embeddings.php
+++ b/src/platform/src/Bridge/Gemini/Embeddings.php
@@ -20,11 +20,8 @@ use Symfony\AI\Platform\Model;
  */
 class Embeddings extends Model
 {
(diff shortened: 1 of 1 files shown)


valtzu:
I originally added these, because with at least mariadb store, the column definition must exactly match the model dimensions. Since it's not possible to define dimensions with all models, it may not be super obvious what's the issue when you get something like `"'?=!.."O?"#!#+12!?"!?...' is not a valid vector"` error from the database – that's why I though a little comment could improve the DX.

Anyway, not a problem for me anymore :v:

OskarStark:
Thanks for your feedback

PR #415, opened 2025-09-03
[AI Bundle] Improve `ChatCommand`

OskarStark:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | Follows #395 
| License       | MIT

cc @stof

### DIFF
--- a/src/ai-bundle/src/Command/ChatCommand.php
+++ b/src/ai-bundle/src/Command/ChatCommand.php
@@ -36,9 +36,6 @@ use Symfony\Component\DependencyInjection\ServiceLocator;
 )]
 final class ChatCommand extends Command
 {
-    private AgentInterface $agent;
-    private string $agentName;
-
     /**
      * @param ServiceLocator<AgentInterface> $agents
      */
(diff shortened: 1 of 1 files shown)


src/ai-bundle/src/Command/ChatCommand.php
```diff
@@ -126,29 +115,29 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
         }
 
         // If we still don't have an agent name at this point, something went wrong
-        if (!$this->agentName) {
+        if (!$agentName) {
             throw new InvalidArgumentException(\sprintf('Agent name is required. Available agents: "%s"', implode(', ', $availableAgents)));
         }
 
-        $this->agent = $this->agents->get($this->agentName);
+        $this->agent = $this->agents->get($agentName);
```
stof: this should also be a local variable

src/ai-bundle/src/Command/ChatCommand.php
```diff
@@ -126,29 +115,29 @@ protected function execute(InputInterface $input, OutputInterface $output): int
         }
 
         $agentArg = $input->getArgument('agent');
-        $this->agentName = \is_string($agentArg) ? $agentArg : '';
+        $agentName = \is_string($agentArg) ? $agentArg : '';
 
         // In non-interactive mode, agent is required
-        if (!$this->agentName && !$input->isInteractive()) {
+        if (!$agentName && !$input->isInteractive()) {
```
stof: this check can be removed entirely. The argument is always required by the time `execute` is called, but it is validated by Symfony.

PR #416, opened 2025-09-03
[Platform] Add Ollama prompt cache

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | None
| License       | MIT

Hi 👋🏻 

This PR aim to introduce a caching layer for `Ollama` platform (like OpenAI, Anthropic and more already does).


examples/ollama/chat-llama-with-cache.php
```diff
@@ -0,0 +1,39 @@
...
+    'prompt_cache_key' => 'chat',
+]);
+
+echo $result->getContent().\PHP_EOL;
+
+$secondResult = $agent->call($messages, [
+    'prompt_cache_key' => 'chat',
+]);
+
+echo $result->getContent().\PHP_EOL;
```
OskarStark: How can we ensure, that it really uses the cache and does not just return the exact same answer twice?

src/ai-bundle/config/options.php
```diff
@@ -88,6 +88,7 @@
                     ->arrayNode('ollama')
                         ->children()
                             ->scalarNode('host_url')->defaultValue('http://127.0.0.1:11434')->end()
+                            ->scalarNode('cache')->end()
```
VincentLanglet: We might end with the same cache repeated again and again in every platform.

Should we introduce a cache config key at a higher level ?

Guikingone: That's a good question, IMHO, we should introduce a "root" key for it and allowing the override "per platform", @OskarStark @chr-hertel, any idea?

examples/ollama/chat-llama-with-cache.php
```diff
@@ -0,0 +1,45 @@
...
+assert('chat' === $result->getMetadata()->get('prompt_cache_key'));
+
+$secondResult = $agent->call($messages, [
+    'prompt_cache_key' => 'chat',
+]);
+
+echo $secondResult->getContent().\PHP_EOL;
+
+assert($secondResult->getMetadata()->get('cached'));
+assert('chat' === $secondResult->getMetadata()->get('prompt_cache_key'));
```
junaidbinfarooq: **Claude**, IIRC, exposes extra cache information, e.g., `cache_creation_input_tokens` and `cache_read_input_tokens`, when you use _prompt caching_. If it is also true for ollama, exposing (and printing here) that information is probably a good option.

Guikingone: I added two keys to access those informations 🙂

src/platform/src/Bridge/Ollama/OllamaResultConverter.php
```diff
@@ -73,7 +74,18 @@ public function doConvertCompletion(array $data): ResultInterface
...
-        return new TextResult($data['message']['content']);
+        $result = new TextResult($data['message']['content']);
+
+        if (\array_key_exists('prompt_cache_key', $options)) {
+            $metadata = $result->getMetadata();
+
+            $metadata->add('cached', true);
+            $metadata->add('prompt_cache_key', $options['prompt_cache_key']);
+            $metadata->add('cached_prompt_count', $data['prompt_eval_count']);
+            $metadata->add('cached_completion_count', $data['eval_count']);
```
junaidbinfarooq: Wouldn't it make sense to group this data into a DTO, like there is `TokenUsage`, and then add that DTO to _metadata_ or perhaps even reuse the said DTO?

src/platform/tests/Bridge/Ollama/OllamaClientTest.php
```diff
@@ -175,4 +176,64 @@ public function testStreamingConverterWithDirectResponse()
...
+                ],
+            ],
+            'model' => 'llama3.2',
+        ], [
+            'prompt_cache_key' => 'foo',
+        ]);
+
+        $secondResult = $secondCall->getResult();
+
+        $this->assertSame('Hello world', $secondResult->getContent());
```
junaidbinfarooq:
```suggestion
        $firstCall = $platform->invoke(new Ollama(Ollama::LLAMA_3_2), [
            'messages' => [
                [
                    'role' => 'user',
                    'content' => 'Say hello world',
                ],
            ],
            'model' => 'llama3.2',
        ], [
            'prompt_cache_key' => 'foo',
        ]);
        $secondCall = $platform->invoke(new Ollama(Ollama::LLAMA_3_2), [
            'messages' => [
                [
                    'role' => 'user',
                    'content' => 'Say hello world',
                ],
            ],
            'model' => 'llama3.2',
        ], [
            'prompt_cache_key' => 'foo',
        ]);

        $firstResult = $firstCall->getResult();
        $secondResult = $secondCall->getResult();

        $this->assertSame('Hello world', $firstResult->getContent());
        $this->assertSame(10, $firstResult->getMetadata()->get('cached_prompt_count'));
        $this->assertSame(10, $firstResult->getMetadata()->get('cached_completion_count'));
        $this->assertSame('Hello world', $secondResult->getContent());
```

src/platform/src/Bridge/Ollama/OllamaResultConverter.php
Guikingone: Not convinced about the benefits of using an object here, we're only storing an integer, I don't see the benefits to be honest 🤔 

@OskarStark @chr-hertel Any thoughts?

chr-hertel:
Let's zoom a bit out here, for two reasons:
1. doesn't ollama caching already?
2. if we want to have it user land, why only ollama?

Guikingone:
> doesn't ollama caching already?

Ollama does a "context caching" and/or a K/V caching, it stores the X latest messages for the model window (or pending tokens to speed TTFT), it's not a cache that returns the generated response if the request already exist.

> if we want to have it user land, why only Ollama?

Well, because that's the one that I use the most and the easiest to implement first but we can integrate it for every platform if that's the question, we just need to use the API contract, both Anthropic and OpenAI already does it natively 🤔 

If the question is: Could we implement it at the platform layer for every platform without relying on API calls, well, that's not a big deal to be honest and we could easily integrate it 🙂

chr-hertel:
What do you think about having it as decorator `CachedPlatform` or similar?

Guikingone:
I like the idea of `CachedPlatform`, looks and sound like `HttpCache`, I'll rewrite it 👍🏻

src/ai-bundle/src/AiBundle.php
```diff
@@ -415,6 +415,16 @@ private function processPlatformConfig(string $type, array $platform, ContainerB
         }
 
         if ('ollama' === $type) {
+            $arguments = [
+                $platform['host_url'],
+                new Reference('http_client', ContainerInterface::NULL_ON_INVALID_REFERENCE),
```
```suggestion
                new Reference('http_client', ContainerInterface::NULL_ON_INVALID_REFERENCE),
                new Reference('ai.platform.model_catalog.ollama'),
```

src/ai-bundle/src/AiBundle.php
```diff
@@ -415,6 +415,16 @@ private function processPlatformConfig(string $type, array $platform, ContainerB
...
+                new Reference('ai.platform.contract.ollama'),
+            ];
+
+            if (\array_key_exists('cache', $platform)) {
+                $arguments[] = new Reference($platform['cache'], ContainerInterface::NULL_ON_INVALID_REFERENCE);
+            }
+
```
OskarStark, line 612: $arguments is not used

src/platform/src/Bridge/Ollama/OllamaResultConverter.php
chr-hertel, line 1: changes here would also belong into `CachedPlatform` so every bridge can benefit from this decorator

src/platform/src/Bridge/Ollama/OllamaResultConverter.php
chr-hertel: i agree, it would be great to have an object like `CacheUsage` similar to `TokenUsage`

docs/bundles/ai-bundle.rst
```diff
@@ -136,6 +136,28 @@ Advanced Example with Multiple Agents
                 vectorizer: 'ai.vectorizer.mistral_embeddings'
                 store: 'ai.store.memory.research'
 
+Cached platform
```
OskarStark:
```suggestion
Cached Platform
```

docs/bundles/ai-bundle.rst
```diff
@@ -136,6 +136,28 @@ Advanced Example with Multiple Agents
...
+---------------
+
+Thanks to Symfony's Cache component, platforms can be decorated and use any cache adapter,
+this platform  allows to reduce network calls / resource comsumption:
```
OskarStark:
```suggestion
this platform allows to reduce network calls / resource consumption:
```

docs/components/platform.rst
```diff
@@ -377,6 +377,29 @@ which can be useful to speed up the processing::
...
+    use Symfony\Component\Cache\Adapter\TagAwareAdapter;
+
+    $platform = PlatformFactory::create($apiKey, eventDispatcher: $dispatcher);
+    $cachedPlatform = new CachedPlatform($platform, new TagAwareAdapter(new ArrayAdapter());
+
+    $firstResult = $cachedPlatform->invoke('gpt-4o-mini', new MessageBag(Message::ofUser('What is the capital of France?')));
+
+    echo $firstResult->getContent().\PHP_EOL;
+
+    $secondResult = $cachedPlatform->invoke('gpt-4o-mini', new MessageBag(Message::ofUser('What is the capital of France?')));
```
OskarStark, line 427: maybe something like:

```suggestion
    // This call will not be executed against the API
    $secondResult = $cachedPlatform->invoke('gpt-4o-mini', new MessageBag(Message::ofUser('What is the capital of France?')));
```

examples/.env
```diff
@@ -16,7 +16,7 @@ VOYAGE_API_KEY=
 REPLICATE_API_KEY=
 
 # For using Ollama
-OLLAMA_HOST_URL=http://localhost:11434
+OLLAMA_HOST_URL=http://127.0.0.1:11434
```
OskarStark: Can we move this to an extra PR please?

src/ai-bundle/config/options.php
```diff
@@ -74,6 +74,16 @@
                             ->end()
                         ->end()
                     ->end()
+                    ->arrayNode('cache')
+                        ->useAttributeAsKey('name')
+                        ->arrayPrototype()
+                            ->children()
+                                ->stringNode('platform')->isRequired()->end()
+                                ->stringNode('service')->isRequired()->end()
+                                ->stringNode('cache_key')->end()
```
OskarStark: Can/should we provide a default key here?

src/ai-bundle/config/options.php
OskarStark: Thanks

src/platform/composer.json
```diff
@@ -65,12 +65,16 @@
...
         "phpunit/phpunit": "^11.5",
         "symfony/ai-agent": "@dev",
+        "symfony/cache": "^7.3|^8.0",
         "symfony/console": "^7.3|^8.0",
         "symfony/dotenv": "^7.3|^8.0",
         "symfony/finder": "^7.3|^8.0",
         "symfony/process": "^7.3|^8.0",
         "symfony/var-dumper": "^7.3|^8.0"
     },
+    "suggest": {
```
OskarStark: Please remove, in Symfony we decided against `suggest` section in `composer.json` files

src/platform/src/Bridge/Ollama/OllamaResultConverter.php
OskarStark: Yes `CacheUsage` would be a good fit

docs/components/platform.rst
Guikingone: Fixed 👍🏻

src/ai-bundle/config/options.php
```diff
@@ -164,6 +176,7 @@
                             ->booleanNode('api_catalog')
                                 ->info('If set, the Ollama API will be used to build the catalog and retrieve models information, using this option leads to additional HTTP calls')
                             ->end()
+                            ->scalarNode('cache')->end()
```
chr-hertel: this is not used anymore, is it?

Guikingone: Fixed 👍🏻

Guikingone:
IMHO, the token usage must be added in a separate PR, it requires to update existing platforms to send the informations in the result.

Out of the current scope for this PR IMO.

Guikingone:
@junaidbinfarooq I created an issue for the `TokenUsage` DTO 🙂

docs/components/platform.rst
```diff
@@ -405,6 +405,29 @@ which can be useful to speed up the processing::
         echo $result->asText().PHP_EOL;
     }
 
+Cached Platform Calls
+---------------------
+
+Thanks to Symfony's Cache component, platform's calls can be cached to reduce network/resources calls/consumption::
```
OskarStark:
```suggestion
Thanks to Symfony's Cache component, platform calls can be cached to reduce calls and resources consumption::
```

docs/components/platform.rst
OskarStark: unresolved, as I don't see the comment in the latest changeset or is it a Github issue? 🤔

src/ai-bundle/config/options.php
```diff
@@ -75,6 +75,18 @@
                             ->end()
                         ->end()
                     ->end()
+                    ->arrayNode('cache')
+                        ->useAttributeAsKey('name')
+                        ->arrayPrototype()
+                            ->children()
+                                ->stringNode('platform')->isRequired()->end()
+                                ->stringNode('service')->isRequired()->end()
```
OskarStark: lets add `->info('The cache service id or so......)`

docs/components/platform.rst
Guikingone: It is: 

```php
# ...

// Thanks to the cache adapter and the "prompt_cache_key" key, this call will not trigger any network call

$secondResult = $agent->call($messages, [
    'prompt_cache_key' => 'chat',
]);

assert($secondResult->getMetadata()->has('cached'));

echo $secondResult->getContent().\PHP_EOL;
```

OskarStark:
Thank you @Guikingone.

PR #423, opened 2025-09-03
[AiBundle][VertexAI] Fix service definition

### LINKED ISSUE #419: [Platform][VertexAI] Invalid service config
Hi, I tired to use Vertex AI:

```
ai:
    platform:
        vertexai:
            location: '%env(GOOGLE_CLOUD_LOCATION)%'
            project_id: '%env(GOOGLE_CLOUD_PROJECT)%'
```

but I've got

```
Unable to dump a service container if a parameter is an object or a resource, got "Symfony\Component\HttpClient\EventSourceHttpClient".
```

it seems it comes from

https://github.com/symfony/ai/commit/ff296bf8e5aefa07a8c7e032e34c21d2a878cf31#diff-f2828d6c9ecf4603ca1695c4958952ae24994a6e697e0a6e5b57f7f1c336ee47R261-R267

We should define a factory here instead, don't we?

valtzu:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #419
| License       | MIT

Fix a couple of VertexAI issues:

1. Bundle config was not working at all since instances were created build-time instead of building `Definition`s
2. Access token was fetched build-time

To add tests, we'd need to add `google/auth` to dev dependencies. If that's ok, I'll do that and remove exclusion from `phpstan.dist.neon`.

### DIFF
--- a/src/ai-bundle/composer.json
+++ b/src/ai-bundle/composer.json
@@ -25,6 +25,7 @@
         "symfony/string": "^6.4 || ^7.0"
     },
     "require-dev": {
+        "google/auth": "^1.47",
         "phpstan/phpstan": "^2.1",
         "phpunit/phpunit": "^11.5",
         "symfony/expression-language": "^6.4 || ^7.0",
--- a/src/ai-bundle/src/AiBundle.php
+++ b/src/ai-bundle/src/AiBundle.php
@@ -12,6 +12,7 @@
 namespace Symfony\AI\AiBundle;
 
 use Google\Auth\ApplicationDefaultCredentials;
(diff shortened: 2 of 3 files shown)


chr-hertel:
> To add tests, we'd need to add google/auth to dev dependencies.

Yes please :+1:

chr-hertel:
@1ed can you confirm this patch works for you?

OskarStark:
Thanks for fixing this bug @valtzu.

1ed:
> @1ed can you confirm this patch works for you?

It works, thank you!

PR #426, opened 2025-09-03
[AI Bundle] Fix agent name check while registering processors

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | 
| License       | MIT

The reason why tool calling was broken ... follow up of #352

### DIFF
--- a/src/ai-bundle/src/DependencyInjection/ProcessorCompilerPass.php
+++ b/src/ai-bundle/src/DependencyInjection/ProcessorCompilerPass.php
@@ -17,6 +17,8 @@ use Symfony\Component\DependencyInjection\Reference;
 
 class ProcessorCompilerPass implements CompilerPassInterface
 {
+    private const AGENT_PREFIX = 'ai.agent.';
+
     public function process(ContainerBuilder $container): void
     {
         $inputProcessors = $container->findTaggedServiceIds('ai.agent.input_processor');
@@ -32,7 +34,7 @@ class ProcessorCompilerPass implements CompilerPassInterface
                     }
 
                     $agent = $tag['agent'] ?? null;
-                    if (null === $agent || $agent === $serviceId) {
+                    if (null === $agent || self::AGENT_PREFIX.$agent === $serviceId) {
                         $priority = $tag['priority'] ?? 0;
                         $agentInputProcessors[] = [$priority, new Reference($processorId)];
                     }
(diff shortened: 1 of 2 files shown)


chr-hertel:
@VincentLanglet makes sense?

chr-hertel:
cc @OskarStark 
image

VincentLanglet:
> @VincentLanglet makes sense?

Not really, I think.

Attribute constructor was documented as
```
/**
     * @param string|null $agent the service id of the agent which will use this processor,
     *                           null to register this processor for all existing agents
     */
    public function __construct(
        public ?string $agent = null,
        public int $priority = 0,
    ) {
    }
```
So it's explicit it should be the serviceId.

I expect to use it with `#[AsInputProcessor(agent: 'ai.agent.foo')]` especially because I could register my own agent, without the `ai.agent` prefix.

The issue is in the config instead, it should have been
```
$container->setDefinition('ai.tool.agent_processor.'.$name, $toolProcessorDefinition)
                    ->addTag('ai.agent.input_processor', ['agent' => 'ai.agent.'.$name, 'priority' => -10])
                    ->addTag('ai.agent.output_processor', ['agent' => 'ai.agent.'.$name, 'priority' => -10]);
```
and so on...

chr-hertel:
@VincentLanglet well, actually you also documented it like this
https://github.com/symfony/ai/blob/299e1a7c11939f0f35b92bcdbea498444ccfc53f/src/ai-bundle/doc/index.rst?plain=1#L166-L168

Which makes sense to me, since the identifier a user might think of, is the one they used in the config for it :thinking:

VincentLanglet:
This is what I had in mind when implementing it https://github.com/symfony/ai/pull/428

That's also why I used `my_agent_id` and not `my_agent_name` in the doc example.
But maybe it's unclear ?

I was wondering if someone declare his own agent like any other service...
```
    my_agent_id:
        class: Symfony\AI\Agent
```

Also I dunno if we communicate a lot in symfony about the "name" of an agent rather than his id (which can be easily found with a container dump).

Maybe it's better if I update the doc from 
```
#[AsOutputProcessor(agent: 'my_agent_id')]
```
to
```
#[AsOutputProcessor(agent: 'ai.agent.my_agent_name')]
```
?

VincentLanglet:
@chr-hertel It seems legit for me to work on the service id rather than the name directly because we already do this for the platform.

The config is 
```
ai:
    platform:
        openai:
    agent:
        openai:
            platform: 'ai.platform.openai' // HERE THE SERVICE ID
```
and not
```
ai:
    platform:
        openai:
    agent:
        openai:
            platform: 'openai'
```

So the same way people will work with `ai.platform.openai` they should work with `ai.agent.openai` rather than just `openai`.

chr-hertel:
Alright, fair point, please fix the docs in #428 than :+1:

PR #453, opened 2025-09-04
[Platform][HuggingFace] Add tests for Bridge components

OskarStark:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | --
| License       | MIT

### DIFF
--- a/src/platform/tests/Bridge/HuggingFace/ApiClientTest.php
+++ b/src/platform/tests/Bridge/HuggingFace/ApiClientTest.php
@@ -1,0 +1,118 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Platform\Tests\Bridge\HuggingFace;
+
+use PHPUnit\Framework\Attributes\CoversClass;
+use PHPUnit\Framework\Attributes\Small;
+use PHPUnit\Framework\Attributes\TestDox;
+use PHPUnit\Framework\Attributes\UsesClass;
+use PHPUnit\Framework\TestCase;
+use Symfony\AI\Platform\Bridge\HuggingFace\ApiClient;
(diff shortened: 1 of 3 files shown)


src/platform/src/Bridge/HuggingFace/ResultConverter.php
```diff
@@ -79,7 +79,7 @@ public function convert(RawResultInterface|RawHttpResult $result, array $options
             Task::AUTOMATIC_SPEECH_RECOGNITION => new TextResult($content['text'] ?? ''),
             Task::CHAT_COMPLETION => new TextResult($content['choices'][0]['message']['content'] ?? ''),
             Task::FEATURE_EXTRACTION => new VectorResult(new Vector($content)),
-            Task::TEXT_CLASSIFICATION => new ObjectResult(ClassificationResult::fromArray(reset($content) ?? [])),
+            Task::TEXT_CLASSIFICATION => new ObjectResult(ClassificationResult::fromArray($content)),
```
OskarStark: > The ResultConverter incorrectly used reset() for text classification, causing type errors. Text classification should use the full content array like other classification tasks.

@chr-hertel what do you think?

src/platform/tests/Bridge/HuggingFace/ApiClientTest.php
```diff
@@ -0,0 +1,120 @@
...
+            $this->assertArrayHasKey('query', $options);
+            $this->assertNull($options['query']['inference_provider']);
+            $this->assertNull($options['query']['pipeline_tag']);
+
+            return new JsonMockResponse([]);
+        });
+
+        $apiClient = new ApiClient($httpClient);
+        $apiClient->models(null, null);
+    }
```
OskarStark: Does this make sense @chr-hertel or should we validate, that at least one needs to be set?

src/platform/src/Bridge/HuggingFace/ResultConverter.php
chr-hertel: which type error does claude get? :joy: 

no, never had that. looks wrong to me. llm gaslighting here.

PR #465, opened 2025-09-05
Feature/document indexing pipeline

### LINKED ISSUE #429: [Store][AI Bundle] Document Indexing Pipeline
A common scenario is to
* **load** documents from a specific source (e.g. folder of PDFs, set of entities, URLs, ...)
* **transform** them (e.g. split into chunks, normalize language, summarize, ...)
* **vectorize** them with an embedding model
* **store** them in a vector store
* **query** for them incl. keeping track of original sources while retrieval

A configuration for this could for example look like this:
```yaml
ai:
    indexer:
         my_indexer:
              loader: 'service_id'
              source: 'string' or ['array', 'of', 'string']
              transformer:
                  - 'service_id'
                  - 'service_id'
              embeddings:
                   class: 'Symfony\AI\Platform\Bridge\OpenAi\Embeddings'
                   name: 'text-embedding-ada-002'
              store: 'ai.store.chroma_db.my_store'
```

And we could also add a command here:
```bash
$ bin/console ai:store:index my_indexer
```
---
This should sit on top / rework
* `Symfony\AI\Store\Document\LoaderInterface`
* `Symfony\AI\Store\Document\TransformerInterface`
* `Symfony\AI\Store\Document\Vectorizer`
* `Symfony\AI\Store\Indexer`

OskarStark:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | Fixes #429 
| License       | MIT


examples/indexer/movies.php
```diff
@@ -0,0 +1,47 @@
...
+$store = new InMemoryStore();
+$indexer = new Indexer(
+    new TextFileLoader(),
+    [
+        new ReplaceTextTransformer(search: '## Plot', replace: '## Synopsis'),
+        new TextSplitTransformer(chunkSize: 500, overlap: 100),
+    ],
+    new Vectorizer($platform, new Embeddings('text-embedding-3-small')),
+    $store
+);
```
chr-hertel: :heart_eyes:

chr-hertel:
mixed feelings about the `__invoke` in the interfaces ... i know i started that, but it just looks awful when chaining ...

for example:
https://github.com/chr-hertel/nca-chat/blob/8f8366ffe9cde07ea53ed7cf728614c3bf4b53b1/src/Store/Indexer.php#L21-L26

OskarStark:
Allright let's go with real method names then

OskarStark:
> mixed feelings about the __invoke in the interfaces

fixed in
* #476

demo/src/Blog/Command/EmbedCommand.php
```diff
@@ -11,18 +11,20 @@
...
-use App\Blog\Embedder;
+use Symfony\AI\Store\IndexerInterface;
 use Symfony\Component\Console\Attribute\AsCommand;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Style\SymfonyStyle;
+use Symfony\Component\DependencyInjection\Attribute\Autowire;
 
 #[AsCommand('app:blog:embed', description: 'Create embeddings for Symfony blog and push to ChromaDB.')]
```
OskarStark: can be removed once we have a dedicated index command

demo/config/packages/ai.yaml
```diff
@@ -59,7 +59,8 @@ ai:
                 class: 'Symfony\AI\Platform\Bridge\OpenAi\Embeddings'
                 name: !php/const Symfony\AI\Platform\Bridge\OpenAi\Embeddings::TEXT_ADA_002
     indexer:
-        default:
+        blog:
```
OskarStark, line 62: lets be more specific here

src/store/src/Document/LoaderInterface.php
```diff
@@ -17,10 +17,10 @@
...
 {
     /**
-     * @param string               $source  Identifier for the loader to load the documents from, e.g. file path, folder, or URL.
+     * @param string|null          $source  Identifier for the loader to load the documents from, e.g. file path, folder, or URL. Can be null for InMemoryLoader.
      * @param array<string, mixed> $options loader specific set of options to control the loading process
      *
      * @return iterable<TextDocument> iterable of TextDocuments loaded from the source
      */
-    public function load(string $source, array $options = []): iterable;
+    public function load(?string $source, array $options = []): iterable;
```
OskarStark, line 25: Don't like the nullable source here @chr-hertel

demo/src/Blog/Command/EmbedCommand.php
```diff
@@ -32,7 +34,7 @@ protected function execute(InputInterface $input, OutputInterface $output): int
         $io = new SymfonyStyle($input, $output);
         $io->title('Loading RSS of Symfony blog as embeddings into ChromaDB');
 
-        $this->embedder->embedBlog();
+        $this->indexer->index('https://feeds.feedburner.com/symfony/blog');
```
chr-hertel: would be the `source` - as a next step

src/store/src/Document/LoaderInterface.php
chr-hertel: me neither, let's start more restrictive without nullable - can open up later as well

chr-hertel: or is there already a reason we'd need it for?

src/store/src/IndexerInterface.php
```diff
@@ -11,18 +11,18 @@
...
     /**
-     * @param TextDocument|iterable<TextDocument> $documents
-     * @param int                                 $chunkSize number of documents to vectorize and store in one batch
+     * Process sources through the complete document pipeline: load → transform → vectorize → store.
+     * 
+     * @param null|string|array<string> $source  Source identifier (file path, URL, etc.), array of sources, or null if the loader does not need a source, e.g. InMemoryLoader
+     * @param array{chunk_size?: int} $options Processing options
      */
-    public function index(TextDocument|iterable $documents, int $chunkSize = 50): void;
+    public function index(null|string|array $source = null, array $options = []): void;
```
chr-hertel: would also argue for removing `null` here

src/store/src/Indexer.php
```diff
@@ -13,31 +13,62 @@
...
-     * @param TextDocument|iterable<TextDocument> $documents
-     * @param int                                 $chunkSize number of documents to vectorize and store in one batch
-     */
-    public function index(TextDocument|iterable $documents, int $chunkSize = 50): void
+    public function index(null|string|array $source = null, array $options = []): void
     {
-        if ($documents instanceof TextDocument) {
-            $documents = [$documents];
+        // Prevent conflicting sources
+        if (null !== $source && null !== $this->source) {
```
OskarStark: this way we have both, we can set a default on config level, but cannot overwrite then, as one wants to overwrite, another one wants to merge. You can define/configure an indexer now with source = null, in this case you can provide it via the method.

src/store/src/Document/LoaderInterface.php
OskarStark: Yes the `InMemoryLoader` gets populated via its constructor.

chr-hertel:
Thank you @OskarStark.

PR #421, opened 2025-09-03
[Platform][Gemini] Fix choice conversion logic for `executableCode` and `codeExecutionResult`

pentiminax:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #422
| License       | MIT

This PR fixes a crash in the **Gemini** ResultConverter when handling responses that only contained executableCode and codeExecutionResult parts.

Before: 
- RuntimeException: Unsupported finish reason "STOP"

After:
- Proper conversion into a TextResult containing formatted code blocks and the last successful OUTCOME_OK output.
- No more exceptions for valid Gemini responses.

What’s Changed
- Enhanced choice conversion logic in ResultConverter::convertChoice():
- Single part → keep legacy behavior (functionCall, text, or one code/output part).
- Multiple parts → aggregate:
- executableCode → Markdown code blocks.
- codeExecutionResult (OUTCOME_OK) → output block.
- Parts flagged as thought: true are ignored for text/code, but their successful execution results are preserved.

New unit tests:
- Covers the failing payload where all parts were marked thought: true.
- Ensures the result is a readable TextResult instead of an exception.

Why
- Prevents crashes when Gemini outputs reasoning steps as thought: true.
- Keeps only useful information (code + last valid output).
- Maintains backward compatibility for simpler responses.

Impact
- No BC breaks — existing behavior for single-part responses unchanged.
- Improved UX — users now see clean Markdown code + output instead of runtime errors.

PR #466, opened 2025-09-05
[Platform][VertexAI] Update `ResultConverter`

pentiminax:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| License       | MIT

Same as #421 but for VertexAI bridge

This PR fixes a crash in the Gemini ResultConverter when handling responses that only contained executableCode and codeExecutionResult parts.

Rather than duplicating the code, I created a trait that is used in both bridges.
I also reused the same test fixtures.

### DIFF
--- a/src/platform/src/Bridge/VertexAi/Gemini/ResultConverter.php
+++ b/src/platform/src/Bridge/VertexAi/Gemini/ResultConverter.php
@@ -31,6 +31,10 @@ use Symfony\Contracts\HttpClient\ResponseInterface as HttpResponse;
  */
 final readonly class ResultConverter implements ResultConverterInterface
 {
+    public const OUTCOME_OK = 'OUTCOME_OK';
+    public const OUTCOME_FAILED = 'OUTCOME_FAILED';
+    public const OUTCOME_DEADLINE_EXCEEDED = 'OUTCOME_DEADLINE_EXCEEDED';
+
     public function supports(BaseModel $model): bool
     {
         return $model instanceof Model;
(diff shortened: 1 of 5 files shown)


src/platform/src/Bridge/VertexAi/Gemini/ResultConverter.php
```diff
@@ -11,138 +11,45 @@
 
 namespace Symfony\AI\Platform\Bridge\VertexAi\Gemini;
 
+use Symfony\AI\Platform\Bridge\Gemini\Gemini\ResultConverterTrait;
```
OskarStark: We cannot use classes from other bridges, we need to duplicate the code

pentiminax: I was not sure, I will do this so

src/platform/tests/Bridge/VertexAi/Gemini/ResultConverterTest.php
```diff
@@ -46,8 +46,54 @@ public function testItConvertsAResponseToAVectorResult()
...
         // Assert
-
         $this->assertInstanceOf(TextResult::class, $result);
         $this->assertSame('Hello, world!', $result->getContent());
     }
+
+    public function testItReturnsAggregatedTextOnSuccess()
+    {
+        $response = $this->createStub(ResponseInterface::class);
+        $responseContent = file_get_contents(\dirname(__DIR__, 6).'/fixtures/Bridge/Gemini/code_execution_outcome_ok.json');
```
OskarStark:
```suggestion
        $responseContent = file_get_contents(\dirname(__DIR__, 6).'/fixtures/Bridge/VertexAi/code_execution_outcome_ok.json');
```
same for the others

OskarStark:
... and one more bug fixed, thank you @pentiminax.

PR #474, opened 2025-09-05
[Store][Qdrant] Allow setting filters on a query

natewiebe13:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| License       | MIT

Allow setting filters on queries

### DIFF
--- a/src/store/src/Bridge/Qdrant/Store.php
+++ b/src/store/src/Bridge/Qdrant/Store.php
@@ -65,6 +65,7 @@ final readonly class Store implements ManagedStoreInterface, StoreInterface
 
     /**
      * @param array{
+     *     filter?: array<string, mixed>,
      *     limit?: positive-int,
      *     offset?: positive-int
      * } $options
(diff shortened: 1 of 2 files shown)


natewiebe13:
I see the code quality error is due to the new option. Qdrant's filtering is pretty complex, not sure exactly the best way to express the option and how useful it would be to try to match it exactly (especially if it changes)

src/store/src/Bridge/Qdrant/Store.php
```diff
@@ -65,6 +65,7 @@ public function add(VectorDocument ...$documents): void
 
     /**
      * @param array{
+     *     filter?: array<string, array>,
```
Guikingone: What about: 

```
filter?: array<string, mixed>,
```

?

natewiebe13: Works for me. Tests are passing now too

OskarStark:
Can you add tests for this new filter?

natewiebe13:
> Can you add tests for this new filter?

Added

chr-hertel (approved):

chr-hertel:
Thank you @natewiebe13.

PR #478, opened 2025-09-07
[Agent] Allow exposing exceptions to AI

### LINKED ISSUE #255: [Agent] Allow exposing some tool exceptions to the LLM
Currently we always return the same static response in https://github.com/symfony/ai/blob/9f0eec0c4cf79caafc2538d5314d3a85f0cc3636/src/agent/src/Toolbox/FaultTolerantToolbox.php#L41 

For example, if a tool execution is prevented due to `AccessDeniedException`, the LLM still sees this static error. 

I think we could have some exception interface, like

```php
interface ToolCallExceptionInterface 
{
    public function getToolCallResult(): ?string;
}
```
and then return that as tool call response instead of static message.

Use cases
---

1. The LLM would be able to tell the user "Sorry mate, permission denied to perform the requested action".
2. The LLM could auto-fix the issue of providing wrong parameters to the function call

valtzu:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | Fix #255
| License       | MIT

Make it possible to throw custom exceptions that expose the error to the LLM when using `FaultTolerantToolbox`.


fixtures/Tool/ToolCustomException.php
OskarStark, line 1: I think this should be part of the fixtures in agent component

src/agent/doc/index.rst
```diff
@@ -223,6 +223,8 @@ to the LLM::
 
     $agent = new Agent($platform, $model, inputProcessor: [$toolProcessor], outputProcessor: [$toolProcessor]);
 
+If you want to expose the underlying error to the LLM, you can throw a custom exception that implements `ToolExecutionExceptionInterface`.
+
```
OskarStark, line 256: Let's add an example here?

src/agent/tests/Toolbox/ToolboxTest.php
```diff
@@ -18,11 +18,13 @@
 use Symfony\AI\Agent\Toolbox\Attribute\AsTool;
 use Symfony\AI\Agent\Toolbox\Exception\ToolConfigurationException;
 use Symfony\AI\Agent\Toolbox\Exception\ToolExecutionException;
+use Symfony\AI\Agent\Toolbox\Exception\ToolExecutionExceptionInterface;
 use Symfony\AI\Agent\Toolbox\Exception\ToolNotFoundException;
 use Symfony\AI\Agent\Toolbox\Toolbox;
 use Symfony\AI\Agent\Toolbox\ToolFactory\ChainFactory;
 use Symfony\AI\Agent\Toolbox\ToolFactory\MemoryToolFactory;
 use Symfony\AI\Agent\Toolbox\ToolFactory\ReflectionToolFactory;
+use Symfony\AI\Fixtures\Tool\ToolCustomException;
```
OskarStark, line 27: Hmm, same for the other fixtures?

valtzu: You mean all fixtures in `Symfony\AI\Fixtures\Tool` namespace? I think they were only recently moved to root fixtures :thinking:

OskarStark: Can you check where they are used? If they are only used in agent then we should move them. 

cc @chr-hertel

valtzu: 70% of `Symfony\AI\Fixtures\Tool` usage is in Agent component and the rest 30% is in Platform

I'd probably keep them as they are now instead of having them in several locations. Or remove all these from the root (and move to the packages), and just duplicate the ones that are needed in several components. I think it is better DX-wise if a developer can add the new fixture where there already are existing ones, instead of needing to figure out which directory you should choose (in case we had both, root + in the package).

src/agent/doc/index.rst
valtzu: Added

OskarStark:
Shall we rename the PR title?

```diff
-[Agent] Allow exposing exceptions to AI
+[Agent] Allow exposing tool exceptions to AI
```

src/agent/doc/index.rst
```diff
@@ -223,6 +223,37 @@ to the LLM::
...
+        {
+        }
+
+        public function getToolCallResult(): mixed
+        {
+            return \sprintf('No %s found with id %d', $this->entityName, $this->id);
+        }
+    }
+
+    #[AsTool('get_user_age', 'Get age by user id')]
```
OskarStark, line 242: not sure if possible, but wouldn't it be better if you would be able to provide this info via callable in the attribute to not couple you code with the 3rd party exception?

```suggestion
    #[AsTool('get_user_age', 'Get age by user id', exceptions: [EntityNotFoundException::class => static fn(THE TOOL CALL ARGUMENTS HERE)....])]
```

Or is this not possible? Or is it a bad idea?

valtzu: Interesting idea,

but my reasoning for this feature is actually about throwing from an event listener of `ToolCallArgumentsResolved` – therefore I don't see this as a good place for it as you'd have to put it into every tool definition then.

Maybe this is just a bad example in the docs?

---

Another possibility to implement what I need could be an ability to define a custom error handler service for particular toolbox (or maybe even globally for all agents). How that could be implemented, f.e. instead of current `fault_tolerant_toolbox: false/true` you could set `fault_tolerant_toolbox: 'app.my_abstract_fault_tolerant_toolbox'` which would then use your service definition instead of builtin `FaultTolerantToolbox`.

I find the exception interface more convenient, however, it does have the downside of not being able to use/inject services to produce the result

OskarStark: I am fine with the interface, the error handler seems to complicated for such a "normal" use case, WDYT @chr-hertel ?

chr-hertel: I'm fine with that interface as well - valid use-case too me - and opt-in :+1:

OskarStark:
Thank you @valtzu.

PR #489, opened 2025-09-08
[Platform] Prepend `ImageNormalizer` in `OpenAIContract`

### LINKED ISSUE #488: [Platform][Azure OpenAI] Image message content serialization incorrect due to Normalizer ordering
# Summary
When I try to send an image to Azure OpenAI endpoint, I get an error of `Response does not contain choices.`. 

This is a red herring to the actual problem, which is an error response from the API:

```json
"error": {
  "message": "Invalid file data: 'messages[0].content[1].file.file_data'. Expected a base64-encoded data URL with an application/pdf MIME type (e.g. 'data:application/pdf;base64,SGVsbG8sIFdvcmxkIQ=='), but got unsupported MIME type 'image/png'.",
  "param": "messages[0].content[1].file.file_data",
}
```

Even though I am attaching an `Image`, it gets serialized as a plain `File`. This is because the `OpenAiContract` prepends its own `FileNormalizer`, which also matches instances of `Image` (because subclass and `is_a()` check).

A quick fix is to explicitly prepend the default `ImageNormalizer` in front. With that change the reproducer below starts working again. There are no other `File` subtypes that would be affected (`AudioNormalizer` is already prepended there). 

# Reproducer

```php
#[AsCommand(
    'debug:ai:chat-image',
)]
class ImageTestChatCommand
{
    public function __construct(
        #[Autowire(service: 'ai.agent.tester')] private readonly AgentInterface $agent,
    ) {
    }

    public function __invoke(InputInterface $input, OutputInterface $output) : int
    {
        $io = new SymfonyStyle($input, $output);

        $messages = new MessageBag();
        $messages->add(Message::ofUser(
            'What is in the image?',
            Image::fromFile('/Users/melka/cat.png'),
        ));

        $response = $this->agent->call(
            $messages,
        );

        $io->success($response->getContent());

        return Command::SUCCESS;
    }
}
```

melkamar:
Fixes serialization of Image messages as Image
rather than its parent File.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Docs?         | no 
| Issues        | Fix #488 
| License       | MIT

### DIFF
--- a/src/platform/src/Bridge/OpenAi/Contract/OpenAiContract.php
+++ b/src/platform/src/Bridge/OpenAi/Contract/OpenAiContract.php
@@ -13,6 +13,7 @@ namespace Symfony\AI\Platform\Bridge\OpenAi\Contract;
 
 use Symfony\AI\Platform\Bridge\OpenAi\Whisper\AudioNormalizer;
 use Symfony\AI\Platform\Contract;
+use Symfony\AI\Platform\Contract\Normalizer\Message\Content\ImageNormalizer;
 use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
 
 /**
@@ -24,6 +25,7 @@ final readonly class OpenAiContract extends Contract
     {
         return parent::create(
             new AudioNormalizer(),
+            new ImageNormalizer(),
             new FileNormalizer(),
             ...$normalizer
         );


melkamar:
Narrowing down the `FileNormalizer` makes this work for the `Image` case. But I can't comment on whether the normalizer was really intended only for `Document`.

There is a unit test in the [commit introducing it](https://github.com/symfony/ai/commit/e621538c419483406e910d917fe6fe7cf3270806) which asserts that the normalizer can also handle `File` instances.

chr-hertel:
> Narrowing down the `FileNormalizer` makes this work for the `Image` case. But I can't comment on whether the normalizer was really intended only for `Document`.
> 
> There is a unit test in the [commit introducing it](https://github.com/symfony/ai/commit/e621538c419483406e910d917fe6fe7cf3270806) which asserts that the normalizer can also handle `File` instances.

@glengemann can you comment? what was intended there?

glengemann:
Actually, the original intention was only to support `Document`.

chr-hertel:
Thanks @glengemann for the fast feedback!

Let's go with limiting the `FileNormalizer` to be a `DocumentNormalizer` for a fix than. If that works out at least.

OskarStark:
And renameing `FileNormalizer` to `DocumentNormalizer`, right?

glengemann:
Oskar, yes, that would be fine. We named it **file** to match the JSON key ("type": "file") in the request. However, now it seems more appropriate to call it Document.

```json
{
    "type": "file",
    "file": {
        "file_id": "file-6F2ksmvXxt4VdoqmHRw6kL"
    }
},
```

OskarStark:
I renamed the class in
https://github.com/symfony/ai/pull/493

chr-hertel:
Closed in favor of #496

PR #494, opened 2025-09-08
[Platform] Add automatic enum validation for backed enums

### LINKED ISSUE #492: [Platform] Backed Enums Support for Tool Arguments and Structured Output
Json Schema supports defining [enums](https://json-schema.org/understanding-json-schema/reference/enum) on properties, which is even supported with our `#[With]` attribute to be configured explicitly. But it would also be great to support this without attribute, when a Backed Enum is used as type:

```php
enum Mode: string
{
    case AND = 'and';
    case OR = 'or';
}

class MySearchTool
{
    public function __invoke(array $searchTerms, Mode $mode): array
    {
        // ...
    }
}
```

OskarStark:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | Fix #492
| License       | MIT

Add automatic enum validation for PHP's backed enum types (string and int) without requiring the `#[With]` attribute. This enables tool methods to use enums as parameters with automatic JSON schema generation that includes the enum values for validation.

### DIFF
--- a/fixtures/Tool/EnumMode.php
+++ b/fixtures/Tool/EnumMode.php
@@ -1,0 +1,19 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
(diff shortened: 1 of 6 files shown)


chr-hertel:
Thank you @OskarStark.

PR #503, opened 2025-09-09
[Examples] Fix `env()` function to check for empty string values

OskarStark:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | --
| License       | MIT

The env() function now checks both isset() and empty string to properly handle environment variables defined in .env but with empty values.


chr-hertel:
Thank you @OskarStark.

PR #506, opened 2025-09-09
[Agent] Show system message first in chat command

### LINKED ISSUE #434: [AI Bundle] `ai:chat` command does not show system prompt

xersion22:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #434 
| License       | MIT

This PR introduces the `getSystemMessage()` method to `AgentInterface` and implements it in the `Agent` class.
It also uses this method in `ChatCommand` to display the system message at the start of the chat session.

Tests were added to verify the new behaviour.

### DIFF
--- a/src/agent/src/Agent.php
+++ b/src/agent/src/Agent.php
@@ -122,4 +122,13 @@ final readonly class Agent implements AgentInterface
 
         return $processors instanceof \Traversable ? iterator_to_array($processors) : $processors;
     }
+
+    public function getSystemMessage(): ?string
+    {
+        $input = new Input($this->model, new MessageBag(), []);
+
+        array_map(fn (InputProcessorInterface $processor) => $processor->processInput($input), $this->inputProcessors);
+
+        return $input->messages->getSystemMessage()?->content;
+    }
 }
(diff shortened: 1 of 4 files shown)


OskarStark:
Thanks for your PR here, much appreciated.
I think this is not needed, we have the input processors in place, but somehow they are not executed when injecting the $agent via DI, can you maybe dig a little deeper? or check if it is registered in a correct way?

While this PR might work, its the wrong direction to me

xersion22:
> ot needed, we have the input processors in place, but somehow they are not executed when injecting the $agent via DI, can you maybe dig a little de

@OskarStark 
I am a bit confused with the comment. The code to display the system message appears in `ChatCommand` from `lines 159 to 163`.
The prompt  is fetched and displayed after the **first successful call** to the agent.
Inside the agent call, `processInput` is called inside an `array_map` which in turn populates the system message.

So in the chat command it is only possible to fetch the message once the agent call is executed (or `processInput` executed). The input processor is in place, no doubt. But are you hinting at calling the `processor` before/outside of the agent call ?

chr-hertel:
i guess the issue is with the processor adding that on the fly without mutating the original MessageBag but cloning it. i would favor to use some kind of platform decorator maybe to solve this - if even possible.

but i'm not convinced to extend the AgentInterface based on some debugging use case.

Issue #19, opened 2025-06-07
[Agent] Multi-Agent Runner


chr-hertel:
Currently the agent is only a slim wrapper around the Platform and basically only has a few features
* structured output (should that be in Platform?)
* input/output extension points (should that be in Platform?)
* validation of input & capabilities (should that be in Platform?)
* handling of multi-step tool calling <- the only feature atm?

One big feature we should enable is multi-agent with different characteristics:
1. orchestration agent
  one agent that is the main agent and basically uses other agents as tools (should be possible already)
2. agent hand-off
  multiple agents handing off the responsibility and context between each other

CashWilliams:
I'd love to see a2a support for the agent built in.

junaidbinfarooq:
@chr-hertel @OskarStark 
I would like to work on this one. Do you have any particular idea on its implementation? Following is a rough sketch of how it would work IMO.

The multi-agent system operates in a sequential handoff pattern where specialized agents collaborate to complete complex tasks:
### Setup Phase
**Agent Definition:** Each agent is created with specific capabilities (research, writing, review)
**System Prompts:** Agents are given specialized instructions and told when to hand off
**Handoff Rules:** The system creates rules for when agents should transfer control

### Execution Flow
`User Request → Orchestrator/Initial Agent → Specialized Agent 1 → Agent 2 → Agent 3 → Final Result`

### Two Execution Patterns
**Pattern A: Orchestrated Workflow**
An orchestrator agent receives the user request
It coordinates and delegates to specialized agents
Each agent completes its part and suggests the next handoff

**Pattern B: Direct Collaboration**
Agents work directly with each other without a central orchestrator
Each agent analyzes the conversation and decides when to hand off

### Handoff Triggers
The system detects handoffs through:
**Explicit mentions:** Agent says "handoff to writer" or "transfer to reviewer"
**Capability detection:** Agent mentions capabilities it doesn't have
**Automatic rules:** Pre-configured rules based on agent capabilities

This creates a collaborative pipeline where each agent contributes its expertise, resulting in higher-quality outputs than a single general-purpose agent could produce.

Let me know your thoughts.

OskarStark:
I don't have any concrete plans yet and Chris is off for some days. So maybe give it a try and draft a PR with a working example?

junaidbinfarooq:
[...]

PR #509, opened 2025-09-10
[Agent] Add `getName()` method to `AgentInterface`

OskarStark:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | Follows https://github.com/symfony/ai/issues/19#issuecomment-3273423359 and https://github.com/symfony/ai/issues/19#issuecomment-3273585513
| License       | MIT


chr-hertel:
to ease the implementation first, we could go with `interface NamedAgentInterface extends AgentInterface` and see where this is going - so you don't have to patch all existing things here, but play around with the `getName` until we're confident to promote it to the `AgentInterface` - would be a different approach to nullable at least

OskarStark:
Hmm, I would make it mandatory in the state we are, we could ofc go with: 

```diff
public function __construct(
    private PlatformInterface $platform,
    private Model $model,
    iterable $inputProcessors = [],
    iterable $outputProcessors = [],
-    private ?string $name = null,
+    private string $name = 'agent',
    private LoggerInterface $logger = new NullLogger(),
) {
```
to always have a string

chr-hertel:
Same for the instructions?

OskarStark:
> Same for the instructions?

Lets discuss this in #19

PR #514, opened 2025-09-11
[Agent] Support translations for system input processor

### LINKED ISSUE #370: [Agent] Support Translations with System Prompt Processor
Currently it is only possible to define the system prompt directly in the `ai.yaml` of the bundle, but it would be great to reference a translation token, which requires the `SystemPromptProcessor` to have an optional dependency to the `TranslatorInterface`.

paulinevos:
So that system prompts can be written in any locale

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Docs?         | no (there's no documentation for system input, so not sure where to put it. Open to ideas! 
| Issues        | Fix #370  
| License       | MIT

This adds an optional translator to the `SystemInputProcessor` which, if set, will translate the configured system prompt.

### DIFF
--- a/src/agent/composer.json
+++ b/src/agent/composer.json
@@ -41,7 +41,8 @@
         "symfony/css-selector": "^6.4 || ^7.1",
         "symfony/dom-crawler": "^6.4 || ^7.1",
         "symfony/event-dispatcher": "^6.4 || ^7.1",
-        "symfony/http-foundation": "^6.4 || ^7.1"
+        "symfony/http-foundation": "^6.4 || ^7.1",
+        "symfony/translation-contracts": "^3.6"
     },
     "autoload": {
         "psr-4": {
--- a/src/agent/src/InputProcessor/SystemPromptInputProcessor.php
+++ b/src/agent/src/InputProcessor/SystemPromptInputProcessor.php
@@ -13,11 +13,13 @@ namespace Symfony\AI\Agent\InputProcessor;
 
 use Psr\Log\LoggerInterface;
 use Psr\Log\NullLogger;
+use Symfony\AI\Agent\Exception\RuntimeException;
 use Symfony\AI\Agent\Input;
 use Symfony\AI\Agent\InputProcessorInterface;
 use Symfony\AI\Agent\Toolbox\ToolboxInterface;
 use Symfony\AI\Platform\Message\Message;
 use Symfony\AI\Platform\Tool\Tool;
+use Symfony\Contracts\Translation\TranslatorInterface;
 
 /**
  * @author Christopher Hertel <mail@christopher-hertel.de>
@@ -31,8 +33,14 @@ final readonly class SystemPromptInputProcessor implements InputProcessorInterfa
     public function __construct(
         private \Stringable|string $systemPrompt,
         private ?ToolboxInterface $toolbox = null,
+        private ?TranslatorInterface $translator = null,
+        private bool $enableTranslation = false,
+        private ?string $translationDomain = null,
         private LoggerInterface $logger = new NullLogger(),
     ) {
+        if ($this->enableTranslation && !$this->translator) {
+            throw new RuntimeException('Prompt translation is enabled but no translator was provided.');
+        }
     }
 
     public function processInput(Input $input): void
@@ -45,7 +53,9 @@ final readonly class SystemPromptInputProcessor implements InputProcessorInterfa
             return;
         }
 
-        $message = (string) $this->systemPrompt;
+        $message = $this->enableTranslation
+            ? $this->translator->trans((string) $this->systemPrompt, [], $this->translationDomain)
+            : (string) $this->systemPrompt;
 
         if ($this->toolbox instanceof ToolboxInterface
             && [] !== $this->toolbox->getTools()
@@ -61,7 +71,7 @@ final readonly class SystemPromptInputProcessor implements InputProcessorInterfa
             ));
 
             $message = <<<PROMPT
-                {$this->systemPrompt}
+                {$message}
 
                 # Tools
 
--- a/src/agent/tests/InputProcessor/SystemPromptInputProcessorTest.php
+++ b/src/agent/tests/InputProcessor/SystemPromptInputProcessorTest.php
@@ -29,6 +29,7 @@ use Symfony\AI\Platform\Message\UserMessage;
 use Symfony\AI\Platform\Result\ToolCall;
 use Symfony\AI\Platform\Tool\ExecutionReference;
 use Symfony\AI\Platform\Tool\Tool;
+use Symfony\Contracts\Translation\TranslatorInterface;
 
 #[CoversClass(SystemPromptInputProcessor::class)]
 #[UsesClass(Gpt::class)]
@@ -89,7 +90,7 @@ final class SystemPromptInputProcessorTest extends TestCase
                 {
                     return null;
                 }
-            }
+            },
         );
 
         $input = new Input(new Gpt(), new MessageBag(Message::ofUser('This is a user message')));
@@ -105,7 +106,7 @@ final class SystemPromptInputProcessorTest extends TestCase
     public function testIncludeToolDefinitions()
     {
         $processor = new SystemPromptInputProcessor(
-            'This is a system prompt',
+            'This is a',
             new class implements ToolboxInterface {
                 public function getTools(): array
                 {
@@ -127,7 +128,9 @@ final class SystemPromptInputProcessorTest extends TestCase
                 {
                     return null;
                 }
-            }
+            },
+            $this->getTranslator(),
+            true,
         );
 
         $input = new Input(new Gpt(), new MessageBag(Message::ofUser('This is a user message')));
@@ -138,7 +141,7 @@ final class SystemPromptInputProcessorTest extends TestCase
         $this->assertInstanceOf(SystemMessage::class, $messages[0]);
         $this->assertInstanceOf(UserMessage::class, $messages[1]);
         $this->assertSame(<<<PROMPT
-            This is a system prompt
+            This is a cool translated system prompt
 
             # Tools
 
@@ -169,7 +172,7 @@ final class SystemPromptInputProcessorTest extends TestCase
                 {
                     return null;
                 }
-            }
+            },
         );
 
         $input = new Input(new Gpt(), new MessageBag(Message::ofUser('This is a user message')));
@@ -189,5 +192,67 @@ final class SystemPromptInputProcessorTest extends TestCase
             ## tool_no_params
             A tool without parameters
             PROMPT, $messages[0]->content);
+    }
+
+    public function testWithTranslatedSystemPrompt()
+    {
+        $processor = new SystemPromptInputProcessor('This is a', null, $this->getTranslator(), true);
+
+        $input = new Input(new Gpt(), new MessageBag(Message::ofUser('This is a user message')), []);
+        $processor->processInput($input);
+
+        $messages = $input->messages->getMessages();
+        $this->assertCount(2, $messages);
+        $this->assertInstanceOf(SystemMessage::class, $messages[0]);
+        $this->assertInstanceOf(UserMessage::class, $messages[1]);
+        $this->assertSame('This is a cool translated system prompt', $messages[0]->content);
+    }
+
+    public function testWithTranslationDomainSystemPrompt()
+    {
+        $processor = new SystemPromptInputProcessor(
+            'This is a',
+            null,
+            $this->getTranslator(),
+            true,
+            'prompts'
+        );
+
+        $input = new Input(new Gpt(), new MessageBag(), []);
+        $processor->processInput($input);
+
+        $messages = $input->messages->getMessages();
+        $this->assertCount(1, $messages);
+        $this->assertInstanceOf(SystemMessage::class, $messages[0]);
+        $this->assertSame('This is a cool translated system prompt with a translation domain', $messages[0]->content);
+    }
+
+    public function testWithMissingTranslator()
+    {
+        $this->expectExceptionMessage('Prompt translation is enabled but no translator was provided');
+
+        new SystemPromptInputProcessor(
+            'This is a',
+            null,
+            null,
+            true,
+        );
+    }
+
+    private function getTranslator(): TranslatorInterface
+    {
+        return new class implements TranslatorInterface {
+            public function trans(string $id, array $parameters = [], ?string $domain = null, ?string $locale = null): string
+            {
+                $translated = \sprintf('%s cool translated system prompt', $id);
+
+                return $domain ? $translated.' with a translation domain' : $translated;
+            }
+
+            public function getLocale(): string
+            {
+                return 'en';
+            }
+        };
     }
 }
--- a/src/ai-bundle/config/options.php
+++ b/src/ai-bundle/config/options.php
@@ -18,6 +18,7 @@ use Symfony\AI\Platform\Bridge\OpenAi\PlatformFactory;
 use Symfony\AI\Platform\PlatformInterface;
 use Symfony\AI\Store\Document\VectorizerInterface;
 use Symfony\AI\Store\StoreInterface;
+use Symfony\Contracts\Translation\TranslatorInterface;
 
 return static function (DefinitionConfigurator $configurator): void {
     $configurator->rootNode()
@@ -154,6 +155,12 @@ return static function (DefinitionConfigurator $configurator): void {
                                 })
                                 ->thenInvalid('The "text" cannot be empty.')
                             ->end()
+                            ->validate()
+                                ->ifTrue(function ($v) {
+                                    return \is_array($v) && $v['enabled'] && !interface_exists(TranslatorInterface::class);
+                                })
+                                ->thenInvalid('System prompt translation is enabled, but no translator is present. Try running `composer require symfony/translation`.')
+                            ->end()
                             ->children()
                                 ->scalarNode('text')
                                     ->info('The system prompt text')
@@ -161,6 +168,14 @@ return static function (DefinitionConfigurator $configurator): void {
                                 ->booleanNode('include_tools')
                                     ->info('Include tool definitions at the end of the system prompt')
                                     ->defaultFalse()
+                                ->end()
+                                ->booleanNode('enable_translation')
+                                    ->info('Enable translation for the system prompt')
+                                    ->defaultFalse()
+                                ->end()
+                                ->scalarNode('translation_domain')
+                                    ->info('The translation domain for the system prompt')
+                                    ->defaultNull()
                                 ->end()
                             ->end()
                         ->end()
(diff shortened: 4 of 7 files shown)


src/agent/composer.json
```diff
@@ -30,6 +30,7 @@
         "symfony/property-access": "^6.4 || ^7.1",
         "symfony/property-info": "^6.4 || ^7.1",
         "symfony/serializer": "^6.4 || ^7.1",
+        "symfony/translation": "^6.4 || 7.3",
```
OskarStark:
```suggestion
        "symfony/translation": "^6.4 || ^7.1",
```
like for the others

src/agent/tests/InputProcessor/SystemPromptInputProcessorTest.php
```diff
@@ -186,4 +187,37 @@ public function execute(ToolCall $toolCall): mixed
             A tool without parameters
             PROMPT, $messages[0]->content);
     }
+
+    public function testWithTranslatedSystemPrompt()
+    {
+        $processor = new SystemPromptInputProcessor(
+            systemPrompt: 'This is a',
+            toolbox: null,
```
OskarStark:
```suggestion
```

src/agent/tests/InputProcessor/SystemPromptInputProcessorTest.php
```diff
@@ -186,4 +187,37 @@ public function execute(ToolCall $toolCall): mixed
...
+            translator: $this->getTranslator()
```
OskarStark: lets inline the translator here, we use it just in one testcase

OskarStark:
Can we please make the translator configurable in the bundle? Thanks

src/ai-bundle/tests/DependencyInjection/AiBundleTest.php
```diff
@@ -471,11 +475,15 @@ public function testMultipleAgentsWithProcessors()
         $firstSystemPrompt = $container->getDefinition('ai.agent.first_agent.system_prompt_processor');
         $firstSystemTags = $firstSystemPrompt->getTag('ai.agent.input_processor');
         $this->assertSame($firstAgentId, $firstSystemTags[0]['agent']);
+        $this->assertCount(2, array_filter($firstSystemPrompt->getArguments()));
```
paulinevos: Wasn't super sure how to test this, let me know if this is sufficient.

src/ai-bundle/config/options.php
```diff
@@ -125,13 +125,25 @@
...
+                                ->end()
+                                ->scalarNode('translator')
+                                    ->defaultNull()
+                                    ->info('Service name of translator for the system prompt')
+                                ->end()
                             ->end()
-                            ->defaultNull()
-                            ->info('The default system prompt of the agent')
                         ->end()
                         ->booleanNode('include_tools')
```
OskarStark: basically this needs to be moved also within the system_prompt, but can be a follow up.

src/ai-bundle/src/AiBundle.php
```diff
@@ -585,11 +585,12 @@ private function processAgentConfig(string $name, array $config, ContainerBuilde
         }
 
         // SYSTEM PROMPT
-        if (\is_string($config['system_prompt'])) {
+        if (\array_key_exists('system_prompt', $config)) {
             $systemPromptInputProcessorDefinition = (new Definition(SystemPromptInputProcessor::class))
                 ->setArguments([
-                    $config['system_prompt'],
+                    $config['system_prompt']['system_prompt'],
```
```suggestion
                    $config['system_prompt']['prompt'],
```
shall we name it prompt then if it is nested?

src/ai-bundle/tests/DependencyInjection/AiBundleTest.php
```diff
@@ -447,7 +448,10 @@ public function testMultipleAgentsWithProcessors()
                         'tools' => [
                             ['service' => 'tool_two', 'description' => 'Tool for second agent'],
                         ],
-                        'system_prompt' => 'Second agent prompt',
```
OskarStark: we should keep this testcase to see if it is working on the first level, right?

src/ai-bundle/config/options.php
```diff
@@ -125,13 +125,25 @@
...
+                                    ->validate()
+                                        ->ifTrue(fn ($v) => null !== $v && '' === trim($v))
+                                        ->thenInvalid('The default system prompt must not be an empty string')
+                                    ->end()
+                                    ->defaultNull()
+                                    ->info('The default system prompt of the agent')
+                                ->end()
+                                ->scalarNode('translator')
+                                    ->defaultNull()
+                                    ->info('Service name of translator for the system prompt')
```
VincentLanglet: @chr-hertel @OskarStark 

I feel like we shouldn't give a translator class (it should inject automatically the `TranslatorInterface` which is registered like in any constructor of a Symfony application)

What should be passed instead is a translation domain with the value
string|false|null.

Then the code will be
```
if ($domain === false) {
     $message = (string) $this->systemPrompt;
} else {
     $message = $this->translator->trans((string) $this->systemPrompt, domain: $domain);
}
```

Also, it might require some reflection about how we pass parameters...

src/ai-bundle/src/AiBundle.php
paulinevos: Yeah agree

src/ai-bundle/tests/DependencyInjection/AiBundleTest.php
paulinevos: There's a `First agent prompt` that covers that, so I wanted to change this one to ensure both syntaxes work

src/ai-bundle/config/options.php
paulinevos: By domain do you mean `locale`?

VincentLanglet: No I mean the translation domain
```
public function trans(string $id, array $parameters = [], ?string $domain = null, ?string $locale = null): string;
```

By defaut it's `messages.yaml` but I can prefer to use the domain `prompt.yaml` (or any other domain) for my prompt for exemple.

paulinevos: Ah right okay. I'll wait for the general consensus to go and change things, as making the translator configurable was a request on the PR, but what you say makes sense.

chr-hertel:
so, what we just discussed as config structure:
```yaml
system_prompt:
    prompt: '...',
    enable_translation: true|false, # default: true if translator is present
    translation_domain: '...', # default: "messages"
    include_tools: true|false # default same as now
```

VincentLanglet:
> so, what we just discussed as config structure:
> 
> ```yaml
>     enable_translation: true|false, # default: true if translator is present
>     translation_domain: '...', # default: "messages"
> ```

@chr-hertel I feel like this is kinda redundant, you could have only one key by allowing `string|null (default to message)|false` like lot of other places where translation is used

OskarStark:
You are right, but it's some kind of a hidden feature. It's not clear, that string or null means, the translator is enabled. So lets go with @chr-hertel proposal

OskarStark:
Fyi 
* https://github.com/symfony/ai/pull/576
* https://github.com/symfony/ai/pull/579

> no (there's no documentation for system input, so not sure where to put it. Open to ideas!

I also added initial documentation in #576 which can be extended now

VincentLanglet:
> Fyi
> 
> * [[AI Bundle] Restructure `system_prompt` configuration to nest `include_tools` #576](https://github.com/symfony/ai/pull/576)
> * [[AI Bundle] Rename `system_prompt` to `prompt` and `prompt` key to `text` #579](https://github.com/symfony/ai/pull/579)
> 
> > no (there's no documentation for system input, so not sure where to put it. Open to ideas!
> 
> I also added initial documentation in #576 which can be extended now

I fee like that with a `prompt.text` config key, it shouldn't be translated by default. So I expect
```
enable_translation: true|false, # default: false
translation_domain: '...', # default: "messages"
```

paulinevos:
> it shouldn't be translated by default

Why not?

OskarStark:
Yeah we can disable by default

VincentLanglet:
> > it shouldn't be translated by default
> 
> Why not?

I dunno if there is a lot of example in the existing Symfony config for such "translated text" (I only have tanslation domain in mind like validators). But it's called `"text"` not translation key or anything similar/related to any translation.

OskarStark called it a "hidden feature." and if it is, I feel like it's odd having a "hidden modification of the prompt given", which will ends in unexpected results.

But more important, even if I agree it's a useful feature, I feel like the translated prompt is something which will more often not used rather than used. 
- Because I can have question in which the answer does not depend on the local: "What is the sum of 2+3" 
- Or because I can also directly ask for a specific locale "What is ... . Please answer in french".
- And because prompt is a really delicate thing, something adding one extra word or removing one will ends up with a different result/answer ; it requires lot of testing to get the right prompt. And it will be ten times harder to have the same quality of prompt in different local.

paulinevos:
@VincentLanglet My reasoning is, if there's translation tokens in the text then that would be intentional. And if there are none we'd get the untranslated result. Then for the edge case where translation token syntax might somehow need to be in the prompt, there's an option to disable translation. 

But I've now implemented it with `false` as the default.

VincentLanglet:
> @VincentLanglet My reasoning is, if there's translation tokens in the text then that would be intentional. 

I agree

> And if there are none we'd get the untranslated result. Then for the edge case where translation token syntax might somehow need to be in the prompt, there's an option to disable translation.

It depends if you're concern about
- having a useless trans call
- having a "missing translation" in your profiler etc

I personally try to avoid useless call and missing translations ; but this is maybe not the symfony philosophy

> But I've now implemented it with `false` as the default.

👍

src/ai-bundle/config/options.php
```diff
@@ -162,6 +162,14 @@
                                     ->info('Include tool definitions at the end of the system prompt')
                                     ->defaultFalse()
                                 ->end()
+                                ->booleanNode('enable_translation')
+                                    ->info('Enable translation for the system prompt')
+                                    ->defaultFalse()
+                                ->end()
+                                ->scalarNode('translation_domain')
+                                    ->info('The translation domain for the system prompt')
+                                    ->defaultNull()
```
chr-hertel, line 178: let's use the default domain as default here - makes it easier to get started:
```suggestion
                                    ->defaultValue('messages')
```

src/ai-bundle/src/AiBundle.php
```diff
@@ -611,6 +612,9 @@ private function processAgentConfig(string $name, array $config, ContainerBuilde
                 ->setArguments([
                     $config['prompt']['text'],
                     $includeTools ? new Reference('ai.toolbox.'.$name) : null,
+                    TranslatorInterface::class,
```
chr-hertel: needs to be a service reference instead:
```suggestion
                    new Reference('translator', ContainerInterface::IGNORE_ON_INVALID_REFERENCE),
```

src/agent/src/InputProcessor/SystemPromptInputProcessor.php
```diff
@@ -45,7 +49,9 @@ public function processInput(Input $input): void
             return;
         }
 
-        $message = (string) $this->systemPrompt;
+        $message = $this->translator && $this->enableTranslation
```
chr-hertel: this var needs to be used in L70 as well for the combined prompt with tools

chr-hertel (changes requested):

src/agent/src/InputProcessor/SystemPromptInputProcessor.php
```diff
@@ -31,6 +32,9 @@
     public function __construct(
         private \Stringable|string $systemPrompt,
         private ?ToolboxInterface $toolbox = null,
+        private ?TranslatorInterface $translator = null,
+        private ?bool $enableTranslation = false,
```
VincentLanglet:
```suggestion
        private bool $enableTranslation = false,
```
no need for null ?

src/ai-bundle/src/AiBundle.php
```diff
@@ -611,6 +612,9 @@ private function processAgentConfig(string $name, array $config, ContainerBuilde
...
+                    $config['prompt']['enable_translation'],
```
VincentLanglet, line 615: Does it make sens to have both
- Translator
- enable_translation

in the constructor @chr-hertel ?

I would expect to just pass the translator with something like
```
$config['prompt']['enable_translation'] ? new Reference('translator') : null
```

And maybe if you use enable_translation = true, but have no registered translator we should throw an exception asking the user to install symfony/translator instead of a being "silent" about the error.

chr-hertel: i'd keep it separate, but you're right - we could get an error here.
i guess in that case we should throw an exception, that people need to install the translator. something like
```php
if (!interface_exists(TranslatorInterface::class)) {
    throw new RuntimeException('For using the translations with the prompt, the translator is required. Try running "composer require symfony/translation".');
}
```

OskarStark: Yes this check is needed

src/ai-bundle/config/options.php
OskarStark: but `null` uses the default translation domain, so I would stick to `defaultNull()` here

src/ai-bundle/config/options.php
```diff
@@ -142,7 +142,7 @@
                                 ->ifArray()
                                 ->then(function (array $v) {
                                     if (!isset($v['text']) && !isset($v['include_tools'])) {
-                                        throw new \InvalidArgumentException('Either "text" must be configured for prompt.');
+                                        throw new \InvalidArgumentException('Either "text" or "tools" must be configured for prompt.');
```
```suggestion
                                        throw new \InvalidArgumentException('Either "text" or "include_tools" must be configured for prompt.');
```

src/ai-bundle/config/options.php
paulinevos: Exactly. Passing in `messages` here makes assumptions about what the translator dictates as the default domain. It would make us "know" about something outside of our concern.

OskarStark: null is fine

src/agent/src/InputProcessor/SystemPromptInputProcessor.php
```diff
@@ -45,7 +49,9 @@ public function processInput(Input $input): void
             return;
         }
```
OskarStark, line 55: lets add a validation and throw if translator is null and enable translation is true, WDYT?

paulinevos: Yes, agree. I will make the changes on Monday

src/ai-bundle/src/AiBundle.php
paulinevos: Correct me if I'm wrong, but isn't it always going to exist as it's a dependency of `symfony-ai/agent`?

VincentLanglet: > Correct me if I'm wrong, but isn't it always going to exist as it's a dependency of `symfony-ai/agent`?

Indeed with the current state of the PR.
But maybe it should be an optional dependency (?)

src/agent/composer.json
```diff
@@ -30,6 +30,7 @@
         "symfony/property-access": "^6.4 || ^7.1",
         "symfony/property-info": "^6.4 || ^7.1",
         "symfony/serializer": "^6.4 || ^7.1",
+        "symfony/translation": "^6.4 || ^7.1",
```
VincentLanglet: It could be in `suggest` section (and require-dev for tests) rather than in the require section if we consider it should be an optional dependency.

I let the call on @OskarStark @chr-hertel side

BTW where do you use `symfony/translation` @paulinevos ?

I only see
Symfony\Contracts\Translation\TranslatorInterface;

So you could just require symfony/translation-contracts (and suggest symfony/translation)

src/ai-bundle/src/AiBundle.php
paulinevos: Can it be? It's imported into `SystemPromptInputProcessor` for type hinting the constructor. Either way there's a null check in that class now so it will be validated on that level at least :)

src/agent/composer.json
OskarStark: I would vote for for an optional one, but in symfony, we don't use `suggest` section

paulinevos: That's true. It still wouldn't make the `interface_exists(TranslatorInterface::class` make sense, but it possibly saves a bit of space on the `translations` dependency.

I also don't think I even need to `require-dev` it <del>if</del> because I just inject a mock implementation of the interface.

OskarStark: But your mock implements the interface, which would not be available otherwise

paulinevos: @OskarStark It is, because I would still require `symfony/translation-contracts` instead of the entire `translations` package

paulinevos:
Hmm, I'm not super convinced about using `TranslatableMessage` in this way, as it seems to add some complexity for a fairly simple use case and it puts a lot of behavior in the bundle that (in my eyes) "belongs to" `SystemPromptInputProcessor`

src/agent/src/InputProcessor/SystemPromptInputProcessor.php
```diff
@@ -73,4 +78,15 @@ public function processInput(Input $input): void
 
         $input->messages = $messages->prepend(Message::forSystem($message));
     }
+
+    private function translateIfEnabled(): string
+    {
+        if ($this->enableTranslation && !$this->translator) {
```
OskarStark: Shall we do this validation in the constructor?

src/ai-bundle/src/AiBundle.php
OskarStark: My point is, as long you are not using the SystemPromptInputProcessor, you don't need the translation package, do you? Sonia the bundle when we decide, someone wants/needs to use this processor, he needs to bring the translation stuff. Otherwise not and it totally optional

VincentLanglet:
> Hmm, I'm not super convinced about using `TranslatableMessage` in this way, as it seems to add some complexity for a fairly simple use case and it puts a lot of behavior in the bundle that (in my eyes) "belongs to" `SystemPromptInputProcessor`

We could argue the opposite.

Currently we're adding complexity in SystemPromptInputProcessor which has to be aware about
- if we want enable translation or not
- the translation domain

Such things should be "prompt" knownledge rather than SystemPromptInputProcessor one. Quoting the original PR https://github.com/symfony/symfony/pull/37670
> The service shouldn't need to care how the message is going to be handled, but just needs to create it in a way that it can be translated if needed.

TranslatableMessage is a way to avoid putting this knownledge inside the SystemPromptInputProcessor

It's also a way to allow using translation parameters.

src/ai-bundle/src/AiBundle.php
paulinevos: The dependency was already removed because of the other parallel thread

OskarStark (approved):

OskarStark:
It took time, but here we go, this is in now. Thank you very much @paulinevos.

VincentLanglet:
@OskarStark would you be interested with a Poc with translatableMessage ?

OskarStark:
Why not, shouldn't be to much code, right?

PR #515, opened 2025-09-11
[Platform] Add Docker Model Runner

### LINKED ISSUE #531: [Platform] Add Docker Model Runner
I'm on it!

https://docs.docker.com/ai/model-runner/

- [x] Implement basic support to run simple models
- [x] Add example with a model supporting tool calling
- [ ] ~~Add example with a model supporting image / audio / etc~~ Impossible for now because of an error from DMR: `"message" => "image input is not supported - hint: if this is unexpected, you may need to provide the mmproj"`, I postpone this example.
- [x] Wire in AiBundle

welcoMattic:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Docs?         | no 
| Issues        | Fix https://github.com/symfony/ai/issues/531 
| License       | MIT

This PR adds Docker Model Runner as supported platform to host and run models.

I base it on LmStudio bridge as it works similarly.

DMR docs:
- Getting Started https://docs.docker.com/ai/model-runner/get-started/
- API Reference https://docs.docker.com/ai/model-runner/api-reference/

### Examples output:

image


welcoMattic:
Consider this first implementation as basic bridge for basic usage. It could be improved during the hackathon by:

- Adding examples (tool calling, image, etc)
- Add guard about model presence (using "Get model" endpoint https://docs.docker.com/ai/model-runner/api-reference/#available-dmr-endpoints) to avoid calling a non-present model

examples/dockermodelrunner/embeddings.php
```diff
@@ -0,0 +1,24 @@
...
+require_once dirname(__DIR__).'/bootstrap.php';
+
+$platform = PlatformFactory::create(env('DOCKER_MODEL_RUNNER_HOST_URL'), http_client());
+$response = $platform->invoke(new Embeddings(Embeddings::NOMIC_EMBED_TEXT), <<<TEXT
+    Once upon a time, there was a country called Japan. It was a beautiful country with a lot of mountains and rivers.
+    The people of Japan were very kind and hardworking. They loved their country very much and took care of it. The
+    country was very peaceful and prosperous. The people lived happily ever after.
+    TEXT);
+
+echo 'Dimensions: '.$response->asVectors()[0]->getDimensions().\PHP_EOL;
```
OskarStark: we have a print_vectors()

src/platform/src/Bridge/DockerModelRunner/Completions.php
```diff
@@ -0,0 +1,72 @@
...
+    ) {
+        $capabilities = [];
+
+        foreach (self::TOOL_PATTERNS as $pattern => $possibleCapabilities) {
+            if (1 === preg_match($pattern, $name)) {
+                foreach ($possibleCapabilities as $capability) {
+                    $capabilities[] = $capability;
+                }
+            }
+        }
```
OskarStark: What if we just don't do it? we would get an error from the API, right?

src/platform/tests/Bridge/DockerModelRunner/Completions/ModelClientTest.php
```diff
@@ -0,0 +1,108 @@
...
+
+    public function testItIsExecutingTheCorrectRequest()
+    {
+        $resultCallback = static function (string $method, string $url, array $options): MockResponse {
+            self::assertSame('POST', $method);
+            self::assertSame('http://localhost:1234/engines/v1/chat/completions', $url);
+            self::assertSame(
+                '{"model":"test-model","messages":[{"role":"user","content":"Hello, world!"}]}',
+                $options['body']
+            );
```
OskarStark: lets use one line for such things

src/platform/src/Bridge/DockerModelRunner/Completions.php
welcoMattic: That's being part of things we should test and improve today

OskarStark: Otherwise the user would be locked, once a model will gain new capabilities, and would need for a new release. Same thing, we didn't validate to much in user land for the notifier bridges like max message length etc.

welcoMattic: So, we provide all Capabilities by default?

OskarStark: I would say we provide all capabilities we provide and are able to handle.

welcoMattic:
> [!NOTE]
> PHPStan failure is expected as example code use not yet merged Bridge in Platform package.

chr-hertel:
oh, that's due to #535 and not using the components from the branch but via composer install, right?
cc @VincentLanglet

VincentLanglet:
> oh, that's due to #535 and not using the components from the branch but via composer install, right? cc @VincentLanglet

Indeed, I'll take a look.

welcoMattic:
@VincentLanglet if you need any help about setting up DMR, I'm on Slack and WorkAdventure (main hall)

VincentLanglet:
> @VincentLanglet if you need any help about setting up DMR, I'm on Slack and WorkAdventure (main hall)

I feel like this can be easily fixed with https://github.com/symfony/ai/pull/552/commits/93464f6a3e6e4db4520bf155852cd4525402e63a

But I get a infinite loop 
https://github.com/symfony/ai/actions/runs/17674610626/job/50233715432?pr=552

I assume it's because of https://github.com/symfony/ai/issues/527 cc @chr-hertel 

Not sure if we can do something until...

VincentLanglet:
I'm getting something working https://github.com/symfony/ai/pull/568

proof https://github.com/symfony/ai/pull/552

examples/dockermodelrunner/chat.php
```diff
@@ -0,0 +1,31 @@
...
+$model = new Completions();
+
+$agent = new Agent($platform, $model, logger: logger());
+$messages = new MessageBag(
+    Message::forSystem('You are a pirate and you write funny.'),
+    Message::ofUser('What is the Symfony framework?'),
+);
+$result = $agent->call($messages, [
+    'max_tokens' => 500, // specific options just for this call
+]);
```
chr-hertel:
```suggestion
$result = $agent->call($messages);
```

examples/.env
```diff
@@ -19,6 +19,9 @@ REPLICATE_API_KEY=
 OLLAMA_HOST_URL=
 OLLAMA_MODEL=
 
+# For using Docker Model Runner
+DOCKER_MODEL_RUNNER_HOST_URL=
```
chr-hertel, line 24: does it make sense to provide the default here as well?
```suggestion
DOCKER_MODEL_RUNNER_HOST_URL=http://127.0.0.1:12434
```

welcoMattic: Port is customizable, I would prefer let it empty to "force" users to set it properly matching their settings in Docker Desktop.

chr-hertel:
still unable to install and test locally - sorry - can someone else test and confirm?

src/platform/src/Bridge/DockerModelRunner/Completions.php
```diff
@@ -0,0 +1,47 @@
...
+    public const GPT_OSS = 'ai/gpt-oss';
+    public const SMOLLM_2 = 'ai/smollm2';
+    public const SMOLLM_3 = 'ai/smollm3';
+
+    public function __construct(
+        string $name = self::SMOLLM_2,
+        array $options = [],
+    ) {
+        // All capabilities are assumed to be supported since we cannot know in advance
+        // whether Docker Model Runner and/or each model allows for a particular capability.
```
OskarStark:
```suggestion
```

src/platform/src/Bridge/DockerModelRunner/Embeddings.php
```diff
@@ -0,0 +1,35 @@
...
+    public const MXBAI_EMBED_LARGE = 'ai/mxbai-embed-large';
+    public const EMBEDDING_GEMMA = 'ai/embeddinggemma';
+    public const GRANITE_EMBEDDING_MULTI = 'ai/granite-embedding-multilingual';
+
+    public function __construct(
+        string $name = self::NOMIC_EMBED_TEXT,
+        array $options = [],
+    ) {
+        // All capabilities are assumed to be supported since we cannot know in advance
+        // whether Docker Model Runner and/or each model allows for a particular capability.
```
OskarStark:
```suggestion
```

welcoMattic:
> still unable to install and test locally - sorry - can someone else test and confirm? 

What do you need to be able to test? More doc about setting up Docker Model Runner?

src/ai-bundle/config/options.php
```diff
@@ -106,6 +106,11 @@
                             ->scalarNode('api_key')->isRequired()->end()
                         ->end()
                     ->end()
+                    ->arrayNode('dockermodelrunner')
```
OskarStark, line 158: Following #604 , can you make it configurable?

welcoMattic: Done

src/platform/src/Bridge/DockerModelRunner/Embeddings.php
```diff
@@ -0,0 +1,33 @@
...
+ */
+class Embeddings extends Model
+{
+    public const NOMIC_EMBED_TEXT = 'ai/nomic-embed-text-v1.5';
+    public const MXBAI_EMBED_LARGE = 'ai/mxbai-embed-large';
+    public const EMBEDDING_GEMMA = 'ai/embeddinggemma';
+    public const GRANITE_EMBEDDING_MULTI = 'ai/granite-embedding-multilingual';
+
+    public function __construct(
+        string $name = self::NOMIC_EMBED_TEXT,
```
OskarStark:
```suggestion
        string $name,
```

src/platform/src/Bridge/DockerModelRunner/Completions.php
```diff
@@ -0,0 +1,45 @@
...
+    public const MISTRAL_NEMO = 'ai/mistral-nemo';
+    public const PHI_4 = 'ai/phi4';
+    public const DEEPSEEK_R_1 = 'ai/deepseek-r1-distill-llama';
+    public const SEED_OSS = 'ai/seed-oss';
+    public const GPT_OSS = 'ai/gpt-oss';
+    public const SMOLLM_2 = 'ai/smollm2';
+    public const SMOLLM_3 = 'ai/smollm3';
+
+    public function __construct(
+        string $name = self::SMOLLM_2,
```
OskarStark:
```suggestion
        string $name,
```

examples/.env
OskarStark: I am in favor of adding it here, but add an output to the examples like "Calling on....."

OskarStark (approved):
CleanShot 2025-09-17 at 09 27 53@2x

Good to merge after my comment, it works seamlessly on my Mac

OskarStark:
I forgot to install one model and ended up with:
CleanShot 2025-09-17 at 09 20 45@2x

If possible, it would be nice to throw a `MissingModelException`, if we can check that beforehand. Can be a follow up PR

OskarStark:
Thanks Mathieu for working on this feature, this is much appreciated.

examples/.env
OskarStark: Fixed in
* https://github.com/symfony/ai/pull/610

PR #537, opened 2025-09-12
[AiBundle][Perplexity] Add platform configuration support

### LINKED ISSUE #534: [AiBundle] Wire Perplexity bridge
Perplexity bridge has been merged lately, but it's not yet wired in the AiBundle. Let's dot it!

To do so, take example on already wired bridges in https://github.com/symfony/ai/blob/main/src/ai-bundle/src/AiBundle.php#L211

matyo91:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Docs?         | yes 
| Issues        | Fix #534 
| License       | MIT


src/ai-bundle/CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+0.2
+---
+
+ * Add Perplexity platform support with API key configuration
+
```
welcoMattic: 0.1 is not yet released, so you can move your change under 0.1

src/ai-bundle/doc/index.rst
```diff
@@ -334,6 +342,67 @@ To disable token usage tracking for an agent, set the ``track_token_usage`` opti
                     class: 'Symfony\AI\Platform\Bridge\OpenAi\Gpt'
                     name: !php/const Symfony\AI\Platform\Bridge\OpenAi\Gpt::GPT_4O_MINI
 
+Perplexity Platform
```
welcoMattic: I'm not sure that this is the right place for such a documentation.
As we are looking for splitting bridges in dedicated packages (like for Mailer and Notifier in Symfony), maybe the future README.md of `symfony/ai-perplexity-platform` package will be the right place.

matyo91: I'm lookin forward instructions about this one to put the documentation into the right place.

src/ai-bundle/doc/index.rst
```diff
@@ -93,6 +95,12 @@ Configuration
                 tools: # If undefined, all tools are injected into the agent, use "tools: false" to disable tools.
                     - 'Symfony\AI\Agent\Toolbox\Tool\Wikipedia'
                 fault_tolerant_toolbox: false # Disables fault tolerant toolbox, default is true
+            search_agent:
+                platform: 'ai.platform.perplexity'
+                model:
+                    class: 'Symfony\AI\Platform\Bridge\Perplexity\Perplexity'
+                    name: !php/const Symfony\AI\Platform\Bridge\Perplexity\Perplexity::SONAR
+                tools: false # Perplexity has built-in search capabilities
```
OskarStark:
```suggestion
                tools: false
```

src/ai-bundle/doc/index.rst
OskarStark: Lets remove this completely

src/ai-bundle/src/AiBundle.php
```diff
@@ -453,6 +454,24 @@ private function processPlatformConfig(string $type, array $platform, ContainerB
             return;
         }
 
+        if ('perplexity' === $type && isset($platform['api_key'])) {
```
OskarStark: I know we are doing it on all the other platforms, but can this be the case that `isset($platform['api_key'])` returns `false`?

kniziol: Instead of using the `perplexity` string, let’s think about using `const` or `enum`. This way, we can make sure that the values are consistent. What do you think?

**Before:**
```php
if ('perplexity' === $type && isset($platform['api_key'])) {
```
**After:**
```php
enum Platform: string
{
    case Perplexity = 'perplexity';
    case Voyage = 'voyage';
}
```
```php
if (Platform:: Perplexity->value === $type && isset($platform['api_key'])) {
```

matyo91: I put this in a separated commit https://github.com/symfony/ai/pull/537/commits/52143ffa5a4e026d6d180281b12a94e32282f182#diff-d619aee53c1e2e62adb789e1bc05048cb3956974560414d85b0d3905b5bda9f7, in case of refactoring, i need a proper namespace name : eg `Symfony\AI\AiBundle\Type\PlatformType`

src/ai-bundle/src/PlatformType.php
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+namespace Symfony\AI\AiBundle;
+
+/**
+ * @author Christopher Hertel <mail@christopher-hertel.de>
+ */
+enum PlatformType: string
+{
+    case Anthropic = 'anthropic';
```
welcoMattic: As this change is not (only) related to Perplexity bridge, let's move it in another PR to discuss the topic separately. Same for StoreType.

src/ai-bundle/src/AiBundle.php
matyo91: If the configuration validation is working correctly, this check should never be false.

I did update the code : https://github.com/symfony/ai/pull/537/commits/3a8544e1c991f0df2ed5417504c04b630dd5ab23

kniziol: As @welcoMattic mentioned, it’s best to create a separate PR for that. We also need to refactor other code blocks that aren’t related to Perplexity. Thanks.

src/ai-bundle/src/PlatformType.php
matyo91: I did a new related issue https://github.com/symfony/ai/issues/542 and PR https://github.com/symfony/ai/pull/546

src/ai-bundle/src/AiBundle.php
matyo91: cf https://github.com/symfony/ai/pull/537#discussion_r2343714938

OskarStark:
Thanks Mathieu for working on this feature, this is much appreciated.

welcoMattic:
Just figured out that Perplexity Bridge comes with a `TokenOutputProcessor` and it was not wired.

See https://github.com/symfony/ai/blob/main/src/ai-bundle/config/services.php#L139

There also is the `SearchResultProcessor`, should we wire it too?

cc @matyo91 @OskarStark

OskarStark:
> There also is the SearchResultProcessor, should we wire it too?

We should, yes

Issue #528, opened 2025-09-11
[Platform] Expose API errors for better Developer Experience


chr-hertel:
Currently the Platform does not really handle API errors well, which is quite a pain for developers when running into errors. It's basically invisible what the API complains about.

For example:
> PHP Fatal error:  Uncaught Symfony\AI\Platform\Exception\RuntimeException: Response does not contain any content. in /home/christopher/Projects/SymfonyAI/repo/src/platform/src/Bridge/Anthropic/ResultConverter.php:48
Stack trace:
#0 /home/christopher/Projects/SymfonyAI/repo/src/platform/src/Result/ResultPromise.php(48): Symfony\AI\Platform\Bridge\Anthropic\ResultConverter->convert()
#1 /home/christopher/Projects/SymfonyAI/repo/src/platform/src/Result/ResultPromise.php(37): Symfony\AI\Platform\Result\ResultPromise->await()
#2 /home/christopher/Projects/SymfonyAI/repo/src/agent/src/Agent.php(93): Symfony\AI\Platform\Result\ResultPromise->getResult()
#3 /home/christopher/Projects/SymfonyAI/repo/examples/anthropic/image-input-binary.php(32): Symfony\AI\Agent\Agent->call()
#4 {main}
  thrown in /home/christopher/Projects/SymfonyAI/repo/src/platform/src/Bridge/Anthropic/ResultConverter.php on line 48

See #167 for inspiration.

floranpagliai:
I often run into rate limits when using Claude. I’d be happy to contribute a RateExceededException (or similar) that would expose the Retry-After header value. This would allow devs to implement proper retry logic based on the API’s response.

Sound goods ?

vitalii-kyktov:
We are going to look into improving the 401 error handling for OpenAI

chr-hertel:
We just discussed that it makes sense to have both: a high level that basically reacts to all kind of 4xx/5xx and at least converts the result to string instead of hiding it - and the possibility to through specific ones in ModelClient implementations

floranpagliai:
> We just discussed that it makes sense to have both: a high level that basically reacts to all kind of 4xx/5xx and at least converts the result to string instead of hiding it - and the possibility to through specific ones in ModelClient implementations

Marked my PR as ready for review. It introduces a rate limit exception for three providers, all of which return a 429 error but with different specificities:
	•	Gemini: no documented headers (I’ve reached out to our Gemini engineer for clarification).
[...]

PR #538, opened 2025-09-12
[Platform] Implement `RateLimitExceededException` handling across AI platforms

floranpagliai:
| Q             | A
| ------------- | ---
| Bug fix?      |no
| New feature?  |no 
| Docs?         | no 
| Issues        | #528 
| License       | MIT

  - Created a `RateLimitExceededException` to handle 429 across all AI platforms
  - Unified the handling of rate limit responses with platform-specific retry information
  - Used platform-specific header extraction for each provider:
    - Anthropic: Standard retry-after HTTP header (seconds)
    - Gemini: No retry information provided
    - OpenAI: Custom headers x-ratelimit-reset-requests and x-ratelimit-reset-tokens

### DIFF
--- a/src/platform/src/Bridge/Anthropic/ResultConverter.php
+++ b/src/platform/src/Bridge/Anthropic/ResultConverter.php
@@ -11,6 +11,7 @@
 
 namespace Symfony\AI\Platform\Bridge\Anthropic;
 
+use Symfony\AI\Platform\Exception\RateLimitExceededException;
 use Symfony\AI\Platform\Exception\RuntimeException;
 use Symfony\AI\Platform\Model;
 use Symfony\AI\Platform\Result\RawHttpResult;
@@ -38,8 +39,16 @@ class ResultConverter implements ResultConverterInterface
 
     public function convert(RawHttpResult|RawResultInterface $result, array $options = []): ResultInterface
     {
+        $response = $result->getObject();
+
+        if (429 === $response->getStatusCode()) {
+            $retryAfter = $response->getHeaders(false)['retry-after'][0] ?? null;
+            $retryAfterValue = $retryAfter ? (float) $retryAfter : null;
+            throw new RateLimitExceededException($retryAfterValue);
+        }
+
         if ($options['stream'] ?? false) {
-            return new StreamResult($this->convertStream($result->getObject()));
+            return new StreamResult($this->convertStream($response));
         }
 
         $data = $result->getData();
--- a/src/platform/src/Bridge/OpenAi/Gpt/ResultConverter.php
+++ b/src/platform/src/Bridge/OpenAi/Gpt/ResultConverter.php
@@ -14,6 +14,7 @@ namespace Symfony\AI\Platform\Bridge\OpenAi\Gpt;
 use Symfony\AI\Platform\Bridge\OpenAi\Gpt;
 use Symfony\AI\Platform\Exception\AuthenticationException;
 use Symfony\AI\Platform\Exception\ContentFilterException;
+use Symfony\AI\Platform\Exception\RateLimitExceededException;
 use Symfony\AI\Platform\Exception\RuntimeException;
 use Symfony\AI\Platform\Model;
 use Symfony\AI\Platform\Result\ChoiceResult;
@@ -50,8 +51,21 @@ final class ResultConverter implements ResultConverterInterface
             throw new AuthenticationException($errorMessage);
         }
 
+        if (429 === $response->getStatusCode()) {
+            $headers = $response->getHeaders(false);
+            $resetTime = null;
+
+            if (isset($headers['x-ratelimit-reset-requests'][0])) {
+                $resetTime = self::parseResetTime($headers['x-ratelimit-reset-requests'][0]);
+            } elseif (isset($headers['x-ratelimit-reset-tokens'][0])) {
+                $resetTime = self::parseResetTime($headers['x-ratelimit-reset-tokens'][0]);
+            }
+
+            throw new RateLimitExceededException($resetTime);
+        }
+
         if ($options['stream'] ?? false) {
-            return new StreamResult($this->convertStream($result->getObject()));
+            return new StreamResult($this->convertStream($response));
         }
 
         $data = $result->getData();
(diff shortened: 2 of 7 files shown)


src/platform/src/Bridge/Anthropic/ResultConverter.php
```diff
@@ -38,8 +39,16 @@ public function supports(Model $model): bool
 
     public function convert(RawHttpResult|RawResultInterface $result, array $options = []): ResultInterface
     {
+        $response = $result->getObject();
+
+        if (429 === $response->getStatusCode()) {
+            $retryAfter = $response->getHeaders(false)['retry-after'][0] ?? null;
+            $retryAfterValue = $retryAfter ? (float) $retryAfter : null;
+            throw new RateExceededException('Rate limit exceeded', $retryAfterValue);
```
VincentLanglet: If the message is a generic one (and I think it should be a generic one), it might be better to define it directly in the exception constructor, so
```
throw new RateExceededException($retryAfterValue);
```
and
```
public function __construct(
        public readonly ?float $retryAfter = null,
    ) {
        parent::__construct('Rate limit exceeded');
    }
```

src/platform/src/Bridge/Anthropic/ResultConverter.php
```diff
@@ -38,8 +39,16 @@ public function supports(Model $model): bool
 
     public function convert(RawHttpResult|RawResultInterface $result, array $options = []): ResultInterface
     {
+        $response = $result->getObject();
+
+        if (429 === $response->getStatusCode()) {
```
kniziol, line 44:
```suggestion
        if (Response::HTTP_TOO_MANY_REQUESTS === $response->getStatusCode()) {
```
Plus this import at the top:
```php
use Symfony\Component\HttpFoundation\Response;
```

floranpagliai: I hesited to use Response as all other check I found in the repo where not using it.

OskarStark: Lets stick to the 429

src/platform/src/Bridge/OpenAi/Gpt/ResultConverter.php
```diff
@@ -186,4 +202,25 @@ private function convertToolCall(array $toolCall): ToolCall
...
+
+    /**
+     * Converts OpenAI's reset time format (e.g. "1s", "6m0s", "2m30s") into seconds.
+     *
+     * Supported formats:
+     * - "1s"
+     * - "6m0s"
+     * - "2m30s"
+     */
+    private function parseResetTime(string $resetTime): float
```
OskarStark: can be static?

src/platform/src/Exception/RateExceededException.php
```diff
@@ -0,0 +1,24 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Platform\Exception;
+
+/**
+ * @author Floran Pagliai <floran.pagliai@gmail.com>
+ */
+final class RateExceededException extends \RuntimeException implements ExceptionInterface
```
OskarStark:
```suggestion
final class RateLimitExceededException extends \RuntimeException implements ExceptionInterface
```

src/platform/src/Exception/RateExceededException.php
```diff
@@ -0,0 +1,24 @@
...
+{
+    public function __construct(
+        public readonly ?float $retryAfter = null,
```
OskarStark: lets make it private and add a getter

OskarStark:
lets call it `RateLimitExceededException`

floranpagliai:
@OskarStark installation of PHPStan is flakky, don't know if you can launch again the CI.

floranpagliai:
> Could be nice to document ResultConverterInterface::convert that it can throws exception then.
> 
> With something like
> 
> ```
>     /**
>      * @param array<string, mixed> $options
>      *
>      * @throws RateExceededException when rate limit exceeeded
>      * @throws RuntimeException when another error occured
>      */
>     public function convert(RawResultInterface $result, array $options = []): ResultInterface;
> ```

Agreed, will include it in a try to have a higher level ResultConverter.

src/platform/src/Exception/RateLimitExceededException.php
```diff
@@ -0,0 +1,29 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Platform\Exception;
+
+/**
+ * @author Floran Pagliai <floran.pagliai@gmail.com>
+ */
+final class RateLimitExceededException extends \RuntimeException implements ExceptionInterface
```
chr-hertel: don't see this as a RuntimeException tho
```suggestion
final class RateLimitExceededException extends \InvalidArgumentException implements ExceptionInterface
```

floranpagliai: I get your point, but I’m not sure \InvalidArgumentException is the best fit here. A rate limit being exceeded isn’t really about invalid input, it’s more of an external condition depending of usage. That’s why I went with \RuntimeException.

Another option could be to extend directly from a generic \Exception or define a dedicated base exception for API-related errors ?

VincentLanglet: Maybe my point of view, but i consider
- a LogicException being an exception which should never happen unless bad code is written, and the dev should fix his code. (Bad param, bad call, case which shouldn't exist, etc)
- a RuntimeException being an exception which is not avoidable and might occurs under some circonstances

Here, it seems to be a Runtime one to me.

VincentLanglet: But what about extending directly the symfony/ai RuntimeException
https://github.com/symfony/ai/blob/main/src/platform/src/Exception/RuntimeException.php

rather than extending `\RuntimeException` and implementing the ExceptionInterface ?

floranpagliai: Went with directly extending https://github.com/symfony/ai/blob/main/src/platform/src/Exception/RuntimeException.php

chr-hertel (approved):

chr-hertel:
Thank you @floranpagliai.

PR #562, opened 2025-09-12
[Ollama] Add RAG examples and a small README file

damijanc:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | yes
| Examples | yes
| License       | MIT

Ollama RAG examples and a readme file.

### DIFF
--- a/examples/ollama/Ollama.md
+++ b/examples/ollama/Ollama.md
@@ -1,0 +1,30 @@
+# Symfony Ollama Examples
+
+This directory contains various examples of how to use the Symfony AI with [Ollama](https://ollama.com/).
+
+## Getting started
+
+To get started with Ollama please check their [Quickstart guide](https://github.com/ollama/ollama/blob/main/README.md#quickstart).
+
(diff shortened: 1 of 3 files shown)


examples/ollama/Ollama.md
```diff
@@ -0,0 +1,30 @@
...
+
+Once models are downloaded you can run them with
+```bash
+ollama run <model-name>
+```
+for example
+
+```bash
+ollama run llama3.2
+```
```
welcoMattic, line 22: Running models is not mandatory, but serving them is. Using `ollama serve` command. BTW running an embedding model is not possible using Ollama, output is not human readable.

chr-hertel: true, so to run the examples it would be:
```
ollama pull llama3.2
ollama pull nomic-embed-text
ollama serve
```

chr-hertel:
Thank you @damijanc.

PR #552, opened 2025-09-12
 Try

VincentLanglet:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no 
| Docs?         | yes/no 
| Issues        | Fix #... 
| License       | MIT

PR #568, opened 2025-09-12
 Fix running PHPStan on example

VincentLanglet:
Proof https://github.com/symfony/ai/pull/552


VincentLanglet:
cc @OskarStark @chr-hertel ; this should unblock multiple PR with a red phpstan build

welcoMattic:
Thank you @VincentLanglet for taking care of this regression.

Issue #17, opened 2025-06-07
[Agent] Agent Execution Trace


chr-hertel:
While working with an agent, especially if it includes tools, it is really useful to access details and metadata of execution steps. The bundle already tries to improve that situation with traceable decorators and an UI.

But I was wondering if we should introduce some kind of trace/protocol of the chain execution (incl. raw messages and potential tool details) out of the box.

MLFlow solves that quite impressive:
https://mlflow.org/docs/latest/tracing/

carsonbot:
Thank you for this issue.
There has not been a lot of activity here for a while. Has this been resolved?

carsonbot:
Friendly ping? Should this still be open? I will close if I don't hear anything.

carsonbot:
Hey,

I didn't hear anything so I'm going to close it. Feel free to comment if this is still relevant, I can always reopen!

PR #570, opened 2025-09-12
[AI Bundle] Trace and display `AgentInterface` calls in Profiler

Griffon-Weglot:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Docs?         | no 
| Issues        | #17 
| License       | MIT

sample2

### DIFF
--- a/demo/src/Mcp/Tools/CurrentTimeTool.php
+++ b/demo/src/Mcp/Tools/CurrentTimeTool.php
@@ -12,12 +12,18 @@
 namespace App\Mcp\Tools;
 
 use Mcp\Capability\Attribute\McpTool;
+use Psr\Log\LoggerInterface;
 
 /**
  * @author Tom Hart <tom.hart.221@gmail.com>
  */
 class CurrentTimeTool
 {
+    public function __construct(
+        private readonly LoggerInterface $logger,
+    ) {
+    }
+
     /**
      * Returns the current time in UTC.
      *
@@ -26,6 +32,8 @@ class CurrentTimeTool
     #[McpTool(name: 'current-time')]
     public function getCurrentTime(string $format = 'Y-m-d H:i:s'): string
     {
+        $this->logger->info('CurrentTimeTool called', ['format' => $format]);
+
         return (new \DateTime('now', new \DateTimeZone('UTC')))->format($format);
     }
 }
--- a/docs/bundles/mcp-bundle.rst
+++ b/docs/bundles/mcp-bundle.rst
@@ -113,6 +113,33 @@ Dynamic resources with parameters:
 
 All capabilities are automatically discovered in the ``src/`` directory when the server starts.
 
+Attribute Placement Patterns
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The MCP SDK, and therefore the MCP Bundle, supports two patterns for placing attributes on your capabilities:
+
+**Invokable Pattern** - Attribute on a class with ``__invoke()`` method::
+
+    #[McpTool(name: 'my-tool')]
+    class MyTool
+    {
+        public function __invoke(string $param): string
+        {
+            // Implementation
+        }
+    }
+
+**Method-Based Pattern** - Multiple attributes on individual methods::
+
+    class MyTools
+    {
+        #[McpTool(name: 'tool-one')]
+        public function toolOne(): string { }
+
+        #[McpTool(name: 'tool-two')]
+        public function toolTwo(): string { }
+    }
+
 Transport Types
 ...............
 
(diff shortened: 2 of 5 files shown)


OskarStark:
Can you please add a screenshot to the PR header?

src/ai-bundle/templates/data_collector.html.twig
```diff
@@ -12,6 +12,10 @@
 
         {% set text %}
             <div class="sf-toolbar-info-piece">
+                <div class="sf-toolbar-info-piece">
+                    <b class="label">Chat Calls</b>
```
OskarStark:
```suggestion
                    <b class="label">Agent Calls</b>
```
?

src/ai-bundle/config/services.php
```diff
@@ -127,6 +129,13 @@
             ])
             ->tag('kernel.event_listener')
 
+        ->set('ai.traceable_agent', TraceableAgent::class)
+        ->decorate(AgentInterface::class, priority: 5)
+        ->args([
+            service('.inner'),
+            service('ai.data_collector'),
+        ])
```
OskarStark:
```suggestion
            ->decorate(AgentInterface::class, priority: 5)
            ->args([
                service('.inner'),
                service('ai.data_collector'),
            ])
```

src/ai-bundle/config/services.php
```diff
@@ -127,6 +129,13 @@
...
+
         // profiler
```
OskarStark, line 167: shall we move this below this comment?

src/ai-bundle/src/Profiler/DataCollector.php
```diff
@@ -37,6 +37,11 @@ final class DataCollector extends AbstractDataCollector implements LateDataColle
      */
     private readonly array $toolboxes;
 
+    /**
+     * @var list<array{method: string, duration: float, input: mixed, result: mixed, error: ?\Throwable}>
+     */
+    private array $collectedChatCalls = [];
```
OskarStark:
```suggestion
    private array $collectedCalls = [];
```
?

src/ai-bundle/src/Profiler/DataCollector.php
```diff
@@ -52,7 +57,6 @@ public function __construct(
 
     public function collect(Request $request, Response $response, ?\Throwable $exception = null): void
     {
-        $this->lateCollect();
```
OskarStark, line 53: why this change?

src/ai-bundle/templates/data_collector.html.twig
```diff
@@ -235,11 +296,11 @@
                     </tr>
                     <tr>
                         <th>Arguments</th>
-                        <td>{{ dump(call.call.arguments) }}</td>
+                        <td><pre>{{ dump(call.call.arguments) }}</pre></td>
```
OskarStark: can we have the real dump, including the colors an being able to open close nodes inside the dump?

chr-hertel (changes requested):
Thanks for working on this @Griffon-Weglot!

One thing that would help is to move the input/output html conversion that is used on the platform calls to some macro, that we can use it on the agent and on the platform. so we wouldn't need to just dump the those.

src/ai-bundle/src/Profiler/DataCollector.php
Griffon-Weglot: It seemed redundant to call late collect since it will be called on `kernel.terminate` so i've removed it, also it's going againts the idea of implementing `LateDataCollectorInterface` to me, do you think it's relevant ?

Griffon-Weglot:
> One thing that would help is to move the input/output html conversion that is used on the platform calls to some macro, that we can use it on the agent and on the platform. so we wouldn't need to just dump the those.

@chr-hertel to simplify the display i've used `profiler_dump` since it's doing that very well

src/ai-bundle/src/Profiler/DataCollector.php
chr-hertel: It was introduced in #173 with the reason - does this change break the profiler on streaming?

src/ai-bundle/src/Profiler/DataCollector.php
```diff
@@ -94,28 +144,16 @@ public function getToolCalls(): array
...
-            $result = $call['result']->await();
-
-            if (isset($platform->resultCache[$result])) {
-                $call['result'] = $platform->resultCache[$result];
-            } else {
-                $call['result'] = $result->getContent();
-            }
-
-            $calls[$key] = $call;
-        }
```
chr-hertel, line 126: is that necessary? also see #173 for reasoning

chr-hertel:
please bring up the metrics again to the top:
image

Griffon-Weglot:
> please bring up the metrics again to the top

Sure ! Thanks for your feedbacks i'll make those changes asap

Griffon-Weglot:
I made a mistake with my fork by commiting in my main branch, this issue is closed for now but i'll try to redo it

PR #573, opened 2025-09-12
[AI Bundle][Agent][Platform] Rewrite output processors from Agent side used by Platform

### LINKED ISSUE #527: [Platform] Decouple Token Processors from Agent component
Currently the Platform component has dependencies to the Agent component - which is wrong from an architecture point of view, since the Platform component is the lower level component, and the dependency should be uni-directional from Agent to Platform.

Besides #519 there is also the dependency from out current token output solution in Platform to Agent because it is using the OutputProcessorInterface extension point.

Does the Platform component need a similar extension point?
Potentially should be aligned with implementation of #519

glady:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no 
| Issues        | Fix #527
| License       | MIT

- Introduced a separate output processor, that accepts a result handler and gives the result of the Output data container to this handler.
- removed dependency to symfony/ai-agent from composer file within platform subtree
- added deptrac check

### DIFF
--- a/deptrac.yaml
+++ b/deptrac.yaml
@@ -20,7 +20,6 @@ deptrac:
     Platform: ~
     Agent:
         - Platform
-        - orm
         - Store
     Store:
         - Platform


chr-hertel:
Thanks @glady for that proposal! The downside of introducing the new extension point `ResultHandlerInterface` like that is with standalone usage of the Platform component. So deptrac is great proof for having a proper decoupling, but imagine someone is not using the Agent component and wants to leverage the `TokenUsageResultHandler` as well - of course that's possible but only with their own glue code sticking the Platform's classes together.

An event-based solution didn't fly or was too much of hustle?

Edit:Maybe this could sit on top of #575?

glady:
@chr-hertel The event based solution had some more things to think about, I didn't want to introduce an additional dependency without thinking about more proper events. Not everyone is interested in using that, so he will need to write _some_ glue code (like we also need with the agent usage and the processors) in order to "activate" it, so he will at least need to add a subscriber that still would call the result handler (so this class or something similar would be needed anyway). Thats why I decided to build an result output processor entry point like its used in the examples (and in the bundle as well).

My opinion: to introduce an event dispatcher with proper events (for more events than just result) is another (bigger) scope I didn't want to touch yesterday.

chr-hertel:
> to introduce an event dispatcher with proper events (for more events than just result) is another (bigger) scope I didn't want to touch yesterday.

fair enough - and i agree that we could try to keep the event dispatcher dependency optional

other than that we could have input/output extension points instead of events, similar to the agent - to skip the event-dispatcher dependency - cc @OskarStark @RamyHakam any opinion on event dispatcher vs explicit lib-specific extension point?

OskarStark:
The explanation sounds valid so let's try the extension points 🙌

chr-hertel:
alright, just lost my mind, was close to proposing `OutputConverterInterface` and realized that we have a `ResultConverterInterface` already :joy:

So we have already an explicit extension point - but we can't directly use it with additional result converters for the tokens - now i wonder if we can extend the current ones without bringing even more concerns to the existing implementations.

glady:
@chr-hertel yes, I recognized this result converter interface too, but it is "first match" and just handles the model-based conversions (raw result -> result), which is something that should work out-of-the-box not like the optional token usage. Maybe the token usage can be an option to be put to those converters (as option given to Platform::invoke call). In this case my classes would be obsolete, indeed.

chr-hertel:
Thanks @glady for this proposal - we picked the deptrac setup as foundation and the idea of an extra interface - implemented with #1125.

Closing in favor of that - thanks again! :+1:

PR #574, opened 2025-09-12
[Platform] Add generic exceptions

junaidbinfarooq:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | Fix #528 
| License       | MIT

# Changes proposed:
The PR aims to add the following:
- Exceptions for 404 and 503
- A generic HTTP error handler

### DIFF
--- a/src/platform/src/Bridge/Anthropic/ResultConverter.php
+++ b/src/platform/src/Bridge/Anthropic/ResultConverter.php
@@ -12,7 +12,6 @@
 namespace Symfony\AI\Platform\Bridge\Anthropic;
 
 use Symfony\AI\Platform\Exception\HttpErrorHandler;
-use Symfony\AI\Platform\Exception\RateLimitExceededException;
 use Symfony\AI\Platform\Exception\RuntimeException;
 use Symfony\AI\Platform\Model;
 use Symfony\AI\Platform\Result\RawHttpResult;
@@ -41,12 +40,6 @@ class ResultConverter implements ResultConverterInterface
     public function convert(RawHttpResult|RawResultInterface $result, array $options = []): ResultInterface
     {
         $response = $result->getObject();
-
-        if (429 === $response->getStatusCode()) {
-            $retryAfter = $response->getHeaders(false)['retry-after'][0] ?? null;
-            $retryAfterValue = $retryAfter ? (float) $retryAfter : null;
-            throw new RateLimitExceededException($retryAfterValue);
-        }
 
         HttpErrorHandler::handleHttpError($response);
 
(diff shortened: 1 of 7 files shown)


src/platform/src/Bridge/Anthropic/ResultConverter.php
```diff
@@ -38,8 +39,11 @@ public function supports(Model $model): bool
 
     public function convert(RawHttpResult|RawResultInterface $result, array $options = []): ResultInterface
     {
+        $response = $result->getObject();
+        HttpErrorHandler::handleHttpError($response);
```
VincentLanglet, line 44: If you start to introduce a generic way to handle http error (which is a good thing), I think you should use it in all existing ResultConverter to prove it works for every one.

Currently you look for `$data['error']['message']` or `$data['details]`, but what if another API use another format ?
Should the `handleHttpError` accept a callback instead to extract the error message ?

Also, how you plan to deal 429, see the existing work from @floranpagliai in https://github.com/symfony/ai/pull/538, introducing a RateLimitExceededException ; it seems to be needed to compute the retryAfter value differently and therefor it's not easy to add this logic in a generic handler...
`

junaidbinfarooq: > If you start to introduce a generic way to handle http error (which is a good thing), I think you should use it in all existing ResultConverter to prove it works for every one.

Yep, that is what is supposed to be done but doing it in a single PR is a bit hectic and maybe an overkill. So, it could be done in another PR(s).

> Currently you look for $data['error']['message'] or $data['details], but what if another API use another format ?
Should the handleHttpError accept a callback instead to extract the error message ?

Right. Perhaps. I'd need to look into it first.

> Also, how you plan to deal 429, see the existing work from @floranpagliai in https://github.com/symfony/ai/pull/538, introducing a RateLimitExceededException ; it seems to be needed to compute the retryAfter value differently and therefor it's not easy to add this logic in a generic handler...

This is something that needs thought and perhaps more opinion(s) especially from the maintainers.

floranpagliai: > This is something that needs thought and perhaps more opinion(s) especially from the maintainers.

@chr-hertel suggested to handle generic error inside `Symfony\AI\Platform\Result\RawHttpResult::getData`

https://github.com/symfony/ai/pull/561#issuecomment-3288496099

My point of view, which aligns with @VincentLanglet , is that to build a generic handler, it needs to be able to account for the specificities of each bridge, whether through callbacks, inheritance, or other mechanisms.

junaidbinfarooq:
@chr-hertel @OskarStark 
Please review this and let me know what you think.

junaidbinfarooq:
@chr-hertel @OskarStark
Could you have a look?

chr-hertel:
@junaidbinfarooq I'm currently not really convinced that it should be the role of result converters to handle this stuff. see #646 #136 and #167

please don't spend more time here - not closing for now, since we might come back to this approach.

PR #575, opened 2025-09-12
[Platform] Add event system for string input conversion

### LINKED ISSUE #326: [Platform] Fatal Error: array_merge() called with string payload in ModelClient implementations

RamyHakam:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| Docs?         | no
| Issues        | Fix #326
| License       | MIT

## Description

Implements event-based input processing to resolve fatal errors when string payloads are passed to ModelClient implementations using `array_merge($payload, $options)`.

**Solution:**
- `PlatformInvokationEvent`: Dispatched before platform invocation
- `StringToMessageBagListener`: Converts strings to MessageBag for models with INPUT_MESSAGES capability
- Enables simple usage: `$platform->invoke($model, 'Hello')`

**Backward Compatible:** EventDispatcher is optional, existing code unchanged.

### DIFF
--- a/src/platform/composer.json
+++ b/src/platform/composer.json
@@ -48,6 +48,7 @@
         "phpstan/phpdoc-parser": "^2.1",
         "psr/log": "^3.0",
         "symfony/clock": "^7.3|^8.0",
+        "symfony/event-dispatcher": "^7.3|^8.0",
         "symfony/http-client": "^7.3|^8.0",
         "symfony/property-access": "^7.3|^8.0",
         "symfony/property-info": "^7.3|^8.0",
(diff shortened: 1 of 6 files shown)


chr-hertel:
It took time, but here we go, this is in now. Thank you very much @RamyHakam.

PR #583, opened 2025-09-13
[AI Bundle] Add memory provider configuration support

OskarStark:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | Follows #117
| License       | MIT

This PR introduces comprehensive memory provider configuration capabilities for AI agents, enabling both simple static memory and advanced dynamic memory scenarios.

### Smart Detection Logic
The system automatically detects the intent:
- If the value matches an existing service/alias → uses that service directly  
- If no matching service exists → creates `StaticMemoryProvider` with the string as static content

### Memory as System Prompt
- When only `memory` is provided (no `prompt`), memory serves as the system prompt
- When both `memory` and `prompt` are provided, memory is prepended to the prompt

### Configuration Examples

**Static Memory (Most Common):**
```yaml
ai:
  agent:
    my_agent:
      model:
      memory: 'You have access to user preferences and conversation history'
      prompt:
        text: 'You are a helpful assistant.'
```

**Dynamic Memory (Advanced):**
```yaml
ai:
  agent:
    my_agent:
      model:
      memory: 'my_custom_memory_service'  # References existing service
      prompt:
        text: 'You are a helpful assistant.'
```

**Memory as System Prompt:**
```yaml
ai:
  agent:
    my_agent:
      model:
      memory: 'You are a helpful assistant with access to user context.'
      # No prompt needed - memory serves as system prompt
```

This enhancement significantly improves the developer experience while maintaining full backward compatibility and adding powerful new memory capabilities for AI agents.

cc @DZunke


chr-hertel:
hmm, i know that it is injected in the prompt, but still i feel it should be standalone config, not below the prompt

OskarStark:
> hmm, i know that it is injected in the prompt, but still i feel it should be standalone config, not below the prompt

I thought the same, but it can only work with the SystemInputProcessor and this is only registered, once a prompt is defined 🤷‍♂️ Imagine, you don't add a system prompt but memory, then the memory will be your system prompt, which is weird to me.

chr-hertel:
> but it can only work with the SystemInputProcessor

i think it works on every MessageBag with system prompt - no matter if pushed by the user or the SystemPromptProcessor:
https://github.com/symfony/ai/blob/d8dfb30470551224f7004411c8d32b6a9e600622/src/agent/src/Memory/MemoryInputProcessor.php#L72-L83

OskarStark:
I updated my comment, it's about the result

DZunke:
> hmm, i know that it is injected in the prompt, but still i feel it should be standalone config, not below the prompt

Yeah, feels a bit weird but i was inspired by other libraries doing the same. The memory was always injected to the prompt message type as the memory do not have an own message type. I had shortly also seen some internal OpenAI / Antrophic internal prompt collections where there memory was also injected into the prompt. 

Another approach i have seen was to have `bio` or `memory` tools to give the LLM more possibilitied to manage the memory. That approach was also combined with the "static" memory approach. 

I, for example, have replaced a static configured fantasy calendar that was a "forced" tooling for each LLM call with this static memory. Also some other personal information are injected here from different sources. I understand the tooling approach more as a dynamic memory to let the LLM decide information from conversation to be stored into a memory that could then be injected again. 

An interesting overview about how complex prompts can be filled, for me, was https://github.com/asgeirtj/system_prompts_leaks/tree/main - There is a lot going on in the system prompts that is going beyond just the "System Prompt".

OskarStark:
In my case I have a demo where I have a system prompt and adding all activities and apartments of the client. Roughly 15 apartments an 19 activities. I would love to be able to split the "data" front he prompt, so this feels handy.

chr-hertel:
No, don't get me wrong, i'm not questioning that this belongs in the system prompt when looking at the actual payload. I'm questioning if from the configuration point of view, it should go there.
tbh, i would hide the entire prompt thing going forward, making it more like:

```yaml
my_agent:
    platform: 'service_id'
    model: 'model_name'
    instruction:
        text: '...'
        ...
    memory: '...'
    tools: [...]
```

I think the memory is a feature of the agent, not a feature of the prompt - even tho from an implementation point of view, it is implemented via the system prompt.

OskarStark:
Fine, but validating, that you can only use memory once a prompt is set, agree?

chr-hertel:
> Fine, but validating, that you can only use memory once a prompt is set, agree?

The memory itself has a prompt - I think even if no system prompt is around, we should add this - and this is actually what the `MemoryInputProcessor` does, if i get it right

PR #585, opened 2025-09-14
DRAFT Added support for polymporphic* types by using DiscriminatorMap

### LINKED ISSUE #559: Union type support for structured outputs
Unfortunately `\Symfony\AI\Platform\Contract\JsonSchema\Factory::getTypeSchema` doesn't support for `\Symfony\Component\TypeInfo\Type\UnionType`.

What I'm trying to achieve:

```php
<?php

declare(strict_types=1);

namespace App\Agent\Application\StructuredOutput;

use Webmozart\Assert\Assert;

final class ChatStructuredOutput
{
    public function __construct(
        public string $title,
        public string $finalAnswer,
        /** @var array<int, ChartPart | TablePart | TextPart> */
        public array $parts = [],
    ) {
        Assert::allIsInstanceOf($this->parts, StructuredOutputPart::class);;
    }
}
```

It's quite possible I'm using the structured output wrong, and I should deal with the problem differently. I'm all ears how could I make it work.

HaKIMus:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Docs?         | yes/no 
| Issues        | Fix #559  
| License       | MIT

It's a draft, because, primarily, I want to ask if the DiscriminatorMap approach is ok with you.

**Why the discriminator map usage?**
I've achieved the native Foo|Bar union type support for agent json schema output, but Symfony Serializer wasn't working with `list<TableDataVisualizationDto|ChartDataVisualizationDto>` correctly, outputting an array in `\Symfony\Component\Serializer\SerializerInterface::deserialize::112`. If you know how to make the serializer work with the "nested" union types correctly, then let me know and we might achieve a native support without the DiscriminatorMap usage.

**Example**
This is an example from my project I want the feature to work on.

Model I tried it with: `gpt-5-nano` with the `Symfony\AI\Platform\Bridge\OpenAi\Gpt`model class.

Prompt:
```
{
    "message": "Provide me with a table representing my latest orders and a chart representing amount of orders for last 10 months."
}
```

Schema produced:
```
 array:2 [
  "type" => "json_schema"
  "json_schema" => array:3 [
    "name" => "ChatStructuredOutput"
    "schema" => array:4 [
      "type" => "object"
      "properties" => array:4 [
        "title" => array:1 [
          "type" => "string"
        ]
        "finalAnswer" => array:1 [
          "type" => "string"
        ]
        "parts" => array:2 [
          "type" => "array"
          "items" => array:1 [
            "anyOf" => array:3 [
              0 => array:4 [
                "type" => "object"
                "properties" => array:6 [
                  "title" => array:1 [
                    "type" => "string"
                  ]
                  "chartType" => array:2 [
                    "type" => "string"
                    "enum" => array:3 [
                      0 => "bar"
                      1 => "line"
                      2 => "pie"
                    ]
                  ]
                  "labels" => array:2 [
                    "type" => "array"
                    "items" => array:1 [
                      "type" => "string"
                    ]
                  ]
                  "datasets" => array:2 [
                    "type" => "array"
                    "items" => array:4 [
                      "type" => "object"
                      "properties" => array:3 [
                        "label" => array:1 [
                          "type" => "string"
                        ]
                        "data" => array:2 [
                          "type" => "array"
                          "items" => array:1 [
                            "type" => "number"
                          ]
                        ]
                        "bgColor" => array:2 [
                          "type" => "string"
                          "description" => "Optional param, default to empty string."
                        ]
                      ]
                      "required" => array:3 [
                        0 => "label"
                        1 => "data"
                        2 => "bgColor"
                      ]
                      "additionalProperties" => false
                    ]
                  ]
                  "options" => array:2 [
                    "type" => "array"
                    "items" => array:1 [
                      "type" => "string"
                    ]
                  ]
                  "type" => array:2 [
                    "type" => "string"
                    "pattern" => "^chart$"
                  ]
                ]
                "required" => array:6 [
                  0 => "title"
                  1 => "chartType"
                  2 => "labels"
                  3 => "datasets"
                  4 => "options"
                  5 => "type"
                ]
                "additionalProperties" => false
              ]
              1 => array:4 [
                "type" => "object"
                "properties" => array:4 [
                  "columns" => array:2 [
                    "type" => "array"
                    "items" => array:1 [
                      "type" => "string"
                    ]
                  ]
                  "rows" => array:2 [
                    "type" => "array"
                    "items" => array:2 [
                      "type" => "array"
                      "items" => array:1 [
                        "type" => "string"
                      ]
                    ]
                  ]
                  "meta" => array:2 [
                    "type" => "array"
                    "items" => array:1 [
                      "type" => "string"
                    ]
                  ]
                  "type" => array:2 [
                    "type" => "string"
                    "pattern" => "^table$"
                  ]
                ]
                "required" => array:4 [
                  0 => "columns"
                  1 => "rows"
                  2 => "meta"
                  3 => "type"
                ]
                "additionalProperties" => false
              ]
              2 => array:4 [
                "type" => "object"
                "properties" => array:2 [
                  "text" => array:1 [
                    "type" => "string"
                  ]
                  "type" => array:2 [
                    "type" => "string"
                    "pattern" => "^text$"
                  ]
                ]
                "required" => array:2 [
                  0 => "text"
                  1 => "type"
                ]
                "additionalProperties" => false
              ]
            ]
          ]
        ]
        "reasoningStep" => array:2 [
          "type" => "array"
          "items" => array:4 [
            "type" => "object"
            "properties" => array:1 [
              "reasoning" => array:1 [
                "type" => "string"
              ]
            ]
            "required" => array:1 [
              0 => "reasoning"
            ]
            "additionalProperties" => false
          ]
        ]
      ]
      "required" => array:4 [
        0 => "title"
        1 => "finalAnswer"
        2 => "parts"
        3 => "reasoningStep"
      ]
      "additionalProperties" => false
    ]
    "strict" => true
  ]
]
```

The description is long enough, so I'd skip the PHP DTOs structure. They look almost identical to the ones I used in tests in the PR.

Output (the data are fake):
```
 App\Agent\Application\StructuredOutput\ChatStructuredOutput {#2041
  +title: "Latest orders and orders by month (last 10 months)"
  +finalAnswer: "Here is a table with your latest orders and a bar chart showing the number of orders per month for the last 10 months."
  +parts: array:3 [
    0 => 
App\Agent\Application\StructuredOutput
\
TablePart
 {#1596
      +type: "table"
      +columns: array:5 [
        0 => "Order #"
        1 => "Date Created"
        2 => "Status"
        3 => "Total PLN"
        4 => "Customer Email"
      ]
      +rows: array:10 [
        0 => array:5 [
          0 => "115"
          1 => "2025-09-14 11:46:11"
          2 => "failed"
          3 => "24902.21"
          4 => "N/A"
        ]
        1 => array:5 [
          0 => "117"
          1 => "2025-09-14 11:46:11"
          2 => "completed"
          3 => "28029.39"
          4 => "ellis.freya@example.com"
        ]
        2 => array:5 [
          0 => "124"
          1 => "2025-09-14 11:46:11"
          2 => "completed"
          3 => "28424.05"
          4 => "mckenzie.britney@example.com"
        ]
        3 => array:5 [
          0 => "122"
          1 => "2025-09-14 11:46:11"
          2 => "completed"
          3 => "8692.10"
          4 => "maxie.wiegand@wiegand.com"
        ]
        4 => array:5 [
          0 => "121"
          1 => "2025-09-14 11:46:11"
          2 => "completed"
          3 => "10583.07"
          4 => "frami.fabiola@example.com"
        ]
        5 => array:5 [
          0 => "120"
          1 => "2025-09-14 11:46:11"
          2 => "completed"
          3 => "14509.36"
          4 => "user@example.com"
        ]
        6 => array:5 [
          0 => "119"
          1 => "2025-09-14 11:46:11"
          2 => "completed"
          3 => "36850.59"
          4 => "user@example.com"
        ]
        7 => array:5 [
          0 => "116"
          1 => "2025-09-14 11:46:11"
          2 => "completed"
          3 => "19252.16"
          4 => "ellis.freya@example.com"
        ]
        8 => array:5 [
          0 => "118"
          1 => "2025-09-14 11:46:11"
          2 => "on-hold"
          3 => "33717.00"
          4 => "N/A"
        ]
        9 => array:5 [
          0 => "123"
          1 => "2025-09-14 11:46:11"
          2 => "completed"
          3 => "33392.61"
          4 => "bechtelar.davonte@example.com"
        ]
      ]
      +meta: []
    }
    1 => 
App\Agent\Application\StructuredOutput
\
ChartPart
 {#2100
      +type: "chart"
      +title: "Orders in the last 10 months"
      +chartType: 
App\Agent\Application\StructuredOutput
\
ChartPartType
 {#1107
        +name: "BAR"
        +value: "bar"
      }
      +labels: array:10 [
        0 => "Dec 2024"
        1 => "Jan 2025"
        2 => "Feb 2025"
        3 => "Mar 2025"
        4 => "Apr 2025"
        5 => "May 2025"
        6 => "Jun 2025"
        7 => "Jul 2025"
        8 => "Aug 2025"
        9 => "Sep 2025"
      ]
      +datasets: array:1 [
        0 => 
App\Agent\Application\StructuredOutput
\
DatasetPart
 {#2079
          +label: "Number of orders"
          +data: array:10 [
            0 => 0
            1 => 0
            2 => 0
            3 => 0
            4 => 0
            5 => 0
            6 => 0
            7 => 0
            8 => 0
            9 => 85
          ]
          +bgColor: "#4e79a7"
        }
      ]
      +options: array:2 [
        0 => "responsive"
        1 => "scales: {y: {beginAtZero: true}}"
      ]
    }
    2 => 
App\Agent\Application\StructuredOutput
\
TextPart
 {#2120
      +type: "text"
      +text: "Note: The chart shows order counts per month for the last 10 months. In this dataset, only Sep 2025 contains orders (85 orders)."
    }
  ]
  +reasoningStep: array:2 [
    0 => 
App\Agent\Application\StructuredOutput
\
ReasoningStep
 {#2069
      +reasoning: "I retrieved the latest orders using the WooCommerce orders endpoint and prepared a compact table with key fields: Order #, Date Created, Status, Total PLN, and Customer Email. Since the dataset appears to be current (all entries dated 2025-09-14), I listed the ten most recent orders as they appear from the data source."
    }
    1 => 
App\Agent\Application\StructuredOutput
\
ReasoningStep
 {#2148
      +reasoning: "For the chart, I computed a 10-month window ending in Sep 2025. Based on the provided data, all orders fall in Sep 2025, so Sep 2025 has 85 orders and the preceding months have 0 orders. This yields a bar chart with a single non-zero bar for Sep 2025. If you’d like a different window (e.g., moving 10 months from a different start date) I can adjust the chart accordingly."
    }
  ]
}
```
As you can see all data have been correctly mapped from schema to php objects.


src/platform/src/Contract/JsonSchema/Factory.php
```diff
@@ -47,6 +49,7 @@
  *         minProperties?: int,
  *         maxProperties?: int,
  *         dependentRequired?: bool,
+ *         anyOf?: list<mixed>,
```
OskarStark, line 52: This and oneOf seems not to be supported by OpenAI, can you confirm?

Which LLM do you use?

HaKIMus: > This and oneOf seems not to be supported by OpenAI, can you confirm?

While oneOf wasn't supported, and threw an error, anyOf works perfectly well.

> Which LLM do you use?

gpt-5-nano.

image

OskarStark: Good to know, did you saw the other PR using by union types?

HaKIMus: As a matter of fact, I didn't. I started the PR 'cause I thought no one is working on it. I can't find the PR, can I ask you for a link to it?

There is a chance you're referring to the PR description. In such case I have the code in the local history of my phpstorm. If that's what you meant, I can send the factory code here that was working with union types but was failing during the AgentProcessor due to deserialization issues.

OskarStark: Sorry it was your issue I associated with a PR because of the code examples 🤓

OskarStark:
Maybe I miss something, but why do we need the `DiscriminatorMap`?

```php
public function __invoke(
    Foo|Bar $myVar,
) {}
```

Doesn't this contain enough information to resolve it?

HaKIMus:
> Maybe I miss something, but why do we need the `DiscriminatorMap`?
> 
> ```
> public function __invoke(
>     Foo|Bar $myVar,
> ) {}
> ```
> 
> Doesn't this contain enough information to resolve it?

It might be, but [the problematic case](https://github.com/symfony/ai/pull/585/files#diff-f1025dae4c475c116140ae00e19644b7014145191621c2629839f49dd3df0e8fR17) is a `List<Foo|Bar>`. The discriminator map is here primarily for the Agent Processor to know how to deserialize data produced by llm according to our json schema correctly.

If Symfony Serializer would understand how to work with annotated `list<Foo|Bar>` then we could get rid of the DiscriminatorMap. Because it lacks the strategy of deserializing a list of union types we'd get during the deserialization an array structurally reflecting our desired DTO, and not the DTO itself.

OskarStark:
I have one question, can we simplify the fixture classes to a bare minimum and use simple terminology? I now this is from your real world use case, but for tests, maintenance too much complexity for me. if we need the enum inside the union type, fine, if not, lets remove it or lets only keep 2 values etc.

Thanks

chr-hertel (changes requested):
Implementation looks good to me as well, agree on the complexity issue with the fixture, and would love to see an example here - basically as to proof also for its functionality

HaKIMus:
@OskarStark I'll simplify and try to make the new DTOs examples more comprehensible max 'till Tomorrow.

@chr-hertel when you ask for examples - you mean like the ones in PR description or something like `examples/openai/structured-output-math.php:24`? I think the second one could be more useful long term.

OskarStark:
lets add `examples/openai/structured-output-union-types.php`

fixtures/StructuredOutput/PolymorphicType/ListItemDiscriminator.php
```diff
@@ -0,0 +1,33 @@
...
+
+namespace Symfony\AI\Fixtures\StructuredOutput\PolymorphicType;
+
+use Symfony\Component\Serializer\Attribute\DiscriminatorMap;
+
+#[DiscriminatorMap(
+    typeProperty: 'type',
+    mapping: [
+        'foo' => ListItemName::class,
+        'bar' => ListItemAge::class,
```
OskarStark:
```suggestion
        'name' => ListItemName::class,
        'age' => ListItemAge::class,
```
?

fixtures/StructuredOutput/PolymorphicType/ListItemName.php
```diff
@@ -0,0 +1,14 @@
...
+
+namespace Symfony\AI\Fixtures\StructuredOutput\PolymorphicType;
+
+use Symfony\AI\Platform\Contract\JsonSchema\Attribute\With;
+
+class ListItemName implements ListItemDiscriminator
+{
+    public function __construct(
+        public string $name,
+        #[With(pattern: '^name$')] public string $type = 'name',
```
OskarStark:
```suggestion
        #[With(pattern: '^name$')]
        public string $type = 'name',
```

fixtures/StructuredOutput/PolymorphicType/ListItemAge.php
```diff
@@ -0,0 +1,14 @@
...
+
+namespace Symfony\AI\Fixtures\StructuredOutput\PolymorphicType;
+
+use Symfony\AI\Platform\Contract\JsonSchema\Attribute\With;
+
+final class ListItemAge implements ListItemDiscriminator
+{
+    public function __construct(
+        public int $age,
+        #[With(pattern: '^age$')] public string $type = 'age',
```
OskarStark:
```suggestion
        #[With(pattern: '^age$')]
        public string $type = 'age',
```

HaKIMus:
I've simplified the examples to simpler objects and organized them in separated namespaces.

I'm not sure why the pipeline fails with:
```
1) Symfony\AI\Platform\Tests\Contract\JsonSchema\FactoryTest::testBuildPropertiesForListOfPolymorphicTypesDto
Error: Call to undefined method Symfony\Component\Serializer\Attribute\DiscriminatorMap::getMapping()
```
Locally the test runs correctly, and I'm on version 8.4.
```
php -v
PHP 8.4.12 (cli) (built: Aug 26 2025 13:36:28) (NTS)
```
None of us wants to introduce a ghost error the project, it'd be good to know why it has happened, and I, at the moment, have no clue.

----
**UPDATE**
I know why the error is occurring. The pipeline installs the version 8 for symfony serializer, where the DiscriminatorMap doesn't include the method. I'll think of a workaround. 
https://github.com/symfony/serializer/blob/8.0/Attribute/DiscriminatorMap.php

HaKIMus:
@OskarStark @chr-hertel in the release 8.0 of the symfony/serializer package DiscriminatorMap doesn't implement `getMapping()` in favor of public properties. How do you handle such cases in the project? The pipeline builds against 7.* and 8.* version of the serializer.

I'm thinking of the following solutions:
* Reflectively change the property access
* Detect the PHP version and access the property based on it (phpstan will not be happy about calling a private property)
* Do not build against 8.* of symfony serializer
* Asking nicely to Symfony/Serializer maintainers to support for both getters and access properties

Wdyt?

7.3: https://github.com/symfony/serializer/blob/7.3/Attribute/DiscriminatorMap.php
8.0: https://github.com/symfony/serializer/blob/8.0/Attribute/DiscriminatorMap.php

EDIT
I've chosen the reflective approach.

src/agent/CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ CHANGELOG
 0.1
 ---
 
+ * Add support for polymorphic types via DiscriminatorMap
+ * Add support for union types
```
```suggestion
 * Add support for union types and polymorphic types via DiscriminatorMap
```

OskarStark:
One minor and good to merge

OskarStark:

chr-hertel:
@HaKIMus does the example `openai/structured-output-union-types.php` work for you?

HaKIMus:
I used the `structured-output-math.php` example as a template, and it fails to me with the same issue as my example about not existing constant: `Fatal error: Uncaught Error: Undefined constant Symfony\AI\Platform\Bridge\OpenAi\Gpt::GPT_4O_MINI`.

When I set the name explicitly as 'gpt-5-nano', I started to get errors about unknown named parameter logger. To me it looks like an error with autoloading or my local set up (I didn't run docker compose to save resources on my machine), as the constant name and the logger param definitely exist.

I must say I never run the example, I've assumed the "compiling" correctness of the `structured-output-math.php` file. I never run it, because I was using the feature - even during the development - in my project all the time, so at the very least I can say the functionality works.

chr-hertel:
found this issue on `main` and even on the branch of this PR it doesn't work for me:
image

HaKIMus:
From what I've deduct - we're not setting nullable properties in the required list, but all structure keys must be present in it.

Because the $time property is accepting a null the field is not required. If we remove the null as an option - it's present in the required list.

The behavior is strange, however, because examples like `\Symfony\AI\Fixtures\StructuredOutput\ExampleDto` should fail when prompted as well. Furthermore `\Symfony\AI\Fixtures\StructuredOutput\User` because of its nulalble age property wouldn't pass as well.

I've actually tested the cases myself, and that's true.

------

To summarize

Nullable fields must be present in the required list of their parent output structure, otherwise platforms (like openai) will reject the call.

------

Fix

The fix could be as follow:

```php
            if ($type->isNullable()) {
                // anyOf already contains the null variant when applicable; do nothing
                if (!isset($schema['anyOf'])) {
                    $schema['type'] = [$schema['type'], 'null'];
                }
            }

            if (!($element instanceof \ReflectionParameter && $element->isOptional())) {
                $result['required'][] = $name;
            }
```
image


We've assumed the nullable fields aren't required, and the fix separates the logic of setting an element as required from the fact if it's null or not.

Before the fix:
image

After the fix:
image

Namespaces are different, because it was easier for me to run the example in my project.

HaKIMus:
@chr-hertel  I was about to open a PR, but I see you've this one open: https://github.com/symfony/ai/pull/802. Do you want me to implement the feature of treating nullables as required or you're going to handle it?

chr-hertel:
@HaKIMus no no, feel free to go ahead - didn't address that issue and didn't change the schema part

HaKIMus:
Additionally, I'd like to clarify the discordance, I wrote:

> I never run it, because I was using the feature - even during the development - in my project all the time, so at the very least I can say the functionality works.

I've just checked, and none of my properties of the 7 possible structured outputs are nullable, and that's why - aside from not running the example - I've never stumbled upon the issue.

I apologize for my mistake.

aszenz:
@HaKIMus This only works for structure outputs right, not tool result normalizers?

I'm trying to make union types work for tool calls

HaKIMus:
I can't remember. I can say, however, I wasn't making the solution with tool calls in mind.

If you had the time to test it yourself, and your results are negative, it's very likely that I didn't cover this case.

aszenz:
> I can't remember. I can say, however, I wasn't making the solution with tool calls in mind.
> 
> If you had the time to test it yourself, and your results are negative, it's very likely that I didn't cover this case.

Thanks, json schema changes work, I created an issue with denormalization for discriminated unions.

PR #543, opened 2025-09-12
[Platform][OpenAI] Add AuthenticationException for better error handling

vitalii-kyktov:
| Q             | A
| ------------- | ---
| Branch?       | main
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #528
| License       | MIT

When OpenAI API returns 401 (authentication error), users now get a clear AuthenticationException with the actual API error message instead of the confusing "Response does not contain choices" error.

Changes:
- Add AuthenticationException class for API authentication failures
- Update OpenAI ResultConverter to check HTTP 401 status before processing
- Add test coverage for authentication error scenarios

PR #586, opened 2025-09-15
[platform] Add status code 429 to rate limit exceeded exception

junaidbinfarooq:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #528 Builds on https://github.com/symfony/ai/pull/538
| License       | MIT

### DIFF
--- a/src/platform/src/Exception/RateLimitExceededException.php
+++ b/src/platform/src/Exception/RateLimitExceededException.php
@@ -19,7 +19,7 @@ final class RateLimitExceededException extends RuntimeException
     public function __construct(
         private readonly ?float $retryAfter = null,
     ) {
-        parent::__construct('Rate limit exceeded.');
+        parent::__construct('Rate limit exceeded.', 429);
     }
 
     public function getRetryAfter(): ?float
--- a/src/platform/tests/Bridge/Anthropic/ResultConverterRateLimitTest.php
+++ b/src/platform/tests/Bridge/Anthropic/ResultConverterRateLimitTest.php
@@ -39,6 +39,7 @@ final class ResultConverterRateLimitTest extends TestCase
         $handler = new ResultConverter();
 
         $this->expectException(RateLimitExceededException::class);
+        $this->expectExceptionCode(429);
         $this->expectExceptionMessage('Rate limit exceeded');
 
         try {
@@ -61,6 +62,7 @@ final class ResultConverterRateLimitTest extends TestCase
         $handler = new ResultConverter();
 
         $this->expectException(RateLimitExceededException::class);
+        $this->expectExceptionCode(429);
         $this->expectExceptionMessage('Rate limit exceeded');
 
         try {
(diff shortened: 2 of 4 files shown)


OskarStark:
Is it really common to use the status code as exception code?

junaidbinfarooq:
> Is it really common to use the status code as exception code?

Generally speaking, it does make sense to have the code present, and application-level code can be interested in the exception code, which in this case would map to the 429 status code.

VincentLanglet:
> Is it really common to use the status code as exception code?

Definitely no.
Even the HttpException from symfony are using a different value for the status code and the code
https://github.com/symfony/symfony/blob/7.4/src/Symfony/Component/HttpKernel/Exception/HttpException.php

This PR start to creating a confusion between both IMHO. 

> application-level code can be interested in the exception code, which in this case would map to the 429 status code.

You can manually do the mapping
```
match ($exception::class) {
     RateLimitExceededException::class => 429,
     // ...
}
```

junaidbinfarooq:
> > Is it really common to use the status code as exception code?
> 
> Definitely no. Even the HttpException from symfony are using a different value for the status code and the code https://github.com/symfony/symfony/blob/7.4/src/Symfony/Component/HttpKernel/Exception/HttpException.php
> 
> This PR start to creating a confusion between both IMHO.
> 
> > application-level code can be interested in the exception code, which in this case would map to the 429 status code.
> 
> You can manually do the mapping
> 
> ```
> match ($exception::class) {
>      RateLimitExceededException::class => 429,
>      // ...
> }
> ```

There are libraries where the two are kept separate, and likely some where the code is directly mapped to the status code, especially in the context of HTTP, but PHP doesn't care how you decide to use the exception code property, AFAIK.

> You can manually do the mapping

Why? You shared the example of Symfony's HttpException class. Does it not have a status code field so that the application users get the status code OOTB?

Perhaps, we can have a _status code_ field here too.
WDYT @chr-hertel @OskarStark ?

OskarStark:
a statusCode field makes much more sense to me

VincentLanglet:
> a statusCode field makes much more sense to me

Since we have mutiple exception, like https://github.com/symfony/ai/pull/543
This would certainly mean introducing some HttpException base exception with a status code field.

I dunno @chr-hertel @OskarStark if we should
- Try to reuse the HttpException from Symfony
- Or recreating them in the symfony/ai context like we're starting to do

junaidbinfarooq:
> a statusCode field makes much more sense to me

@chr-hertel @OskarStark 
Should we, in this case, create something like an `HttpException` base class with the status code field that extends from the platform's `RuntimeException`?

junaidbinfarooq:
@chr-hertel @OskarStark
Could you have a look?

OskarStark:
a statusCode would work

OskarStark:
Closing here for now, thanks for proposing

VincentLanglet:
Maybe you could look at the attribute `WithHttpStatus` ?

PR #595, opened 2025-09-16
[AI Bundle] Simplify memory configuration structure

### LINKED ISSUE #591: [Ai Bundle][Agent] Memory as service does not work
I registered a custom service as memory provider like this:

```yaml
ai:
    platform:
        openai:
    agent:
        default:
            model:
                name: 'gpt-4o-mini'
            prompt: 'You are a pirate and you write funny.'
            memory: 'App\MyMemory'
```

And the implementation looks like this:
```php
<?php

declare(strict_types=1);

namespace App;

use Symfony\AI\Agent\Input;
use Symfony\AI\Agent\Memory\Memory;
use Symfony\AI\Agent\Memory\MemoryProviderInterface;

final readonly class DevDayMemory implements MemoryProviderInterface
{
    public function __construct(
        private string $myMemory,
    ) {
    }

    public function load(Input $input): array
    {
        return [
            new Memory(file_get_contents($this->myMemory)),
        ];
    }
}
```

But this is always false for `App\MyMemory`:
https://github.com/symfony/ai/blob/f69152356dce1b92289405da401c16702a723c4f/src/ai-bundle/src/AiBundle.php#L628

OskarStark:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | Fix #591 
| License       | MIT

Remove guessing logic for memory keys and introduce explicit syntax:
```yaml
memory: 'text' # always creates static memory processor
```

```yaml
memory:
    service: 'service_name' # uses service reference
```

This makes the configuration more predictable and eliminates the need to check if a string refers to an existing service.

cc @DZunke

### DIFF
--- a/src/ai-bundle/config/options.php
+++ b/src/ai-bundle/config/options.php
@@ -138,12 +138,26 @@ return static function (DefinitionConfigurator $configurator): void {
                             ->end()
                         ->end()
                         ->booleanNode('structured_output')->defaultTrue()->end()
-                        ->scalarNode('memory')
-                            ->info('Plain string or service name of the memory provider implementing MemoryProviderInterface')
+                        ->variableNode('memory')
+                            ->info('Memory configuration: string for static memory, or array with "service" key for service reference')
                             ->defaultNull()
                             ->validate()
-                                ->ifTrue(function ($v) { return \is_string($v) && '' === $v; })
+                                ->ifTrue(function ($v) {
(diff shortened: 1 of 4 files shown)


chr-hertel:
Thanks @OskarStark for the quick solution - will directly use that in my Dresden workshop 😎

PR #600, opened 2025-09-16
[MCP SDK] Handle empty message queue in CachePoolStore::pop()

MartkCz:
Previously, pop() would call array_shift() even if the cached queue was empty,
leading to unnecessary operations. This change ensures that pop() returns null
early when no messages are available.


chr-hertel:
That would also go to the new, official sdk: https://github.com/modelcontextprotocol/php-sdk

Thanks for the PR tho - hope to see you on the other repo!

Issue #401, opened 2025-09-02
Allow runtime model capability declaration and options via DSN syntax


OskarStark:
## Problem

When AI models gain new capabilities that are already supported by our platform, users cannot use them without a library release. Example: Platform supports `tool-calling`, model initially doesn't. When model gains `tool-calling` support, users must wait for a library update to use it.
## Proposed Solution
Allow users to declare model capabilities at runtime using DSN syntax:
```
model-name@capability1,capability2,capability3
```
As a bonus we could get rid of all the model classes.

### Scenario

Platform supports: `input-image`, `tool-calling`, `output-streaming`  
Model GPT-4 initially supports: `input-image`  
Model GPT-4 later adds: `tool-calling` (no library release needed!)
```php
// Before: Limited to hardcoded capabilities in library
$model = 'gpt-4';  // Only uses capabilities defined in library code

// After: User can declare newly supported capabilities
$model = 'gpt-4@tool-calling';  // Uses tool-calling without library update
```

### Usage Examples

```php
// Enable single capability
'gpt-4@tool-calling'
// Enable multiple capabilities  
'claude-3@tool-calling,output-streaming'
// Enable multiple capabilities
'gpt-4o@input-image,tool-calling,output-streaming'
// No additional capabilities (default behavior)
'gpt-4'
```

### Available Capabilities (from `Capability` enum)

- `input-audio`, `input-image`, `input-messages`, `input-multiple`, `input-pdf`, `input-text`
- `output-audio`, `output-image`, `output-streaming`, `output-structured`, `output-text`
- `tool-calling`
- `text-to-speech`, `speech-to-text`
## Benefits
- **Zero-day capability adoption**: Use new model capabilities immediately when API supports them
- **No library releases needed** for capability updates
- **Forward compatible**: Platform already knows the capability, just needs model declaration
- **Backward compatible**: Models without `@` work as before

### Bundle

In the bundle it could look like:

```diff
ai:
    agent:
        my_agent:
            platform: 'ai.platform.openai'
-            model:
-                class: 'Symfony\AI\Platform\Bridge\OpenAi\Gpt'
-                name: 'gpt-4.1-mini'
+            model: 'gpt-4.1-mini@tool-calling'
```

## Note

- API validates if model supports requested capabilities and returns meaningful errors if not supported.
- You cannot forbid certain capabilities or limit the model to some specific capabilitites as those are implemented on platform level.

---

**EDIT:**

## Updated Proposal: Extended Configuration Support

[...]

PR #601, opened 2025-09-16
[Platform][Mistral][Gemini] Add some missing models

Guite:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | 
| License       | MIT

This MR adds constants for some missing models. While the Gemini models could also just be used with string literals, the Mistral Medium model does not work because it needs to get its its capabilities properly specified.

### DIFF
--- a/src/platform/src/Bridge/Mistral/Mistral.php
+++ b/src/platform/src/Bridge/Mistral/Mistral.php
@@ -20,13 +20,13 @@ use Symfony\AI\Platform\Model;
 final class Mistral extends Model
 {
     public const CODESTRAL = 'codestral-latest';
-    public const CODESTRAL_MAMBA = 'open-codestral-mamba';
     public const MISTRAL_LARGE = 'mistral-large-latest';
+    public const MISTRAL_MEDIUM = 'mistral-medium-latest';
     public const MISTRAL_SMALL = 'mistral-small-latest';
     public const MISTRAL_NEMO = 'open-mistral-nemo';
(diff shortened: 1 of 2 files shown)


src/platform/src/Bridge/Mistral/Mistral.php
```diff
@@ -22,6 +22,7 @@ final class Mistral extends Model
     public const CODESTRAL = 'codestral-latest';
     public const CODESTRAL_MAMBA = 'open-codestral-mamba';
     public const MISTRAL_LARGE = 'mistral-large-latest';
+    public const MISTRAL_MEDIUM = 'mistral-medium-latest';
```
petski, line 24: Can you review all models please? :angel: 

See

```bash
curl -s -q --location "https://api.mistral.ai/v1/models" --header 'Content-Type: application/json' --header 'Accept: application/json' --header "Authorization: Bearer $MISTRAL_API_KEY" | jq | grep '"id"'
```

I.e. `mistral-3b-latest` does not exist, it should be `ministral-3b-latest` instead.

Guite: added corresponding changes in 9b94730

OskarStark:
>  the Mistral Medium model does not work because it needs to get its its capabilities properly specified.

This PR would not be needed once we implement
*  https://github.com/symfony/ai/issues/401

OskarStark:
Thank you @Guite.

PR #635, opened 2025-09-18
[Agent] Allow translatable prompts in `SystemPromptInputProcessor`

VincentLanglet:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes/no
| Docs?         |  no 
| Issues        | Fix #... 
| License       | MIT

This is a follow up of https://github.com/symfony/ai/pull/514 with a suggestion of using TranslatableMessage instance for prompts rather than the need in SystemPromptInputProcessor constructor of
- A boolean enableTranslation
- A domain

Since we're passing a TranslatableMessage this also allow to use translation parameters.

cc @OskarStark that was the suggestion in https://github.com/symfony/ai/pull/514#issuecomment-3299514958

### DIFF
--- a/src/agent/composer.json
+++ b/src/agent/composer.json
@@ -42,6 +42,7 @@
         "symfony/dom-crawler": "^7.3|^8.0",
         "symfony/event-dispatcher": "^7.3|^8.0",
         "symfony/http-foundation": "^7.3|^8.0",
+        "symfony/translation": "^7.3|^8.0",
         "symfony/translation-contracts": "^3.6"
     },
     "autoload": {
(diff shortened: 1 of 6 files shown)


VincentLanglet:
> LGTM after CI is green and do we need some docs?

Indeed,  I forgot the ci, it's fixed now @OskarStark.

For the doc, https://github.com/symfony/ai/pull/514 didn't introduce any so there is nothing to update.
It's mainly a change about how the ai-bundle transform the config.
Also I feel like the constructor is self-documenting himself

OskarStark:
Thank you @VincentLanglet.

PR #641, opened 2025-09-21
[AI Bundle] Improvement on alias for store

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix none
| License       | MIT

Hi 👋🏻 

A small improvement for the store alias, this approach allows to use the same name for various store (like `foo` for a memory one, a qdrant one, etc) without injection issues.

### DIFF
--- a/src/ai-bundle/tests/DependencyInjection/AiBundleTest.php
+++ b/src/ai-bundle/tests/DependencyInjection/AiBundleTest.php
@@ -29,6 +29,7 @@ use Symfony\AI\Store\Document\Filter\TextContainsFilter;
 use Symfony\AI\Store\Document\Loader\InMemoryLoader;
 use Symfony\AI\Store\Document\Transformer\TextTrimTransformer;
 use Symfony\AI\Store\Document\Vectorizer;
+use Symfony\AI\Store\StoreInterface;
 use Symfony\Component\Config\Definition\Exception\InvalidConfigurationException;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\ContainerInterface;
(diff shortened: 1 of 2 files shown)


chr-hertel:
So I would have to use `StoreInterface $chromadbBlogStore` instead of `StoreInterface $blogStore`?

src/ai-bundle/tests/DependencyInjection/AiBundleTest.php
```diff
@@ -100,6 +101,24 @@ public function testInjectionAgentAliasIsRegistered()
...
+                        'main' => [
+                            'strategy' => 'cosine',
+                        ],
+                    ],
+                ],
+            ],
+        ]);
+
+        $this->assertTrue($container->hasAlias(StoreInterface::class));
+        $this->assertTrue($container->hasAlias(StoreInterface::class.' $memorymainStore'));
```
OskarStark:
```suggestion
        $this->assertTrue($container->hasAlias(StoreInterface::class.' $memoryMainStore'));
```
?

OskarStark:
What about registering both? The long and the short version? I would argue, that this is not really common use case, right?

Also docs would need an update

src/ai-bundle/tests/DependencyInjection/AiBundleTest.php
Guikingone: Yes and no, your version is the one I'd like to have but sadly, Sf is not returning it that way 😅

Guikingone:
@chr-hertel Yes

@OskarStark Yes, agree on the fact that it's not a common approach (even if it can be used when you're migrating from one to another), registering both is fine for me if @chr-hertel agreed on it 👍🏻

src/ai-bundle/tests/DependencyInjection/AiBundleTest.php
OskarStark: Shall we open an issue on Symfony?

Guikingone: Would be a good idea, I bet this behavior is what Sf is expecting but using it is a real pain, maybe it can be improved (I'll be glad to help on this one if it can improve the code that we already merged 😅 )

OskarStark: Ping me in the issue once opened

Guikingone: Issue opened: https://github.com/symfony/symfony/issues/61810, thanks for the help 🙂

OskarStark:
Thank you @Guikingone.

PR #642, opened 2025-09-22
Add support for query parameters in model configuration

OskarStark:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | --
| License       | MIT


chr-hertel:
Is this for the DSN thingy or what are you aiming for?

OskarStark:
And for the model catalog to be able to modify options, as the bundle will not provide a model object anymore, just a plain string

PR #650, opened 2025-09-23
[AI Bundle] Document store dependency injection improvement

OskarStark:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | yes
| Issues        | Follows #641 
| License       | MIT

Add documentation for the improved store alias functionality that allows multiple stores with the same name across different types. The feature provides both simple aliases and type-prefixed aliases for flexible dependency injection without conflicts.

### DIFF
--- a/src/ai-bundle/doc/index.rst
+++ b/src/ai-bundle/doc/index.rst
@@ -145,6 +145,55 @@ Configuration
                 vectorizer: 'ai.vectorizer.mistral_embeddings'
                 store: 'ai.store.memory.research'
 
+Store Dependency Injection
+--------------------------
+
+When using multiple stores in your application, the AI Bundle provides flexible dependency injection through store aliases. 
+This allows you to inject specific stores into your services without conflicts, even when stores share the same name across different types.
+
+For each configured store, the bundle automatically creates two types of aliases:
+
+1. **Simple alias**: ``StoreInterface $storeName`` - Direct reference by store name
+2. **Type-prefixed alias**: ``StoreInterface $typeStoreName`` - Reference with store type prefix in camelCase
+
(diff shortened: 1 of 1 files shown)


src/ai-bundle/doc/index.rst
```diff
@@ -145,6 +145,69 @@ Configuration
...
+
+You can inject stores into your services using the generated aliases::
+
+    use Symfony\AI\Store\StoreInterface;
+
+    final readonly class DocumentService
+    {
+        public function __construct(
+            private StoreInterface $main,              // Uses memory store (first occurrence)
+            private StoreInterface $weaviate_main,     // Explicitly uses weaviate store
```
OskarStark: I am wondering if this should be `$weaviateMain` @Guikingone

Guikingone: Agree, in the version you pushed, it could be:

```php
private StoreInterface $chromaDbMain # Not sure about `Db`
```

Guikingone: I was working on the documentation before you merged, I'm slow 😅

OskarStark: Sorry 😄

OskarStark: > Agree, in the version you pushed, it could be:

So should I change that?

Guikingone: I would change it to `chromaDbMain`, it respects the PSRs and Symfony style guideline, not a fan of the `foo_bar` syntax 🤔

OskarStark: If it is working, ok 😄

PR #654, opened 2025-09-23
[Platform] Remove default values from model constructors

OskarStark:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | Follows #653 
| License       | MIT

Remove hardcoded max_tokens defaults from Claude and Nova model classes to prevent potential BC breaks when these values might change in future versions. Users should explicitly specify max_tokens if needed.

### DIFF
--- a/src/platform/src/Bridge/Anthropic/Claude.php
+++ b/src/platform/src/Bridge/Anthropic/Claude.php
@@ -36,7 +36,7 @@ class Claude extends Model
      */
     public function __construct(
         string $name,
-        array $options = ['max_tokens' => 1000],
+        array $options = [],
     ) {
         $capabilities = [
             Capability::INPUT_MESSAGES,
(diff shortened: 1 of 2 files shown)


src/platform/src/Bridge/Anthropic/Claude.php
```diff
@@ -36,7 +36,7 @@ class Claude extends Model
      */
     public function __construct(
         string $name,
-        array $options = ['max_tokens' => 1000],
```
chr-hertel, line 39: setting `max_tokens` is actually needed - if you mind the value here, we could set it in the ModelClient instead, but we need to have a default to have the API similar to the other models.

OskarStark: Can you explain why this is needed?

chr-hertel: it's not mee - it's them :D

https://docs.claude.com/en/api/messages#body-max-tokens

PR #655, opened 2025-09-23
[Store] Introduce a Failover store

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | Fix None
| License       | MIT

Hi 👋🏻 

This PR is an idea, I was searching for something that can easily handle errors and ended looking at the `Mailer` transports, here's a naive implementation for stores.

Thanks for the feedback 👍🏻 

PS: If the idea is interesting, I'll update the docs, no need to work on it if we don't merge or if the API change.

### DIFF
--- a/src/store/src/FailoverStore.php
+++ b/src/store/src/FailoverStore.php
@@ -71,6 +71,6 @@ final readonly class FailoverStore implements ManagedStoreInterface, StoreInterf
             }
         }
 
-        throw new RuntimeException('No store available');
+        throw new RuntimeException('No store available.');
     }
 }


OskarStark:
So when using this while indexing, I could end up with n stores, each containing some documents?

Guikingone:
I'm totally honest, yes, that's a risk, we can't control if a store fails during the storage of the second / third document.

What we can do, is implementing a small strategy to retry the operation after a small period (called quorum on certain implementations) and see if it works, not perfect but might solve the consistency issues 🤔

OskarStark:
Hmm, I am not convinced yet, or are you facing this issue right now?

Guikingone:
> are you facing this issue right now?

Which one? The error during query or the need for the failover store? 🤔

OskarStark:
The need for failover

Guikingone:
In this case, let's close this PR and focus on pending ones 🙂

PR #659, opened 2025-09-23
 Set model only once

OskarStark:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | --
| License       | MIT

No need to set and then overwrite the model IMHO

### DIFF
--- a/examples/perplexity/token-metadata.php
+++ b/examples/perplexity/token-metadata.php
@@ -19,7 +19,7 @@ use Symfony\AI\Platform\Message\MessageBag;
 require_once dirname(__DIR__).'/bootstrap.php';
 
 $platform = PlatformFactory::create(env('PERPLEXITY_API_KEY'), http_client());
-$model = new Perplexity(Perplexity::SONAR);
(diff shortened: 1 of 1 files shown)


examples/perplexity/token-metadata.php
```diff
@@ -19,15 +19,14 @@
...
-$model = new Perplexity(Perplexity::SONAR);
+$model = new Perplexity(Perplexity::SONAR_DEEP_RESEARCH);
 $agent = new Agent($platform, $model, outputProcessors: [new TokenOutputProcessor()], logger: logger());
 
 $messages = new MessageBag(
     Message::forSystem('You are a pirate and you write funny.'),
     Message::ofUser('What is the Symfony framework?'),
 );
 $result = $agent->call($messages, [
-    'model' => Perplexity::SONAR_DEEP_RESEARCH,
```
chr-hertel, line 30: not for the example, but i need the option in general :laughing:

OskarStark: Yes the options stays for sure

PR #663, opened 2025-09-24
[PHPStan] Add rule to forbid empty() usage and replace all existing empty() calls

OskarStark:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | --
| License       | MIT

### DIFF
--- a/.phpstan/ForbidEmptyRule.php
+++ b/.phpstan/ForbidEmptyRule.php
@@ -1,0 +1,68 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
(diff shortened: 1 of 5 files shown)


derrabus:
You might want to consider using PHPStan's strict rules plugin, see https://github.com/phpstan/phpstan-strict-rules/blob/2.0.x/src/Rules/DisallowedConstructs/DisallowedEmptyRule.php

OskarStark:
> You might want to consider using PHPStan's strict rules plugin, see https://github.com/phpstan/phpstan-strict-rules/blob/2.0.x/src/Rules/DisallowedConstructs/DisallowedEmptyRule.php

Thanks for the hint and I am aware of it, but right now I want to allow empty in `ai-bundle/config/options.php` as it can be quite complex. 👍

PR #669, opened 2025-09-24
[Platform][Ollama] Better regex for qwen tool pattern

sonnymilton:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes/no 
| Issues        |  - 
| License       | MIT

The current regex for QWEN models does not cover model size notation like `qwen3:32b` or `qwen3-coder:32b`.
I changed the regular expression to support notations with model sizes.


chr-hertel:
Thank you @sonnymilton.

PR #675, opened 2025-09-25
[Chat] Introduce a new component

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | Related to #254 
| License       | MIT

Hi 👋🏻 

Ok, time to tackle the big work on the #254 issue, the current structure doesn't allows to introduce new stores neither handling the storage of messages, mainly due to the fact that we can't intervene in the agent component without breaking the SRP of it.

This PR aims to split the `Chat` into a new component and ease the work on storing messages (with a new component, we can easily add storages here rather than in the `agent` component), plus, it helps splitting the responsibilities inside the initiative as we should be allowed to start new agents without relying on any chats.

This PR is a draft, feel free to open debates about it 😄

### DIFF
--- a/examples/chat/persistent-chat.php
+++ b/examples/chat/persistent-chat.php
@@ -10,8 +10,8 @@
  */
 
 use Symfony\AI\Agent\Agent;
-use Symfony\AI\Agent\Chat;
-use Symfony\AI\Agent\Chat\MessageStore\InMemoryStore;
+use Symfony\AI\Chat\Bridge\Local\InMemoryStore;
+use Symfony\AI\Chat\Chat;
 use Symfony\AI\Platform\Bridge\OpenAi\PlatformFactory;
 use Symfony\AI\Platform\Message\Message;
 use Symfony\AI\Platform\Message\MessageBag;
--- a/examples/composer.json
+++ b/examples/composer.json
@@ -17,6 +17,7 @@
         "probots-io/pinecone-php": "^1.0",
         "psr/http-factory-implementation": "*",
         "symfony/ai-agent": "@dev",
+        "symfony/ai-chat": "@dev",
         "symfony/ai-platform": "@dev",
         "symfony/ai-store": "@dev",
         "symfony/cache": "^7.3|^8.0",
--- a/src/chat/composer.json
+++ b/src/chat/composer.json
@@ -1,0 +1,54 @@
+{
+    "name": "symfony/ai-chat",
+    "description": "PHP library for building chats with agents.",
+    "license": "MIT",
+    "type": "library",
+    "keywords": [
+        "ai",
+        "llm",
+        "agent",
+        "chat"
+    ],
+    "authors": [
+        {
+            "name": "Christopher Hertel",
+            "email": "mail@christopher-hertel.de"
+        },
+        {
+            "name": "Oskar Stark",
+            "email": "oskarstark@googlemail.com"
+        }
+    ],
+    "require": {
+        "php": ">=8.2",
+        "symfony/ai-agent": "@dev"
+    },
+    "require-dev": {
+        "phpstan/phpstan": "^2.0",
+        "phpstan/phpstan-strict-rules": "^2.0",
+        "phpunit/phpunit": "^11.5.13",
+        "symfony/http-foundation": "^7.3|^8.0",
+        "psr/cache": "^3.0"
+    },
+    "autoload": {
+        "psr-4": {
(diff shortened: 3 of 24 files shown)


src/chat/composer.json
```diff
@@ -0,0 +1,55 @@
...
+        },
+        {
+            "name": "Oskar Stark",
+            "email": "oskarstark@googlemail.com"
+        }
+    ],
+    "require": {
+        "php": ">=8.2",
+        "symfony/ai-agent": "@dev",
+        "symfony/ai-platform": "@dev"
```
OskarStark: Wait agent should already rely on platform, is this not the case?

examples/composer.json
```diff
@@ -17,6 +17,7 @@
         "probots-io/pinecone-php": "^1.0",
         "psr/http-factory-implementation": "*",
         "symfony/ai-agent": "@dev",
+        "symfony/ai-chat": "@dev",
```
OskarStark, line 20: Either we don't need it or we need to adjust an example to use the new namespace

src/chat/composer.json
Guikingone: Both `agent` and `chat` relies on `platform` as messages are defined in `platform`, should we change this structure? 🤔

examples/composer.json
Guikingone: It's needed for `examples/misc/persistent-chat`, the namespaces import are already adjusted.

src/chat/composer.json
OskarStark: So require the platform is not needed, right?

OskarStark:
I like having a new component for this

src/chat/composer.json
Guikingone: For now, it's required for the `Chat` to use messages, their classes are defined in `Platform` so either we stay as it or we move the messages in `chat`: 

```php
# ...
{

    $messages->add($message);

    $this->store->save($messages);

    return $assistantMessage;
}
```

My approach would be to let as it for now (as messages are used outside of `chat`).

OskarStark: Ok sorry, I mean, as we require the agent, we also require the platform indirectly, which is fine, right?

Guikingone: Oh ok, got it, yes, absolutely, we can remove this dependency 🙂

src/chat/src/Bridge/Symfony/SessionStore.php
```diff
@@ -9,10 +9,10 @@
  * file that was distributed with this source code.
  */
 
-namespace Symfony\AI\Agent\Chat\MessageStore;
+namespace Symfony\AI\Chat\Bridge\Symfony;
```
chr-hertel: maybe rather `HttpFoundation`?

```suggestion
namespace Symfony\AI\Chat\Bridge\HttpFoundation;
```

examples/chat/persistent-chat.php
chr-hertel, line 1: let's introduce a `chat` folder then

chr-hertel (changes requested):
missing a readme, right?

chr-hertel:
alright, let's go - will add some comments on #254 as well - building on top of this i guess.

Guikingone:
@chr-hertel Yes, the idea is to start by splitting the chat, from that point, we can work on the storage part then handle the final implementation 🙂

Guikingone:
@OskarStark @chr-hertel Regarding this PR, I can't fix the tests / quality as the `ai-chat` repository is required, is there any solution to fix/validate them before merging or should we wait until its merged? 🤔

OskarStark:
I had the same problem for a new package in the brave tool PR.
I am fine merging as is and taking care of the CI afterwards, but immediately 👌🏻😝

Guikingone:
Ok, no problem 😅 

Let me know when you have time to review the PR and merging it, on my side, most of the work is done, I can work on other PRs until you merge it.

OskarStark:
@fabpot we would need a new package here please

fabpot:
@OskarStark We need to merge the PR first :) Ping me when done.

OskarStark:
Thank you @Guikingone.

OskarStark:
@fabpot done ✅ 

@Guikingone please keep an 👁️ on the CI, thanks 🙏

fabpot:
New package created and published on Packagist: https://packagist.org/packages/symfony/ai-chat

OskarStark:
Thanks

PR #677, opened 2025-09-26
 Refactor examples to use `print_stream` helper function

OskarStark:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | --
| License       | MIT

Extract repeated streaming pattern into reusable print_stream function in examples/bootstrap.php and update all streaming examples to use it.

### DIFF
--- a/examples/anthropic/stream.php
+++ b/examples/anthropic/stream.php
@@ -25,7 +25,4 @@ $messages = new MessageBag(
 );
 $result = $platform->invoke($model, $messages, ['stream' => true]);
 
-foreach ($result->getResult()->getContent() as $word) {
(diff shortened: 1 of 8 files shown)


chr-hertel:
naaah, not sure tbh, we're now extracting code from the example that is actually helpful to understand what's going on - i feel like this harms the meaning of the example

same for other `print_` functions maybe

OskarStark:
Ok, so shall I revert?

chr-hertel:
we can discuss in #693 - it's not the only one, but let's see how we move forward.

PR #683, opened 2025-09-26
[AI Bundle] Rename `ai:chat` command to `ai:agent:call`

OskarStark:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | Follows #395 
| License       | MIT

I think we can have ai:chat command soon after #675


src/ai-bundle/src/Command/AgentCallCommand.php
```diff
@@ -32,10 +32,11 @@
  * @author Oskar Stark <oskarstark@googlemail.com>
  */
 #[AsCommand(
-    name: 'ai:chat',
-    description: 'Chat with an agent',
+    name: 'ai:agent:call',
+    aliases: ['ai:chat'],
```
OskarStark, line 36: I keep the alias for now

PR #640, opened 2025-09-19
[AI Bundle][Platform] Add `ModelCatalog`

OskarStark:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | --
| License       | MIT

Recipe-PR:
* https://github.com/symfony/recipes/pull/1452

# ModelCatalog Architecture Refactoring

### BC Breaks
* :warning: All model constants removed
* :warning: Child model class constructors were changed and now have an `$capabilities` parameter
* ⚠️ `PlatformFactory` constructor changed for all bridges to be able to receive a `ModelCatalogInterface`
* ⚠️ `PlatformInterface` now has an additional `getModelCatalog` method
* ⚠️ `PlatformInterface::__invoke` no receives an model as string (before `Model`)
* ⚠️ `Vectorizer::__construct` no receives an model as string (before `Model`)
* ⚠️ `Agent::__construct` no receives an model as string (before `Model`)

## Overview
This PR refactors how models are defined and managed in the Symfony AI Platform, moving from constants in Model classes to centralized ModelCatalogs.

### Defining Models
Models were defined as constants in Model classes:

```php
// src/platform/src/Bridge/OpenAi/Gpt.php
class Gpt extends Model
{
    public const GPT_4O = 'gpt-4o';
    public const GPT_4O_MINI = 'gpt-4o-mini';
    // ... more constants
    
    public function __construct(string $name = self::GPT_4O, array $options = [])
    {
        $capabilities = [
            Capability::INPUT_MESSAGES,
            Capability::OUTPUT_TEXT,
            // ... capabilities determined in constructor
        ];
        
        if (self::GPT_4O_AUDIO === $name) {
            $capabilities[] = Capability::INPUT_AUDIO;
        }
        // ... more logic
        
        parent::__construct($name, $capabilities, $options);
    }
}
```

### Using Models
```php
use Symfony\AI\Platform\Bridge\OpenAi\Gpt;
use Symfony\AI\Platform\Bridge\Voyage\Voyage;

// Had to create Model objects
$model = new Gpt('gpt-4o-mini');
$result = $platform->invoke($model, $input);

// Or with constants
$embeddings = new Voyage(Voyage::V3);
$result = $platform->invoke($embeddings, $text);
```

### Adding Custom Models
There was no standardized way to add custom models. You had to:
- Either extend a Model class and override the constructor
- Or create a Model instance with hardcoded capabilities

### Defining Models
[...]

PR #696, opened 2025-09-28
[AI Bundle] Fix service registration

OskarStark:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Follows #640 
| License       | MIT

### DIFF
--- a/src/ai-bundle/config/services.php
+++ b/src/ai-bundle/config/services.php
@@ -73,7 +73,7 @@ return static function (ContainerConfigurator $container): void {
             ->factory([OpenAiContract::class, 'create'])
         ->set('ai.platform.contract.anthropic', Contract::class)
             ->factory([AnthropicContract::class, 'create'])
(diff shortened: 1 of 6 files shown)


src/platform/src/Bridge/Voyage/PlatformFactory.php
OskarStark, line 1: only line breaks here

src/ai-bundle/config/services.php
```diff
@@ -73,7 +73,7 @@
             ->factory([OpenAiContract::class, 'create'])
         ->set('ai.platform.contract.anthropic', Contract::class)
             ->factory([AnthropicContract::class, 'create'])
-        ->set('ai.platform.contract.google', Contract::class)
+        ->set('ai.platform.contract.gemini', Contract::class)
```
OskarStark, line 76: it is also aliased as `GeminiContract`

PR #700, opened 2025-09-29
[AI Bundle] Fix data_collector template for multiple parameter types

franzwilding:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | 
| License       | MIT

The tool call parameter type can be an array (for example, when it is nullable). In this case, the entire profiler tab breaks. This fix addresses the issue by properly handling multiple types.


chr-hertel:
Thank you @franzwilding.

PR #701, opened 2025-09-29
[Agent] Passing filtered tools option for internal agent calls

franzwilding:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | 
| License       | MIT

When passing agent options to an internal agent call, the tools array is already resolved and is no longer just an array of strings. As a result, the input processor no longer filters the available agent tools. I’m not sure if this behavior is intentional, but in my case it causes issues: with each internal call, all tools suddenly become available, which confuses the model.


chr-hertel:
Hey, okay, sounds fishy, are you referring to an agent using a different agent as a tool or what is an "internal agent"?

franzwilding:
No, just regular tool calling. Once, one model does not successfully run, the inner agent->call is triggered and at this time, the tools are already resolved.

franzwilding:
@chr-hertel I think this only happens, if we have an event without an result. In this case, the AgentProcessor will call the agent again: 
```$result = $event->hasResponse() ? $event->result : $this->agent->call($messages, $output->options);```

chr-hertel:
Do you have a way how I could reproduce this?

chr-hertel:
Hey @franzwilding - is this still relevant?

OskarStark:
friendly ping 😃

franzwilding:
I could not reproduce the problem anymore in recent versions

PR #703, opened 2025-09-29
[Platform][Gemini] Do tool call if any `contentPart` is a `functionCall`

DrauzJu:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        |
| License       | MIT

The Gemini API may return multiple content parts even if a tool call is desired. For example, one content part can be an explanation why the tool was called. This PR adds support for this by first checking if any content part is a `functionCall`. If so, all other content parts are discarded and the function/tool call is processed.

### DIFF
--- a/src/platform/tests/Bridge/Gemini/Gemini/ResultConverterTest.php
+++ b/src/platform/tests/Bridge/Gemini/Gemini/ResultConverterTest.php
@@ -15,6 +15,8 @@ use PHPUnit\Framework\TestCase;
 use Symfony\AI\Platform\Bridge\Gemini\Gemini\ResultConverter;
 use Symfony\AI\Platform\Exception\RuntimeException;
 use Symfony\AI\Platform\Result\RawHttpResult;
+use Symfony\AI\Platform\Result\ToolCall;
+use Symfony\AI\Platform\Result\ToolCallResult;
 use Symfony\Contracts\HttpClient\ResponseInterface;
 
 /**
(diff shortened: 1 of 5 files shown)


src/platform/tests/Bridge/Gemini/Gemini/ResultConverterTest.php
```diff
@@ -40,4 +42,38 @@ public function testConvertThrowsExceptionWithDetailedErrorInformation()
...
+                ],
+            ],
+        ]);
+
+        $result = $converter->convert(new RawHttpResult($httpResponse));
+        $this->assertInstanceOf(ToolCallResult::class, $result);
+        $this->assertCount(1, $result->getContent());
+        $toolCall = $result->getContent()[0];
+        $this->assertInstanceOf(ToolCall::class, $toolCall);
+        $this->assertEquals('1234', $toolCall->id);
```
OskarStark:
```suggestion
        $this->assertSame('1234', $toolCall->id);
```

OskarStark:
I think the same needs to be done for the vertexai bridge, right?

DrauzJu:
> I think the same needs to be done for the vertexai bridge, right?

@OskarStark Probably yes, but I never used it so far. I can try it though.

OskarStark:
The code should look quite similar, can you try adding your fix there too? Thanks

OskarStark (approved):

OskarStark:
Thanks for fixing this bug Julian.

PR #707, opened 2025-09-30
[Store] Introduce `EmbeddableDocumentInterface`

### LINKED ISSUE #18: [Store] Support user land objects for embedding instead of TextDocument

paulinevos:
| Q             | A
| ------------- | ---
| Bug fix?      |no
| New feature?  | no 
| Docs?         | no 
| Issues        | Fix #18 
| License       | MIT

In an effort to support user land objects for the store, this PR introduces an `Embeddable` interface instead of relying on the `TextDocument` class. Please note this interface would require php ^8.4, as it uses interface properties. I chose this approach so as to introduce minimal change to the existing `TextDocument` class and usages. Instead of `content`, an `Embeddable` document has a `data` property. Any type embeddable document class can implement a hook for the `data` property if needed.

This PR also introduces union types for the document IDs so as to not make assumptions about the ID type.

### DIFF
--- a/src/store/src/Document/EmbeddableDocumentInterface.php
+++ b/src/store/src/Document/EmbeddableDocumentInterface.php
@@ -1,0 +1,21 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Store\Document;
+
+interface EmbeddableDocumentInterface
+{
+    public function getId(): mixed;
+
+    public function getContent(): string;
+
+    public function getMetadata(): Metadata;
+}
--- a/src/store/src/Document/TextDocument.php
+++ b/src/store/src/Document/TextDocument.php
@@ -17,12 +17,12 @@ use Symfony\Component\Uid\Uuid;
 /**
  * @author Christopher Hertel <mail@christopher-hertel.de>
  */
-final readonly class TextDocument
+final readonly class TextDocument implements EmbeddableDocumentInterface
 {
     public function __construct(
-        public Uuid $id,
-        public string $content,
-        public Metadata $metadata = new Metadata(),
+        private Uuid $id,
+        private string $content,
+        private Metadata $metadata = new Metadata(),
     ) {
         if ('' === trim($this->content)) {
             throw new InvalidArgumentException('The content shall not be an empty string.');
@@ -32,5 +32,20 @@ final readonly class TextDocument
     public function withContent(string $content): self
     {
         return new self($this->id, $content, $this->metadata);
+    }
+
+    public function getId(): Uuid
+    {
+        return $this->id;
+    }
+
+    public function getContent(): string
+    {
+        return $this->content;
+    }
+
+    public function getMetadata(): Metadata
+    {
+        return $this->metadata;
     }
 }
--- a/src/store/src/Document/VectorizerInterface.php
+++ b/src/store/src/Document/VectorizerInterface.php
@@ -14,7 +14,7 @@ namespace Symfony\AI\Store\Document;
 use Symfony\AI\Platform\Vector\Vector;
 
 /**
- * Interface for converting a collection of TextDocuments into VectorDocuments
+ * Interface for converting a collection of Embeddable documents into VectorDocuments
  * and for vectorizing individual strings.
  *
  * @author Oskar Stark <oskarstark@googlemail.com>
@@ -22,12 +22,12 @@ use Symfony\AI\Platform\Vector\Vector;
 interface VectorizerInterface
 {
     /**
-     * @param TextDocument[]       $documents
-     * @param array<string, mixed> $options   Options to pass to the underlying platform
+     * @param EmbeddableDocumentInterface[] $documents
+     * @param array<string, mixed>          $options   Options to pass to the underlying platform
      *
      * @return VectorDocument[]
      */
-    public function vectorizeTextDocuments(array $documents, array $options = []): array;
+    public function vectorizeEmbeddableDocuments(array $documents, array $options = []): array;
 
     /**
      * Vectorizes a single string or Stringable object into a Vector.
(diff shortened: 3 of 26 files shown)


paulinevos:
Not sure why PHPStan fails on 8.4 -- is it a linking issue? The argument it's complaining about seems fine to me.

How do we feel about the rest? < 8.4 is going to fail because of the interface properties. Should we support 8.4 or should I solve this with getter methods on the interface instead of interface properties to accommodate for the earlier versions?

src/store/composer.json
```diff
@@ -32,7 +32,7 @@
         }
     ],
     "require": {
-        "php": ">=8.2",
+        "php": ">=8.4",
```
chr-hertel: is there are reason besides those property hooks for bumping the php version?

in general i don't think we should do that

paulinevos: No, that's the only reason. Hence my previous comment, I wasn't sure if it's worth leveraging this new features since we haven't officially released yet or if I should find another solution.

paulinevos: Maybe an attribute is more appropriate. An attribute that can optionally take values for the `id` and `data` properties? Maybe that's better. Also in terms of not having to implement any hooks in the user land classes or changing their structure for this "embeddable" concept they probably shouldn't be aware of.

paulinevos: Decided against attributes because these objects are passed through a bunch of services. Just went with old-school interfaces instead.

src/store/src/Document/TextDocument.php
```diff
@@ -33,4 +33,14 @@ public function withContent(string $content): self
     {
         return new self($this->id, $content, $this->metadata);
     }
+
+    public function getId(): Uuid|string|int
+    {
+        return $this->id;
+    }
+
+    public function getContent(): string
```
OskarStark, line 42: lets also add a getMetadata() here, not to the interface, and make the properties private, this is anyway better for BC handling in the future.

src/store/src/Document/VectorizerInterface.php
```diff
@@ -14,20 +14,20 @@
...
 {
     /**
-     * @param TextDocument[]       $documents
+     * @param Embeddable[]         $documents
      * @param array<string, mixed> $options   Options to pass to the underlying platform
      *
      * @return VectorDocument[]
      */
-    public function vectorizeTextDocuments(array $documents, array $options = []): array;
+    public function vectorizeEmbeddableDocuments(array $documents, array $options = []): array;
```
OskarStark, line 30: What do you think of a generic method instead of 2 methods?
```php
// with a PHPStan if (string or string able, return Vector, if array, return array<VectorDocument>)
public function vectorize(string|\Stringable|array $subject, array $options = []): array|Vector;
```

Would that make sense? cc @GromNaN

src/store/composer.json
chr-hertel: Yea, old-school getters are better, and we still use `Interface` suffix with Symfony - meaning it'll be `EmbeddableInterface` i guess

src/store/src/Document/Embeddable.php
```diff
@@ -0,0 +1,20 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Store\Document;
+
+use Symfony\Component\Uid\Uuid;
+
+interface Embeddable
+{
+    public function getId(): Uuid|string|int;
```
GromNaN: This can return `mixed`, the type of the id is not meaningful as soon as it's consistent.
It can be an array of key-value pairs.

src/store/src/Document/TextDocument.php
```diff
@@ -33,4 +33,14 @@ public function withContent(string $content): self
     {
         return new self($this->id, $content, $this->metadata);
     }
+
+    public function getId(): Uuid|string|int
```
GromNaN: The implementation can be more restrictive than the interface.

```suggestion
    public function getId(): Uuid
```

src/store/src/Document/VectorizerInterface.php
GromNaN: The ability to process an array of documents is very good as the embedding backend can support batching or concurrent requests.
I like your idea of supporting both a single or a collection of texts. With a conditional return type it's possible to specify the actual return type of a given call based on its arguments; even if it's not a native feature.

paulinevos: There is already a generic method, haha. You wrote it! But it lacks the phpstan stuff, I'll add. I might move the `TextDocument` specific method into its own implementation of `VectorizerInterface::vectorize()`? Seems like it might be a bit more logical.

OskarStark: Maybe, but let's keep it in one class for now, can be a private method. Otherwise you need to know which vectorizer you want to inject where as it has a different service id

examples/openai/agent-as-tool.php
OskarStark, line 1: this should not be part of this PR, rebase broken? 🤔

src/store/src/Document/VectorizerInterface.php
paulinevos: As discussed in Slack, the generic implementation will be done in a separate PR

src/store/src/Document/EmbeddableDocumentInterface.php
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Store\Document;
+
+interface EmbeddableDocumentInterface
```
OskarStark, line 14: If this is the final name, PR title needs an update

paulinevos: Done and yeah, this name feels more "Symfony" imo

OskarStark:
Thank you Pauline.

PR #708, opened 2025-09-30
[Platform][Ollama] Add support for models with size in model catalog

sonnymilton:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Docs?         | no 
| Issues        | - 
| License       | MIT

This PR extends the `Ollama\ModelCatalog` to support model names that include a `size` suffix after a colon (e.g. `qwen3:32b`, `all-minilm:33m`).

- The `getModel()` method now resolves the model by its base name (before the `:` suffix).
- The actual name including the suffix is still passed to the model constructor.
- Additional test cases have been added to ensure models with size suffixes are correctly resolved.

### DIFF
--- a/src/platform/src/Bridge/Ollama/ModelCatalog.php
+++ b/src/platform/src/Bridge/Ollama/ModelCatalog.php
@@ -227,7 +227,7 @@ final class ModelCatalog extends AbstractModelCatalog
         $baseModelName = explode(':', $actualModelName, 2)[0];
 
         if (!isset($this->models[$actualModelName]) && !isset($this->models[$baseModelName])) {
-            throw new ModelNotFoundException(\sprintf('Model "%s" not found.', $baseModelName));
+            throw new ModelNotFoundException(\sprintf('Model "%s" not found.', $actualModelName));
         }
 
         $modelConfig = $this->models[$actualModelName] ?? $this->models[$baseModelName];
--- a/src/platform/src/Bridge/Ollama/ModelCatalog.php
+++ b/src/platform/src/Bridge/Ollama/ModelCatalog.php
@@ -12,9 +12,6 @@
 namespace Symfony\AI\Platform\Bridge\Ollama;
 
 use Symfony\AI\Platform\Capability;
-use Symfony\AI\Platform\Exception\InvalidArgumentException;
-use Symfony\AI\Platform\Exception\ModelNotFoundException;
-use Symfony\AI\Platform\Model;
 use Symfony\AI\Platform\ModelCatalog\AbstractModelCatalog;
 
 /**
@@ -23,7 +20,7 @@ use Symfony\AI\Platform\ModelCatalog\AbstractModelCatalog;
 final class ModelCatalog extends AbstractModelCatalog
 {
     /**
-     * @param array<string, array{class: class-string<Model>, capabilities: list<Capability>}> $additionalModels
+     * @param array<string, array{class: class-string<\Symfony\AI\Platform\Model>, capabilities: list<Capability>}> $additionalModels
      */
     public function __construct(array $additionalModels = [])
     {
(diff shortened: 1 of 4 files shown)


src/platform/src/Bridge/Ollama/ModelCatalog.php
```diff
@@ -210,4 +213,35 @@ public function __construct(array $additionalModels = [])
 
         $this->models = array_merge($defaultModels, $additionalModels);
     }
+
+    public function getModel(string $modelName): Model
```
OskarStark, line 217: Let's modify `AbstractModelCatalog` and interpret either everything as modelName or everything before `?`, if a `?` is present.

src/platform/tests/Bridge/Ollama/ModelCatalogTest.php
```diff
@@ -30,6 +30,7 @@ public static function modelsProvider(): iterable
         yield 'llama3' => ['llama3', Ollama::class, [Capability::INPUT_MESSAGES, Capability::OUTPUT_TEXT, Capability::OUTPUT_STRUCTURED, Capability::TOOL_CALLING]];
         yield 'mistral' => ['mistral', Ollama::class, [Capability::INPUT_MESSAGES, Capability::OUTPUT_TEXT, Capability::OUTPUT_STRUCTURED, Capability::TOOL_CALLING]];
         yield 'qwen3' => ['qwen3', Ollama::class, [Capability::INPUT_MESSAGES, Capability::OUTPUT_TEXT, Capability::OUTPUT_STRUCTURED, Capability::TOOL_CALLING]];
+        yield 'qwen3:32b' => ['qwen3:32b', Ollama::class, [Capability::INPUT_MESSAGES, Capability::OUTPUT_TEXT, Capability::OUTPUT_STRUCTURED, Capability::TOOL_CALLING]];
```
OskarStark, line 33: if they are officially supported, you can add them also to the Ollama\ModelCatalog.

src/platform/tests/Bridge/Ollama/ModelCatalogTest.php
```diff
@@ -45,6 +46,7 @@ public static function modelsProvider(): iterable
         yield 'nomic-embed-text' => ['nomic-embed-text', Ollama::class, [Capability::INPUT_MESSAGES, Capability::OUTPUT_TEXT, Capability::OUTPUT_STRUCTURED, Capability::INPUT_MULTIPLE]];
         yield 'bge-m3' => ['bge-m3', Ollama::class, [Capability::INPUT_MESSAGES, Capability::OUTPUT_TEXT, Capability::OUTPUT_STRUCTURED, Capability::INPUT_MULTIPLE]];
         yield 'all-minilm' => ['all-minilm', Ollama::class, [Capability::INPUT_MESSAGES, Capability::OUTPUT_TEXT, Capability::OUTPUT_STRUCTURED, Capability::INPUT_MULTIPLE]];
+        yield 'all-minilm:33m' => ['all-minilm:33m', Ollama::class, [Capability::INPUT_MESSAGES, Capability::OUTPUT_TEXT, Capability::OUTPUT_STRUCTURED, Capability::INPUT_MULTIPLE]];
```
OskarStark, line 49: same for this model

src/platform/src/ModelCatalog/AbstractModelCatalog.php
```diff
@@ -89,6 +90,7 @@ protected static function parseModelName(string $modelName): array
 
         return [
             'name' => $actualModelName,
+            'baseName' => explode(':', $actualModelName, 2)[0],
```
OskarStark: why do we need the basename?

sonnymilton: We use the basename fallback to avoid duplicating identical model registrations for every size suffix (e.g. `qwen3:32b`, `qwen3:64m`). 
Almost every model supported by OLLAMA contains variants with size suffix. For example, the qwen3 model supports over 50 different variants. https://ollama.com/library/qwen3/tags
Registering each one of them would lead to a huge increase of the catalog

src/platform/tests/Bridge/Ollama/ModelCatalogTest.php
sonnymilton: Ollama qwen3 supports 56 different variants.  
We don't want to register them all.

OskarStark: got it

src/platform/src/ModelCatalog/AbstractModelCatalog.php
OskarStark: Ok, but what are you doing with the baseName in user land?

sonnymilton: We don’t expose or use the baseName in user land.  
It’s only an internal fallback inside `AbstractModelCatalog::getModel()`.
From the perspective of user land, they always pass the full name (e.g. `qwen3:32b`) and get back a model with that exact name. The baseName is just used internally to resolve the model configuration when the variant isn’t explicitly registered.

OskarStark:
```suggestion
```
so we don't need it here, right? we will end up with model name `qwen3:32b` which is fine

OskarStark:
Ah now I get it, it is working, but you need to create the model to be able to use it, which is fine IMHO. the "sizes" are not common enough through all platforms IMHO. So you should define the model on your end and use it

src/platform/src/ModelCatalog/AbstractModelCatalog.php
sonnymilton: Since `:size` suffixes are not a cross-vendor pattern (only use Ollama), it might be more appropriate to keep the fallback logic in the `Symfony\AI\Platform\Bridge\Ollama\ModelCatalog::getModel()` (like in the first version of this PR https://github.com/symfony/ai/pull/708/commits/853d90256c34e347e590c1268e8c9725a99f6b04).

sonnymilton:
From the user perspective you still need to explicitly register the model you want to use. The fallback only helps when a variant isn’t registered but the base model is.

sonnymilton:
Also worth mentioning: before the recent refactoring (when `PlatformCatalog` was introduced), this already worked as expected in the Ollama bridge. At that time, the matching of a suitable config was done via regular expressions, which allowed different variants to resolve correctly to the base model. With the refactor this logic was lost, so this PR just reintroduces the same behavior.

OskarStark:
Fixed in #712 with a slightly different approach. Thanks for proposing.

PR #709, opened 2025-10-01
[Agent] Dispatch more tool call events

### LINKED ISSUE #680: [Agent] Streaming ResultInterface parts
I'm trying to use symfony/ai as a backend for https://ai-sdk.dev. ai-sdk requires the backend to send the content in a structured way (also when streaming). This allows the frontend to display status information, internal tool calling, reasoning etc. (like in le chat or chat-gpt). 

Basic content streaming of one text part works well, by wrapping each chunk of the content response in a small envelope: 

```php
$result = $defaultAgent->call($messages, ['stream' => true]);
$response = new StreamedResponse(function () use ($result) {

        $send = function (string $type, array $data = []) {
            echo 'data: ' . json_encode(array_merge(['type' => $type], $data), JSON_UNESCAPED_SLASHES) . "\n\n";
            @ob_flush(); @flush();
        };

        $send('start');
        $textId = (string) new Ulid();
        $send('text-start', ['id' => $textId]);
        
        foreach ($result->getContent() as $part) {
            $send('text-delta', ['id' => $textId, 'delta' => $part]);
        }

        $send('text-end', ['id' => $textId]);
        $send('finish');
        echo "data: [DONE]\n\n";
        @ob_flush();
        @flush();
}
```

However when i want to stream tool calling status or internal reasoning, things are getting more complicated. 

I tried to implement a custom OutputProcessor, however the internal agentic flow breaks, when replacing tool_calling result with a custom envenlope. 

The best result I got, was writing to the output-stream directly from an output-processor, however this is not very elegant or maintanable. 

Basically what I need would be: 
- a better $result->getResultPart()s generator, that will not return simple content, but instances of `ResultInterface` or mybe a new `FinalResultInterface`? 
- A new TextChunkResult to not get a mixture of string and `ResultInterface` objects. This would allow me to add my envelopes in the `$result->getResultParts()` loop depending on the part class. 

I tink this would be a very nice feature for any UI that wants to display whats going on inside one agent call.

franzwilding:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | Fix #680
| License       | MIT

This PR dispatches three new events around tool calling.

### DIFF
--- a/src/agent/doc/index.rst
+++ b/src/agent/doc/index.rst
@@ -331,6 +331,24 @@ to skip the next LLM call by setting a result yourself::
         }
     });
 
+Tool Call Lifecycle Events
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If you need to react more granularly to the lifecycle of individual tool calls, you can listen to the
+``ToolCallArgumentsResolved``, ``ToolCallSucceeded`` and ``ToolCallFailed`` events. These are dispatched at different stages::
+
+    $eventDispatcher->addListener(ToolCallArgumentsResolved::class, function (ToolCallArgumentsResolved $event): void {
+        // Let the client know, that the tool $event->toolCall->name was executed
+    });
+
+    $eventDispatcher->addListener(ToolCallSucceeded::class, function (ToolCallSucceeded $event): void {
+        // Let the client know, that the tool $event->toolCall->name successfully returned the result $event->result
+    });
+
+    $eventDispatcher->addListener(ToolCallFailed::class, function (ToolCallFailed $event): void {
+        // Let the client know, that the tool $event->toolCall->name failed with the exception: $event->exception
+    });
+
 Keeping Tool Messages
 ~~~~~~~~~~~~~~~~~~~~~
 
--- a/src/agent/src/Toolbox/Event/ToolCallFailed.php
+++ b/src/agent/src/Toolbox/Event/ToolCallFailed.php
@@ -1,0 +1,31 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Agent\Toolbox\Event;
+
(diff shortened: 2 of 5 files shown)


src/agent/src/Toolbox/Event/ToolCallExecute.php
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+namespace Symfony\AI\Agent\Toolbox\Event;
+
+use Symfony\AI\Platform\Tool\Tool;
+
+/**
+ * Dispatched just before invoking a tool.
+ */
+final readonly class ToolCallExecute
```
OskarStark:
```suggestion
final readonly class ExecuteToolCall
```
?

src/agent/src/Toolbox/Event/ToolCallSuccess.php
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+namespace Symfony\AI\Agent\Toolbox\Event;
+
+use Symfony\AI\Platform\Tool\Tool;
+
+/**
+ * Dispatched after successfully invoking a tool.
+ */
+final readonly class ToolCallSuccess
```
OskarStark: either `ToolCallSuccessful` or `SuccessfulToolCall`? Same for the failed one.

WDYT @chr-hertel ?

OskarStark:
Thanks for tackling this topic, can you please add some docs?

chr-hertel (changes requested):
I think in general this totally makes sense, and I'd only see some fine-tuning here.

1. I think it's a bit odd to me having basically two events with the same payload dispatched line after line, referring to `ToolCallArgumentsResolved`  and `ExecuteToolCall` - do i miss something here?
2. I would propose to move to an event-style of naming, and we're half way there

proposed changes:
* drop `ExecuteToolCall`
* rename `ToolCallSuccess` to `ToolCallSucceeded`

franzwilding:
@chr-hertel I renamed the success event and dropped ExecuteToolCall.

OskarStark:
Thank you @franzwilding, I rebased your PR to get rid of the merge commit 👍

chr-hertel:
Nice one @franzwilding - thank you! :+1:

PR #718, opened 2025-10-02
[Platform] Support boolean string conversion in `AbstractModelCatalog::parseModelName()`

### LINKED ISSUE #713: [Platform] Boolean options not supported in AbstractModelCatalog::parseModelName()
Currently, `AbstractModelCatalog::parseModelName()` relies on `AbstractModelCatalog::convertNumericStrings()` which only converts numeric strings to integers/floats.

However, boolean-like query parameters (e.g. `?think=false`) are always parsed as strings. This makes it harder to pass boolean options to models, since they need to be manually casted.

**Example:**

```php
$catalog = new \Symfony\AI\Platform\Bridge\Ollama\ModelCatalog();
$model = $catalog->getModel('qwen3:0.6b?think=false');
$client = new \Symfony\AI\Platform\Bridge\Ollama\OllamaClient(
    Symfony\Component\HttpClient\HttpClient::create(),
    $_ENV['OLLAMA_HOST']
);
$options = $model->getOptions(); // ['think' => 'false']

// will throw an exception because ollama responds with 400 Bad request
// {"error":"invalid think value: \"false\" (must be \"high\", \"medium\", \"low\", true, or false)"}
$client->request(
    $model,
    [new \Symfony\AI\Platform\Message\UserMessage(
        new \Symfony\AI\Platform\Message\Content\Text('Hello world'),
    )],
    $model->getOptions()
);
```

Here `think=false` is passed as a string (`'false'`), but Ollama server expects a real boolean `false`.

## Expected behavior:
Boolean-like strings should be automatically converted to booleans.

## Proposed solution:
Introduce a more general converter, e.g. `convertScalarStrings()`, that handles both numeric and boolean strings.

sonnymilton:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Docs?         | no 
| Issues        | Fix #713  
| License       | MIT

This PR extends the query parameter normalization logic in `AbstractModelCatalog` to handle boolean-like strings (`"true"` / `"false"`) in addition to numeric strings:

- Query parameters such as `?think=true` or `?think=false` are now correctly converted to PHP booleans instead of remaining plain strings.  
- Conversion is applied recursively, so nested arrays produced by `parse_str()` also normalize their boolean values.  
- Numeric string handling is preserved as before.  

### Tests

New test cases were added to ensure:

- Boolean string parameters are converted properly (`true`/`false`).  
- Recursive conversion inside nested query arrays works as expected.  
- Mixed parameters with numbers, booleans, and plain strings remain consistent.

### DIFF
--- a/src/platform/src/ModelCatalog/AbstractModelCatalog.php
+++ b/src/platform/src/ModelCatalog/AbstractModelCatalog.php
@@ -86,7 +86,7 @@ abstract class AbstractModelCatalog implements ModelCatalogInterface
 
             parse_str($queryString, $options);
 
-            $options = self::convertNumericStrings($options);
+            $options = self::convertScalarStrings($options);
         }
 
         // Determine catalog key: try exact match first, then fall back to base model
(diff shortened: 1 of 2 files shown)


OskarStark:
Thank you @sonnymilton for taking care of this regression.

PR #719, opened 2025-10-02
[AI Bundle] Preserve boolean options in agent/vectorizer model confguration

sonnymilton:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Docs?         | no 
| Issues        | - 
| License       | MIT

## Problem
When configuring models in `symfony/ai-bundle` via the `options` section, boolean values were **misrepresented**.  
This happens because `http_build_query()` converts booleans into integers:
- `true` → `1`  
- `false` → `0`

So configuration like:

```yaml
ai:
  agent:
    my_agent:
      model:
        name: gpt-4o-mini
        options:
          stream: false
```

Converted to `gpt-4o-mini?stream=0`
Expected: `gpt-4o-mini?stream=false`

This makes it impossible to distinguish between a booleans and an integers during parsing.

## Changes
 - Inline normalization of options before calling `http_build_query()`. Booleans are explicitly converted to the strings `"true"` / `"false"`.
- Added regression tests for both agents and vectorizers.


OskarStark:
Good catch, thanks @sonnymilton.

PR #720, opened 2025-10-02
[Agent][Platform] Improve `ResultConverter`PHPDoc

VincentLanglet:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Docs?         | yes
| Issues        | Fix #...
| License       | MIT

Since https://github.com/symfony/ai/pull/538 exceptions thrown by `ResultConverterInterface::convert` are more useful and should be documented.
I personally use them when doing `Agent::call`, and they are accessible because
Agent::call use `getResult` which use `await` which use `convert`.

I added some basic documentation about them then.


OskarStark:
Thank you @VincentLanglet.

PR #723, opened 2025-10-02
Use int for retry after

VincentLanglet:
| Q             | A
| ------------- | ---
| Bug fix?      | not really
| New feature?  | no
| Docs?         | no 
| Issues        | Fix #... 
| License       | MIT

According to RFC 7231 should be an integer
https://datatracker.ietf.org/doc/html/rfc7231#page-69

So let's avoid a useless float here.


OskarStark:
Thank you Vincent.

PR #726, opened 2025-10-04
[AI Bundle] Update documentation for indexer

jarek-lach:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | yes
| License       | MIT

Added example loader class to indexer configuration because it's required and without it indexer could not be instantiated


chr-hertel:
Thank you @jarek-lach.

PR #728, opened 2025-10-05
[Agent] Hardening & Slim down of Agent

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | 
| License       | MIT

Basically hardening the agent by
* getting rid of public properties in favor of getter/setter
* removing unnecessary error handling
* removing capability checks

### DIFF
--- a/examples/aimlapi/toolcall.php
+++ b/examples/aimlapi/toolcall.php
@@ -24,7 +24,7 @@ $platform = PlatformFactory::create(env('AIMLAPI_API_KEY'), http_client());
 $wikipedia = new Wikipedia(http_client());
 $toolbox = new Toolbox([$wikipedia], logger: logger());
 $processor = new AgentProcessor($toolbox);
-$agent = new Agent($platform, 'google/gemini-2.5-flash', [$processor], [$processor], logger: logger());
+$agent = new Agent($platform, 'google/gemini-2.5-flash', [$processor], [$processor]);
 
 $messages = new MessageBag(Message::ofUser('Who is the current chancellor of Germany?'));
 $result = $agent->call($messages);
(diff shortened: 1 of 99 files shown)


OskarStark:
Thanks 🚀

VincentLanglet:
Hi @chr-hertel I'm curious, why did you remove the `Capability::OUTPUT_STRUCTURED` from the `StructuredOutput/AgentProcessor` ?

I would find useful to have the processor failing if the option is not supported. Especially to avoid a weird crash later.

chr-hertel:
@VincentLanglet Sure, i can share. If there's a weird crash later let's optimize for reducing the weirdness instead.

In general the take is that we can reduce some of our validations since they're only duplicates of what APIs would flag later as well. And worst case it could lead to false negative - PHP level exception even If the API would support that. We don't own model and API so this could easily diverge.

If we one day go more into the direction of capability based routing or some other logic, we can reintroduce similar logic, but for now it's not beneficial.

Were your code relying on that, did you have a use case or are we talking theory here?

VincentLanglet:
> Were your code relying on that, did you have a use case or are we talking theory here?

I would say it's more about helping the user.

I was using gemini-2.0 which seems to not (fully ?) support structured output even if I was trying to.
I end up with a MissingConstructor serializer error (cf https://github.com/symfony/ai/issues/786), which was really not easy to debug. Instead of having to debug my vendor and the stack trace, looking for docs, etc it would have been way more helpful having an InvalidArgumentException.

Also, a second reason for having the InvalidArgumentException is the fact it happens BEFORE the api call and not after. This way I avoid consuming PAID token for nothing.

chr-hertel:
Fair, if it's actually useful to you as user, we should revert

PR #732, opened 2025-10-06
[Store] Unify `Vectorizer` methods into single `vectorize()` method

OskarStark:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | --
| License       | MIT

Replace separate vectorizeTextDocuments() and vectorize(string) methods with a single vectorize() method that handles:
- string -> Vector
- array<string> -> array<Vector>
- TextDocument -> VectorDocument
- array<TextDocument> -> array<VectorDocument>

Add PHPStan conditional return types for proper type inference. Add validation to ensure arrays contain only strings or TextDocuments. Update all usages across codebase including tests and examples.

cc @paulinevos

### DIFF
--- a/src/store/src/Document/Vectorizer.php
+++ b/src/store/src/Document/Vectorizer.php
@@ -27,7 +27,129 @@ final readonly class Vectorizer implements VectorizerInterface
     ) {
     }
 
-    public function vectorizeEmbeddableDocuments(array $documents, array $options = []): array
+    public function vectorize(string|\Stringable|EmbeddableDocumentInterface|array $values, array $options = []): Vector|VectorDocument|array
+    {
+        if (\is_string($values) || $values instanceof \Stringable) {
+            return $this->vectorizeString($values, $options);
+        }
+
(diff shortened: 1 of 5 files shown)


src/store/src/Document/Vectorizer.php
```diff
@@ -27,7 +27,125 @@ public function __construct(
     ) {
     }
 
-    public function vectorizeTextDocuments(array $documents, array $options = []): array
+    public function vectorize(string|TextDocument|array $values, array $options = []): Vector|VectorDocument|array
```
paulinevos: Might it be an idea to base this on my branch and typehint `EmbeddableDocumentInterface`? Otherwise we'll have to change the types here later anyway.

paulinevos: We don't want to keep supporting `\Stringable`?

OskarStark: Nice catch, thanks

PR #734, opened 2025-10-06
[AI Bundle][Chat] Add lifecycle commands

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | Fix none
| License       | MIT

Hi 👋🏻 

This PR aims to introduce commands to handle message store lifecycle, it also introduce the configuration for message stores.


chr-hertel:
I feel like this is coming in early - is it motivated by an actual need of yours or rather a hypothesis?
(just a curious question :laughing: )

Guikingone:
It's the continuation of the `Chat` migration process, as stores can be initialized/dropped, we need to have commands to handle the booting phase of them, I also integrated the configuration to ease the usage of stores in current applications (until we merge the `Chat` refactoring), IMHO, merging commands before having "external" bridges is not an issue (Meilisearch should be ready later this week BTW).

And yes, I'm currently hitting a wall in a project that requires a message store 😅

OskarStark (approved):

OskarStark:
Thank you Guillaume.

PR #735, opened 2025-10-06
[Platform][Gemini][VertexAI] Handle `inlineData`

### LINKED ISSUE #730: Vertex AI - gemini-2.5-flash-image-preview intergration
Hi, while using the vertex platform, i want to use the gemini-2.5-flash-image-preview.

So i add it to the catalog, add the different capabilities, (set up the google cloud adc..).

When i want to fetch the result, the only way for me to get the image binary is to use the getRawResult method of the reponse.

In fact in Bridge/VertexAi/Gemini/ResultConverter.php::128 (convertChoice) cannot handle inline_data response as content part and cant return BinaryResult.

I managed to make it work easily adding the case 

            if (isset($contentPart['inlineData'])) {
                return new BinaryResult($contentPart['inlineData']['data'], $contentPart['inlineData']['mimeType'] ?? null);
            }

and adding BinaryResult as function return type.

See you, thanks for the work

lochmueller:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #730
| License       | MIT

Fix handling of inlineData as Binary response in ResultConverter of Google Gemini.

### DIFF
--- a/src/platform/src/Bridge/Gemini/Gemini/ResultConverter.php
+++ b/src/platform/src/Bridge/Gemini/Gemini/ResultConverter.php
@@ -15,6 +15,7 @@ use Symfony\AI\Platform\Bridge\Gemini\Gemini;
 use Symfony\AI\Platform\Exception\RateLimitExceededException;
 use Symfony\AI\Platform\Exception\RuntimeException;
 use Symfony\AI\Platform\Model;
+use Symfony\AI\Platform\Result\BinaryResult;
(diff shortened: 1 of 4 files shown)


OskarStark:
Looks good to me, can you please add tests?

OskarStark:
Thanks for fixing this bug Tim, I pushed some tests

PR #745, opened 2025-10-06
[Platform] Rename to `DeferredResult` and explicit contract instead of closure

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | 
| License       | MIT

* Renaming `ResultPromise` to `DeferredResult` since the interface is not really promise-like
* Replacing the `\Closure` with `ResultConverterInterface` in constructor

Alternative names
* `ResultFactory` => is it a factory?
* `ResultBuilder` => is it a builder?
* `LazyResult`  => would expect the same interface as the Result
* `ResultProxy` => would expect the same interface as the Result
* `WrappedResult` => i like wraps
* `ResultContainer` => might confuse with _the_ container
* `ResultHandle` => ?

### DIFF
--- a/examples/bootstrap.php
+++ b/examples/bootstrap.php
@@ -15,7 +15,7 @@ use Symfony\AI\Agent\Exception\ExceptionInterface as AgentException;
 use Symfony\AI\Platform\Exception\ExceptionInterface as PlatformException;
 use Symfony\AI\Platform\Metadata\Metadata;
 use Symfony\AI\Platform\Metadata\TokenUsage;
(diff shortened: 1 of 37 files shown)


OskarStark:
ResultHandle is also nice

lyrixx (approved):
LGTM 👍🏼 

Thanks

OskarStark (approved):

OskarStark:
Thank you @chr-hertel.

PR #754, opened 2025-10-07
[Agent][Platform][OpenAI] Add stream usage support

### LINKED ISSUE #753: [Agent][Platform][OpenAI] Add support to include usage on streams
Hi !  

OpenAI GPT added a functionality to receive the usage token using stream https://platform.openai.com/docs/api-reference/chat/create#chat-create-stream_options-include_usage.

The current implementation just ignore these messages as is not part of the delta content.  https://github.com/symfony/ai/blob/2c4d20d7b2865e0772ce03069542fbb9e0da478b/src/platform/src/Bridge/OpenAi/Gpt/ResultConverter.php#L110

I’ve been working on an [implementation](https://github.com/symfony/ai/pull/754) to parse and return the usage object through the generator, but I’m not really sure if this is the best approach. Alternatively, I could consider using an event listener or adding a usage property in the StreamResult that would be accessible after the stream has been consumed.

So far, this is how the PR looks in action:

```php
$options = [
    'stream' => true,
    'stream_options' => [
        'include_usage' => true,
    ],
];

$stream = $agent->call($messages, $options)->getContent();

if (!is_iterable($stream)) {
    throw new \RuntimeException('Invalid result type, expected iterable');
}

foreach ($stream as $chunk) {
    if (is_string($chunk)) {
        $context->setDelta($chunk);
        continue;
    }

    if ($chunk instanceof TokenUsage) {
        $context->addUsedTokens(
            $chunk->totalTokens ?: 0,
            $chunk->cachedTokens ?: 0
        );
    }
}
```

I'll be so glad to know your opinion about these approaches or any other alternatives.

Thanks !

oesteve:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | Fix #753
| License       | MIT

```php

    'stream' => true,
]);

foreach ($result->getContent() as $textChunk) {
    echo $textChunk;
}

print_token_usage($result->getMetadata()->get('token_usage'));
```

### DIFF
--- a/examples/openai/stream-token-metadata.php
+++ b/examples/openai/stream-token-metadata.php
@@ -1,0 +1,38 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use Symfony\AI\Agent\Agent;
+use Symfony\AI\Platform\Bridge\OpenAi\PlatformFactory;
+use Symfony\AI\Platform\Message\Message;
+use Symfony\AI\Platform\Message\MessageBag;
+
+require_once dirname(__DIR__).'/bootstrap.php';
+
+$platform = PlatformFactory::create(env('OPENAI_API_KEY'), http_client());
+
+$agent = new Agent($platform, 'gpt-4o-mini');
+$messages = new MessageBag(
+    Message::forSystem('You are a pirate and you write funny.'),
+    Message::ofUser('What is the Symfony framework?'),
+);
+$result = $agent->call($messages, [
+    'stream' => true,
+]);
+
+foreach ($result->getContent() as $textChunk) {
+    echo $textChunk;
+}
+
+echo \PHP_EOL;
+
+print_token_usage($result->getMetadata()->get('token_usage'));
+
+echo \PHP_EOL;
--- a/src/agent/src/Toolbox/AgentProcessor.php
+++ b/src/agent/src/Toolbox/AgentProcessor.php
@@ -81,7 +81,7 @@ final class AgentProcessor implements InputProcessorInterface, OutputProcessorIn
         if ($result instanceof GenericStreamResponse) {
             $output->setResult(
                 new ToolboxStreamResponse(
-                    $result->getContent(),
+                    $result,
                     fn (ToolCallResult $result, ?AssistantMessage $streamedAssistantResponse = null) => $this->handleToolCallsCallback($output, $result, $streamedAssistantResponse)
                 )
             );
--- a/src/agent/src/Toolbox/StreamResult.php
+++ b/src/agent/src/Toolbox/StreamResult.php
@@ -12,8 +12,12 @@
 namespace Symfony\AI\Agent\Toolbox;
 
 use Symfony\AI\Platform\Message\Message;
+use Symfony\AI\Platform\Metadata\Metadata;
 use Symfony\AI\Platform\Result\BaseResult;
+use Symfony\AI\Platform\Result\StreamResult as PlatformStreamResult;
 use Symfony\AI\Platform\Result\ToolCallResult;
+use Symfony\AI\Platform\TokenUsage\TokenUsageAggregation;
+use Symfony\AI\Platform\TokenUsage\TokenUsageInterface;
 
 /**
  * @author Denis Zunke <denis.zunke@gmail.com>
@@ -21,7 +25,7 @@ use Symfony\AI\Platform\Result\ToolCallResult;
 final class StreamResult extends BaseResult
 {
     public function __construct(
-        private readonly \Generator $generator,
+        private readonly PlatformStreamResult $streamResult,
         private readonly \Closure $handleToolCallsCallback,
     ) {
     }
@@ -29,14 +33,9 @@ final class StreamResult extends BaseResult
     public function getContent(): \Generator
     {
         $streamedResult = '';
-        foreach ($this->generator as $value) {
+        foreach ($this->streamResult->getContent() as $value) {
             if ($value instanceof ToolCallResult) {
                 $innerResult = ($this->handleToolCallsCallback)($value, Message::ofAssistant($streamedResult));
-
-                // Propagate metadata from inner result to this result
-                foreach ($innerResult->getMetadata()->all() as $key => $metadataValue) {
-                    $this->getMetadata()->add($key, $metadataValue);
-                }
 
                 $content = $innerResult->getContent();
                 // Strings are iterable in PHP but yield from would iterate character-by-character.
@@ -48,12 +47,29 @@ final class StreamResult extends BaseResult
                     yield from $content;
                 }
 
+                // Propagate metadata from inner result to this result
+                $this->propagateMetadata($innerResult->getMetadata());
+
                 break;
             }
 
             $streamedResult .= $value;
 
             yield $value;
+        }
+
+        $this->propagateMetadata($this->streamResult->getMetadata());
+    }
+
+    private function propagateMetadata(Metadata $source): void
+    {
+        foreach ($source->all() as $key => $value) {
+            if ('token_usage' === $key && $this->getMetadata()->get('token_usage') instanceof TokenUsageInterface && $value instanceof TokenUsageInterface) {
+                $this->getMetadata()->add('token_usage', new TokenUsageAggregation($this->getMetadata()->get('token_usage'), $value));
+                continue;
+            }
+
+            $this->getMetadata()->add($key, $value);
         }
     }
 }
--- a/src/platform/src/Bridge/OpenAi/Gpt/ResultConverter.php
+++ b/src/platform/src/Bridge/OpenAi/Gpt/ResultConverter.php
@@ -27,6 +27,7 @@ use Symfony\AI\Platform\Result\TextResult;
 use Symfony\AI\Platform\Result\ToolCall;
 use Symfony\AI\Platform\Result\ToolCallResult;
 use Symfony\AI\Platform\ResultConverterInterface;
+use Symfony\AI\Platform\TokenUsage\TokenUsage;
 
 /**
  * @author Christopher Hertel <mail@christopher-hertel.de>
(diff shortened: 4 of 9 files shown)


examples/openai/toolcall-stream-with-usage.php
```diff
@@ -0,0 +1,65 @@
...
+
+// Output text chunks
+/** @var TextChunk $textChunk */
+foreach ($result->getContent() as $textChunk) {
+
+    // $textChunk implement \Stringable
+    echo $textChunk->getContent();
+
+    // Chunk also contain metadata
+    // $textChunk->getMetadata()->get('id'); // Stream id
```
OskarStark: remove?

examples/openai/toolcall-stream-with-usage.php
```diff
@@ -0,0 +1,65 @@
...
+    'stream_options' => [
+        'include_usage' => true, // include usage in the response
+    ],
+]);
+
+// Output text chunks
+/** @var TextChunk $textChunk */
+foreach ($result->getContent() as $textChunk) {
+
+    // $textChunk implement \Stringable
```
OskarStark: remove

examples/openai/toolcall-stream-with-usage.php
```diff
@@ -0,0 +1,65 @@
...
+
+$clock = new Clock();
+$openMeteo = new OpenMeteo(http_client());
+$toolbox = new Toolbox([$clock, $openMeteo], logger: logger());
+$processor = new AgentProcessor($toolbox);
+
+$tokenOutputProcessor = new TokenOutputProcessor();
+
+$agent = new Agent($platform, 'gpt-4o-mini', [$processor], [$processor]);
+$messages = new MessageBag(Message::ofUser(<<<TXT
```
OskarStark: why a multiline text here?

examples/openai/toolcall-stream-with-usage.php
```diff
@@ -0,0 +1,65 @@
...
+    Tell me the time and the weather in Dublin.
+    TXT));
+$result = $agent->call($messages, [
+    'stream' => true, // enable streaming of response text
+    'stream_options' => [
+        'include_usage' => true, // include usage in the response
+    ],
+]);
+
+// Output text chunks
```
OskarStark:
```suggestion

```

src/agent/src/Toolbox/AgentProcessor.php
```diff
@@ -76,19 +76,24 @@ public function processInput(Input $input): void
...
-        if ($output->getResult() instanceof GenericStreamResponse) {
+        $result = $output->getResult();
+
+        if ($result instanceof GenericStreamResponse) {
             $output->setResult(
-                new ToolboxStreamResponse($output->getResult()->getContent(), $this->handleToolCallsCallback($output))
+                new ToolboxStreamResponse(
+                    $result,
+                    $this->handleToolCallsCallback($output)
+                )
```
OskarStark, line 86: lets keep one line

examples/openai/toolcall-stream-with-usage.php
```diff
@@ -0,0 +1,65 @@
...
+require_once dirname(__DIR__).'/bootstrap.php';
+
+$platform = PlatformFactory::create(env('OPENAI_API_KEY'), http_client());
+
+$clock = new Clock();
+$openMeteo = new OpenMeteo(http_client());
+$toolbox = new Toolbox([$clock, $openMeteo], logger: logger());
+$processor = new AgentProcessor($toolbox);
+
+$tokenOutputProcessor = new TokenOutputProcessor();
```
OskarStark: is it used?

oesteve: It was not

examples/openai/toolcall-stream-with-usage.php
oesteve: It made sense at some point, but it doesn't anymore.

oesteve:
@chr-hertel , thanks for the feedback. Your approach with `TokenUsageAggregation` looks much cleaner and allows for the reuse of `token_usage`. 

On the other side, the `TokenOutputProcessor` works well with regular results, but I can’t see how to make an `OutputProcessorInterface` work with streams. The usage is received at the end of the stream after the `processOutput` has been called.

chr-hertel:
Yeah, makes sense - that looks tricky - you needed to do the metadata handling at various places.
Can we go for some kind of middle way tho - maybe skipping the processor but dropping that extra option (is it interpreted anyways?) and provide an aggregated object as `token_usage` on high level for the devs?

OskarStark:
open to finish this PR @oesteve ?

oesteve:
@OskarStark  Yes, I'd like to get back to this. I'll have some free time this season to make the changes. If it's a priority for you and you want to open an alternative PR, that's fine to me.

oesteve:
@OskarStark @chr-hertel Finally, I had some time to come back to this. I've practically rewritten the PR. I hope this approach is interesting for you. What do you think?

I also have some doubts about whether the example provides real value or just noise, since the approach is quite similar for the non-stream calls.

chr-hertel (changes requested):
Looks like with `.ai` and `.idea` some unintended folders got commited with rebase

examples/openai/stream-token-metadata.php
chr-hertel, line 1: Works well :+1:

src/agent/src/Toolbox/AgentProcessor.php
```diff
@@ -79,12 +79,7 @@ public function processOutput(Output $output): void
         $result = $output->getResult();
 
         if ($result instanceof GenericStreamResponse) {
-            $output->setResult(
-                new ToolboxStreamResponse(
-                    $result->getContent(),
-                    fn (ToolCallResult $result, ?AssistantMessage $streamedAssistantResponse = null) => $this->handleToolCallsCallback($output, $result, $streamedAssistantResponse)
-                )
-            );
+            $output->setResult(new ToolboxStreamResponse($result, fn (ToolCallResult $result, ?AssistantMessage $streamedAssistantResponse = null) => $this->handleToolCallsCallback($output, $result, $streamedAssistantResponse)));
```
chr-hertel: change is unrelated, right? in that case please revert to keep the diff small :+1:

src/platform/src/Bridge/OpenAi/Gpt/ResultConverter.php
```diff
@@ -71,7 +72,9 @@ public function convert(RawResultInterface|RawHttpResult $result, array $options
         }
 
         if ($options['stream'] ?? false) {
-            return new StreamResult($this->convertStream($result));
+            $streamResult = new StreamResult($this->convertStream($result, $streamResult));
+
+            return $streamResult;
```
chr-hertel, line 78: that kind of binding is pretty much mind bending, but i currently have no better idea

chr-hertel: we should at least comment what is happening here

src/agent/src/Toolbox/StreamResult.php
chr-hertel, line 1: i wonder if this is too late and should happen directly after conversion - we won't propagate the metadata for usage without agent and toolbox then, right?

chr-hertel (changes requested):
Thanks for wrapping your head around this - though nut :pray: - left some comments, but this is on a good track, thanks again!

src/agent/src/Toolbox/AgentProcessor.php
oesteve: Subtle, but it is. Maybe I didn’t fully catch @OskarStark 's [comment](https://github.com/symfony/ai/pull/754#discussion_r2573126312). 

I kept the previous indention.

src/platform/src/Bridge/OpenAi/Gpt/ResultConverter.php
oesteve: Yes it really is. Another approach that came to my mind was to yield the TokenUsage and propagate to the metadata in the StreamResult

```php
namespace Symfony\AI\Platform\Result;

final class StreamResult extends BaseResult
{
    public function getContent(): \Generator
    {
        foreach ($this->generator as $chunk) {
            if ($chunk instanceof TokenUsageInterface) {
                $this->getMetadata()->add('token_usage', $chunk);
                continue;
            }

            yield $chunk;
        }
    }
}
```

That makes this approach available for all providers, but honestly, I don't know if this is a good or bad approach.

src/agent/src/Toolbox/StreamResult.php
oesteve: I'm not sure I understood your comment. This is the earliest point I found at the Agent level. Do you have a better approach in mind?

At the platform level, the result includes the metadata along with the propagated usage.

https://github.com/oesteve/ai/blob/0497a1951f0428ff7f951b4469a3175bbe5ac70a/src/platform/src/Bridge/OpenAi/Tests/Gpt/ResultConverterTest.php#L222

oesteve:
Thank you for your comments, being able to contribute is truly inspiring 🙂

chr-hertel:
Hi @oesteve, your work, especially the your comment about handling the cases while streaming, solved it for me:

this snippet:
```php
        foreach ($this->generator as $chunk) {
            if ($chunk instanceof TokenUsageInterface) {
                $this->getMetadata()->add('token_usage', $chunk);
                continue;
            }

            yield $chunk;
        }
```

in the end became: https://github.com/oesteve/ai/compare/main...chr-hertel:ai:stream-listener
(not completely done yet, but i'm quite happy how that transforms the issue, and makes it easier managable for tokens and toolcalls)

---

So much about "truly inspiring" - thanks for that! :pray:

oesteve:
@chr-hertel, that looks really promising. An event-driven interface add more flexiblility from my point of view.

It would be great to have access from the Agent layer to add a listener and receive Platform events. 

I'm not really sure in which direction the Agent component is going ... but implementing certain uncommon use cases like this requires forking and sending a PR. Access to these low-level events enables external implementation and simplifies the Agent logic.

One last thought, could it make sense to be close to symfony/event-dispatcher-contracts?

chr-hertel:
@oesteve yea, thought about the EventDispatcher interface, but was a bit reluctant - injecting the dispatcher to the stream would also add a larger service dependency. maybe it's a rather small step, but i feels dangerous to enable easy container integration since the stream is something quite easy to break and performance sensitive ... but maybe that's only a documentation concern and provide users with the freedom to shoot themselves in the foot :grimacing: 

anyhow, that could be an easy follow-up. will wrap that up, merge your work and mine as follow up :+1:

chr-hertel:
Thank you @oesteve.

Issue #755, opened 2025-10-08
[MCP Bundle] Pattern support decision + ServiceLocator bug


camilleislasse:
### Description

Two issues with MCP tool dependency injection:

1. **Design decision needed**: Bundle currently only supports invokable pattern, but SDK supports both patterns (invokable + methods).
2. **Bug**: `McpPass::process()` passes tag arrays instead of References to `ServiceLocatorTagPass`

### Current Situation

**MCP SDK supports two patterns:**

1. **Invokable**: `#[McpTool]` on class + `__invoke()` method
2. **Methods**: `#[McpTool]` on public methods

**MCP Bundle currently:**
- Uses `registerAttributeForAutoconfiguration()`
- Only detects class-level attributes (invokable pattern)
- Method pattern not supported with auto-configuration
- Tools with dependencies crash due to ServiceLocator bug

### Pattern Support Decision

**Option 1: Keep invokable only**

Uses `registerAttributeForAutoconfiguration()` which only detects class-level attributes. Users need to use `#[McpTool]` on class with `__invoke()` method.

**Option 2: Support both patterns**

Find a way to detect and auto-tag methods with `#[McpTool]` attributes. Would provide full compatibility with SDK examples and support multiple tools per class.

### ServiceLocator Bug (must fix regardless)

**Current code in `McpPass::process()` (line 36):**

```php
public function process(ContainerBuilder $container): void
{
    // ... collect services ...

    $allMcpServices = [
        'App\MCP\Tools\WeatherTool' => [['priority' => 0]], // Tag metadata array
        // ...
    ];

    // BUG: Passes tag arrays instead of References
    $serviceLocatorRef = ServiceLocatorTagPass::register($container, $allMcpServices);
}
```

**What `ServiceLocatorTagPass::register()` expects:**

```php
[
    'App\MCP\Tools\WeatherTool' => new Reference('App\MCP\Tools\WeatherTool'),
    // ...
]
```

**Fix (required for invokable pattern to work):**

```php
// src/mcp-bundle/src/DependencyInjection/McpPass.php
use Symfony\Component\DependencyInjection\Reference; // Add import

public function process(ContainerBuilder $container): void
{
    if (!$container->hasDefinition('mcp.server.builder')) {
        return;
    }

    $allMcpServices = [];
    $mcpTags = ['mcp.tool', 'mcp.prompt', 'mcp.resource', 'mcp.resource_template'];

    foreach ($mcpTags as $tag) {
[...]

PR #757, opened 2025-10-08
[MCP Bundle] Fix dependency injection and method-based / invokable support

camilleislasse:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Docs?         | yes
| Issues        | https://github.com/symfony/ai/issues/755
| License       | MIT

** [MCP Bundle]  Fix dependency injection and method-based / invokable support**

  This commit fixes critical issues preventing MCP tools with constructor
  dependencies from working and enables method-based attributes support.

  Changes:
  - Fix McpPass to use Reference objects for ServiceLocator registration
    Previously passed tag arrays directly, causing crashes when tools had
    constructor dependencies

  - Fix McpBundle::registerMcpAttributes() closure signature
    Added missing object $attribute and \Reflector $reflector parameters.
    Without these parameters, Symfony's AttributeAutoconfigurationPass only
    scans class-level attributes and skips method-level ones entirely

  - Add LoggerInterface to CurrentTimeTool to demonstrate DI works

  - Update documentation to reflect both patterns are supported
    Added examples showing invokable and method-based patterns both work
    with full DI support

  - Improve McpPassTest to verify ServiceLocator uses References
    Previous tests only checked presence, not type of values

### DIFF
--- a/docs/bundles/mcp-bundle.rst
+++ b/docs/bundles/mcp-bundle.rst
@@ -113,6 +113,33 @@ Dynamic resources with parameters:
 
 All capabilities are automatically discovered in the ``src/`` directory when the server starts.
 
+Attribute Placement Patterns
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The MCP SDK, and therefore the MCP Bundle, supports two patterns for placing attributes on your capabilities:
+
+**Invokable Pattern** - Attribute on a class with ``__invoke()`` method::
+
+    #[McpTool(name: 'my-tool')]
+    class MyTool
+    {
+        public function __invoke(string $param): string
+        {
+            // Implementation
+        }
+    }
+
+**Method-Based Pattern** - Multiple attributes on individual methods::
+
+    class MyTools
+    {
+        #[McpTool(name: 'tool-one')]
+        public function toolOne(): string { }
+
+        #[McpTool(name: 'tool-two')]
+        public function toolTwo(): string { }
+    }
+
 Transport Types
 ...............
 
--- a/demo/src/Mcp/Tools/CurrentTimeTool.php
+++ b/demo/src/Mcp/Tools/CurrentTimeTool.php
@@ -12,12 +12,18 @@
 namespace App\Mcp\Tools;
 
 use Mcp\Capability\Attribute\McpTool;
+use Psr\Log\LoggerInterface;
 
 /**
  * @author Tom Hart <tom.hart.221@gmail.com>
  */
 class CurrentTimeTool
 {
+    public function __construct(
+        private readonly LoggerInterface $logger,
+    ) {
+    }
+
     /**
      * Returns the current time in UTC.
      *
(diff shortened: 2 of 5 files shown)


src/mcp-bundle/config/services.php
```diff
@@ -21,8 +21,8 @@
             ->args(['mcp'])
             ->tag('monolog.logger', ['channel' => 'mcp'])
 
-        ->set('mcp.server.builder', Builder::class)
-            ->factory([Server::class, 'builder'])
+        ->set('mcp.server.builder', ServerBuilder::class)
+            ->factory([Server::class, 'make'])
```
chr-hertel: this is basically reverting #697 - could be an outdated mcp/sdk version or branched off an old main?

chr-hertel:
Hi @camilleislasse,

thanks for working on this - is the best next step! :+1: 

Can you add your findings about when `#[Mcp*]` work on methods vs. classes with the bundle to the docs? i think for now that should be a warning there.

camilleislasse:
> Hi @camilleislasse,
> 
> thanks for working on this - is the best next step! 👍
> 
> Can you add your findings about when `#[Mcp*]` work on methods vs. classes with the bundle to the docs? i think for now that should be a warning there.

https://github.com/symfony/ai/pull/757/commits/80da7cbb3e6a2e28b01282b34e8569c5369edada

OskarStark:
I get the note, but couldn't we make it work to support both?

camilleislasse:
> I get the note, but couldn't we make it work to support both?

I have find a way :

By changing the closure signature from:
```
static function (ChildDefinition $definition) use ($tag): void {
```
To:
```
static function (ChildDefinition $definition, object $attribute, \Reflector $reflector) use ($tag): void {
```
Symfony now scans both class-level attributes (invokable pattern) and method-level attributes (method-based pattern). The `\Reflector` parameter tells Symfony to also check  methods, properties, and parameters - not just classes.

chr-hertel:
That's great news - thanks @camilleislasse!

docs/bundles/mcp-bundle.rst
```diff
@@ -113,6 +113,33 @@ Dynamic resources with parameters:
...
+
+        class MyTools
+        {
+            #[McpTool(name: 'tool-one')]
+            public function toolOne(): string { }
+
+            #[McpTool(name: 'tool-two')]
+            public function toolTwo(): string { }
+        }
+
```
chr-hertel: This should be a regular section like "Defining Capabilities" or "Attribute Usage" or other, instead of a `warning`

chr-hertel:
does it work for you though?
i currently get this error while running the branch locally with inspector:
image

camilleislasse:
> i currently get this error while running the branch locally with inspector:



> does it work for you though? i currently get this error while running the branch locally with inspector: image

Yes working, I look 

Capture d’écran 2025-10-15 à 10 04 46

1. Do you have the MonologBundle installed?
2. Did you run composer install and cache:clear after pulling the branch?

chr-hertel (approved):
Needed to rebase the branch for making it work - but it works :+1: 

Thanks @camilleislasse!

camilleislasse:
> Needed to rebase the branch for making it work - but it works 👍
> 
> Thanks @camilleislasse!

Cool thank's 👍

PR #764, opened 2025-10-09
[Agent] Enable tools to add sources

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | 
| License       | MIT

image

- [x] tests
- [x] provide way without keep messages

### DIFF
--- a/src/agent/src/Toolbox/AgentProcessor.php
+++ b/src/agent/src/Toolbox/AgentProcessor.php
@@ -18,6 +18,7 @@ use Symfony\AI\Agent\InputProcessorInterface;
 use Symfony\AI\Agent\Output;
 use Symfony\AI\Agent\OutputProcessorInterface;
 use Symfony\AI\Agent\Toolbox\Event\ToolCallsExecuted;
+use Symfony\AI\Agent\Toolbox\Source\Source;
 use Symfony\AI\Agent\Toolbox\StreamResult as ToolboxStreamResponse;
 use Symfony\AI\Platform\Message\AssistantMessage;
 use Symfony\AI\Platform\Message\Message;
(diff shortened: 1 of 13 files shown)


src/agent/src/Toolbox/Source/UsesSourcesInterface.php
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Agent\Toolbox\Source;
+
+interface UsesSourcesInterface
```
OskarStark: SourcesAwareInterface?

chr-hertel:
not yet happy how sources are transported to the surface.
i like the slim API on tool level, but the other side is not that great with `keepToolMessages` and double-foreach

thinking of shifting the addition of `AssistantMessage` to the `MessageBag` into the `Agent` instead of user land code

src/agent/src/Toolbox/Source/UsesSourcesInterface.php
chr-hertel: that somehow inverses the relation. the service is not aware of an external thing, but exposes or uses sources

OskarStark: HasSources?

OskarStark: ProvidesSources?

src/agent/src/Toolbox/Source/Source.php
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Agent\Toolbox\Source;
+
+readonly class Source
+{
+    public function __construct(
```
OskarStark, line 16: Are all of them required?

chr-hertel: by default i don't use nullable or defaults. when there's a reason to lower that, we can do so.

so my question would be: is one of them not required? :D

OskarStark: fine for me :-)

OskarStark:
Thank you Christopher.

src/agent/src/Toolbox/AgentProcessor.php
```diff
@@ -109,6 +126,12 @@ private function handleToolCallsCallback(Output $output): \Closure
                 $result = $event->hasResult() ? $event->getResult() : $this->agent->call($messages, $output->getOptions());
             } while ($result instanceof ToolCallResult);
 
+            --$this->nestingLevel;
+            if ($this->keepToolSources && 0 === $this->nestingLevel) {
+                $result->getMetadata()->add('sources', $this->sources);
+                $this->sources = [];
```
stof, line 132: shouldn't the resetting be done even when `keepToolSources` is false, to free memory (and avoid leaking them to the next call) ?

PR #766, opened 2025-10-09
[Platform][Ollama] Introduce `OllamaApiCatalog`

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| Docs?         | yes
| Issues        | None
| License       | MIT

Hi 👋🏻 

It looks like when we're dealing with platforms that can list / detain multiple models at once, the default catalog is "a fixed one", when trying to add a `FallbackModelCatalog`, the platform fails to retrieve the model as most of them use a custom `Model` class, this PR aims to introduce / fix the issue with an extra argument.

I know that the solution is not perfect, my main concern is the problem we're facing as we have custom models classes, that's the only way I found to prevent the error:

Capture d’écran 2025-10-09 à 17 52 28

This approach allows to use any model stored in the platform (especially as the ones listed in the default catalog are not up-to-date).


Guikingone:
Another approach would be to define a catalog for each platform when needed (Ollama requires a custom one, it's probably the only one with ElevenLabs) and inject it as the default implementation for the platform.

OskarStark:
I think it's better to open up the FallbackModelCatalog to be able to receive additional models configured from the outside (like for the real ModelCatalog's), check these first and if not go with the Model::class as default. WDYT?

Guikingone:
Not sure about the implementation but I get the idea of accepting additional models 🤔 

My main concern is about the capability to add those models at runtime, my issue here is that if I'm using models stored in Ollama but not in the catalog (so I can't configure them / add them in the catalog directly), the catalog is returning a `Model` rather than `Ollama`, of course, it triggers an error later when trying to see if the model is supported 😅

Guikingone:
Another approach would be to build the catalog from the Ollama API directly, something similar to what we're doing in the `OllamaClient`, the API already know if the model is supported, IMHO, listing models in the class is a bad idea at the end 🤔

OskarStark:
Yes, but not all models are available and in the end you cannot be up to date in the end always, so we need to have the option to be able to provide custom models. Take Azure for example, you need to deploy some models, so no model catalog for azure will look the same for a user...

Guikingone:
What if we're building the catalog directly from the API for specific platform?

What I mean is that in the catalog, we can query the API (when available) and build the model list from it then if a list of models is provided in the constructor, we can merge both at the end of the method, WDYT? 🤔

OskarStark:
I am not convinced about doing this on runtime, it can be costly, we would need to introduce caches etc. WDYT @chr-hertel

Guikingone:
I pushed a POC, the idea is to move the code that we already have in the `OllamaClient` to the `OllamaCatalog`, this way, the model is built once then you can query the client as many times as needed.

src/platform/src/Bridge/Ollama/OllamaCatalog.php
```diff
@@ -0,0 +1,39 @@
...
+
+    public function getModel(string $modelName): Model
+    {
+        $response = $this->httpClient->request('POST', \sprintf('%s/api/show', $this->host), [
+            'json' => [
+                'model' => $modelName,
+            ],
+        ]);
+
+        $payload = $response->toArray();
```
OskarStark: now you would need to implement the option handling etc here as well

Guikingone: Hum, maybe we can do `parent::getModel()` to retrieve an object with its options then transform it into the correct model 🤔

Guikingone: Something like: 

```php
    public function getModel(string $modelName): Model
    {
        $model = parent::getModel($modelName);

        $response = $this->httpClient->request('POST', \sprintf('%s/api/show', $this->host), [
            'json' => [
                'model' => $model->getName(),
            ],
        ]);

        $payload = $response->toArray();

        if ([] === $payload['capabilities'] ?? []) {
            throw new InvalidArgumentException('The model information could not be retrieved from the Ollama API. Your Ollama server might be too old. Try upgrade it.');
        }

        return new Ollama($modelName, $payload['capabilities']);
    }
```

chr-hertel:
> I am not convinced about doing this on runtime, it can be costly, we would need to introduce caches etc. WDYT @chr-hertel

I don't think it's an issue here. The first call is a fraction of the time compared to the main call later, and we're talking local execution only anyways.

but i agree that at least the `OllamaCatalog` should have a local cache var so the single instance of that object doesn't need to do it twice - an optional psr cache would be sugar on top but also introduces new questions like cachekey&invalidation after download of a new model etc

Guikingone:
@chr-hertel What about using the `Cache` component? Is there any real issue with it? I mean, we can make it optional (along with a TTL), if no cache is provided, the calls are performed each time and if provided, we store it for 10 minutes (or an hour?) 🤔

chr-hertel:
same like PSR cache. sugar on top & introducing new concerns. not needed from my side for a first PR

OskarStark:
For reference:
* #777 

I already updated the PR header and title with the new term `FallbackModelCatalog`.

I will come up with a proposal shortly

OskarStark:
Sorry, when checking the code, Ollama has its own ModelCatalog, where you can configure the models you need. I don't see the problem here, why would you want to use FallbackModelCatalog?

Guikingone:
@OskarStark 👋🏻 

I agree on the presence of a `ModelCatalog`, the problem is not "totally there", you can configure new models when using `Ollama` but it requires either to extends the class / decorate it, something not "developer friendly" IMHO when you're using multiple models / tags for a specific models and you don't want to configure again each time, the benefits of using a `FallbackModelCatalog` along with the retrieval of the model from the API is that you can tag a new model, deploy it and use it "out of the box" without building the container again and so on.

The second benefit is that you're not forced to keep the AI codebase "up-to-date" with the latest models (what we're currently doing and even with the work that we're doing, we don't support out of the box the latest ones), we can just rely on the catalog to retrieve the model, perform our `invoke` call and it works, no need to configure the catalog or submit a new PR (same thing goes for `ElevenLabs` BTW).

The goal behind this PR is to ease the usage of Ollama (especially when you're using a "cloud" version of it and/or you're using dozen of models) while removing the "hardcoded" aspect of it to keep the API as the "source of truth".

OskarStark:
Ok, but could we use the decorator pattern here, so the user can decide if he want to have the optional call or not? I talk about decorating the FallbackModelCatalog

Guikingone:
So maybe with an option like `retrieve_catalog_from_api`?

If set to `true`, then we decorate and perform the call on the API?

OskarStark:
yes, don't like the name, but we can rename it later

src/ai-bundle/config/options.php
```diff
@@ -160,6 +160,7 @@
                                 ->defaultValue('http_client')
                                 ->info('Service ID of the HTTP client to use')
                             ->end()
+                            ->booleanNode('use_api_as_catalog')->end()
```
```suggestion
                            ->booleanNode('api_catalog')->end()
```
?
and add a description what it does and what can be side effects

Also update the docs please

src/platform/src/Bridge/Ollama/OllamaApiCatalog.php
```diff
@@ -0,0 +1,83 @@
...
+use Symfony\AI\Platform\Capability;
+use Symfony\AI\Platform\Exception\InvalidArgumentException;
+use Symfony\AI\Platform\Model;
+use Symfony\AI\Platform\ModelCatalog\FallbackModelCatalog;
+use Symfony\Contracts\HttpClient\HttpClientInterface;
+
+/**
+ * @author Guillaume Loulier <personal@guillaumeloulier.fr>
+ */
+final class OllamaApiCatalog extends FallbackModelCatalog
```
OskarStark, line 22: Hmm, if I am using an API Catalog, I think we should not fallback, should we?

Guikingone: Agree, I was playing with the decorating approach but wasn't convinced about the fallback, will remove it.

src/ai-bundle/config/options.php
```diff
@@ -160,6 +160,9 @@
                                 ->defaultValue('http_client')
                                 ->info('Service ID of the HTTP client to use')
                             ->end()
+                            ->booleanNode('api_catalog')
+                                ->info('If set, the Ollama API will be used to build the catalog and retrieve models informations, using this option lead to additional HTTP calls')
```
```suggestion
                                ->info('If set, the Ollama API will be used to build the catalog and retrieve models informations, using this option leads to additional HTTP calls')
```

examples/ollama/chat-llama-api-catalog.php
chr-hertel, line 1: i think your docs addition is sufficient and we don't need this example

Guikingone: Removed 👍🏻

OskarStark:
Thank you @Guikingone.

src/platform/src/Bridge/Ollama/OllamaApiCatalog.php
OskarStark: Sorry I missed this comment, the Fallback is still there, can you please fix it in a follow up? Thanks

Guikingone: Yes, no problem 👍🏻

PR #771, opened 2025-10-09
[AI Bundle][Demo] Add config options for keeping tool call messages and sources + expose metadata in profiler

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | 
| License       | MIT

image

image


demo/config/packages/ai.yaml
```diff
@@ -25,10 +25,11 @@ ai:
                 options:
                     temperature: 0.5
             prompt:
-                text: 'Please answer the users question based on Wikipedia and provide a link to the article.'
+                text: 'Please answer the users question based on Wikipedia, only use information provided in the messages.'
                 include_tools: true
             tools:
                 - 'Symfony\AI\Agent\Toolbox\Tool\Wikipedia'
+            keep_tool_sources: true
```
OskarStark:
```suggestion
            include_sources: true
```
? Just an idea

chr-hertel: Makes more sense, yes

demo/config/packages/ai.yaml
```diff
@@ -25,10 +25,11 @@ ai:
                 options:
                     temperature: 0.5
             prompt:
-                text: 'Please answer the users question based on Wikipedia and provide a link to the article.'
+                text: 'Please answer the users question based on Wikipedia, only use information provided in the messages.'
                 include_tools: true
             tools:
                 - 'Symfony\AI\Agent\Toolbox\Tool\Wikipedia'
+            include_sources: true
```
OskarStark, line 32: Only thing is, that this should be under the tools key, or not?

chr-hertel: Yes, thought the same, but wasn't able to find think of a good target structure. Should also include the `fault_toletant_toolbox`.
Buuuuut, i like the direct `tools` array. Maybe an additional `tool_settings` or just `settings` node?

OskarStark: lets keep it for now, can be optimized later. Ready to merge from my side

PR #787, opened 2025-10-16
[Platform][Mistral] Add `ImageUrlNormalizer`

franzwilding:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| Docs?         | no
| License       | MIT

The mistral platform factory gets a contract per default as an argument, so all the internal normalizers will not be used. This PR changes this by adding a custom MistralContract that will be used (like in the openAI birdge). 

This PR also adds an imageUrl normalizer, since the mistral chat api supports image_urls as well.

### DIFF
--- a/src/platform/src/Bridge/Mistral/PlatformFactory.php
+++ b/src/platform/src/Bridge/Mistral/PlatformFactory.php
@@ -13,6 +13,7 @@ namespace Symfony\AI\Platform\Bridge\Mistral;
 
 use Symfony\AI\Platform\Bridge\Mistral\Contract\DocumentNormalizer;
 use Symfony\AI\Platform\Bridge\Mistral\Contract\DocumentUrlNormalizer;
+use Symfony\AI\Platform\Bridge\Mistral\Contract\ImageUrlNormalizer;
 use Symfony\AI\Platform\Bridge\Mistral\Contract\ToolNormalizer;
 use Symfony\AI\Platform\Contract;
 use Symfony\AI\Platform\ModelCatalog\ModelCatalogInterface;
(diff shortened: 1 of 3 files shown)


src/platform/src/Bridge/Mistral/PlatformFactory.php
```diff
@@ -37,11 +35,7 @@ public static function create(
             [new Embeddings\ModelClient($httpClient, $apiKey), new Llm\ModelClient($httpClient, $apiKey)],
             [new Embeddings\ResultConverter(), new Llm\ResultConverter()],
             $modelCatalog,
-            $contract ?? Contract::create(
-                new ToolNormalizer(),
-                new DocumentNormalizer(),
-                new DocumentUrlNormalizer(),
-            ),
+            $contract ?? MistralContract::create(),
```
OskarStark: is this described bug the same problem for all other bridges as well?

franzwilding: @OskarStark I’m not entirely sure, but I believe that’s the case. I haven’t fully figured out how it’s supposed to work yet. What I noticed is that for all bridges without their own contract, the default contract is passed as an argument to the platform factories. Since it’s already set, Contract::create() is never called. One possible solution might be to avoid passing a default contract at all—what do you think?

OskarStark: you mean via the bundle config, yes lets not inject the default one there, can you open a separate PR please?

franzwilding: @OskarStark https://github.com/symfony/ai/pull/789

franzwilding:
@OskarStark I removed the dedicated mistral contract. Once https://github.com/symfony/ai/pull/789 is merged, this PR can also be merged and the new imageUrlNormalizer gets added directly in the platform factory.

OskarStark:
Thank you @franzwilding.

PR #789, opened 2025-10-16
 Fix contract injection issue

franzwilding:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | No
| Docs?         | No
| License       | MIT

This PR removes the injection of the default contract in order to allow the platform factories to create their own contracts per default.

### DIFF
--- a/src/ai-bundle/src/AiBundle.php
+++ b/src/ai-bundle/src/AiBundle.php
@@ -297,7 +297,6 @@ final class AiBundle extends AbstractBundle
                     $platform['host'],
                     new Reference($platform['http_client'], ContainerInterface::NULL_ON_INVALID_REFERENCE),
                     new Reference('ai.platform.model_catalog.elevenlabs'),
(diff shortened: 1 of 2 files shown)


src/ai-bundle/src/AiBundle.php
```diff
@@ -297,7 +297,6 @@ private function processPlatformConfig(string $type, array $platform, ContainerB
                     $platform['host'],
                     new Reference($platform['http_client'], ContainerInterface::NULL_ON_INVALID_REFERENCE),
                     new Reference('ai.platform.model_catalog.elevenlabs'),
-                    new Reference('ai.platform.contract.default'),
```
OskarStark, line 300: so should we remove the `ai.platform.contract.default` definition, too?

franzwilding:
@OskarStark you are right! The service isn't used anywhere anymore. I removed it.

OskarStark:
Thank you @franzwilding.

PR #794, opened 2025-10-21
[Platform] Fix structured output capability check for unsupported models

### LINKED ISSUE #791: [Bug] Capability::OUTPUT_STRUCTURED is incorrectly listed for Gemini < 2.5 and lacks a proper exception
Hello @OskarStark, @chr-hertel,

I'm encountering an issue with the output_structure option (which maps to Capability::OUTPUT_STRUCTURED) when using Gemini models, specifically versions 2.0.

Problem Description
When trying to use the AI Agent with a gemini-2.0 model and the output_structure option to get a formatted JSON response, the call does not work as expected.

Upon investigation, I checked the official Google AI documentation regarding structured output (JSON Schema): https://ai.google.dev/gemini-api/docs/structured-output?hl=en

This documentation is very clear and states (regarding JSON Schema support via responseJsonSchema):

"It only works with Gemini 2.5."

This confirms that Gemini 2.0 models (like gemini-2.0-flash, etc.) do not support this feature.

Issues Identified in symfony/ai
The current problem is twofold:

Incorrect Model Catalog: The ModelCatalog (likely for the Google/VertexAI bridge) appears to list Capability::OUTPUT_STRUCTURED for Gemini 2.0 models. This is in direct contradiction with Google's official documentation.

Lack of Exception: The most problematic part is that the AgentProcessor (or a similar class) does not throw any exception (e.g., LogicException or UnsupportedPlatformException) when a capability (OUTPUT_STRUCTURED) is requested but not actually supported by the model. This check seems to have been removed (see [this comment in #728](https://github.com/symfony/ai/pull/728#issuecomment-3415418506)). The call fails silently or in an unexpected way, which makes diagnosis and debugging very difficult.

Suggested Improvements
To fix this behavior and improve the component's robustness, I suggest the following actions:

Update the Catalog: Remove Capability::OUTPUT_STRUCTURED from all Gemini models with a version lower than 2.5 in the relevant ModelCatalog files (Google, VertexAI, etc.).

Restore/Add an Exception: Add a check back into the AgentProcessor (or wherever the options are validated) to throw a clear exception if the user tries to use an option (like output_structure) that is not declared in the selected model's capabilities ($model->getCapabilities()).

This would provide "fail-fast" behavior and immediately inform the developer of the incompatibility between their request and the chosen model.

Thank you for your work on this component!

Griffon-Weglot:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #791 Related to #802
| License       | MIT


src/agent/src/Agent.php
```diff
@@ -77,6 +78,13 @@ public function call(MessageBag $messages, array $options = []): ResultInterface
         $messages = $input->getMessageBag();
         $options = $input->getOptions();
 
+        if (isset($options['output_structure'])) {
```
OskarStark: IIRC we had some kind of check before, right @chr-hertel ?

VincentLanglet: The way it was checked before is directly in the AgentProcessor
https://github.com/symfony/ai/pull/728/files#diff-212a36313084d7a7912d6ed09c262e978715148ce2d045aee724cfc3ec9b342cL77-L79

I think it would be better like it was before @Griffon-Weglot

VincentLanglet: The only issue is that it requires to call `$this->platform->getModelCatalog()->getModel($input->getModel());` in the OutputProcessor (and therefor inject the platform inside the OutputProcessor).

I don't know if it's was a good win to change `$input->getModel` from the object to the string.

Griffon-Weglot: Ok i've made the changes, i agree with @VincentLanglet, injecting platform is maybe a problem

src/agent/src/StructuredOutput/AgentProcessor.php
```diff
@@ -77,6 +80,11 @@ public function processInput(Input $input): void
             throw new InvalidArgumentException('Streamed responses are not supported for structured output.');
         }
 
+        $modelObject = $this->platform->getModelCatalog()->getModel($input->getModel());
+        if (!\in_array(Capability::OUTPUT_STRUCTURED, $modelObject->getCapabilities(), true)) {
+            throw MissingModelSupportException::forStructuredOutput($modelObject->getName());
```
chr-hertel: this would be enough for me
```suggestion
        $model = $this->platform->getModelCatalog()->getModel($input->getModel());
        if (!\in_array(Capability::OUTPUT_STRUCTURED, $model->getCapabilities(), true)) {
            throw MissingModelSupportException::forStructuredOutput($model->getName());
```

src/agent/tests/StructuredOutput/AgentProcessorTest.php
```diff
@@ -25,18 +27,31 @@
...
 use Symfony\AI\Platform\Result\ObjectResult;
 use Symfony\AI\Platform\Result\TextResult;
 use Symfony\Component\Serializer\SerializerInterface;
 
+if (!class_exists(__NAMESPACE__.'\ConfigurableResponseFormatFactory')) {
+    class ConfigurableResponseFormatFactory implements ResponseFormatFactoryInterface {
+        public function __construct(private array $format = []) {}
+        public function create(string $structure): array { return $this->format; }
+    }
+}
```
chr-hertel: do we need this? there is already `Symfony\AI\Agent\Tests\StructuredOutput\ConfigurableResponseFormatFactory`

src/agent/src/StructuredOutput/AgentProcessor.php
VincentLanglet: I'm not familiar with the new pattern.

Is it ok to inject the platform every time we need the capabilities ; or would it be better to have the model instance directly in the input like it was before @chr-hertel ?

chr-hertel:
Let's please put this to hold - most likely gonna be replaced with #802

Thanks anyways @Griffon-Weglot - also for second the necessity to bring back the validation!

VincentLanglet:
> Let's please put this to hold - most likely gonna be replaced with #802
> 
> Thanks anyways @Griffon-Weglot - also for second the necessity to bring back the validation!

Am I wrong or @Griffon-Weglot could/should still open a PR after #802 in order to update/fix the supported capabilities of Gemini ?  (And this can be done already since it shouldn't conflict with your work @chr-hertel)

chr-hertel:
@VincentLanglet that is correct

chr-hertel (changes requested):
Hey @Griffon-Weglot, i just merged #802 - can you please rebase and reduce the scope of this PR to fix the capabilities in the Gemini and VertexAI model catalog?

Griffon-Weglot:
> Hey @Griffon-Weglot, i just merged #802 - can you please rebase and reduce the scope of this PR to fix the capabilities in the Gemini and VertexAI model catalog?

Sure !

Griffon-Weglot:
@chr-hertel Done, sorry for the delay !

chr-hertel:
Thank you @Griffon-Weglot.

PR #802, opened 2025-10-22
[Agent][Platform] Shift structured output from Agent to Platform component

### LINKED ISSUE #791: [Bug] Capability::OUTPUT_STRUCTURED is incorrectly listed for Gemini < 2.5 and lacks a proper exception

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | 
| License       | MIT

The main differentiator between the Agent and the Platform component is that the Agent is designed for multi-step model interaction and this is not the case for structured output. In the beginning the feature was also a bit more heavy than the light platform, since it was using the Serializer, and the Platform not having extension points for it.

Nowadays the Platform uses the Serializer anyways for the contract handling, and relies on a optional EventDispatcher, that now also can provide the needed extension points.

Secondary goals where to introduce a `ResultEvent` and bring back user-land validation for structured output capability of model.

Fixes #791, replaces #794

image


OskarStark:
Rebase needed, but from what I can see 👍

OskarStark (approved):
Definitely makes more sense in the platform from an architectural point of view 👍🏻

PR #808, opened 2025-10-23
[MCP Bundle] Add configurable tool discovery options 

### LINKED ISSUE #801: [MCP Bundle] Add options for tool discovery
The container configuration currently hard codes the directory to be scanned for tools:

```php
# vendor/symfony/mcp-bundle/config/services.php
->call('setDiscovery', [param('kernel.project_dir'), ['src']])
```

The MCP SDK although allows setting multiple `scanDirs` and even `excludeDirs`:

```php
# vendor/mcp/sdk/src/Server/Builder.php
    public function setDiscovery(
        string $basePath,
        array $scanDirs = ['.', 'src'],
        array $excludeDirs = [],
        ?CacheInterface $cache = null,
    ): self {}
```

It would be nice to configure those directories via the bundle configuration. Especially if you define your tools in a small subset of all directories.

## Background

I ran into a problem that I ship the `src/DataFixtures` directory into production where the classes inherit from a base Fixture that is not installed when doing a composer install --no-dev for prod. This has never led to any problems but because the MCP SDK is now running a discovery in full `/src` it fails because of the missing base class.

OskarStark:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | Fix #801
| License       | MIT

## Summary
- Add support for configurable `scan_dirs` and `exclude_dirs` in tool discovery
- Allows users to specify multiple directories to scan for MCP tools
- Allows users to exclude specific directories from discovery

## Configuration Example
```yaml
# config/packages/mcp.yaml
mcp:
    discovery:
        scan_dirs: ['src', 'lib']
        exclude_dirs: ['src/DataFixtures']
```


chr-hertel:
Thank you @OskarStark.

PR #816, opened 2025-10-25
 Fix json_decode calls

VincentLanglet:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Docs?         | no 
| Issues        | Fix #... 
| License       | MIT

`JSON_THROW_ON_ERROR` was passed as the third arg instead of the fourth one.


OskarStark (approved):

chr-hertel:
Thank you @VincentLanglet.

VincentLanglet:
> Oh nice catch, couldn't PHPStan spot such issues?

Not that easy since `JSON_THROW_ON_ERROR` is an integer (4194304) so it's technically a valid value for the `depth`.

What you could use is maybe rector:
https://getrector.com/rule-detail/json-throw-on-error-rector

PR #820, opened 2025-10-27
[Store] Support `string|object` for `getContent` method

paulinevos:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Docs?         | no 
| License       | MIT

As the Vectorizer currently makes assumptions about having to pass string data to `PlatformInterface::invoke`, when that method can actually take about any sort of input.

`Platform` implementations decide what data types they can handle through their normalizers. Therefore, it should be up to `platform` to handle types and not up to `store` to decide that it should only pass strings.

### DIFF
--- a/src/store/src/Document/EmbeddableDocumentInterface.php
+++ b/src/store/src/Document/EmbeddableDocumentInterface.php
@@ -15,7 +15,7 @@ interface EmbeddableDocumentInterface
 {
     public function getId(): mixed;
 
-    public function getContent(): string;
+    public function getContent(): string|object;
 
     public function getMetadata(): Metadata;
 }
--- a/src/store/src/Document/Vectorizer.php
+++ b/src/store/src/Document/Vectorizer.php
@@ -14,6 +14,7 @@ namespace Symfony\AI\Store\Document;
 use Psr\Log\LoggerInterface;
 use Psr\Log\NullLogger;
 use Symfony\AI\Platform\Capability;
+use Symfony\AI\Platform\Exception\ExceptionInterface;
 use Symfony\AI\Platform\PlatformInterface;
 use Symfony\AI\Platform\Vector\Vector;
 use Symfony\AI\Store\Exception\RuntimeException;
@@ -75,6 +76,8 @@ final class Vectorizer implements VectorizerInterface
 
     /**
      * @param array<string, mixed> $options
+     *
+     * @throws ExceptionInterface
      */
     private function vectorizeString(string|\Stringable $string, array $options = []): Vector
     {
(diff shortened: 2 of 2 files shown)


src/store/src/Document/Vectorizer.php
```diff
@@ -18,12 +18,12 @@
...
 use Symfony\AI\Store\Exception\RuntimeException;
 
-final readonly class Vectorizer implements VectorizerInterface
+final class Vectorizer implements VectorizerInterface
 {
     public function __construct(
-        private PlatformInterface $platform,
-        private string $model,
-        private LoggerInterface $logger = new NullLogger(),
+        private readonly PlatformInterface $platform,
```
paulinevos, line 25: PHPStan complained about the `readonly` class btw. I guess this is a new rule?

src/store/src/Document/EmbeddableDocumentInterface.php
```diff
@@ -15,7 +15,7 @@ interface EmbeddableDocumentInterface
 {
     public function getId(): mixed;
 
-    public function getContent(): string;
+    public function getContent(): mixed;
```
paulinevos: Maybe `string|object` is a good alternative if we want to be a bit stricter? Because the `platform` normalizers also mostly assume that you're passing either strings or data classes.

src/store/src/Document/Vectorizer.php
chr-hertel: yes, it is, see https://github.com/symfony/symfony/pull/62168#issuecomment-3450307615

src/store/src/Document/EmbeddableDocumentInterface.php
OskarStark: If possible I would vote to use `string|object` instead of `mixed`, WDYT @chr-hertel ?

chr-hertel: Yes, i would agree, currently the `invoke` also is explicit on the `$input` with`array|string|object` so it would be more consistent to be explicit here as well => `string|object`

paulinevos: Ok, will change! Re: arrays, I haven't come across a use case where arrays would be used. Maybe leave it `string|object` until proven otherwise?

src/store/src/Document/Vectorizer.php
```diff
@@ -14,16 +14,17 @@
...
+        private readonly string $model,
+        private readonly LoggerInterface $logger = new NullLogger(),
     ) {
     }
```
chr-hertel, line 29: a bit confused about this diff, since it is already fixed on main with #829 

https://github.com/symfony/ai/blob/5f498f5d46af215f9a4eb3b26cefb8a4d67d5fb1/src/store/src/Document/Vectorizer.php#L21-L28

src/store/src/Document/EmbeddableDocumentInterface.php
chr-hertel: sounds good, thanks!

src/store/src/Document/Vectorizer.php
paulinevos: Yeah weird, I'll try rebase although that shouldn't have any bearing on this diff... Maybe GH being buggy.

OskarStark:
PR title would need an update

src/store/src/Document/Vectorizer.php
paulinevos: Not sure why this is still showing up, I'll chalk it up to GH being buggy.

chr-hertel:
Thank you @paulinevos.

PR #824, opened 2025-10-29
[Platform] Add support for Cartesia

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | None
| License       | MIT

Hi 👋🏻 

Here's the PR for the support of `Cartesia` as a TTS / STT platform.


Guikingone:
Unpopular opinion to discuss @OskarStark @chr-hertel but IMHO, the CI should not fail if we're using `readonly` classes. 

Small example with the `Symfony\AI\Platform\Contract` defined as `readonly`, if I remove the `readonly` from sub-classes, PHP will throw an error.

Using `readonly` is not a bad approach IMHO, what's the goal behind this restriction? 🤔 

PS: There's little to no chance that one day, we store informations in the `*Client` classes.

chr-hertel:
I don't think that this is an unpopular opinion, it's just something that is not easy to have while having to keep up with the BC promise, see https://github.com/symfony/symfony/pull/62168#issuecomment-3450307615 for reasoning.

examples/.env
```diff
@@ -52,6 +52,11 @@ OPENROUTER_KEY=
 ELEVEN_LABS_URL=https://api.elevenlabs.io/v1
 ELEVEN_LABS_API_KEY=
 
+# For using Cartesia
+CARTESIA_URL=https://api.cartesia.ai
+CARTESIA_API_KEY=
+CARTESIA_API_VERSION=2025-04-16
```
chr-hertel: please keep in mind that we only need env vars for stuff users need to configure - does the URL or API_VERSION change?

Guikingone: The `version` is a list on the API so you can use the latest or previous ones, the URL should not change over time (no versioning on it), I'll remove the URL from here.

src/ai-bundle/src/AiBundle.php
```diff
@@ -291,6 +292,26 @@ private function processPlatformConfig(string $type, array $platform, ContainerB
...
+                ->setFactory(CartesiaPlatformFactory::class.'::create')
+                ->setLazy(true)
+                ->addTag('proxy', ['interface' => PlatformInterface::class])
+                ->setArguments([
+                    $platform['api_key'],
+                    $platform['version'],
+                    $platform['host'],
+                    new Reference($platform['http_client'], ContainerInterface::NULL_ON_INVALID_REFERENCE),
+                    new Reference('ai.platform.model_catalog.cartesia'),
+                    new Reference('event_dispatcher'),
```
chr-hertel, line 308: you copied my mistake from #802 which was patched with #831:
```suggestion
                    new Reference('ai.platform.model_catalog.cartesia'),
                    null,
                    new Reference('event_dispatcher'),
```

src/platform/src/Bridge/Cartesia/CartesiaClient.php
```diff
@@ -0,0 +1,92 @@
...
+/**
+ * @author Guillaume Loulier <personal@guillaumeloulier.fr>
+ */
+final class CartesiaClient implements ModelClientInterface
+{
+    public function __construct(
+        private readonly HttpClientInterface $httpClient,
+        #[\SensitiveParameter] private readonly string $apiKey,
+        private readonly string $version,
+        private readonly ?string $hostUrl = 'https://api.cartesia.ai',
```
chr-hertel: this argument is not needed than, right?

src/ai-bundle/config/options.php
```diff
@@ -54,6 +54,17 @@
                             ->end()
                         ->end()
                     ->end()
+                    ->arrayNode('cartesia')
+                        ->children()
+                            ->stringNode('host')->end()
```
chr-hertel: not needed if i got your comment for env vars right

chr-hertel:
Pipeline still open @Guikingone

Guikingone:
Fixed @chr-hertel, sorry for the delay, small issue with HttpClient.

chr-hertel:
Thank you @Guikingone.

PR #827, opened 2025-10-29
[Platform] Extensions to Ollama streaming

leonexcc:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | None
| License       | MIT

Adds the following things to the Ollama streaming feature:
- Some testing
- OllamaMessageChunk supports thinking (Ollama streams the tokens of the reasoning)
- OllamaMessageChunk adds raw data (like stats and token usage information at the end of the stream)

While I added this, I saw that most of the other ResultConverter::convertStream return just strings. Maybe it would be a good idea to make the API for this more consistent and allow string|MessageChunkInterface or only some MessageChunkInterface . That would make it probably easier to switch Platforms.

Since this is neither a bug nor realy a new feature, I didn't check any above. I didn't find any docs about Ollama streaming so I didn't add anything there. If needed I can try to add something. I did add tests, though 😉.

### DIFF
--- a/src/platform/src/Bridge/Ollama/OllamaMessageChunk.php
+++ b/src/platform/src/Bridge/Ollama/OllamaMessageChunk.php
@@ -18,12 +18,14 @@ final class OllamaMessageChunk
 {
     /**
      * @param array<string, mixed> $message
+     * @param array<string, mixed> $raw
      */
     public function __construct(
         public readonly string $model,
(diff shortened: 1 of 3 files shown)


chr-hertel (approved):

chr-hertel:
Thank you @leonexcc.

PR #831, opened 2025-11-04
[Platform] Fix platform creation

franzwilding:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| License       | MIT

At the moment the platform bundle does not test creating of all possible platforms. Because of this, some more exotic models like mistral could not be created (because of a PR of me some time ago). 

This PR fixes the creation of all platforms and also adds an test to really get the instances of the (lazy) services.


chr-hertel (approved):
Oh, i guess that's my bad with #802 - thanks for patching that :pray:

PR #834, opened 2025-11-04
 Qdrant sync

natewiebe13:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| License       | MIT

Currently, Qdrant PUT operations are async. This forces the user to add checking after the request happens to determine if it was successful or not.

This adds an option to make operations sync so you can handle any errors immediately instead of polling after until the operation completes.

### DIFF
--- a/src/ai-bundle/config/options.php
+++ b/src/ai-bundle/config/options.php
@@ -607,6 +607,7 @@ return static function (DefinitionConfigurator $configurator): void {
                                 ->stringNode('collection_name')->cannotBeEmpty()->end()
                                 ->integerNode('dimensions')->end()
                                 ->stringNode('distance')->end()
+                                ->booleanNode('async')->defaultFalse()->end()
                             ->end()
                         ->end()
                     ->end()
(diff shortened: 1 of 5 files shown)


chr-hertel:
But it still makes sense to have that async capabilities as well? sounds like usually one wouldn't want that?
cc @Guikingone do you have an opinion here?

natewiebe13:
@chr-hertel I took the approach of keeping the existing functionality working as-is. But if you're suggesting to always make the PUT synchronous, I'm happy to update the PR. I do find it slightly odd that Qdrant chose to make operations async by default.

Here's a related issue: https://github.com/qdrant/qdrant/issues/2557 which is actually the exact issue I originally ran into when first using Qdrant. But now I'm hitting other issues related to inserts. So now waiting for a response makes way more sense for my situation.

OskarStark:
lets use async default `false`, its more the standard case I would say

chr-hertel:
Yes, think so too 👍

Guikingone:
Agree with @OskarStark on this one, let use `false` as the default case.

natewiebe13:
Updated the default behavior to synchronously add documents

OskarStark:
Thank you @natewiebe13.

PR #837, opened 2025-11-06
 add support for the albert platform configuration

### LINKED ISSUE #836: [Platform] Add `albert` platform configuration support
The underlying support for the Albert API already exists in the codebase, but it is currently not possible to configure it through the `ai.platforms` section, as the `albert` platform name is not included in the list of auto-configured platforms (see [AiBundle::processPlatformConfig](https://github.com/symfony/ai/blob/main/src/ai-bundle/src/AiBundle.php#L247)).

Example configuration that should ideally work:

```yaml
ai:
  platforms:
    albert:
      api_key: '%env(ALBERT_API_KEY)%'
      host_url: '%env(ALBERT_API_URL)%'
```

**Expected behavior:**
The configuration should register the Albert platform similarly to the existing platforms (e.g., openai, mistral, ollama, etc.).

**Request:**
Add "albert" to the list of recognized platform keys in the bundle so it can be configured normally.

Thanks!

jvancoillie:
| Q             | A
| ------------- | ---
| Bug fix?      |no
| New feature?  | yes 
| Docs?         | no 
| Issues        | Fix #836 
| License       | MIT

This PR introduces support for configuring the **Albert** platform under the `ai.platforms` section, following the pattern of existing platforms (OpenAI, Mistral, Ollama, etc.).

The underlying Albert integration was already implemented, but the `albert` platform key was not recognized during configuration parsing.

```yaml
ai:
  platforms:
    albert:
      api_key: '%env(ALBERT_API_KEY)%'
      base_url: '%env(ALBERT_API_URL)%'
```

The configuration uses `base_url` because this is the parameter expected by the Albert platform factory.
Some platforms use `host` instead, which might cause confusion. If needed, I'm open to aligning naming conventions in a follow-up discussion.

I did not define a default `base_url`, but we could consider providing the known Albert endpoint (`https://albert.api.etalab.gouv.fr/v1`) if that aligns with the project’s policy on defaults.

While working on this, I also noticed that the list of Albert models may not be fully aligned with the current API. I can open a separate issue to track that, unless this work is already in progress.


src/ai-bundle/src/AiBundle.php
```diff
@@ -395,6 +396,25 @@ private function processPlatformConfig(string $type, array $platform, ContainerB
...
+            $platformId = 'ai.platform.albert';
+            $definition = (new Definition(Platform::class))
+                ->setFactory(AlbertPlatformFactory::class.'::create')
+                ->setLazy(true)
+                ->addTag('proxy', ['interface' => PlatformInterface::class])
+                ->setArguments([
+                    $platform['api_key'],
+                    $platform['base_url'],
+                    new Reference($platform['http_client'], ContainerInterface::NULL_ON_INVALID_REFERENCE),
+                    new Reference('event_dispatcher'),
```
OskarStark: No need to set a ModelCatalog?

src/ai-bundle/src/AiBundle.php
```diff
@@ -395,6 +396,25 @@ private function processPlatformConfig(string $type, array $platform, ContainerB
...
+                ])
+                ->addTag('ai.platform', ['name' => 'albert']);
+
+            $container->setDefinition($platformId, $definition);
+
+            return;
+        }
```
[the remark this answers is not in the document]
OskarStark: Makes sense

src/ai-bundle/src/AiBundle.php
jvancoillie: I followed the current factory implementation, which instantiates its own ModelCatalog.
I can extend the factory to accept an optional ModelCatalog later if we want to make it configurable.
Also, the existing ModelCatalog for Albert is not aligned with the actual models exposed by the platform, so it will need to be updated as part of that change

jvancoillie: I updated the factory so it now accepts a ModelCatalog. This makes the catalog configurable when needed.

I also saw that the ModelCatalog alignment for Albert has already been merged (#839), so that part is now resolved. Nothing to handle here on my side anymore.

jvancoillie:
Quick update: I added support for providing a custom ModelCatalog in the factory, so the Albert platform configuration is now consistent with the other platforms.

Also, regarding the model list: I saw that the alignment work was already merged, so that part is resolved (#839).

One open point from the original description: Albert currently uses `base_url` while some platforms use `host`. Do we want to standardize on one naming style across platforms, or keep it platform-specific for now?

Related to that: should we provide a default `base_url` for Albert (e.g. https://albert.api.etalab.gouv.fr/v1), or do we prefer requiring it explicitly in configuration?

src/ai-bundle/src/AiBundle.php
```diff
@@ -1112,6 +1133,10 @@ private function processStoreConfig(string $type, array $stores, ContainerBuilde
                     $arguments[5] = $store['distance'];
                 }
 
+                if (\array_key_exists('async', $store)) {
+                    $arguments[6] = $store['async'];
```
OskarStark, line 1137: this should not be part of this PR, maybe a rebase was not correct?

jvancoillie: You're right, it looks like there was an issue during the rebase and some unrelated commits ended up here. I'll clean this up and force-push the PR.

chr-hertel:
See #839 and #840 for other items - i think base_url vs host or default is fine like this

chr-hertel:
could you please squash the two commits?

chr-hertel:
Thank you @jvancoillie.

PR #849, opened 2025-11-09
[MCP Bundle] Fix services when monolog is not installed

### LINKED ISSUE #830: [MCP Bundle] missing requirement to monolog
There is a missing requirement to `symfony/monolog-bundle` in the mcp-bundle. If monolog is not installed, I got the error `Service "monolog.logger.mcp": Parent definition "monolog.logger_prototype" does not exist.` in the installation process, because of this line: https://github.com/symfony/ai/blob/main/src/mcp-bundle/config/services.php#L20

Solution: We should adapt the services.php file or add the requirement to the composer.json?

Or am I making a mistake?

Regards,
Tim

WebMamba:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #830
| License       | MIT

As said on the issue, when monolog is not installed, the container can't compile the services.
The solution here is to register the service only if the bundle is installed

### DIFF
--- a/src/mcp-bundle/config/services.php
+++ b/src/mcp-bundle/config/services.php
@@ -15,12 +15,16 @@ use Mcp\Server;
 use Mcp\Server\Builder;
 
 return static function (ContainerConfigurator $container): void {
-    $container->services()
-        ->set('monolog.logger.mcp')
+    if (class_exists(\Symfony\Bundle\MonologBundle\MonologBundle::class)) {
+        $container->services()
+            ->set('monolog.logger.mcp')
             ->parent('monolog.logger_prototype')
             ->args(['mcp'])
             ->tag('monolog.logger', ['channel' => 'mcp'])
+        ;
+    }
 
+    $container->services()
         ->set('mcp.server.builder', Builder::class)
             ->factory([Server::class, 'builder'])
             ->call('setServerInfo', [param('mcp.app'), param('mcp.version')])


WebMamba:
@chr-hertel tell me what you think about this solution, but while working on it I was thinking that just require monolog could be even better. I don't see any good reason to not use it.

chr-hertel:
I think that solution works for me - is this the only place where we'd need that tho?

I also wonder about the implementation in the `McpBundle` class:
https://github.com/symfony/ai/blob/806d6be183fa06c94e1e61dbe830dd82b86e5e2f/src/mcp-bundle/src/McpBundle.php#L113-L134

@camilleislasse do you recall why it's `Reference('logger')` with `McpCommand`, but `McpController` is with `Reference('monolog.logger.mcp')`?

src/mcp-bundle/config/services.php
```diff
@@ -15,12 +15,16 @@
 use Mcp\Server\Builder;
 
 return static function (ContainerConfigurator $container): void {
-    $container->services()
-        ->set('monolog.logger.mcp')
+    if (class_exists(\Symfony\Bundle\MonologBundle\MonologBundle::class)) {
```
chr-hertel, line 18: please use in import here instead the FCQN:
```suggestion
    if (class_exists(MonologBundle::class)) {
```

camilleislasse:
> @camilleislasse do you recall why it's `Reference('logger')` with `McpCommand`, but `McpController` is with `Reference('monolog.logger.mcp')`?

 @chr-hertel I don't recall why, probably a mistake on my part

OskarStark:
Closing in favor of https://github.com/symfony/ai/pull/864

Thanks

camilleislasse:
thank's @WebMamba

PR #856, opened 2025-11-11
[Store] Fix CacheStore and InMemoryStore to return similarity scores

camilleislasse:
| Q             | A
  | ------------- | ---
  | Bug fix?      | yes
  | New feature?  | no
  | Docs?         | no
  | Issues        | n/a
  | License       | MIT

  ## Description

  The `DistanceCalculator` was calculating distances but not returning them in the `VectorDocument` results. This meant queries to `CacheStore` and `InMemoryStore` returned
  documents without score information, making it impossible to determine result relevance.

  ## Changes

  This fix creates new `VectorDocument` instances with the calculated distance as the `score` property, matching the pattern used by other stores (Redis, MariaDb, Postgres).

### DIFF
--- a/src/store/src/Bridge/Local/DistanceCalculator.php
+++ b/src/store/src/Bridge/Local/DistanceCalculator.php
@@ -58,7 +58,7 @@ final class DistanceCalculator
         }
 
         return array_map(
-            static fn (array $embedding): VectorDocument => $embedding['document'],
+            static fn (array $embedding): VectorDocument => $embedding['document']->withScore($embedding['distance']),
             $currentEmbeddings,
         );
     }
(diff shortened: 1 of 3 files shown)


src/store/src/Bridge/Local/DistanceCalculator.php
```diff
@@ -58,7 +58,12 @@ public function calculate(array $documents, Vector $vector, ?int $maxItems = nul
         }
 
         return array_map(
-            static fn (array $embedding): VectorDocument => $embedding['document'],
+            static fn (array $embedding): VectorDocument => new VectorDocument(
```
OskarStark: Can we implement a withScore method to `VectorDocument` instead?

camilleislasse: 5259e36f27a78dfd01228754997d38617c4581d9

chr-hertel (approved):

chr-hertel:
Thank you @camilleislasse.

PR #864, opened 2025-11-13
[MCP Bundle] Make MonologBundle optional

### LINKED ISSUE #830: [MCP Bundle] missing requirement to monolog

camilleislasse:
| Q             | A
  | ------------- | ---
  | Bug fix?      | yes
  | New feature?  | no
  | Docs?         | no
  | Issues        | Fix #830
  | License       | MIT

This PR makes MonologBundle optional dependency for McpBundle. Without it installed, the container compilation was failing with "Parent definition' monolog.logger_prototype' does not exist."

## Changes

  - Wrap `monolog.logger.mcp` service creation in `class_exists()` check
  - Add `->nullOnInvalid()` to `setLogger()` call in services.php
  - Uniformize logger references: both McpCommand and McpController now use `monolog.logger.mcp` with `NULL_ON_INVALID_REFERENCE`
  - Update test to skip when MonologBundle is not installed

## Credits

Builds on the initial work in #849 by @WebMamba. That PR protected service creation but the service was still referenced without protection. This PR completes the fix by also protecting the service usage.

### DIFF
--- a/src/mcp-bundle/config/services.php
+++ b/src/mcp-bundle/config/services.php
@@ -17,20 +17,15 @@ use Mcp\Server\Builder;
 
 return static function (ContainerConfigurator $container): void {
     $container->services()
-        ->set('monolog.logger.mcp')
-            ->parent('monolog.logger_prototype')
-            ->args(['mcp'])
-            ->tag('monolog.logger', ['channel' => 'mcp'])
-
         ->set('mcp.registry', Registry::class)
-            ->args([service('event_dispatcher'), service('monolog.logger.mcp')])
+            ->args([service('event_dispatcher'), service('logger')])
+            ->tag('monolog.logger', ['channel' => 'mcp'])
 
         ->set('mcp.server.builder', Builder::class)
             ->factory([Server::class, 'builder'])
             ->call('setServerInfo', [param('mcp.app'), param('mcp.version')])
             ->call('setPaginationLimit', [param('mcp.pagination_limit')])
             ->call('setInstructions', [param('mcp.instructions')])
-            ->call('setLogger', [service('monolog.logger.mcp')])
             ->call('setEventDispatcher', [service('event_dispatcher')])
             ->call('setRegistry', [service('mcp.registry')])
             ->call('setSession', [service('mcp.session.store')])
@@ -38,7 +33,10 @@ return static function (ContainerConfigurator $container): void {
             ->call('addNotificationHandlers', [tagged_iterator('mcp.notification_handler')])
             ->call('addLoaders', [tagged_iterator('mcp.loader')])
             ->call('setDiscovery', [param('kernel.project_dir'), param('mcp.discovery.scan_dirs'), param('mcp.discovery.exclude_dirs')])
+            ->call('setLogger', [service('logger')])
+            ->tag('monolog.logger', ['channel' => 'mcp'])
 
         ->set('mcp.server', Server::class)
-            ->factory([service('mcp.server.builder'), 'build']);
+            ->factory([service('mcp.server.builder'), 'build'])
+    ;
 };
--- a/src/mcp-bundle/src/Command/McpCommand.php
+++ b/src/mcp-bundle/src/Command/McpCommand.php
@@ -24,7 +24,7 @@ class McpCommand extends Command
 {
     public function __construct(
         private readonly Server $server,
-        private readonly LoggerInterface $logger,
+        private readonly ?LoggerInterface $logger = null,
     ) {
         parent::__construct();
     }
--- a/src/mcp-bundle/tests/DependencyInjection/McpBundleTest.php
+++ b/src/mcp-bundle/tests/DependencyInjection/McpBundleTest.php
@@ -51,19 +51,6 @@ class McpBundleTest extends TestCase
         $this->assertSame('This server provides weather and calendar tools', $container->getParameter('mcp.instructions'));
     }
 
-    public function testMcpLoggerServiceIsCreated()
-    {
-        $container = $this->buildContainer([]);
-
-        $this->assertTrue($container->hasDefinition('monolog.logger.mcp'));
-
-        $definition = $container->getDefinition('monolog.logger.mcp');
-        $this->assertInstanceOf(\Symfony\Component\DependencyInjection\ChildDefinition::class, $definition);
-        $this->assertSame('monolog.logger_prototype', $definition->getParent());
-        $this->assertSame(['mcp'], $definition->getArguments());
-        $this->assertTrue($definition->hasTag('monolog.logger'));
-    }
-
     #[DataProvider('provideClientTransportsConfiguration')]
     public function testClientTransportsConfiguration(array $config, array $expectedServices)
     {
--- a/src/mcp-bundle/composer.json
+++ b/src/mcp-bundle/composer.json
@@ -28,7 +28,8 @@
     "require-dev": {
         "phpstan/phpstan": "^2.1",
         "phpstan/phpstan-strict-rules": "^2.0",
-        "phpunit/phpunit": "^11.5.46"
+        "phpunit/phpunit": "^11.5.46",
+        "symfony/monolog-bundle": "^3.10 || ^4.0"
     },
     "minimum-stability": "dev",
     "prefer-stable": true,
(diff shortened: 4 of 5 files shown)


src/mcp-bundle/tests/DependencyInjection/McpBundleTest.php
```diff
@@ -49,6 +50,10 @@ public function testCustomConfiguration()
 
     public function testMcpLoggerServiceIsCreated()
     {
+        if (!class_exists(MonologBundle::class)) {
+            $this->markTestSkipped('MonologBundle is not installed');
+        }
+
```
WebMamba: Since monolog is never installed on the ci, this test will be always skipped, so the rest of the test is useless. What do you think about requiring monolog on dev then?

src/mcp-bundle/config/services.php
```diff
@@ -13,20 +13,25 @@
...
+    }
 
+    $container->services()
         ->set('mcp.server.builder', Builder::class)
             ->factory([Server::class, 'builder'])
             ->call('setServerInfo', [param('mcp.app'), param('mcp.version')])
             ->call('setPaginationLimit', [param('mcp.pagination_limit')])
             ->call('setInstructions', [param('mcp.instructions')])
-            ->call('setLogger', [service('monolog.logger.mcp')])
+            ->call('setLogger', [service('monolog.logger.mcp')->nullOnInvalid()])
```
WebMamba: The function setLogger can't have a nullable as argument see:
https://github.com/modelcontextprotocol/php-sdk/blob/7a3b473c405411aca8be14c0c6e12d2493752d61/src/Server/Builder.php#L254.
So I don't think this can work... Maybe to have an if condition before the call can be better ?

camilleislasse: Yes and another error probably here :     

https://github.com/symfony/ai/blob/1e69969207e6657a6adc811b8b0810c6cd4c9bf1/src/mcp-bundle/src/Command/McpCommand.php#L27

camilleislasse: I'm not entirely happy with this solution ecea6d6 - conditionally instantiating transports with/without the logger parameter feels verbose.

The issue: MCP SDK has `LoggerInterface $logger = new NullLogger()` (non-nullable), so we can't pass `null` even though there's a default, if you have a better idea @WebMamba  

Can change if https://github.com/modelcontextprotocol/php-sdk/pull/152 is correct

src/mcp-bundle/tests/DependencyInjection/McpBundleTest.php
camilleislasse: [a8500ca](https://github.com/symfony/ai/pull/864/commits/a8500cab5c1bd00b6f797f23920c8196cb202ced)

OskarStark:
I tend more and more to require the logger

camilleislasse:
> I tend more and more to require the logger

Me too, I feel like we’re making things unnecessarily complicated  😂

src/mcp-bundle/src/Controller/McpController.php
```diff
@@ -35,12 +35,18 @@ public function __construct(
...
-            $this->responseFactory,
-            $this->streamFactory,
-            logger: $this->logger,
-        );
+        $transport = $this->logger
+            ? new StreamableHttpTransport(
+                $this->httpMessageFactory->createRequest($request),
+                $this->responseFactory,
+                $this->streamFactory,
+                logger: $this->logger,
```
WebMamba: Same here you can pass the NullLogger instead ?

WebMamba (commented):
Also, it’s not really the point of this PR, but one of the key parts of this bundle is to build/instantiate the mcp.server.builder, and we don’t have any tests to secure this part. It seems important to me to add tests to cover it. The first commit of this PR had the CI green even if the builder wasn’t working.

src/mcp-bundle/src/Command/McpCommand.php
```diff
@@ -24,14 +24,16 @@ class McpCommand extends Command
...
     ) {
         parent::__construct();
     }
 
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
-        $transport = new StdioTransport(logger: $this->logger);
+        $transport = $this->logger
+            ? new StdioTransport(logger: $this->logger)
+            : new StdioTransport();
```
WebMamba: what do you thinks instead of having the NullLogger logic here: https://github.com/modelcontextprotocol/php-sdk/pull/152/files to have it here ?
```suggestion
        $transport = new StdioTransport(logger:  $this->logger === null ? NullLogger() : $this->logger)
```

OskarStark:
Makes sense, open for a PR bringing in some tests?

src/mcp-bundle/src/Command/McpCommand.php
WebMamba: Sorry the reviews are in the wrong order I miss clicked 😅

WebMamba:
But yes, to sum up my thought, you can either:
- require Monolog
- or if Monolog is not installed, instantiate a NullLogger

After some reflection, I’m leaning more toward the NullLogger option 😇

camilleislasse:
> But yes, to sum up my thought, you can either:
> 
> * require Monolog
> * or if Monolog is not installed, instantiate a NullLogger
> 
> After some reflection, I’m leaning more toward the NullLogger option 😇

  Maybe wait the review of https://github.com/modelcontextprotocol/php-sdk/pull/152 ? If it's merged, I can pass the logger (which can be null via `ContainerInterface::NULL_ON_INVALID_REFERENCE`) directly to the transport without ternary operators.

camilleislasse:
> Makes sense, open for a PR bringing in some tests?
  
@OskarStark Shouldn't lint:container have caught this? I'm not sure if it's possible to run it independently on each bundle though - they each have their own dependencies, right?

chr-hertel:
Hey @camilleislasse, what's the status here? revisit after merge of https://github.com/modelcontextprotocol/php-sdk/pull/152?

camilleislasse:
> Hey @camilleislasse, what's the status here? revisit after merge of [modelcontextprotocol/php-sdk#152](https://github.com/modelcontextprotocol/php-sdk/pull/152)?

Hey, yes exactly, I'll try to watch it tonight, I'm a bit busy

src/mcp-bundle/tests/DependencyInjection/McpBundleTest.php
WebMamba: Since monolog is now installe in dev we don't need this ?

camilleislasse: b920d78289481b3d1594f91063858f8f1d4b055a

src/mcp-bundle/config/services.php
```diff
@@ -13,24 +13,33 @@
...
 use Mcp\Server\Builder;
+use Symfony\Bundle\MonologBundle\MonologBundle;
 
 return static function (ContainerConfigurator $container): void {
-    $container->services()
-        ->set('monolog.logger.mcp')
-            ->parent('monolog.logger_prototype')
-            ->args(['mcp'])
-            ->tag('monolog.logger', ['channel' => 'mcp'])
+    if (class_exists(MonologBundle::class)) {
```
derrabus: The fact that the bundle class exists does not mean that the bundle is loaded/enabled.

derrabus:
Would it make sense to inject the generic `logger` service instead (which should always be present) and use the `monolog.logger` tag to indicate the channel?

OskarStark:
Do we need to wait for a new mcp sdk tag?

camilleislasse:
@derrabus Is the last commit what you had in mind? 
  @OskarStark If so, we no longer need to wait for a new MCP SDK tag, we always inject a valid logger now, never null. And anyway, the nullable logger (mcp/sdk#152) was already included in v0.1.0.

camilleislasse:
@derrabus If you have time to confirm

derrabus:
> @derrabus Is the last commit what you had in mind?

Yes.

OskarStark:
CI failure is related and rebase is needed

OskarStark:
Thank you Camille.

PR #865, opened 2025-11-13
[Store] Add Pinecone ManagedStore

MolloKhan:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| License       | MIT

This is a draft PR for managing Pinecone store.

Instead of having a property for each value (indexName, dimension, etc), I wanted to send them as `$options`, so the service can create different kinds of Pinecone index, but it is not possible without modifying the interface `drop()` method, which requires knowing what index you want to remove.
If we update the interface, I could merge this class into `Symfony\AI\Store\Bridge\Pinecone\Store` without adding more properties to the constructor
What do you think?


MolloKhan:
The second part of this PR would be to improve the `SetupStoreCommand` command. After creating the store, the user needs to know the just created host and update the Pinecone client config.
I'm thinking of returning the response when calling `ManagedStoreInterface::setup()` and adding another method that explains what to do next.
For example: "Update the Pinecone Client host property with: `{$host}`"

chr-hertel:
I'd be in favor of extending the interface here - @OskarStark @Guikingone what do you think?

Guikingone:
Agree with @chr-hertel 👍🏻

chr-hertel (changes requested):
Please tackle your proposed changes to the interface with this PR

MolloKhan:
Yea, I will, unfortunately I won't be available until after the SymfonyCon :)

OskarStark (commented):
I agree on the options for drop(), should be a separate PR, but the indexable should be configurable on the store and therefore be set in the constructor IMHO

MolloKhan:
> but the indexable should be configurable on the store and therefore be set in the constructor IMHO

Yes, I think you're right. To set up a Pinecone server, you only need to define an index name (along with the API key). Upon that, you get a host value that you need to attach to the Pinecone client config. So, in short, those two values should be bound together to avoid unexpected errors

OskarStark:
So do we really need the options array for dropping a pinecone index if we configure the index via constructor in the pinwvone store?

MolloKhan:
Not anymore really. Anyway, I think the `$options` argument in the `drop()` method still makes sense, there may be a service that could benefit from it

OskarStark:
Closed via https://github.com/symfony/ai/pull/1215

PR #866, opened 2025-11-13
[MCP Bundle] Add MCP profiler support to display server capabilities

### LINKED ISSUE #529: [MCP Bundle] Implement Profiler Panel & Toolbar
The MCP Bundle currently focuses only on providing the configuration and preparing services for wiring, but does not provide insights via Profiler Panel or Toolbar.

The profiler could provide information about
  * Configuration
  * Registered Capabilities

camilleislasse:
| Q             | A
| ------------- | ---
| Bug fix?      |no
| New feature?  | yes 
| Docs?         | no 
| Issues        | Fix #529 
| License       | MIT

## Description

This PR adds a Web Profiler integration for the MCP Bundle that displays MCP server capabilities (tools, prompts, resources, resource templates) in the Symfony debug toolbar and profiler panel.

## Features 
- ProfilingLoader to capture Registry reference during server build
- DataCollector to collect and display MCP capabilities

Note: This implementation depends on[ the loader pattern](https://github.com/modelcontextprotocol/php-sdk/pull/111). If [MCP SDK PR #146](https://github.com/modelcontextprotocol/php-sdk/pull/146) (direct Registry injection via DI) is merged, this approach may need adjustment.

Capture d’écran 2025-11-13 à 20 10 00


src/mcp-bundle/templates/icon.svg
chr-hertel, line 1: I think it would be great to copy this one instead:
https://github.com/modelcontextprotocol/inspector/blob/main/client/public/mcp.svg

chr-hertel: but could easily be a follow up

camilleislasse: f4095beffafcda9f316f6dd6c2f24092f34862c4
 
Capture d’écran 2025-11-15 à 08 53 36

OskarStark:
Would love to see some infos around the MCP Profiler and the screenshot in the documentation

chr-hertel (changes requested):
We should be able to use/decorate the Registry directly now and skip the ProfilingLoader, right?

src/mcp-bundle/config/services.php
```diff
@@ -21,18 +24,27 @@
             ->args(['mcp'])
             ->tag('monolog.logger', ['channel' => 'mcp'])
 
+        ->set('ai.mcp.registry.inner', Registry::class)
+            ->args([service('event_dispatcher'), service('monolog.logger.mcp')])
+
+        ->set('ai.mcp.registry', TraceableRegistry::class)
+            ->args([service('ai.mcp.registry.inner')])
```
chr-hertel: let's shift that into `McpBundle`, make it conditional, drop the `ai.` prefix in favor of just `mcp.` and use the decoration feature of the container:
```php
if ($container->getParameter('kernel.debug')) {
    $traceableRegistry = (new Definition('mcp.traceable_registry'))
        ->setClass(TraceableRegistry::class)
        ->setArguments([new Reference('.inner')])
        ->setDecoratedService('mcp.registry');
    $container->setDefinition('mcp.traceable_registry', $traceableRegistry);
}
```

chr-hertel (changes requested):
Looks good & works with the demo - one last thing about the service registration tho :+1: 

Thanks already!

chr-hertel:
Thank you @camilleislasse.

PR #867, opened 2025-11-14
[Store] Allow `id` to be `int|string|Uuid` for `VectorDocument` and `TextDocument`

### LINKED ISSUE #857: [Store] Allow VectorDocument ID to be a string
Hello, I've been migrating a homegrown AI integration onto Symfony AI. I love the initiative and how it simplifies things. However, I hit a wall when migrating my custom Pinecone integration with Symfony Store. 
The problem is that it expects the vector document ID (coming from Pinecone) to be in `UUID` format, in my case, it is not. All the information is already stored in Pinecone with a different ID format, so I can't use it to query my database (I get an `InvalidArgumentException`). 
It would be great to add a way to overcome this. Perhaps the `VectorDocument` id could support `strings`, or is there a reason it must be a `UUID`?

This is the chunk of code I'm referring to: https://github.com/symfony/ai/blob/main/src/store/src/Bridge/Pinecone/Store.php#L71-L78

Thanks in advance!

MolloKhan:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no 
| Issues        | Fix #857
| License       | MIT

I gave this a second thought, and I think @OskarStark idea is the way to go (https://github.com/symfony/ai/issues/857#issuecomment-3522897819)
If the ID coming from the store is in Uuid format the user will likely cast it themself

If you are ok with my changes, I'll proceed to update all the places where `VectorDocuments` are instantiated


OskarStark (approved):

chr-hertel:
that should also be changed in `TextDocument` than, right?

src/store/src/Document/VectorDocument.php
```diff
@@ -19,12 +19,19 @@
...
+        int|string|Uuid $id,
         public readonly VectorInterface $vector,
         public readonly Metadata $metadata = new Metadata(),
         public readonly ?float $score = null,
     ) {
+        if (\is_string($id) && Uuid::isValid($id)) {
+            $this->id = Uuid::fromString($id);
+        } else {
+            $this->id = $id;
+        }
```
MolloKhan: Are we ok with this code duplication?

MolloKhan:
I think this is it

MolloKhan:
Good thinking. I'd need only to remove the Uuid auto-casting from the constructor, right?

MolloKhan:
I removed the Uuid auto-casting and only left the type declaration

OskarStark:
Can we simplify something in the `demo/` folder? Or not needed?

chr-hertel:
> Can we simplify something in the `demo/` folder? Or not needed?

wouldn't know what, don't think so.

but i think a valid follow up would be to kill UUID dependency in store component

OskarStark:
friendly ping @MolloKhan

MolloKhan:
Sorry for the delay! I just got back home from holidays (it was such a long ride 😄) - I searched for all places where we cast strings into `uuid` and the last place is this one (which looks unrelated to me) `Symfony\AI\Store\Document\Loader\RssFeedLoader`

MolloKhan:
I rebased and fixed the conflicts, but I don't understand why the tests are failing (they do pass locally). It seems it is using an old version of the code base or something.
Perhaps I did something wrong when rebasing?

chr-hertel:
Loos like the store bridges don't use the working branch version of the store component :thinking: 

see #1224

OskarStark:
I merged https://github.com/symfony/ai/pull/1224, can you please rebase? Thanks

MolloKhan:
Done! I noticed that the tests failed due to a circular reference. I **think** the same fix from #1224 needs to be applied to the `src/store/src/bridge/` directory

OskarStark:
Do we still need the symfony/uid deps in the composer.json files of the bridges?

OskarStark (commented):
getBinaryMethod is not needed anymore in MongoDB bridge 

Lets use a real dataprovider with a uuid v4 instead of the if check in the test cases

Good to merge afterwards 👍🏻

MolloKhan:
@OskarStark did you mean this? https://github.com/symfony/ai/pull/867/changes/a6a8d62fdbfd486c18796fcc937ec6d9762f41ba
I noticed the MongoDB store assumes the `id` is always a `Uuid` but that's not true anymore, so I cast it to a `string`, but in the case we get a `Uuid`, it won't transform it into a `Binary`. Do you think we need to cover that case too?

MolloKhan:
Hello, are we fine with the changes?

OskarStark:
As we are working on the 0.2 release, all bridges must require `symfony/ai-store` `^0.3` please

.php-cs-fixer.dist.php
```diff
@@ -71,4 +71,4 @@
                 'src/mate/resources/mate/extensions.php',
             ])
     )
-;
+    ;
```
OskarStark: please revert

MolloKhan:
I don't know why the PR history shows a long list of commits, but I think this one is ready

OskarStark:
We can check if everything works by require 0.3 for store bridges in demo/composer.json file and populate the store like described in the demo/README.md

MolloKhan:
I installed the demo and made it require `symfony/ai-chroma-db-store:0.3`, it worked well, but it's not using my changes to the `ai-store` component. I'm not sure how to make the demo use my local changes to that component because I forked the whole `symfony/ai` repository

OskarStark:
Just require it an push, the CI will use your new code. 

On local you can use the link binary in the root of the project.

MolloKhan:
Ohh cool! I wasn't aware of the link binary

MolloKhan:
Alright, this is working locally, and tests are passing

chr-hertel:
Reviewed & tested, but cannot merge with this history - running into conflicts. I guess due to unrelated commits on the branch. Could you please isolate your patch again?

MolloKhan:
I surely screwed up something when rebasing. I'll create a clean PR

MolloKhan:
Closing in favor #1397

PR #868, opened 2025-11-14
[Agent] Fix sources metadata unavailable during streaming

OskarStark:
| Q             | A
  | ------------- | ---
  | Bug fix?      | yes
  | New feature?  | no
  | Docs?         | no
  | Issues        | Fix #833 
  | License       | MIT

Replaces
* https://github.com/symfony/ai/pull/838

Issue #833, opened 2025-11-04
[Agent] Unable to access sources metadata with streaming


JorrinKievit:
I am implementing the Symfony AI Agent in combination with the [ai-sdk](https://ai-sdk.dev/getting-started). I want to display the used sources in the UI, so I want to dispatch these in an event stream using the ai-sdk Data Stream Protocol. However, during streaming, the metadata seems to reset and sources are always `null`. Without the stream option this works fine.

Should this be done in a different order? The sources are present at some point, but seem to reset while actually streaming the content back.

This is basically all that I am doing:

```php
public function __invoke(ChatRequest $request): StreamedResponse
{
    $platform = PlatformFactory::create(env('OPENAI_API_KEY'), HttpClient::create());
    
    $toolBox = new ToolBox([
        new RandomTool(),
    ]);
    $processor = new AgentProcessor($toolBox, includeSources: true);
    $agent = new Agent($platform, 'gpt-4o-mini', [$processor], [$processor]);

    $messages = new MessageBag();
    foreach ($request->messages() as $message) {
        foreach ($message['parts'] as $part) {
            if ($part['type'] !== 'text') {
                continue;
            }
            $messages->add(Message::ofUser($part['text']));
        }
    }

    $result = $agent->call($messages, [
        'stream' => true
    ]);

    $response = new StreamedResponse(function () use ($result) {
        $messageId = 'msg_' . bin2hex(random_bytes(16));
        $textBlockId = 'text_' . bin2hex(random_bytes(16));
        
        $this->sendSSE(['type' => 'start', 'messageId' => $messageId]);
        $this->sendSSE(['type' => 'text-start', 'id' => $textBlockId]);
        
        foreach ($result->getContent() as $chunk) {
            $text = (string)$chunk;
            if ($text !== '') {
                $this->sendSSE([
                    'type' => 'text-delta',
                    'id' => $textBlockId,
                    'delta' => $text
                ]);
            }
        }
        
        $this->sendSSE(['type' => 'text-end', 'id' => $textBlockId]);
        
        // This is always null
        $sources = $result->getMetadata()->get('sources', []);
        foreach ($sources as $source) {
            if (!$source instanceof Source) {
                continue;
            }
            $this->sendSSE([
                'type' => 'source-url',
                'sourceId' => $source->getReference(),
                'url' => $source->getReference(),
                'title' => $source->getName()
            ]);
        }
        
        $this->sendSSE(['type' => 'finish']);
        
[...]

PR #869, opened 2025-11-14
 Add OpenAI example demonstrating streaming with sources

OskarStark:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | Follows #868 
| License       | MIT

This example shows how to use the Agent with OpenAI platform to stream responses while collecting source metadata from tools like Tavily. This demonstrates the fix for issue #833 where sources metadata was unavailable during streaming.

CleanShot 2025-11-14 at 10 11 21@2x

Weird that the clock is used as source, isn't? 🤔

### DIFF
--- a/examples/openai/agent-stream-sources.php
+++ b/examples/openai/agent-stream-sources.php
@@ -1,0 +1,49 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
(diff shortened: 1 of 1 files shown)


chr-hertel:
> Weird that the clock is used as source, isn't? 🤔

it makes sense since the model doesn't know what date it is and needs that for getting an of "latest".

chr-hertel:
this example def proofs that there is still an issue:
image

but that's not the fault of this PR

chr-hertel (approved):

chr-hertel:
Thank you @OskarStark.

PR #872, opened 2025-11-14
[Platform][LiteLLM] Add Bridge

welcoMattic:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Docs?         | no 
| Issues        |  
| License       | MIT

This PR adds support for [LiteLLM](https://docs.litellm.ai/docs/) as Bridge

For now, it is a very simple support (only chat), see it as ready-to-evolve bridge.

I've added a litellm service in Docker Compose file of examples to easily test it without having to install LiteLLM locally on your host.

### DIFF
--- a/examples/compose.yaml
+++ b/examples/compose.yaml
@@ -188,6 +188,16 @@ services:
             - '8080:8080'
             - '50051:50051'
 
+    litellm:
+        image: ghcr.io/berriai/litellm:v1.79.1-stable
+        ports:
+            - "4000:4000"
+        volumes:
+            - ./litellm/config.yaml:/app/config.yaml
+        env_file:
+            - .env
+        command: [ "--config", "/app/config.yaml", "--port", "4000", "--num_workers", "8" ]
+
 volumes:
     typesense_data:
     etcd_vlm:
--- a/examples/litellm/config.yaml
+++ b/examples/litellm/config.yaml
@@ -1,0 +1,5 @@
+model_list:
+    - model_name: mistral-small-latest
+      litellm_params:
+          model: mistral/mistral-small-latest
+          api_key: "os.environ/MISTRAL_API_KEY"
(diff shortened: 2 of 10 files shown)


chr-hertel:
Hey, thanks for that, that's actually an interesting one - the feature of litellm is to be a generic bridge to whatever based on the original OpenAI contract - right?
AFAIK it adds a lot of features like costs, telemetry, fallback on top - but it's basically as generic as it could get.

So i think we should work on the design shift here more - in case you're up for that.

For example the contract got centralized already - and as long as the bridge uses that original openai compliant contract, no additional normalizers need to be registered with the platform. i think the same should apply for some kind generic endpoint handling (model client + result converter)

so i think it would be great that we could use generic services (like litellm, openrouter, replicate) that conform to that first openai contract from payload and endpoint handling point of view without even installing a bridge.

WDYT?

(if that's off-scope for your current context, we can go ahead with this, merge it for now and potentially remove it later in favor of a generalized/centralized solution)

examples/litellm/config.yaml
```diff
@@ -0,0 +1,5 @@
+model_list:
+    - model_name: mistral-small-latest
+      litellm_params:
+          model: mistral/mistral-small-latest
+          api_key: "os.environ/MISTRAL_API_KEY"
```
OskarStark, line 5: 👀 🫢 🐒

OskarStark: Can't we append the `os.environ/` part internally if its not provided?

welcoMattic: It refers to the .env file that must contains `MISTRAL_API_KEY`

examples/compose.yaml
```diff
@@ -188,6 +188,16 @@ services:
...
             - '50051:50051'
 
+    litellm:
+        image: ghcr.io/berriai/litellm:v1.79.1-stable
+        ports:
+            - "4000:4000"
+        volumes:
+            - ./litellm/config.yaml:/app/config.yaml
+        env_file:
+            - .env
```
welcoMattic, line 198: `.env.local` can be added here to read your own keys if needed.

OskarStark:
Thank you @welcoMattic.

PR #863, opened 2025-11-13
[AI Bundle] Autoregister Vectorizer services

MolloKhan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Docs?        | no
| Issues        | Fix #859
| License       | MIT

PR #874, opened 2025-11-14
[AI Bundle] Add more alias for argument: Indexer and Vectorizer

lyrixx:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        |
| License       | MIT

### DIFF
--- a/src/ai-bundle/src/AiBundle.php
+++ b/src/ai-bundle/src/AiBundle.php
@@ -91,6 +91,7 @@ use Symfony\AI\Store\Bridge\SurrealDb\Store as SurrealDbStore;
 use Symfony\AI\Store\Bridge\Typesense\Store as TypesenseStore;
 use Symfony\AI\Store\Bridge\Weaviate\Store as WeaviateStore;
 use Symfony\AI\Store\Document\Vectorizer;
+use Symfony\AI\Store\Document\VectorizerInterface;
 use Symfony\AI\Store\Indexer;
 use Symfony\AI\Store\IndexerInterface;
 use Symfony\AI\Store\StoreInterface;
@@ -1627,7 +1628,9 @@ final class AiBundle extends AbstractBundle
             new Reference('logger', ContainerInterface::IGNORE_ON_INVALID_REFERENCE),
         ]);
         $vectorizerDefinition->addTag('ai.vectorizer', ['name' => $name]);
-        $container->setDefinition('ai.vectorizer.'.$name, $vectorizerDefinition);
+        $serviceId = 'ai.vectorizer.'.$name;
+        $container->setDefinition($serviceId, $vectorizerDefinition);
+        $container->registerAliasForArgument($serviceId, VectorizerInterface::class, (new Target((string) $name))->getParsedName());
     }
 
     /**
@@ -1656,7 +1659,9 @@ final class AiBundle extends AbstractBundle
         ]);
         $definition->addTag('ai.indexer', ['name' => $name]);
 
-        $container->setDefinition('ai.indexer.'.$name, $definition);
+        $serviceId = 'ai.indexer.'.$name;
+        $container->setDefinition($serviceId, $definition);
+        $container->registerAliasForArgument($serviceId, IndexerInterface::class, (new Target((string) $name))->getParsedName());
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/ai-bundle/src/AiBundle.php
```diff
@@ -1612,7 +1613,9 @@ private function processVectorizerConfig(string $name, array $config, ContainerB
             new Reference('logger', ContainerInterface::IGNORE_ON_INVALID_REFERENCE),
         ]);
         $vectorizerDefinition->addTag('ai.vectorizer', ['name' => $name]);
-        $container->setDefinition('ai.vectorizer.'.$name, $vectorizerDefinition);
+        $serciceId = 'ai.vectorizer.'.$name;
```
```suggestion
        $serviceId = 'ai.vectorizer.'.$name;
```
same for the others

src/ai-bundle/src/AiBundle.php
```diff
@@ -1641,7 +1644,9 @@ private function processIndexerConfig(int|string $name, array $config, Container
         ]);
         $definition->addTag('ai.indexer', ['name' => $name]);
 
-        $container->setDefinition('ai.indexer.'.$name, $definition);
+        $serciceId = 'ai.indexer.'.$name;
```
```suggestion
        $serviceId = 'ai.indexer.'.$name;
```
same for the others

lyrixx:
Thank you, Oskar, for your review. Sorry for letting such a stupid mistake slip into the PR!

OskarStark:
> Thank you, Oskar, for your review. Sorry for letting such a stupid mistake slip into the PR!

Ah no worries 😄 but can you add some tests to assert the container has the defined aliases? Thanks

src/ai-bundle/src/AiBundle.php
```diff
@@ -1612,7 +1613,9 @@ private function processVectorizerConfig(string $name, array $config, ContainerB
             new Reference('logger', ContainerInterface::IGNORE_ON_INVALID_REFERENCE),
         ]);
         $vectorizerDefinition->addTag('ai.vectorizer', ['name' => $name]);
-        $container->setDefinition('ai.vectorizer.'.$name, $vectorizerDefinition);
+        $serviceId = 'ai.vectorizer.'.$name;
+        $container->setDefinition($serviceId, $vectorizerDefinition);
+        $container->registerAliasForArgument($serviceId, VectorizerInterface::class, (new Target((string) $name))->getParsedName());
```
chr-hertel, line 1633: For agent, multiagent and store we use that as part of the target, like
```suggestion
        $container->registerAliasForArgument($serviceId, VectorizerInterface::class, (new Target($name.'Vectorizer'))->getParsedName());
```

makes sense to be consistent here I'd say

src/ai-bundle/src/AiBundle.php
```diff
@@ -1641,7 +1644,9 @@ private function processIndexerConfig(int|string $name, array $config, Container
         ]);
         $definition->addTag('ai.indexer', ['name' => $name]);
 
-        $container->setDefinition('ai.indexer.'.$name, $definition);
+        $serviceId = 'ai.indexer.'.$name;
+        $container->setDefinition($serviceId, $definition);
+        $container->registerAliasForArgument($serviceId, IndexerInterface::class, (new Target((string) $name))->getParsedName());
```
chr-hertel, line 1664: same here, or am i off track?

src/ai-bundle/src/AiBundle.php
lyrixx: I saw some of the previous code. In my humble opinion, they are not useful; they are even harmful.

If I use `foobar` as the name of my vectoriser, I would expect to use `foobar` in the `#[Target()]` attribute. Not `foobarVectorizer`. The suffix doesn't provide any additional "namespacing", since the target relies on the name and the type. Without reading the code or running a debug command, I wouldn't know about the suffix.

So I'm in favour of removing all suffixes everywhere.

chr-hertel: for me the most important part here is consistence - i usually opt-out from this feature and use `#[Autowire]`, which makes it harder for me to have an opinion.

cc @OskarStark @Guikingone do you recall why we went with the suffix?

Guikingone: Yes, we removed the suffix (here's the example for store's): 

```php

# ...
$container->setDefinition('ai.store.'.$type.'.'.$name, $definition);
$container->registerAliasForArgument('ai.store.'.$type.'.'.$name, StoreInterface::class, $name);
$container->registerAliasForArgument('ai.store.'.$type.'.'.$name, StoreInterface::class, $type.'_'.$name);
```

It's the same approach as for HttpClient's, no suffix, just the name of the client.

chr-hertel: Ah okay, than it is already inconsistent :laughing: 

Created #878 as follow up

MolloKhan:
@lyrixx you can copy my test from #863 I think

lyrixx:
Pr updated. Thanks for the review

OskarStark:
Thank you @lyrixx.

PR #876, opened 2025-11-14
[Platform] Unify options key for structured output to `response_format`

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | 
| License       | MIT

it is now possible to just use the option `response_format` for everything instead of differentiating between `output_structure` for classes and `response_format` for concrete array structures.


OskarStark:
Thank you Christopher.

PR #634, opened 2025-09-18
[Agent][Brave] Add `symfony/ai-brave-tool`

OskarStark:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | Closes #625 
| License       | MIT

### Needs
- [ ] A new repo `symfony/ai-brave-tool`
- [ ] Add Packagist configuration
- [ ] Create a new label `Brave`
- [x] Create a recipe in `symfony/recipes`
- [ ] Merge the recipe https://github.com/symfony/recipes/pull/1451

PR #879, opened 2025-11-15
[Tools] Extract third party tools to extra component

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | 
| License       | MIT

Only draft for discussion - basically different approach to #634


chr-hertel:
Nope, looking at it, we should go with #634

PR #903, opened 2025-11-19
[Store][MariaDB] Use UUID type instead of Binary

lyrixx:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        |
| License       | MIT

Binary type is really painful
It needs to be casted to be able to read it. There is no native function
to do that in MariaDB. And except for HUGE amount of data, it does not
make a big deference from char(36). I mean, there is a trade off to find
between perf and DX.

ANYWAY, MariaDB comes from a native UUID type that solves all issues!
Let's use it.


chr-hertel:
Thank you @lyrixx.

PR #904, opened 2025-11-19
[Store][AiBundle] Add support for MariaDB

lyrixx:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        |
| License       | MIT


chr-hertel (approved):
Based on the assumption to you doctrine dbal with bundle - but that works for the most common cases right away i guess, and we can make it more flexible with PDO as a follow up.

chr-hertel:
Thank you @lyrixx.

PR #907, opened 2025-11-19
[Chat] Introduce `MongoDb` as message store

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | #16 
| License       | MIT


src/chat/src/Bridge/MongoDb/MessageStore.php
```diff
@@ -0,0 +1,76 @@
...
+        private readonly SerializerInterface&NormalizerInterface&DenormalizerInterface $serializer = new Serializer([
+            new ArrayDenormalizer(),
+            new MessageNormalizer(),
+        ], [new JsonEncoder()]),
+    ) {
+    }
+
+    public function setup(array $options = []): void
+    {
+        $database = $this->client->selectDatabase($this->databaseName);
```
GromNaN: We will deprecate `selectDatabase`, `getDatabase` exists since `mongodb/mongodb: 1.21`, which is the version required.
```suggestion
        $database = $this->client->getDatabase($this->databaseName);
```

src/chat/src/Bridge/MongoDb/MessageStore.php
```diff
@@ -0,0 +1,76 @@
...
+        ));
+    }
+
+    public function load(): MessageBag
+    {
+        $currentCollection = $this->client->getCollection($this->databaseName, $this->collectionName);
+
+        return new MessageBag(...array_map(
+            fn (\ArrayObject $message): MessageInterface => $this->serializer->denormalize($message->getArrayCopy(), MessageInterface::class),
+            $currentCollection->find()->toArray(),
```
GromNaN: I'm surprised there is no filter or sort order, but that seems to be how the interface is designed.

src/chat/src/Bridge/MongoDb/MessageStore.php
```diff
@@ -0,0 +1,76 @@
...
+            'q' => [],
+        ]);
+    }
+
+    public function save(MessageBag $messages): void
+    {
+        $currentCollection = $this->client->getCollection($this->databaseName, $this->collectionName);
+
+        $currentCollection->insertMany(array_map(
+            fn (MessageInterface $message): array => $this->serializer->normalize($message),
```
GromNaN: The `id` field should be mapped to `_id`.

src/chat/src/Bridge/MongoDb/MessageStore.php
Guikingone: In fact, we can sort them thanks to the `added_at` key (which store a timestamp) but by design, no sorting here, maybe something to improve @chr-hertel @OskarStark ?

src/chat/src/Bridge/MongoDb/MessageStore.php
Guikingone: Fixed 👍🏻

src/chat/src/MessageNormalizer.php
```diff
@@ -107,7 +107,7 @@ public function normalize(mixed $data, ?string $format = null, array $context =
         }
 
         return [
-            'id' => $data->getId()->toRfc4122(),
+            $context['_mongodb_identifier'] ?? false ? '_id' : 'id' => $data->getId()->toRfc4122(),
```
GromNaN: You can make it less specific and allow customized id field:
```suggestion
            $context['identifier'] ??  ? 'id' => $data->getId()->toRfc4122(),
```

src/chat/src/Bridge/MongoDb/MessageStore.php
```diff
@@ -0,0 +1,78 @@
...
+        ]);
+    }
+
+    public function save(MessageBag $messages): void
+    {
+        $currentCollection = $this->client->getCollection($this->databaseName, $this->collectionName);
+
+        $currentCollection->insertMany(array_map(
+            fn (MessageInterface $message): array => $this->serializer->normalize($message, context: [
+                '_mongodb_identifier' => true,
```
GromNaN:
```suggestion
                'identifier' => '_id',
```

src/chat/src/Bridge/MongoDb/MessageStore.php
```diff
@@ -0,0 +1,78 @@
...
+        ));
+    }
+
+    public function load(): MessageBag
+    {
+        $currentCollection = $this->client->getCollection($this->databaseName, $this->collectionName);
+
+        return new MessageBag(...array_map(
+            fn (\ArrayObject $message): MessageInterface => $this->serializer->denormalize($message->getArrayCopy(), MessageInterface::class),
+            $currentCollection->find()->toArray(),
```
GromNaN: The default [`typeMap`](https://www.php.net/manual/en/mongodb.persistence.php#mongodb.persistence.typemaps) of the injected client can be anything. You should force the option here, and ensure you get a PHP array. That should avoid the need of acception `ArrayObject` in the denormalization.
```suggestion
            $currentCollection->find([], ['typeMap' => ['root' => 'array', 'document' => 'array', 'array' => 'array']])->toArray(),
```

Guikingone: Wow, didn't know that you can force the type here, I'll investigate this one for future usages.

GromNaN (approved):
Great job! Thank you.

src/chat/src/Bridge/MongoDb/MessageStore.php
GromNaN: If you sort by `added_at`, you would have to create an index.

OskarStark (approved):

chr-hertel:
Thank you @Guikingone.

PR #908, opened 2025-11-19
[Demo][Website] Add demo for structured output about generating recipes

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | 
| License       | MIT

also re-sorting the demos and their descriptions

[Screencast from 2025-11-19 21-51-58.webm](https://github.com/user-attachments/assets/22e926b4-7f05-42a1-9e23-57c61f847439)


demo/src/Recipe/Data/Recipe.php
```diff
@@ -0,0 +1,68 @@
...
+     * @var int Duration in minutes
+     */
+    #[With(minimum: 5, maximum: 240)]
+    public int $duration;
+
+    /**
+     * @var string Difficulty level of the recipe
+     */
+    #[With(enum: ['Beginner', 'Intermediate', 'Advanced'])]
+    public string $level;
```
OskarStark, line 33: What about using a real enum for this property to showcase?

chr-hertel: didn't work right away - not sure what we currently support there

PR #911, opened 2025-11-20
[Store] Add a way to bypass document loading in Indexer

### LINKED ISSUE #875: [Store] Indexer is nice, but hard to use directly
Hello.

It looks good. 

```php
#[AsMessageHandler()]
{
    public function __construct(
        // #[Target('crawl_url')]
    ) {
    }

    {
            id: new Uuid($crawlUrl->id),
            content: $crawlUrl->body,
            metadata: new Metadata([
                // My stuff here
            ]),
        ));
    }
}
```

lyrixx:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         |
| Issues        | Fix #875
| License       | MIT

### DIFF
--- a/src/store/src/Indexer.php
+++ b/src/store/src/Indexer.php
@@ -26,12 +26,18 @@ use Symfony\AI\Store\Document\VectorizerInterface;
 class Indexer implements IndexerInterface
 {
     /**
-     * @var array<string|null>
+     * @var array<string>
      */
     private array $sources = [];
 
     /**
+     * @var array<EmbeddableDocumentInterface>
+     */
+    private array $documents = [];
+
+    /**
      * @param string|array<string>|null $source       Source identifier(s) for data loading (file paths, URLs, etc.)
+     * @param EmbeddableDocumentInterface|array<EmbeddableDocumentInterface>|null $document       Document or array of documents to process directly, bypassing the loading step
      * @param FilterInterface[]         $filters      Filters to apply after loading documents to remove unwanted content
      * @param TransformerInterface[]    $transformers Transformers to mutate documents after filtering (chunking, cleaning, etc.)
      */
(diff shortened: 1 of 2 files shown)


lyrixx:
Before fixing test & co, is this the way to go?

src/store/src/Indexer.php
```diff
@@ -40,32 +46,47 @@ public function __construct(
         private VectorizerInterface $vectorizer,
         private StoreInterface $store,
         string|array|null $source = null,
+        EmbeddableDocumentInterface|array|null $document = [],
```
OskarStark, line 49:
```suggestion
        EmbeddableDocumentInterface|array|null $documents = [],
```
?

src/store/src/Indexer.php
```diff
@@ -40,32 +46,47 @@ public function __construct(
...
+            $this->documents = [];
+        } elseif ($document instanceof EmbeddableDocumentInterface) {
+            $this->documents = [$document];
+        } else {
+            $this->documents = $document;
+        }
     }
 
-    public function withSource(string|array $source): self
+    public function withSource(EmbeddableDocumentInterface|string|array $source): self
```
OskarStark, line 64: Why should we accept the EmbeddableDocumentInterface here too?

src/store/src/Indexer.php
```diff
@@ -40,32 +46,47 @@ public function __construct(
...
     {
-        return new self($this->loader, $this->vectorizer, $this->store, $source, $this->filters, $this->transformers, $this->logger);
+        return new self($this->loader, $this->vectorizer, $this->store, $source, [], $this->filters, $this->transformers, $this->logger);
+    }
+
+    public function withDocument(EmbeddableDocumentInterface|array $document): self
```
OskarStark, line 69:
```suggestion
    public function withDocuments(array $documents): self
```
?

OskarStark:
Lets wait for @chr-hertel here

chr-hertel:
with `LoaderInterface::load(?string $source, array $options = []): iterable` we have something that provides documents, you have documents at hand.

without a source provided, that implementation will be called with `$source: null` - could you use that?

---

but that's basically just another idea, but i just struggle with state in constructors - Oskar knows about that flaw in my mind already :D

I think that solution is alright here, I would just omit the `EmbeddableDocumentInterface` in the withSource or drop the `withDocuments` - having both feels ambiguous to me - or is it not?

lyrixx:
> but that's basically just another idea, but i just struggle with state in constructors - Oskar knows about that flaw in my mind already :D

Actually, I don't like at all the state in this class.

What do you think about :

```diff
diff --git a/src/store/src/Indexer.php b/src/store/src/Indexer.php
index a79523e3..9e1e2c50 100644
--- a/src/store/src/Indexer.php
+++ b/src/store/src/Indexer.php
@@ -25,19 +25,8 @@ use Symfony\AI\Store\Document\VectorizerInterface;
  */
 class Indexer implements IndexerInterface
 {
-    /**
-     * @var array<string>
-     */
-    private array $sources = [];
 
     /**
-     * @var array<EmbeddableDocumentInterface>
-     */
-    private array $documents = [];
-
-    /**
-     * @param string|array<string>|null $source       Source identifier(s) for data loading (file paths, URLs, etc.)
-     * @param EmbeddableDocumentInterface|array<EmbeddableDocumentInterface>|null $document       Document or array of documents to process directly, bypassing the loading step
      * @param FilterInterface[]         $filters      Filters to apply after loading documents to remove unwanted content
      * @param TransformerInterface[]    $transformers Transformers to mutate documents after filtering (chunking, cleaning, etc.)
      */
@@ -45,31 +34,17 @@ class Indexer implements IndexerInterface
         private LoaderInterface $loader,
         private VectorizerInterface $vectorizer,
         private StoreInterface $store,
-        string|array|null $source = null,
-        EmbeddableDocumentInterface|array|null $document = [],
         private array $filters = [],
         private array $transformers = [],
         private LoggerInterface $logger = new NullLogger(),
     ) {
-        $this->sources = null === $source ? [] : (array) $source;
-        if (null === $document) {
-            $this->documents = [];
-        } elseif ($document instanceof EmbeddableDocumentInterface) {
-            $this->documents = [$document];
-        } else {
-            $this->documents = $document;
-        }
     }
 
-    public function withSource(EmbeddableDocumentInterface|string|array $source): self
+    public function loadAndIndex(?array $sources = null, array $options = []): void
     {
     }
 
-    public function withDocument(EmbeddableDocumentInterface|array $document): self
-    {
-    }
-
-    public function index(array $options = []): void
+    public function index(array $documents, array $options = []): void
     {
     }
 
```

1. no more global state in the class
2. `index()` method always need data to run (the array could be empty tho)
3. new function `loadAndIndex()` takes care of loading the data (specific $source can be passed), and the run the indexation

lyrixx:
Okay, I re-started from scratch in https://github.com/symfony/ai/pull/997
The new code is better than this PR, and I also think it's cleaner than `main`.

I don't like when some state lives in a service. It's almost always source of troubles.

PR #910, opened 2025-11-20
[Platform] `OllamaApiCatalog` improved

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | yes
| Issues        | None
| License       | MIT

PR #915, opened 2025-11-20
[Platform] Add new OpenRouter models, ModelApiCatalog and openrouter/auto + presets

lochmueller:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | 
| License       | MIT

Features...
- Update the static ModelCatalog to the current list of models - New models and embeddings are added (e.g. ChatGPT 5.1, Gemini 3)
- Add a ModelApiCatalog to call the OpenRouter API to handle the current list of Models
- Add openrouter/auto and presets for ModelCatalog and ModelApiCatalog

### DIFF
--- a/src/platform/src/Bridge/OpenRouter/ModelApiCatalog.php
+++ b/src/platform/src/Bridge/OpenRouter/ModelApiCatalog.php
@@ -1,0 +1,122 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Platform\Bridge\OpenRouter;
+
+use Symfony\AI\Platform\Capability;
+use Symfony\AI\Platform\Exception\InvalidArgumentException;
+use Symfony\AI\Platform\Model;
+use Symfony\Contracts\HttpClient\HttpClientInterface;
+
+/**
+ * @author Tim Lochmüller <tim@fruit-lab.de>
+ */
+final class ModelApiCatalog extends AbstractOpenRouterModelCatalog
+{
+    protected bool $modelsAreLoaded = false;
+
+    public function __construct(
+        private readonly HttpClientInterface $httpClient,
+    ) {
+        parent::__construct();
+    }
+
+    public function getModel(string $modelName): Model
+    {
+        $this->preloadRemoteModels();
+
+        return parent::getModel($modelName);
+    }
+
+    public function getModels(): array
+    {
+        $this->preloadRemoteModels();
+
+        return parent::getModels();
+    }
+
+    protected function preloadRemoteModels(): void
+    {
+        if (!$this->modelsAreLoaded) {
+            $this->models = [
+                ...$this->models,
+                ...$this->fetchRemoteModels(),
+                ...$this->fetchRemoteEmbeddings(),
+            ];
+            $this->modelsAreLoaded = true;
+        }
+    }
+
+    /**
+     * @return iterable<string, array{class: class-string<Model>, capabilities: list<Capability::*>}>
+     */
+    protected function fetchRemoteModels(): iterable
+    {
+        $responseModels = $this->httpClient->request('GET', 'https://openrouter.ai/api/v1/models');
+        foreach ($responseModels->toArray()['data'] as $model) {
+            $capabilities = [];
+
+            foreach ($model['architecture']['input_modalities'] as $inputModality) {
+                switch ($inputModality) {
+                    case 'text':
+                        $capabilities[] = Capability::INPUT_TEXT;
+                        break;
+                    case 'image':
+                        $capabilities[] = Capability::INPUT_IMAGE;
+                        break;
+                    case 'audio':
+                        $capabilities[] = Capability::INPUT_AUDIO;
+                        break;
+                    case 'file':
+                        $capabilities[] = Capability::INPUT_PDF;
+                        break;
+                    case 'video':
+                        $capabilities[] = Capability::INPUT_MULTIMODAL; // Video?
+                        break;
+                    default:
+                        throw new InvalidArgumentException('Unknown model '.$inputModality.' input modality.', 1763717587);
+                }
+            }
+
+            foreach ($model['architecture']['output_modalities'] as $outputModality) {
+                switch ($outputModality) {
+                    case 'text':
+                        $capabilities[] = Capability::OUTPUT_TEXT;
+                        break;
+                    case 'image':
+                        $capabilities[] = Capability::OUTPUT_IMAGE;
+                        break;
+                    default:
+                        throw new InvalidArgumentException('Unknown model '.$outputModality.' output modality.', 1763717588);
+                }
+            }
+
+            yield $model['id'] => [
+                'class' => Model::class,
+                'capabilities' => $capabilities,
+            ];
+        }
+    }
+
+    /**
+     * @return iterable<string, array{class: class-string<Embeddings>, capabilities: list<Capability::*>}>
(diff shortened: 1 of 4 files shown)


lochmueller:
Or do we prefer to integrate an external model selection directly into the catalog (like here: https://github.com/symfony/ai/pull/910/files )? I try to avoid external API calls in the model catalog... :-/

OskarStark:
Can we move it to the platform in bin/ folder?

chr-hertel:
> Can we move it to the platform in bin/ folder?

should be even part of the OpenRouter bridge i'd say

chr-hertel:
we'd need another solution in the long run here - i would prefer some kind of catalog that is dynamic with a cache.

updating with the code base doesn't really solve the issue here - even if we'd make it part of the pipeline to do it automatically on release or something. the change frequency requires decoupling lib updates and catalog updates.

not scope of this PR maybe, but we'll face that issue over and over again - with multiple bridges.

OskarStark:
> should be even part of the OpenRouter bridge I'd say

yeah, thats what I meant

OskarStark:
We could have libs just for the models getting regular automated updates, but that's future :-)

lochmueller:
OK.... thanks for your feedback @OskarStark & @chr-hertel . how could I adapt the PR for this future need? 

My thoughts on this: I will drop the example script, I will drop the static model list in the ModelCatalog, move the fetch logic for the models and embeddings to the ModelCatalog and add some kind of PSR-6 compatible cache logic to keep the requests to the API endpoint low. I will rework the PR today...

OskarStark:
you can have a look at the ollama api catalog. next step would be to add proper caching to those api catalogs

lochmueller:
So... I created a ModelApiCatalog that fetch the models and embeddings via API (incl. capabilities). The default catalog is still the default ModelCatalog that contains only the Embeddings (because they are created with another Model class) but create the models "on the fly" without capabilities (because they are not used in the ModelClient/ResultConverter). I also switch from json_decode functions to Serializer->JsonDecoder.

There is no cache currently, but I think we should create some general interface (or even abstract integration incl. cache logic) for "remote fetched model information", because several bridges need this (Ollama, AiMlApi, Albert etc.) and every Bridge integrated this in another way currently...

lochmueller:
I will check the broken pipelines shortly...

src/platform/src/Bridge/OpenRouter/ModelCatalog.php
```diff
@@ -23,2709 +24,197 @@ final class ModelCatalog extends AbstractModelCatalog
...
-    {
+    public function __construct(
+        array $additionalModels = [],
+    ) {
+        // OpenRouter provides access to many different models from various providers
+        // This catalog only contains embeddings as default models
+        // All regular models are added dynamically in the format provider/model als Models.
+        // For a full and up-2-date list of models incl. capabilities, use the ModelApiCatalog
         $defaultModels = [
-            // OpenRouter provides access to many different models from various providers
```
OskarStark, line 29: Please lets keep the static catalog for the ones who don't want to use api catalog with extra api calls, it should be opt in in the AiBundle, like done for Ollama.

https://github.com/symfony/ai/pull/766

lochmueller: I added the current list (like in the first commit) incl. a date in the comment, so the user know from which date the current list is.

src/platform/src/Bridge/OpenRouter/ModelApiCatalog.php
```diff
@@ -0,0 +1,106 @@
...
+        // Models are fetched via OpenRouter meta API
+        $this->models = $this->fetchRemoteModels();
+    }
+
+    /**
+     * @return array<string, array{class: string, capabilities: list<Capability>}>
+     */
+    protected function fetchRemoteModels(): array
+    {
+        // Rework based on Olama Model API
```
chr-hertel: not really meaningful to be present in the code base itself - and the implementation differs

src/platform/src/Bridge/OpenRouter/ModelApiCatalog.php
```diff
@@ -0,0 +1,106 @@
...
+
+        $fullResult = [];
+
+        $serializer = new Serializer(encoders: [new JsonEncoder()]);
+
+        // Fetch models
+        $responseModels = $this->httpClient->request('GET', 'https://openrouter.ai/api/v1/models');
+        $models = $serializer->decode($responseModels->getContent(), 'json');
```
chr-hertel: why using the serializer here over just `$responseModels->toArray()`?

src/platform/src/Bridge/OpenRouter/ModelApiCatalog.php
lochmueller: Sorry... old comment. Removed.

src/platform/src/Bridge/OpenRouter/ModelApiCatalog.php
lochmueller: Don't know/saw this function. Serializer is removed now.

src/platform/src/Bridge/OpenRouter/ModelApiCatalog.php
```diff
@@ -0,0 +1,97 @@
...
+use Symfony\AI\Platform\Capability;
+use Symfony\AI\Platform\Exception\InvalidArgumentException;
+use Symfony\AI\Platform\Model;
+use Symfony\AI\Platform\ModelCatalog\FallbackModelCatalog;
+use Symfony\Contracts\HttpClient\HttpClientInterface;
+
+/**
+ * @author Tim Lochmüller <tim@fruit-lab.de>
+ */
+final class ModelApiCatalog extends FallbackModelCatalog
```
OskarStark: I just merged
* #910 

We should here as well not depend on the fallback catalog and implement the interface directly

src/platform/src/Bridge/OpenRouter/ModelCatalog.php
OskarStark: works for me, lets move it a) to the class description, b) use @see to reference the ModelApiCatalog and c) make the last updated date a little bit more outstanding like:
```
description description
description description
description description

Last generated: ........
```

src/platform/src/Bridge/OpenRouter/ModelApiCatalog.php
lochmueller: Hey @OskarStark 
the FallbackModelCatalog was wrong and I changed it to an own abstraction: https://github.com/lochmueller/ai/commit/515b3a04fd326157f0a16307001f81de0a919c12
- Use an own abstraction to handle the `openrouter/auto` and preset model handling in both ModelCatalogs
- Based on the core Abstraction (not the Fallback anymore), because I need the `parseModelName` to split model name and openrouter modifications like :floor :nitro :exactor etc.
- Also switch the fetch functions in the ModelApiCatalog to iterable

Regards,
Tim

chr-hertel (changes requested):
one thing open for me in the `ModelApiCatalog` before merging it:

the constructor cannot to http calls - that's too much of a workload for a `new` => i know it's a bit annoying with that abstraction, but we need to shift that to a later point in time => getters maybe?

OskarStark:
Pleas update PR title + body, thanks

lochmueller:
> one thing open for me in the `ModelApiCatalog` before merging it:
> 
> the constructor cannot to http calls - that's too much of a workload for a `new` => i know it's a bit annoying with that abstraction, but we need to shift that to a later point in time => getters maybe?

Done. Move the loading of the model list to getModel/getModels incl. a flag to avoid double loading of the list.

> Pleas update PR title + body, thanks

Also done.

OskarStark:
Thank you @lochmueller.

PR #917, opened 2025-11-20
[AI Bundle][Ollama] Fix configuration when a custom http_client is set

lyrixx:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix https://github.com/symfony/ai/pull/766
| License       | MIT

cc @Guikingone

### DIFF
--- a/src/ai-bundle/src/AiBundle.php
+++ b/src/ai-bundle/src/AiBundle.php
@@ -575,7 +575,7 @@ final class AiBundle extends AbstractBundle
                     ->setLazy(true)
                     ->setArguments([
                         $platform['host_url'],
-                        new Reference('http_client'),
+                        new Reference($platform['http_client']),
                     ]);
 
                 $container->setDefinition('ai.platform.model_catalog.ollama', $catalogDefinition);


src/ai-bundle/src/AiBundle.php
```diff
@@ -575,7 +575,7 @@ private function processPlatformConfig(string $type, array $platform, ContainerB
                     ->setLazy(true)
                     ->setArguments([
                         $platform['host_url'],
-                        new Reference('http_client'),
+                        new Reference($platform['http_client']),
```
lyrixx, line 578: If we don't do that, we use the default HTTP Client.
And I have the following options:
```yaml
ai:
    platform:
        ollama:
            http_client: http_client.ollama
            # Configured in the HTTP Client
            # see https://github.com/symfony/ai/issues/899
            host_url: ''
            api_catalog: true
```

So host_url is empty, so the underlying code will explode because the host URL is not a valid endpoint

Guikingone: If I'm right `host_url` can be omitted (as it fallback to `http://127.0.0.1:11434` by default) 🤔 

```php
                    ->arrayNode('ollama')
                        ->children()
                            ->stringNode('host_url')->defaultValue('http://127.0.0.1:11434')->end()
                            ->stringNode('http_client')
                                ->defaultValue('http_client')
                                ->info('Service ID of the HTTP client to use')
                            ->end()
                            ->booleanNode('api_catalog')
                                ->info('If set, the Ollama API will be used to build the catalog and retrieve models information, using this option leads to additional HTTP calls')
                            ->end()
                        ->end()
                    ->end()
```

lyrixx: But if someone configure a custom host_url, without a custom HTTP client, the one in the configure **must be** used. So it cannot be omitted

chr-hertel:
Thank you @lyrixx.

PR #928, opened 2025-11-21
[Chat] Add streaming support & persistence to storage through accumulator callback chaining

### LINKED ISSUE #922: [Chat][Store] Chat API & storage support for streaming
Good morning,

I am creating an issue on this as I am about to implement this feature myself (today). I already talked through a proposal on Slack (see: https://symfony-devs.slack.com/archives/C09BAPA9BJP/p1763370944783239?thread_ts=1763369735.762229&cid=C09BAPA9BJP)

**Proposed approach**
1. Add a finally {} block to StreamResult->getContent() with an optional callback.
2. When the generator is exhausted, the callback fires and receives the fully accumulated message.
3. Update the Chat API so that it defines this callback and passes it down to the StreamResult during streaming calls.
4. This callback would then be responsible for writing the accumulated message to the StoreInterface once the stream completes.
5. This approach preserves chunk-by-chunk yielding (so streaming behaviour stays intact) and updates the conversation store at the end.
6. Ultimately this would allow the Chat API to support streaming while automatically persisting to the configured storage.

**Limitations:**
This callback functionality & `SessionStorage` will not work in unison. Due to the nature of how session/header lifecycles work in Symfony, it seems unavoidable that this will cause inherent issues with headers being already sent. This might be circumventable using low-level session functions, but this is very likely to have tons of side-effects in the future. Due to this, I plan to throw an error when it is attempted to combine streaming & `SessionStorage` .

marco-jouwweb:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | No
| Docs?         | No 
| Issues        | Fix #922
| License       | MIT

**Description**
The `Chat::submit()` function didn't support streaming. I have added support for this through:
1. Creating the `AccumulatingStreamResult` wrapper which allows callback chaining and accumulates the full message (callbacks are fired once `Generator` in inner `StreamResult` is exhausted, aka when stream has completed)
2. Wrap `StreamResult` classes with this wrapper inside the `Chat::submit()` function
3. Create a small callback function which updates the `Storage` with a new `AssistantMessage` built from the accumulated message (as non-stream responses are also automatically added to storage in `Chat::submit()`)

**Other changes:**
- Added `StreamableStoreInterface` which indicates a `Store` implementation supports streaming. Currently added to all stores except `SessionStorage` (due to inherent issues with headers being already sent; deferred updates through callbacks go against nature of header/session lifecycles in Symfony; discussed with @Guikingone)
- Fixed bug where metadata was not persisted in `Chat::submit()` (metadata was not transferred to new `AssistantMessage` we created)
- Added tests for streaming and `Chat::submit()` combination


src/agent/src/Toolbox/StreamResult.php
```diff
@@ -33,10 +33,7 @@ public function getContent(): \Generator
             if ($value instanceof ToolCallResult) {
                 $innerResult = ($this->handleToolCallsCallback)($value, Message::ofAssistant($streamedResult));
 
-                // Propagate metadata from inner result to this result
-                foreach ($innerResult->getMetadata()->all() as $key => $metadataValue) {
-                    $this->getMetadata()->add($key, $metadataValue);
-                }
+                $this->getMetadata()->set($innerResult->getMetadata()->all());
```
chr-hertel: this is a behavior change since this is overwriting all metadata instead of extending it, right?

marco-jouwweb: You are right, good catch! :-)

src/agent/src/Toolbox/StreamResult.php
```diff
@@ -37,7 +37,6 @@ public function getContent(): \Generator
                 foreach ($innerResult->getMetadata()->all() as $key => $metadataValue) {
                     $this->getMetadata()->add($key, $metadataValue);
                 }
-
```
OskarStark, line 40: please revert

marco-jouwweb: im sorry im attempting to haha one sec

src/agent/src/Toolbox/StreamResult.php
OskarStark, line 1: please revert the whole file

marco-jouwweb: will do

src/chat/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 CHANGELOG
 =========
+0.2
```
OskarStark: 0.1 is not released yet, so please move it down to 0.1 section, thanks

src/chat/src/Result/AccumulatingStreamResult.php
```diff
@@ -0,0 +1,81 @@
...
+    private ?\Closure $onComplete = null;
+
+    public function __construct(
+        private readonly StreamResult|ToolboxStreamResult $innerResult,
+        ?\Closure $onComplete = null,
+    ) {
+        $this->onComplete = $onComplete;
+    }
+
+    public function addOnComplete(\Closure $callback): void
```
Guikingone: Would renamed it to `onComplete()`

src/chat/src/Result/AccumulatingStreamResult.php
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * @author Marco van Angeren <marco@jouwweb.nl>
+ */
+final class AccumulatingStreamResult
+{
+    private ?\Closure $onComplete = null;
+
+    public function __construct(
+        private readonly StreamResult|ToolboxStreamResult $innerResult,
+        ?\Closure $onComplete = null,
```
Guikingone: Couldn't be declared as a property?

src/chat/CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ CHANGELOG
 0.1
 ---
 
+* Add streaming support to `ChatInterface::submit()`
+   - Add `StreamableStoreInterface` which indicates `StoreInterface` implementation can be configured with streaming 
+   - Add `AccumulatingStreamResult` wrapper class which adds accumulation logic & callback chaining to `StreamResult` implementations (can wrap both `Agent` and `Platform` variants) to return the full message once `Generator` is exhausted
+   - Streamed responses now also create `AssistantMessage` & are added to `Store` in `Chat::submit()`
+   - Bugfixed loss of metadata in `Chat::submit()`
```
Guikingone: Should be added after the list of stores IMO

src/chat/src/Result/AccumulatingStreamResult.php
marco-jouwweb: I assume you meant the `Close $onComplete` through constructor promotion? If so, I've made the change.

src/chat/src/Result/AccumulatingStreamResult.php
marco-jouwweb: Actually, it was not even used as of now. The implementation I wrote passed the callback through the constructor, so I removed the function for now and we can re-add if it is ever necessary.

marco-jouwweb:
Had some trouble with Git as I never worked with a forked repo like this before, hence the force pushes.. However, the branch history should now be fixed, up-to-date & mergeable.

src/chat/src/Chat.php
```diff
@@ -35,18 +39,31 @@ public function initiate(MessageBag $messages): void
...
         $result = $this->agent->call($messages);
 
+        if ($result instanceof StreamResult || $result instanceof ToolboxStreamResult) {
+            if (!$this->store instanceof StreamableStoreInterface) {
+                throw new RuntimeException($this->store::class.' does not support streaming.');
+            }
+
+            return new AccumulatingStreamResult($result, function (AssistantMessage $assistantMessage) use ($messages) {
+                $messages->add($assistantMessage);
+                $this->store->save($messages);
```
Guikingone: Could be done in a single line with `$this->store->save($messages->with($assistantMessage));`

This way, we could use an arrow function.

src/chat/src/Chat.php
```diff
@@ -35,18 +39,31 @@ public function initiate(MessageBag $messages): void
...
     {
         $messages = $this->store->load();
 
         $messages->add($message);
         $result = $this->agent->call($messages);
 
+        if ($result instanceof StreamResult || $result instanceof ToolboxStreamResult) {
+            if (!$this->store instanceof StreamableStoreInterface) {
+                throw new RuntimeException($this->store::class.' does not support streaming.');
+            }
```
Guikingone, line 52: I'm not sure that we should throw an exception here, maybe just returning the complete stream at once?

Or maybe we can just store the message "as it" in the store?

marco-jouwweb: I'm fairly certain throwing is the right call here. Returning the stream "as is" would work, but it's likely to confuse developers and hide the fact that streaming and session storage are incompatible by nature. As a developer, I'd rather get an early exception that forces me to fix a faulty configuration than a silently degraded solution. I think it'll also prevent a lot of GitHub issues down the road.

src/chat/src/Chat.php
marco-jouwweb: Thanks for the review! For future PRs, it'd be helpful to consolidate these kinds of stylistic/opinion-level comments into a single pass when possible. Going through several iterations for non-functional feedback can slow things down. Just a thought though, I'm happy to help. But that'd keep contributors more motivated 🙂

src/chat/tests/ChatTest.php
```diff
@@ -113,4 +117,118 @@ public function testItHandlesEmptyMessageStore()
...
         $this->assertCount(2, $this->store->load());
     }
+
+    public function testItSupportsStreaming()
+    {
+        $userMessage = Message::ofUser('What is your favourite song?');
+        $generator = (function () {
+            yield 'Bitter Sweet';
+            yield ' ';
+            yield 'Symfony';
```
chr-hertel, line 127: i see what you did here - nice :notes: :joy:

chr-hertel:
Hey @marco-jouwweb,

sorry for the delay and thanks for the reminder in Slack - the main issues why this PR is dragging on is that your jumping on a difficult topic :smile: 

Streaming is one of the more mind bending this we have here, and I feel like adding another result _on top_ is not the right way to go - that's mostly a gut feeling, but i would favor going into the lower level to solve this.
What I mean is that we're currently just yielding strings when, but i guess we could structure that better on platform level - ending up with the `StreamResult` itself supporting also a `onComplete` maybe - i like that idea.

regarding the `StreamableStoreInterface` I'm a bit confused - tbh, i would just throw an exception for the session store if used with a stream option - mixing the streaming concern with the store concerns just feels a bit off - especially with an empty interface that doesn't really give it away.

my last point would be: to me it's quite questionable if and how this PR even works - you would need to switch streaming on somehow - which is usually done by `$agent->call($messages, ['stream' => true])` which would be looped through from the `$chat->submit($message , ['stream' => true])` - or am I missing something here?

action items from my point of view:
* drop the `StreamableStoreInterface` and somehow just throw in case streaming is combined with session - well, it might throw itself due to the header issue and we could just document that?
* add an example to demonstrate functionality
* move `onComplete` to lower level of `StreamResult` instead of extra `AccumulatingStreamResult`

Let me know if you have follow up questions - cheers!

chr-hertel:
Hey @marco-jouwweb, we can close this after #1583 got merged?

PR #871, opened 2025-11-14
[Platform] Migrate OpenAI to Responses API

paulinevos:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | Fix #796
| License       | MIT

This PR migrates OpenAI from chat completions (deprecated) to the Responses API. I tried to keep the scope small and just support everything already in place as seamlessly as possible. The Responses API does offer some more functionality that we should look into supporting (e.g. MCP tool calling, expanded reasoning functionality), but I found that OOS for this migration. 

One thing to note it that the Responses API as of now lacks Audio capabilities ("coming soon", according to the docs), so that example is temporarily disabled and so is the `Gpt` audio capability for now. 

**Note: my PRs are best reviewed commit by commit**

Issue #926, opened 2025-11-21
[Chat][Agent][Platform] Rename `StreamResult` classes to have unique names


marco-jouwweb:
I find it really confusing that there are multiple `StreamResult` classes throughout multiple layers of this AI suite. Would it be an idea to at least give them unique names?

This would probably save a lot of future confusion. Especially as someone who hasn't built this, my opinion is that it is confusing. Also, if they serve the same purpose they should likely be unified somehow.

chr-hertel:
Done with #1394

PR #938, opened 2025-11-22
[Platform] Centralize default model clients and converters for OpenAI based standards

### LINKED ISSUE #114: Custom URL ability or generic bridges?

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | **needed, still open**
| Issues        | Fixes #114 
| License       | MIT

We need a short term solution for this issue to unblock merging #871 without destroying a couple of bridges that are currently coupled to that "de facto standard" of OpenAI-based contracts.

This also will make it more obvious that our current ModelCatalog + Model class + neglected Capabilities need some love.

Introducing
* Explicit base models `Symfony\AI\Platform\Model\CompletionsModel` and `Symfony\AI\Platform\Model\EmbeddingsModel`
* Moving `ResultConverterInterface` into new `Symfony\AI\ResultConverter` namespace
* Moving `ModelClientInterface` into new `Symfony\AI\ModelClient` namespace
* Introducing default, OpenAI-based
  * `Symfony\AI\Platform\ModelClient\CompletionsModelClient`
  * `Symfony\AI\Platform\ModelClient\EmbeddingsModelClient`
  * `Symfony\AI\Platform\ResultConverter\CompletionsResultConverter`
  * `Symfony\AI\Platform\ResultConverter\EmbeddingsResultConverter`

Leads to removal of unspecific usage of OpenAI bridge implementations and basic Model subclasses in favor of specific ones if possible.


src/platform/src/Default/Embeddings/ModelClient.php
```diff
@@ -9,36 +9,39 @@
...
  */
-final class EmbeddingsModelClient implements ModelClientInterface
+class ModelClient implements ModelClientInterface
 {
     public function __construct(
-        private readonly HttpClientInterface $httpClient,
         #[\SensitiveParameter] private readonly string $apiKey,
         private readonly string $baseUrl,
+        private readonly string $path = '/v1/embeddings',
+        private readonly ? HttpClientInterface $httpClient = null,
```
OskarStark:
```suggestion
        private readonly ?HttpClientInterface $httpClient = null,
```

chr-hertel:
Alternative namespace structure could be

* `Symfony\AI\Platform\Completions\{Model, ModelClient, ResultConverter}`
* `Symfony\AI\Platform\Embeddings\{Model, ModelClient, ResultConverter}`

lochmueller:
> Alternative namespace structure could be
> 
>     * `Symfony\AI\Platform\Completions\{Model, ModelClient, ResultConverter}`
> 
>     * `Symfony\AI\Platform\Embeddings\{Model, ModelClient, ResultConverter}`

I like the current structure in the PR (and not the suggestion). Keep in mind, that the Completions API is "old" and the new Responses API would be the "next" one, that need to fit in the namespace structure. Perhaps it is easier to understand, if there is NO "Responses" folder directly in the platform component, because this sound very general.

I would prefer ...

- Symfony\AI\Platform\Model\ResponsesModel
- Symfony\AI\Platform\ModelClient\ResponsesModelClient
- Symfony\AI\Platform\ResultConverter\ResponsesResultConverter

instead of...

- Symfony\AI\Platform\Responses\Model
- Symfony\AI\Platform\Responses\ModelClient
- Symfony\AI\Platform\Responses\ResultConverter

Additionally, in the suggested approach all classes would have the same name again (without namespace), and when the namespaces are imported in the concrete implementations, it's not immediately clear which object is currently being used (#autocomplete). There are already issues related to "all classes have the same name" https://github.com/symfony/ai/issues/926

I'm really looking forward to seeing this patch in the component :) - independent of the class name structure :)

chr-hertel:
Closed in favor of #1052

PR #940, opened 2025-11-22
 Use custom serializer for PlatformSubscriber

### LINKED ISSUE #937: [Platform] StructuredOutput use Serializer to hydrate values but rely on ReflectionProperty for the schema
Hi @chr-hertel @OskarStark 

I just found a bug which is easy to reproduce with
```yaml
framework:
    serializer:
        enabled: true
        enable_attributes: true
        name_converter: 'serializer.name_converter.camel_case_to_snake_case'
```

And an output class like 
```php
class Foo
{
    public function __construct(public string $fooBar) {}
}
```

When using 
```php
$content = $this->agent->call($messages, [
            'response_format' => Foo::class,
])->getContent();
```

The schema is generated using the ReflectionProperty
https://github.com/symfony/ai/blob/5e1d943feb5ad3a882585d8fd3fb0a38e8188d3b/src/platform/src/Contract/JsonSchema/Factory.php#L87-L89 so considered the field as `fooBar`
But the data is hydrated with the serializer 
https://github.com/symfony/ai/blob/5e1d943feb5ad3a882585d8fd3fb0a38e8188d3b/src/platform/src/StructuredOutput/ResultConverter.php#L48
so the serializer name converter is used and it will expect a `foo_bar` key instead of `fooBar`.

Currently the workaround is to do
```php
class Foo
{
    public function __construct(
       #[SerializedName('fooBar')]
       public string $fooBar
    ) {}
}
```

Any clue on how to fix this ? Maybe the PlatformSubscriber shouldn't rely on the default serialzer of Symfony ?

VincentLanglet:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #937
| License       | MIT


src/platform/src/Contract/JsonSchema/Factory.php
```diff
@@ -300,7 +300,6 @@ private function findDiscriminatorMapping(string $className): ?array
          * @see https://github.com/symfony/ai/pull/585#issuecomment-3303631346
          */
         $reflectionProperty = new \ReflectionProperty($result, 'mapping');
-        $reflectionProperty->setAccessible(true);
```
VincentLanglet, line 303: Useless for PHP 8.1+

chr-hertel:
That's what I had in mind as well - even better! :+1: 

The only part missing is maybe the `PropertyNormalizer` - have a look at the recipe bot in the demo, see #908

VincentLanglet:
> The only part missing is maybe the `PropertyNormalizer` - have a look at the recipe bot in the demo, see #908

I'm not sure what I need to solve here with the PropertyNormalizer. Seems like the ObjectNormalizer is already covering it...

I tried to add a test with
```php
final class MathReasoningWithProperties
{
    /**
     * @var Step[] $steps
     */
    public array $steps;

    public string $finalAnswer;

    public float $result;
}
```
and it works fine without PropertyNormalizer.

chr-hertel:
oh, true, with this branch it works as well for that demo without explicitly registering that `PropertyNormalizer` as service ... or maybe i was the one hallucinating the other night :man_shrugging: 

Thanks @VincentLanglet - looks good!

chr-hertel (approved):

chr-hertel:
Thank you @VincentLanglet.

PR #960, opened 2025-11-23
 Ease LiteLLM setup for examples

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | 
| License       | MIT

Trying to get this started as example for #938.

@welcoMattic does that make sense to you? was just wondering if there's a different way than another postgres container

### DIFF
--- a/examples/.env
+++ b/examples/.env
@@ -114,6 +114,7 @@ LMSTUDIO_HOST_URL=http://127.0.0.1:1234
 
 # For using LiteLLM
 LITELLM_HOST_URL=http://127.0.0.1:4000
+LITELLM_API_KEY=sk-1234
 
 # Qdrant (store)
 QDRANT_HOST=http://127.0.0.1:6333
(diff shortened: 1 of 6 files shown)


chr-hertel:
Still cannot make the example `litellm/chat.php` work without API key for litellm - but that's mostly me struggling with litellm

image

welcoMattic:
I'm fixing usage of LiteLLM with ot without apikey (defined by master_key config entry). I'll push it as soon as my network connection allow it from my high-speed train!

welcoMattic:
@chr-hertel I pushed a commit here, it should work (it's ok on my setup)

OskarStark:
rebase needed to fix lowest job

OskarStark:
I rebased and squashed the commits into one to be able to use `gh` tool, I hope you don't mind @welcoMattic 😄

OskarStark:
Thank you Christopher.

PR #963, opened 2025-11-24
[AI Bundle][VertexAI] factory get called with wrong arguments

franzwilding:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| License       | MIT

The vertex ai factory `create´method get called with one wrong "null" parameter. This removes the eventDispatcher, because it would be the 5th argument. Since method signature still is valid with null as 4th parameter, my `AiBundleTest::testExtensionLoadDoesNotThrow` test did not catch it.

### DIFF
--- a/src/ai-bundle/src/AiBundle.php
+++ b/src/ai-bundle/src/AiBundle.php
@@ -507,7 +507,6 @@ final class AiBundle extends AbstractBundle
                     $httpClient,
                     new Reference('ai.platform.model_catalog.vertexai.gemini'),
                     new Reference('ai.platform.contract.vertexai.gemini'),
-                    null,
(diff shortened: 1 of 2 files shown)


OskarStark:
Can you please add a test to avoid regressions in the future?

franzwilding:
@OskarStark I already implemented the test some time ago, but the problem here is, that the platform is able to be created. Anyway. I added another check to see if the number of arguments of the factory method is gte the number of arguments. It is not very robust or elegant, but it catches at least some of the signature-mismatching problems.

OskarStark:
Good catch, thanks Franz.

PR #855, opened 2025-11-10
[AI Bundle] Improvements on configuration / tags

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | #841 
| License       | MIT

Summary:

* Improvements on vector stores tests
* Vector stores definitions sorted in both `options.php` and `AiBundle.php`
* Optional arguments set to required in various cases to prevent errors

PR #964, opened 2025-11-24
[AI Bundle] Improvement on tests

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | First split of #855 
| License       | MIT

### DIFF
--- a/src/ai-bundle/src/AiBundle.php
+++ b/src/ai-bundle/src/AiBundle.php
@@ -960,19 +960,23 @@ final class AiBundle extends AbstractBundle
 
         if ('cache' === $type) {
             foreach ($stores as $name => $store) {
+                $distanceCalculatorDefinition = new Definition(DistanceCalculator::class);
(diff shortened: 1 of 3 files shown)


src/ai-bundle/src/AiBundle.php
OskarStark, line 1: this is no tests change, shall we keep it in this PR?

Guikingone: I can move the changes into a new one except that tests are impacted (as I'm testing the fact that we're dealing with a `Reference` and not a `Definition`) 🤔

OskarStark:
Thank you @Guikingone.

src/ai-bundle/src/AiBundle.php
OskarStark: No need, thanks

PR #967, opened 2025-11-24
[Chat] Fix Doctrine DBAL message store

xprojects-de:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | 
| License       | MIT

If the store is applied to a database that already contains other tables, the schema will not be defined correctly. The existing tables will be defined using a CREATE statement, which will result in an error. The pull request respects the existing structure, including the existing tables, and cleanly adds the new table.


xprojects-de:
I don't know why the Fabbot test is failing. It should be `hasTable`, not `hashTable`... 
And I also don't know why PHP 8.2 fails. The test runs fine locally on my system under 8.2.

OskarStark:
You can ignore this check.

@Guikingone friendly ping

Guikingone:
Ah yes, I'm facing the same error locally on an existing database, I missed this part during tests 😓

src/chat/src/Bridge/Doctrine/DoctrineDbalMessageStore.php
```diff
@@ -52,13 +53,25 @@ public function setup(array $options = []): void
             throw new InvalidArgumentException('No supported options.');
         }
 
-        $schema = $this->dbalConnection->createSchemaManager()->introspectSchema();
+        $schemaManager = $this->dbalConnection->createSchemaManager();
+        $schema = $schemaManager->introspectSchema();
 
         if ($schema->hasTable($this->tableName)) {
```
xprojects-de, line 62: The check could also be omitted. Then it would be possible to adjust the table.

Guikingone: Bu if the table already exist, you're not supposed to add it again / alter it? 🤔 

if would require to being able to passe extra fields during `setup()` calls, for now, it's not supported 🤔

xprojects-de: Yes, that would be exactly my use case :-) I just wanted to do it in two steps and fix the bug first.

It would be really nice if you could specify additional fields in the setup... I can gladly update the program so that's possible too...

xprojects-de: But I think the structure isn't quite there yet, since you would also have to adjust the load and save functions.

OskarStark:
rebase to fix lowest build

OskarStark:
The rebase was wrong

xprojects-de:
oh yes sorry, i did a reflog. now it should be ok :-)

Guikingone:
Hi 👋🏻 

Any plan on rebasing/merging this PR? I'm working on #1356 and I don't want to break something if this PR solve the migration problem 😅

xprojects-de:
I think the migration won't work without the change. 
I'd be happy to merge it :-)

Guikingone:
Friendly ping @chr-hertel @OskarStark (when you have time), as you've merged fixes / improvements, can we work on merging this PR (if @xprojects-de is available to rebase / fix tests if needed) in `0.4` to ensure that Doctrine works as expected? 🙏🏻 

PS: I know that I'm responsible for introducing the bug at first 😅

OskarStark:
Fine after rebase and green CI, or you take over in the worst case

OskarStark:
The rebase looks wrong

xprojects-de:
I´m sorry, i was not able to rebase the PR. SO i will close this one and create a new one https://github.com/symfony/ai/pull/1624...

xprojects-de:
see https://github.com/symfony/ai/pull/1624

PR #1001, opened 2025-11-27
[Platform] Improvements on Uuid handling & `MessageNormalizer`

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | None
| License       | MIT

Summary:

* When using `MessageNormalizer::denormalize()`, it looks like we're recreating a new `Uuid::v7()` in the constructor
* Due to this behavior, when persisting messages, they're receiving a new uuid even on older ones
* This PR introduce a `IdentifierAwareTrait` that expose two methods `getId()` and `withId`, the last one allows to either specify an exiting Uuid or use a `v7` one

### DIFF
--- a/src/chat/src/MessageNormalizer.php
+++ b/src/chat/src/MessageNormalizer.php
@@ -28,6 +28,9 @@ use Symfony\AI\Platform\Result\ToolCall;
 use Symfony\Component\Serializer\Exception\InvalidArgumentException;
 use Symfony\Component\Serializer\Normalizer\DenormalizerInterface;
 use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
+use Symfony\Component\Uid\AbstractUid;
+use Symfony\Component\Uid\TimeBasedUidInterface;
+use Symfony\Component\Uid\Uuid;
 
 /**
  * @author Guillaume Loulier <personal@guillaumeloulier.fr>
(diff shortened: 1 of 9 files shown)


src/platform/src/Metadata/IdentifierAwareTrait.php
```diff
@@ -0,0 +1,34 @@
...
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Platform\Metadata;
```
OskarStark:
```suggestion
namespace Symfony\AI\Platform\Message;
```
?

chr-hertel (changes requested):
My assumption as user would be that a `withId` clones the instance - and that would also help here, right?

we could keep the instantiation of id in the constructor but overwrite if needed with that withId

Guikingone:
Make sense, I was focused on the method and not its usages, to be honest, I was looking at the method last night and said to myself: Damn, that's ugly as hell in the constructor 😅

chr-hertel:
> I was looking at the method last night and said to myself: Damn, that's ugly as hell

well, we've all been there - over and over ...

chr-hertel:
Thank you @Guikingone.

PR #1006, opened 2025-11-27
 Change StoreInterface from array to iterable

### LINKED ISSUE #998: [Store] Change StoreInterface to Iterator
Hey everyone,

I suggest changing the StoreInterface of the Stores from in the `query` method from `array` to `iterable`.
Stores that are NOT based on a single HTTP request would benefit from the iterator running more memory-efficiently (also the creating of the VectorDocument could be more efficiently if we `yield` this one-by-one) , since the entire result wouldn't need to be collected in an array. 

Depending on the use case, iterator_to_array could still be used (if necessary).

Should I create a PR?

Regards,
Tim

lochmueller:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | feature
| Docs?         | no
| Issues        | Fix #998
| License       | MIT

- Change StoreInterface
- Use iterable and yield (and yield from) for all storages
- Use iterator_to_array for existing test cases

Refs: #998

### DIFF
--- a/src/store/src/Bridge/Azure/SearchStore.php
+++ b/src/store/src/Bridge/Azure/SearchStore.php
@@ -44,13 +44,15 @@ final class SearchStore implements StoreInterface
         ]);
     }
 
-    public function query(Vector $vector, array $options = []): array
+    public function query(Vector $vector, array $options = []): iterable
     {
         $result = $this->request('search', [
             'vectorQueries' => [$this->buildVectorQuery($vector)],
         ]);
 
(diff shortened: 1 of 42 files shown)


OskarStark (approved):
Fine, @chr-hertel and @Guikingone please check as well, thanks

Guikingone (approved):

Guikingone:
Looks good to me 👍🏻

chr-hertel (approved):
:heart_eyes:

chr-hertel:
Thank you @lochmueller.

PR #1010, opened 2025-11-27
[Agent] Code grooming around tool call

lyrixx:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        |
| License       | MIT

### DIFF
--- a/src/agent/src/Toolbox/AgentProcessor.php
+++ b/src/agent/src/Toolbox/AgentProcessor.php
@@ -80,7 +80,10 @@ final class AgentProcessor implements InputProcessorInterface, OutputProcessorIn
 
         if ($result instanceof GenericStreamResponse) {
             $output->setResult(
-                new ToolboxStreamResponse($result->getContent(), $this->handleToolCallsCallback($output))
+                new ToolboxStreamResponse(
+                    $result->getContent(),
+                    fn (ToolCallResult $result, ?AssistantMessage $streamedAssistantResponse = null) => $this->handleToolCallsCallback($output, $result, $streamedAssistantResponse)
+                )
             );
 
             return;
@@ -90,7 +93,7 @@ final class AgentProcessor implements InputProcessorInterface, OutputProcessorIn
             return;
         }
 
-        $output->setResult($this->handleToolCallsCallback($output)($result));
+        $output->setResult($this->handleToolCallsCallback($output, $result));
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/agent/src/Toolbox/AgentProcessor.php
```diff
@@ -99,40 +102,39 @@ private function isFlatStringArray(array $tools): bool
         return array_reduce($tools, fn (bool $carry, mixed $item) => $carry && \is_string($item), true);
     }
 
-    private function handleToolCallsCallback(Output $output): \Closure
+    private function handleToolCallsCallback(Output $output, ToolCallResult $result, ?AssistantMessage $streamedAssistantResponse = null): ResultInterface
```
lyrixx, line 107: I reduced the indirection. This reduce the depth in the stack.

src/agent/src/Toolbox/ToolFactory/MemoryToolFactory.php
```diff
@@ -33,16 +33,16 @@ public function addTool(string|object $class, string $name, string $description,
     }
 
     /**
-     * @param class-string $reference
+     * @param class-string $className
```
lyrixx, line 36: Let's use only one name for a "thing". Line 30, this is called `className`. 
I think it's better to be consistant, it's easier to understand

src/agent/src/Toolbox/Toolbox.php
```diff
@@ -34,7 +34,7 @@ final class Toolbox implements ToolboxInterface
     /**
      * List of executable tools.
      *
-     * @var list<mixed>
+     * @var list<object>
```
lyrixx, line 37: We call `::class` on item. So it must be an object. See linne 69

src/agent/src/Toolbox/Toolbox.php
```diff
@@ -60,18 +60,18 @@ public function __construct(
 
     public function getTools(): array
```
lyrixx, line 61: There is something a bit hard to understand.
getTools does not return tool, but tools metadata ! 

I would like to rebame `Tool` to `ToolMetadata`, `getTools()` to `getToolsMetadata()`

It will really ease on-boarding I guess.

And it will fix inconsistency. For example :
```php
private function getMetadata(ToolCall $toolCall): Tool
```

OskarStark:
PR title: Tall or Tool?

lyrixx:
Ouch, too many typos 😅 . I edited the PR title.

src/agent/src/Toolbox/AgentProcessor.php
```diff
@@ -99,40 +102,39 @@ private function isFlatStringArray(array $tools): bool
...
-                $results = [];
-                foreach ($toolCalls as $toolCall) {
-                    $results[] = $toolResult = $this->toolbox->execute($toolCall);
-                    $messages->add(Message::ofToolCall($toolCall, $this->resultConverter->convert($toolResult)));
-                    array_push($this->sources, ...$toolResult->getSources());
-                }
+            $results = [];
+            foreach ($toolCalls as $toolCall) {
+                $results[] = $toolResult = $this->toolbox->execute($toolCall);
+                // dd($toolResult);
```
chr-hertel: this needs to go

chr-hertel:
and the PHPStan issue looks real - wonder if that's worth the trouble

src/agent/src/Toolbox/ToolFactory/MemoryToolFactory.php
chr-hertel: doesn't have to  be a class tho, right?

lyrixx: I'm not sure to understand. What could it be, if it's not a class, but the type is class-string?

lyrixx:
> and the PHPStan issue looks real - wonder if that's worth the trouble

There was a real issue with PHPStan. Since the code is now better, PHPStan found the typo. I fixed it.


---

Pr rebases, and cleaned. Ready for review


> [!NOTE]
> there is a [pending question](https://github.com/symfony/ai/pull/1010#discussion_r2569575651) about a bigger cleaning. It would happen in a new PR, But I would love to get an answer.

src/agent/src/Toolbox/AgentProcessor.php
lyrixx: Fixed 👍🏼

src/agent/src/Toolbox/Toolbox.php
OskarStark: can we extract this as a small bugfix please?

src/platform/src/Bridge/OpenAi/Gpt/ResultConverter.php
```diff
@@ -156,14 +156,14 @@ private function isToolCallsStreamFinished(array $data): bool
      *     message: array{
      *         role: 'assistant',
      *         content: ?string,
-     *         tool_calls: array{
+     *         tool_calls: list<array{
```
OskarStark, line 159: this looks like a bug fix to me, which deserves a separate PR

chr-hertel:
Thank you @lyrixx.

PR #1012, opened 2025-11-27
 [Store][ChromaDB] Add documents for text context

xprojects-de:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Docs?         | no 
| Issues        
| License       | MIT

To work properly with embeddings, the text context is also needed, otherwise LLM won't get the overall context. Therefore, it's possible to query the text context from the database too.

### DIFF
--- a/src/store/src/Bridge/ChromaDb/Store.php
+++ b/src/store/src/Bridge/ChromaDb/Store.php
@@ -53,23 +53,40 @@ final class Store implements StoreInterface
     }
 
     /**
-     * @param array{where?: array<string, string>, whereDocument?: array<string, mixed>} $options
+     * @param array{where?: array<string, string>, whereDocument?: array<string, mixed>, include?: array<string>} $options
      */
     public function query(Vector $vector, array $options = []): iterable
     {
+        $include = null;
+        if ([] !== ($options['include'] ?? [])) {
+            $include = array_values(
+                array_unique(
+                    array_merge(['embeddings', 'metadatas', 'distances'], $options['include'])
+                )
+            );
+        }
+
         $collection = $this->client->getOrCreateCollection($this->collectionName);
         $queryResponse = $collection->query(
             queryEmbeddings: [$vector->getData()],
             nResults: 4,
             where: $options['where'] ?? null,
             whereDocument: $options['whereDocument'] ?? null,
+            include: $include,
         );
 
-        for ($i = 0; $i < \count($queryResponse->metadatas[0]); ++$i) {
+        $metaCount = \count($queryResponse->metadatas[0]);
+
+        for ($i = 0; $i < $metaCount; ++$i) {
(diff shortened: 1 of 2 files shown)


src/store/src/Bridge/ChromaDb/Store.php
```diff
@@ -53,24 +53,41 @@ public function add(VectorDocument ...$documents): void
     }
 
     /**
-     * @param array{where?: array<string, string>, whereDocument?: array<string, mixed>} $options
+     * @param array{where?: array<string, string>, whereDocument?: array<string, mixed>, include?: array<string>} $options
```
chr-hertel, line 56: which expectations do we have here for the `include` option?

if i just want `embeddings`, `metadatas` and `distances` - how would i enable that?

xprojects-de: If you don't want to specify anything, simply leave it out. Then the default from the ChromaDB extension will be used. That's how it was before. However, as soon as something is passed by the user, we should ensure that at least the metadata and embeddings are included, which the array merge does (maybe we can remove distance)

chr-hertel: > If you don't want to specify anything, simply leave it out.

yup, got that, that's good :+1: 

i understand that we basically treat `['embeddings', 'metadatas', 'distances']` as the default here, right?
what would i need to do, to enable that default? i'm questioning if that even works with that current implementation. adding a test here would be great

xprojects-de: Thanks for reviewing the code and for the comments.

I've added some tests.
The default `$include = null` should work since CodeWithKyrian/chromadb-php has the correct default (see link).

https://github.com/CodeWithKyrian/chromadb-php/blob/3bf9f8db6ebfbd805774fa81176f5029afab0371/src/Resources/CollectionResource.php#L279

However, we can also define our own default if you like...

xprojects-de:
Do I need to adjust anything, or is it okay as is?

OskarStark:
From the code POV it looks ok, lets wait for the review of @chr-hertel, but please do a proper rebase, we cannot merge it, when a Merge-Commit is present (see Fabbot CI failure)

xprojects-de:
Thanks but I don't think I can manage a clean rebase myself! When I rebase to the upstream, it always includes the commit history in the pull request. I don't think that's desired. Any help would be appreciated :-)

OskarStark:
`git remote add upstream git@github.com:symfony/ai.git && geit fetch upstream && git rebase upstream/main` in your branch does not work?

xprojects-de:
no change. Maybe i have to close the Pull an make an new one!?

xprojects-de:
hard job :-) But i think the pull is fine now :-) Please excuse the inconvenience.

chr-hertel:
Thank you @xprojects-de.

PR #1017, opened 2025-11-28
[AI Bundle][Platform] Integrate template rendering into Message API

### LINKED ISSUE #258: [Agent] Introduce Prompt Templates
**Feature**

It is a common and simple feature to provide and use prompt templates:
* https://python.langchain.com/docs/concepts/prompt_templates/
* https://docs.llamaindex.ai/en/stable/examples/prompts/rich_prompt_template_features/
* https://learn.microsoft.com/en-us/semantic-kernel/concepts/prompts/prompt-template-syntax

Those linked solutions are quite similar, at least at first glance, and I guess we could see some use-case here already with the `SimilaritySearch` tool:

https://github.com/symfony/ai/blob/dd03c20aae11c6dc7a4aa391f30aafe6979b4778/src/agent/src/Toolbox/Tool/SimilaritySearch.php#L50-L53

**Open Questions**

* I feel like this should live in the Platform component and sit on top of a message, like a `TemplateMessage` or something - not sure tho
* Should we go all in with twig here? feels like over the top 🤔

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | Fix #258
| License       | MIT

## Summary

Adds a new **prompt-template** component to the Symfony AI monorepo, filling a gap we identified in our ([Christopher Hertel](https://github.com/chr-hertel) and me) recent discussion about missing core functionality for prompt management.

## Background

This implementation merges and adapts code from both [sulu.ai](sulu.ai) and [modelflow-ai](https://github.com/modelflow-ai) projects, bringing battle-tested prompt templating into the Symfony AI ecosystem with a more extensible architecture.

## Architecture

The component uses a **strategy pattern** for rendering, allowing different template processors to be plugged in:

- **StringRenderer** (default): Simple `{variable}` replacement with zero dependencies
- **ExpressionRenderer**: Advanced expression evaluation using Symfony Expression Language (optional)
- **Extensible**: Ready for additional renderers like Twig, Mustache, or custom implementations

### Key Features

- Zero core dependencies (only PHP 8.2+)
- Factory pattern for exceptions with typed error methods
- Immutable readonly classes throughout
- Interface-first design for maximum flexibility
- Comprehensive test coverage (42 tests, 52 assertions)
- PHPStan level 6 compliant

## Usage Examples

**Simple variable replacement:**
```php
$template = PromptTemplate::fromString('Hello {name}!');
echo $template->format(['name' => 'World']); // "Hello World!"

Expression-based rendering:
$renderer = new ExpressionRenderer();
$template = new PromptTemplate('Total: {price * quantity}', $renderer);
echo $template->format(['price' => 10, 'quantity' => 5]); // "Total: 50"

Custom renderer:
class TwigRenderer implements RendererInterface {
    public function render(string $template, array $values): string {
        // Twig implementation
    }
}
```

This provides a solid foundation for prompt template management across the Symfony AI ecosystem while maintaining the flexibility to adapt to different rendering needs.

### DIFF
--- a/src/ai-bundle/config/services.php
+++ b/src/ai-bundle/config/services.php
@@ -64,6 +64,10 @@ use Symfony\AI\Platform\Bridge\Voyage\ModelCatalog as VoyageModelCatalog;
 use Symfony\AI\Platform\Contract;
 use Symfony\AI\Platform\Contract\JsonSchema\DescriptionParser;
 use Symfony\AI\Platform\Contract\JsonSchema\Factory as SchemaFactory;
+use Symfony\AI\Platform\EventListener\TemplateRendererListener;
+use Symfony\AI\Platform\Message\TemplateRenderer\ExpressionLanguageTemplateRenderer;
+use Symfony\AI\Platform\Message\TemplateRenderer\StringTemplateRenderer;
+use Symfony\AI\Platform\Message\TemplateRenderer\TemplateRendererRegistry;
 use Symfony\AI\Platform\ModelCatalog\ModelCatalogInterface;
 use Symfony\AI\Platform\Serializer\StructuredOutputSerializer;
 use Symfony\AI\Platform\StructuredOutput\PlatformSubscriber;
@@ -73,6 +77,7 @@ use Symfony\AI\Store\Command\DropStoreCommand;
 use Symfony\AI\Store\Command\IndexCommand;
 use Symfony\AI\Store\Command\RetrieveCommand;
 use Symfony\AI\Store\Command\SetupStoreCommand;
+use Symfony\Component\ExpressionLanguage\ExpressionLanguage;
 
 return static function (ContainerConfigurator $container): void {
     $container->services()
@@ -118,6 +123,30 @@ return static function (ContainerConfigurator $container): void {
         ->set('ai.platform.model_catalog.scaleway', ScalewayModelCatalog::class)
         ->set('ai.platform.model_catalog.vertexai.gemini', VertexAiModelCatalog::class)
         ->set('ai.platform.model_catalog.voyage', VoyageModelCatalog::class)
+
+        // message templates
+        ->set('ai.platform.template_renderer.string', StringTemplateRenderer::class)
+            ->tag('ai.platform.template_renderer');
+
+    if (class_exists(ExpressionLanguage::class)) {
+        $container->services()
+            ->set('ai.platform.template_renderer.expression', ExpressionLanguageTemplateRenderer::class)
+                ->args([
+                    service('expression_language'),
+                ])
+                ->tag('ai.platform.template_renderer');
+    }
+
+    $container->services()
+        ->set('ai.platform.template_renderer_registry', TemplateRendererRegistry::class)
+            ->args([
+                tagged_iterator('ai.platform.template_renderer'),
+            ])
+        ->set('ai.platform.template_renderer_listener', TemplateRendererListener::class)
+            ->args([
+                service('ai.platform.template_renderer_registry'),
+            ])
+            ->tag('kernel.event_subscriber')
 
         // structured output
         ->set('ai.agent.response_format_factory', ResponseFormatFactory::class)
--- a/src/ai-bundle/tests/DependencyInjection/AiBundleTest.php
+++ b/src/ai-bundle/tests/DependencyInjection/AiBundleTest.php
@@ -33,6 +33,10 @@ use Symfony\AI\Platform\Bridge\ElevenLabs\ModelCatalog as ElevenLabsModelCatalog
 use Symfony\AI\Platform\Bridge\ElevenLabs\PlatformFactory as ElevenLabsPlatformFactory;
 use Symfony\AI\Platform\Bridge\Ollama\OllamaApiCatalog;
 use Symfony\AI\Platform\Capability;
+use Symfony\AI\Platform\EventListener\TemplateRendererListener;
+use Symfony\AI\Platform\Message\TemplateRenderer\ExpressionLanguageTemplateRenderer;
+use Symfony\AI\Platform\Message\TemplateRenderer\StringTemplateRenderer;
+use Symfony\AI\Platform\Message\TemplateRenderer\TemplateRendererRegistry;
 use Symfony\AI\Platform\Model;
 use Symfony\AI\Platform\ModelCatalog\ModelCatalogInterface;
 use Symfony\AI\Platform\PlatformInterface;
@@ -6976,6 +6980,42 @@ class AiBundleTest extends TestCase
         // The model catalog for openai should not have been modified
         $definition = $container->getDefinition('ai.platform.model_catalog.openai');
         $this->assertSame([], $definition->getArguments());
+    }
+
+    public function testTemplateRendererServicesAreRegistered()
+    {
+        $container = $this->buildContainer([
+            'ai' => [
+                'platform' => [
+                    'anthropic' => [
+                        'api_key' => 'test_key',
+                    ],
+                ],
+            ],
+        ]);
+
+        // Verify string template renderer is registered
+        $this->assertTrue($container->hasDefinition('ai.platform.template_renderer.string'));
+        $stringRendererDefinition = $container->getDefinition('ai.platform.template_renderer.string');
+        $this->assertSame(StringTemplateRenderer::class, $stringRendererDefinition->getClass());
+        $this->assertTrue($stringRendererDefinition->hasTag('ai.platform.template_renderer'));
+
+        // Verify expression template renderer is registered
+        $this->assertTrue($container->hasDefinition('ai.platform.template_renderer.expression'));
+        $expressionRendererDefinition = $container->getDefinition('ai.platform.template_renderer.expression');
+        $this->assertSame(ExpressionLanguageTemplateRenderer::class, $expressionRendererDefinition->getClass());
+        $this->assertTrue($expressionRendererDefinition->hasTag('ai.platform.template_renderer'));
+
+        // Verify template renderer registry is registered
+        $this->assertTrue($container->hasDefinition('ai.platform.template_renderer_registry'));
+        $registryDefinition = $container->getDefinition('ai.platform.template_renderer_registry');
+        $this->assertSame(TemplateRendererRegistry::class, $registryDefinition->getClass());
+
+        // Verify template renderer listener is registered as event subscriber
+        $this->assertTrue($container->hasDefinition('ai.platform.template_renderer_listener'));
+        $listenerDefinition = $container->getDefinition('ai.platform.template_renderer_listener');
+        $this->assertSame(TemplateRendererListener::class, $listenerDefinition->getClass());
+        $this->assertTrue($listenerDefinition->hasTag('kernel.event_subscriber'));
     }
 
     private function buildContainer(array $configuration): ContainerBuilder
--- a/src/platform/src/EventListener/TemplateRendererListener.php
+++ b/src/platform/src/EventListener/TemplateRendererListener.php
@@ -1,0 +1,113 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Platform\EventListener;
+
+use Symfony\AI\Platform\Event\InvocationEvent;
+use Symfony\AI\Platform\Exception\InvalidArgumentException;
+use Symfony\AI\Platform\Message\Content\Text;
+use Symfony\AI\Platform\Message\MessageBag;
+use Symfony\AI\Platform\Message\MessageInterface;
+use Symfony\AI\Platform\Message\SystemMessage;
+use Symfony\AI\Platform\Message\Template;
+use Symfony\AI\Platform\Message\TemplateRenderer\TemplateRendererRegistryInterface;
+use Symfony\AI\Platform\Message\UserMessage;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+/**
+ * Renders message templates when template_vars option is provided.
+ *
+ * @author Johannes Wachter <johannes@sulu.io>
+ */
+final class TemplateRendererListener implements EventSubscriberInterface
+{
+    public function __construct(
+        private readonly TemplateRendererRegistryInterface $rendererRegistry,
+    ) {
+    }
+
+    public function __invoke(InvocationEvent $event): void
+    {
+        $options = $event->getOptions();
+        if (!isset($options['template_vars'])) {
+            return;
+        }
+
+        if (!\is_array($options['template_vars'])) {
+            throw new InvalidArgumentException('The "template_vars" option must be an array.');
+        }
+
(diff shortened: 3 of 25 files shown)


chr-hertel:
Thanks for this @wachterjohannes - would really love to get this in :pray: 

two things that we need to sort here:
1. is it a standalone component? for that we need to sort how we would use and render the template with the Message & Platform
2. terminology: we currently don't use the terminology "prompt", but "message" - i'd prefer to go with that here as well like `MessageTemplate` 

we could have something like:
```php
$template = new Symfony\AI\Platform\Message\Template::expression('Total: {price * quantity}');
$message = Message::forSystem($template);
$messages = new MessageBag($message);

$result = $platform->invoke('gpt-4o-mini', $messages, [
    'template_vars' => ['price' => 10, 'quantity' => 5],
]);
```

and later - while serializing - take care of the rendering - potentially throw an error on unsupported types.

with type being only a string instead of a renderer instance + named construct for promoted use:
```php
class Template
{
    public function __construct(
        private string $template,
        private string $type,
    ) { }

    public static function string(string $template): self;
    public static function expression(string $template): self;
    public static function twig(string $template): self
    {
        return new self($template, 'twig');
    }
}

```

WDYT?

wachterjohannes:
@chr-hertel in my opinion an own component makes sense as it is totally independent to the other component and as  you mentioned we could integrate it into the agent and platform. @OskarStark what do you think.

Regarding the renaming - makes totally sense and i will do that when the decission over the component is final :)

chr-hertel:
My thinking always is: API first, then implementation, then slicing. That's why i would delay that component discussion.

Templates alone are not valuable, they need an integration into the Message I'd say

wachterjohannes:
@chr-hertel both are absolutly thru statements :) so let us wait for feedback from @OskarStark

OskarStark:
Lets go with the platform component first 👍

examples/platform/message-templates.php
```diff
@@ -0,0 +1,117 @@
...
+$eventDispatcher = new EventDispatcher();
+$rendererRegistry = new TemplateRendererRegistry([
+    new StringTemplateRenderer(),
+]);
+$templateListener = new TemplateRendererListener($rendererRegistry);
+$eventDispatcher->addSubscriber($templateListener);
+
+$platform = PlatformFactory::create(env('OPENAI_API_KEY'), http_client(), eventDispatcher: $eventDispatcher);
+
+// Example 1: SystemMessage with template
```
OskarStark:
```suggestion
```

examples/platform/message-templates.php
```diff
@@ -0,0 +1,117 @@
...
+
+$result = $platform->invoke('gpt-4o-mini', $messages, [
+    'template_vars' => ['role' => 'helpful'],
+]);
+
+echo "SystemMessage template: You are a {role} assistant.\n";
+echo "Variables: ['role' => 'helpful']\n";
+echo 'Response: '.$result->asText()."\n\n";
+
+// Example 2: UserMessage with template
```
OskarStark:
```suggestion
```
same for the others

wachterjohannes:
@OskarStark that was luck 😂 pushed a few minutes ago!

examples/platform/message-templates.php
wachterjohannes: @OskarStark is already updated

src/ai-bundle/tests/DependencyInjection/AiBundleTest.php
```diff
@@ -4410,6 +4410,42 @@ public function testModelConfigurationIsIgnoredForUnknownPlatform()
...
+                        'api_key' => 'test_key',
+                    ],
+                ],
+            ],
+        ]);
+
+        // Verify string template renderer is registered
+        $this->assertTrue($container->hasDefinition('ai.platform.template_renderer.string'));
+        $stringRendererDefinition = $container->getDefinition('ai.platform.template_renderer.string');
+        $this->assertSame(\Symfony\AI\Platform\Message\TemplateRenderer\StringTemplateRenderer::class, $stringRendererDefinition->getClass());
```
OskarStark: import classes please

src/platform/src/Message/AssistantMessage.php
OskarStark, line 1: Why is it needed for an assistant message, which comes from the LLM? 🤔

src/platform/src/Message/TemplateRenderer/ExpressionTemplateRenderer.php
```diff
@@ -0,0 +1,52 @@
...
+use Symfony\AI\Platform\Exception\InvalidArgumentException;
+use Symfony\AI\Platform\Message\Template;
+use Symfony\Component\ExpressionLanguage\ExpressionLanguage;
+
+/**
+ * Expression Language template renderer.
+ *
+ * @author Johannes Wachter <johannes@sulu.io>
+ */
+final readonly class ExpressionTemplateRenderer implements TemplateRendererInterface
```
OskarStark:
```suggestion
final readonly class ExpressionLanguageTemplateRenderer implements TemplateRendererInterface
```
and would remove the class comment

src/platform/src/Message/TemplateRenderer/ChainTemplateRenderer.php
```diff
@@ -0,0 +1,60 @@
...
+        foreach ($this->renderers as $renderer) {
+            if ($renderer->supports($type)) {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    public function render(Template $template, array $variables): string
```
OskarStark: I am not convinced about this method in a class named `*Registry`. A registry should return the renderer. if it has the same signature it should also be named `ChainRenderer` etc.

examples/platform/message-templates.php
```diff
@@ -0,0 +1,112 @@
...
+echo "=======================================\n\n";
+
+$template = Template::string('You are a {role} assistant.');
+$messages = new MessageBag(
+    Message::forSystem($template),
+    Message::ofUser('What is PHP?')
+);
+
+$result = $platform->invoke('gpt-4o-mini', $messages, [
+    'template_vars' => ['role' => 'helpful'],
```
OskarStark: helpful is not a role to me

src/platform/src/Message/AssistantMessage.php
wachterjohannes: you are totally right :) will remove that from the AssistantMessage

src/ai-bundle/config/services.php
```diff
@@ -111,6 +115,24 @@
         ->set('ai.platform.model_catalog.vertexai.gemini', VertexAiModelCatalog::class)
         ->set('ai.platform.model_catalog.voyage', VoyageModelCatalog::class)
 
+        // message templates
+        ->set('ai.platform.template_renderer.string', StringTemplateRenderer::class)
+            ->tag('ai.platform.template_renderer')
+        ->set('ai.platform.template_renderer.expression', ExpressionTemplateRenderer::class)
```
OskarStark: lets check if expression language is available and then register this service and remove the nullOnINvalid()

src/platform/src/Message/TemplateRenderer/ChainTemplateRenderer.php
wachterjohannes: your right - should i change the logic to return the renderer or should i rename to ChainTemplateRenderer?

OskarStark: `ChainTemplateRenderer` 👍

OskarStark (approved):

src/platform/src/EventListener/TemplateRendererListener.php
```diff
@@ -0,0 +1,103 @@
...
+
+/**
+ * Renders message templates when template_vars option is provided.
+ *
+ * @author Johannes Wachter <johannes@sulu.io>
+ */
+final readonly class TemplateRendererListener implements EventSubscriberInterface
+{
+    public function __construct(
+        private ChainTemplateRenderer $rendererRegistry,
```
chr-hertel: With readonly classes the BC promise gets messy

```suggestion
final class TemplateRendererListener implements EventSubscriberInterface
{
    public function __construct(
        private readonly ChainTemplateRenderer $rendererRegistry,
```

src/platform/src/EventListener/TemplateRendererListener.php
```diff
@@ -0,0 +1,103 @@
...
+            return;
+        }
+
+        $input = $event->getInput();
+        if (!$input instanceof MessageBag) {
+            return;
+        }
+
+        $templateVars = $options['template_vars'];
+        $renderedMessages = [];
```
chr-hertel, line 54: i'd say that we only now should check if `$templateVars` is an array - basically moving it down from line 46 - and throw an error if not.

if a developer uses `['template_vars' => 'foobar']` the intend for using the feature is given, but their just doing it wrong => exception

or do you think that's too harsh?

src/platform/src/Message/TemplateRenderer/ChainTemplateRenderer.php
```diff
@@ -0,0 +1,60 @@
...
+use Symfony\AI\Platform\Message\Template;
+
+/**
+ * Composite renderer that chains multiple renderers.
+ *
+ * Implements chain of responsibility pattern - tries each renderer until one supports the type.
+ *
+ * @author Johannes Wachter <johannes@sulu.io>
+ */
+final readonly class ChainTemplateRenderer implements TemplateRendererInterface
```
chr-hertel:
```suggestion
final class ChainTemplateRenderer implements TemplateRendererInterface
```

src/platform/src/Message/Template.php
```diff
@@ -0,0 +1,56 @@
...
+ * Supports variable substitution using different rendering types.
+ * Rendering happens externally during message serialization when template_vars are provided.
+ *
+ * @author Johannes Wachter <johannes@sulu.io>
+ */
+final readonly class Template implements \Stringable, ContentInterface
+{
+    public function __construct(
+        private string $template,
+        private string $type,
```
chr-hertel:
```suggestion
final class Template implements \Stringable, ContentInterface
{
    public function __construct(
        private readonly string $template,
        private readonly string $type,
```

src/platform/src/EventListener/TemplateRendererListener.php
wachterjohannes: @chr-hertel let us try that! i think making it clear that the user is doing something wrong makes totally sense

wachterjohannes:
@chr-hertel fixed your comments

src/ai-bundle/tests/DependencyInjection/AiBundleTest.php
```diff
@@ -4410,6 +4415,46 @@ public function testModelConfigurationIsIgnoredForUnknownPlatform()
...
+        $stringRendererDefinition = $container->getDefinition('ai.platform.template_renderer.string');
+        $this->assertSame(StringTemplateRenderer::class, $stringRendererDefinition->getClass());
+        $this->assertTrue($stringRendererDefinition->hasTag('ai.platform.template_renderer'));
+
+        // Verify expression template renderer is registered only when ExpressionLanguage is available
+        if (class_exists(ExpressionLanguage::class)) {
+            $this->assertTrue($container->hasDefinition('ai.platform.template_renderer.expression'));
+            $expressionRendererDefinition = $container->getDefinition('ai.platform.template_renderer.expression');
+            $this->assertSame(ExpressionLanguageTemplateRenderer::class, $expressionRendererDefinition->getClass());
+            $this->assertTrue($expressionRendererDefinition->hasTag('ai.platform.template_renderer'));
```
chr-hertel: how would we ever end up here - it's not installed in the bundle, right? should we do it with `req --dev`?

src/platform/src/Message/AssistantMessage.php
chr-hertel: still in

src/platform/src/Message/Message.php
```diff
@@ -26,23 +26,31 @@ private function __construct()
...
+        }
+
         return new SystemMessage($content instanceof \Stringable ? (string) $content : $content);
     }
 
     /**
      * @param ?ToolCall[] $toolCalls
      */
-    public static function ofAssistant(?string $content = null, ?array $toolCalls = null): AssistantMessage
+    public static function ofAssistant(string|Template|null $content = null, ?array $toolCalls = null): AssistantMessage
```
chr-hertel: also not needed?
```suggestion
    public static function ofAssistant(?string $content = null, ?array $toolCalls = null): AssistantMessage
```

src/platform/src/Message/TemplateRenderer/ExpressionLanguageTemplateRenderer.php
```diff
@@ -0,0 +1,50 @@
...
+namespace Symfony\AI\Platform\Message\TemplateRenderer;
+
+use Symfony\AI\Platform\Exception\InvalidArgumentException;
+use Symfony\AI\Platform\Message\Template;
+use Symfony\Component\ExpressionLanguage\ExpressionLanguage;
+
+/**
+ * @author Johannes Wachter <johannes@sulu.io>
+ */
+final readonly class ExpressionLanguageTemplateRenderer implements TemplateRendererInterface
```
chr-hertel: missed that before
```suggestion
final class ExpressionLanguageTemplateRenderer implements TemplateRendererInterface
```

src/platform/src/Message/TemplateRenderer/StringTemplateRenderer.php
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * Simple string replacement renderer.
+ *
+ * Replaces {variable} placeholders with values from the provided array.
+ * Has zero external dependencies.
+ *
+ * @author Johannes Wachter <johannes@sulu.io>
+ */
+final readonly class StringTemplateRenderer implements TemplateRendererInterface
```
chr-hertel:
```suggestion
final class StringTemplateRenderer implements TemplateRendererInterface
```

src/platform/tests/EventListener/TemplateRendererListenerTest.php
```diff
@@ -0,0 +1,212 @@
...
+    {
+        $registry = new ChainTemplateRenderer([
+            new StringTemplateRenderer(),
+        ]);
+
+        $this->listener = new TemplateRendererListener($registry);
+        $this->model = new Model('gpt-4o');
+    }
+
+    public function testRendersTemplateWhenTemplateVarsProvided(): void
```
chr-hertel: all those `: void` return type declarations on `test*` methods have to go, according to Symfony code style

examples/platform/message-templates.php
chr-hertel, line 1: please rename the folder to `template` and split into 4 different files - locally I'm executing all examples from time to time in parallel - and it will go faster like that

wachterjohannes:
@chr-hertel hopefully all fixed now :)

OskarStark:
TemplateRendererRegistry is better that Chained.... imho

OskarStark:
No need for 03- prefix etc in the example filenames

wachterjohannes:
@OskarStark yeah your right - just changed that - you can take another look

chr-hertel (approved):

chr-hertel:
Thank you @wachterjohannes.

Issue #1018, opened 2025-11-28
[Chat] Question about conversation history support in Symfony AI


EfrosIonelu:
Hello,

I would like to ask whether Symfony AI currently provides built-in support for conversation history or persistent memory, similar to how LangChain/LangGraph implements memory graphs (for example: https://docs.langchain.com/oss/javascript/langgraph/memory
).

More specifically:

Is there any recommended way in Symfony AI to maintain conversational state across multiple user messages?

Is there support for memory persistence (e.g., storing previous messages, tool invocations, or intermediate steps)?

If not yet implemented, is this feature planned on the roadmap, or is there any recommended design pattern to achieve this behavior using the current Symfony AI components?

I am evaluating how best to implement stateful conversations in a Symfony-based application, and any guidance or examples would be greatly appreciated.

Thank you!

lochmueller:
Hey @EfrosIonelu 

structure: In general, there is the Chat Component, that handle the MessageBag of single conversations. There is a central interface https://github.com/symfony/ai/blob/main/src/chat/src/MessageStoreInterface.php that control the internal message store of one conversation. That means, it should be possible to create a unique marker and store multiple conversions on long terms (handled like the identifier in https://github.com/symfony/ai/blob/main/src/chat/src/Bridge/Local/InMemoryStore.php ) and store it in a database. 

idea: If you implement such a PersitedMessageStore-Thing it is very individual. From my point of view, I would develop the Store-Integration by myself to handle all edge-cases of the env/project/app around the integration....

Regards,
Tim

Guikingone:
Hi @EfrosIonelu 👋🏻 

As said by @lochmueller, `Chat` is the main entrypoint, you can use stores that implements `MessageStoreInterface` and `ManagedStoreInterface`, the capability of storing multiple conversations in a single store is something that we've discussed for some months in #254, the "latest implementation" is under work in #821, this PR would allows to store and switch between multiple conversations without defining X stores (which is required for now).

Regarding the storage, everything that travel through messages (tools calls, images, videos and so on) is stored by default in stores thanks to `MessageNormalizer`.

Guikingone:
Keep in mind that `Langgraph` (like `Letta` / `Mem0` and more) is differentiating long and short term messages, for now, SfAI is not separating them in dedicated stores, if you want something that is not persisted, you can use the `InMemoryStore`, if you're using another message store, it's what we call "long term storage".

[...]

PR #1020, opened 2025-11-28
[Platform] Streamline message bag integration

### LINKED ISSUE #962: [Store][Chat] Unique identifier for conversations/MessageBag objects
I saw that every `Message` object has an unique ID. However, `MessageBag` objects do not have this. I wonder if it'd make sense to add this. Could be nice for multi-turn conversations. 

What do you guys think? Or was this deliberate?

lochmueller:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #962
| License       | MIT

What do you think?

- Add and prepend behave same now (do not clone the object, keep the state / avoid that `prepand` also clone the object, because only with* methods should do so)
- Add and prepend, both return the current object for chain-calls
- Add an ID for the message bag
- Renew the ID in case of cloning the message bag
- Merge do NOT clone the object anymore, because it just merge messages (clone can be done manually, if needed)
- Remove the `with` method, because this was used just once (keep the focus to explicit clone and add, if needed)
- Add `IteratorAggregate` interface for easy access the current messages

Fix #962
Related #1018

### DIFF
--- a/src/platform/src/Message/MessageBag.php
+++ b/src/platform/src/Message/MessageBag.php
@@ -18,6 +18,8 @@ use Symfony\Component\Uid\Uuid;
 
 /**
  * @author Christopher Hertel <mail@christopher-hertel.de>
+ *
+ * @implements \IteratorAggregate<MessageInterface>
  */
 class MessageBag implements \IteratorAggregate, \Countable
 {


chr-hertel:
Thanks for that - ID & iterator make def sense and generally I'm in favor of what's going with the cloning here - i only struggle a bit with having the fluent interface for `add` and `prepend` and let's keep the `with` please

OskarStark:
I extracted the array iterator to a separate PR, so please rebase, thanks.

Can you extract the ID to a dedicated PR too?

lochmueller:
hey @chr-hertel 
hey @OskarStark 

thanks for the first steps. I moved the ID part to a separated PR: https://github.com/symfony/ai/pull/1027

For the other stuff... I suggest that only "with*" and "without*" methods should clone the message Bag (like for immutable objects). "Add" and "prepend" are a little bit strange, because the one clone the object and the other not.

I would close this PR and create a new one, if you confirm, that the "prepend" shouldn't clone the object? Next question.... "add & prepend" are not really a good "naming combination". I thought about "append & prepend"? Should I change the Naming also?

Regards,
Tim

chr-hertel:
Don't really have an issue with the naming tbh. The prepend was originally introduced for bringing system prompts in on messagebags without it. And cloning makes sense in that case.

Let's have a look at the use cases for the decision. Do we have others in mind? Do you use prepend in your code?

lochmueller:
I only mentioned the naming because I mentally link "add & remove" and "append & prepend"... but we can keep this. 

The last point out of the PR would be, that "prepend" do not clone the object, and we keep the clone-processes to the `with*` functions only. Should I keep this (or create a last fresh PR out of this)?

chr-hertel:
I'd like to close this one in favor of #1055 - what do you think @lochmueller?

lochmueller:
Good solution. Thanks @chr-hertel 
Closed...

PR #1022, opened 2025-11-29
[Deptrac] Add separate layers for Agent bridge tools

OskarStark:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | --
| License       | MIT

This ensures Agent bridge tools can only depend on the Agent component and not on each other, preventing cross-bridge dependencies.

- Renamed layers to *Component convention
- Added individual layers for each Agent bridge tool
- SimilaritySearchTool can access StoreComponent and PlatformComponent
- Other tools can only access AgentComponent

### DIFF
--- a/deptrac.yaml
+++ b/deptrac.yaml
@@ -2,35 +2,105 @@ deptrac:
   paths:
     - ./src
   exclude_files:
-    - '#.*test.*#'
+    - '#.*test.*#i'
     - '#.*vendor.*#'
   layers:
-    - name: Agent
+    - name: AgentComponent
       collectors:
         - type: classLike
-          value: Symfony\\AI\\Agent.*
-    - name: Chat
+          value: ^Symfony\\AI\\Agent\\(?!Bridge\\).*
+    - name: BraveTool
+      collectors:
+        - type: classLike
+          value: Symfony\\AI\\Agent\\Bridge\\Brave\\.*
+    - name: ClockTool
+      collectors:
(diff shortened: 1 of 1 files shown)


deptrac.yaml
```diff
@@ -2,35 +2,105 @@ deptrac:
...
-      - Platform
-    Platform: ~
-    Store:
-      - Platform
+    AgentComponent:
+      - PlatformComponent
+      - StoreComponent
+    BraveTool:
+      - AgentComponent
+      - PlatformComponent
```
OskarStark, line 74: this is needed for example, as the `#[With]` is used from the Platform component

deptrac.yaml
```diff
@@ -2,35 +2,105 @@ deptrac:
   paths:
     - ./src
   exclude_files:
-    - '#.*test.*#'
+    - '#.*test.*#i'
```
OskarStark, line 5: The original pattern #.*test.*# only matches lowercase "test", but the test files in this codebase are in directories named Tests (capital T), like:
`src/agent/src/Bridge/SimilaritySearch/Tests/SimilaritySearchTest.php`

Without the i flag, deptrac was analyzing test files and reporting violations for test-only dependencies (like the test using Symfony\AI\Platform\Vector\Vector).

With #.*test.*#i, it matches both test and Tests, properly excluding all test files from analysis.

PR #1039, opened 2025-12-02
[Store] Introduce `OpenSearch` support

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | #16 
| License       | MIT


OskarStark:
CI is failing:
```
An error occurred while setting up the "opensearch" store: Empty reply from server for "http://127.0.0.1:9200/symfony".
```

OskarStark:
Hmm, CI is still failing

Guikingone:
Yes, I know 😅 

Looks like OS is too slow during its boot phase, I'm looking for a solution 🙂

OskarStark:
I made it a Draft-PR again until solved

Guikingone:
@OskarStark It's fixed, sorry for the extra delay and the X notifications about the CI 😅

OskarStark:
No worries

examples/commands/message-stores.php
```diff
@@ -103,6 +103,9 @@
     new DropStoreCommand(new ServiceLocator($factories)),
 ]);
 
+$clock = new MonotonicClock();
+$clock->sleep(10);
```
chr-hertel, line 107: Is this the solution now?
> Looks like OS is too slow during its boot phase, I'm looking for a solution 🙂

Guikingone: At least, the only one that I found that isn't triggering errors due to OS slow start.

Calling `http://127.0.0.1:9200` is not enough as it's only a health response, OS is not fully ready when the health response is OK.

OskarStark (approved):

OskarStark:
Thank you Guillaume.

PR #1048, opened 2025-12-05
[Platform][Mistral] Add 14b latest model

welcoMattic:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Docs?         | no 
| Issues        |  
| License       | MIT

Add missing `ministral-14b-latest` model.

NB: All new models (Mistral 3 series), are already supported as they are aliased with `-latest` suffix, which is what we have in Mistral `ModelCatalog` class.


OskarStark:
Thank you Mathieu.

PR #1049, opened 2025-12-05
[Store][ChromaDb] Split store as dedicated package

OskarStark:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | --
| License       | MIT

Similar to how tools were split in the Agent component, this extracts the ChromaDb store into its own standalone package.

### DIFF
--- a/deptrac.yaml
+++ b/deptrac.yaml
@@ -64,7 +64,11 @@ deptrac:
     - name: StoreComponent
       collectors:
         - type: classLike
-          value: Symfony\\AI\\Store.*
+          value: ^Symfony\\AI\\Store\\(?!Bridge\\).*
+    - name: ChromaDbStore
+      collectors:
+        - type: classLike
+          value: Symfony\\AI\\Store\\Bridge\\ChromaDb\\.*
   ruleset:
     AgentComponent:
       - PlatformComponent
@@ -100,6 +104,9 @@ deptrac:
       - PlatformComponent
     PlatformComponent: ~
     StoreComponent:
+      - PlatformComponent
+    ChromaDbStore:
+      - StoreComponent
       - PlatformComponent
   # Baseline of known violations to be skipped for now
   skip_violations:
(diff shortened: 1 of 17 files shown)


src/store/src/Bridge/ChromaDb/composer.json
```diff
@@ -0,0 +1,59 @@
+{
+    "name": "symfony/ai-chromadb-store",
```
OskarStark: @fabpot is the package name correct?

xprojects-de:
@OskarStark nice to split, but what about https://github.com/symfony/ai/pull/1012 
Is this merged before?

OskarStark:
Final review from @chr-hertel needed, otherwise we need to rebase afterwards 👍🏻

src/store/src/Bridge/ChromaDb/Store.php
```diff
@@ -28,9 +27,6 @@ public function __construct(
         private readonly Client $client,
         private readonly string $collectionName,
     ) {
-        if (!class_exists(Client::class)) {
-            throw new RuntimeException('For using the ChromaDB as retrieval vector store, the codewithkyrian/chromadb-php package is required. Try running "composer require codewithkyrian/chromadb-php".');
-        }
```
chr-hertel, line 33: yeah nice :)

OskarStark:
Will fix the deptrac stuff and would like to get the confirmation for the package name from @fabpot, then good to merge 

Edit:
Ah no the frameworkbundle config needs to check for the store class available to allow the options

deptrac.yaml
```diff
@@ -101,6 +105,9 @@ deptrac:
     PlatformComponent: ~
     StoreComponent:
       - PlatformComponent
+    ChromaDbStore:
+      - StoreComponent
+      - PlatformComponent
```
OskarStark, line 110: Because of `Vector` class

chr-hertel (approved):

chr-hertel:
Thank you @OskarStark.

PR #1052, opened 2025-12-06
[Platform][Generic] Introduce generic platform for openai embeddings- and completions-based platforms

### LINKED ISSUE #114: Custom URL ability or generic bridges?

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | Fix #114
| License       | MIT

Replaces #938 and unblocks #871 - diff looks a bit weird due to git mixing up removed classes and new generic bridge.

- [x] Tests
- [x] Docs
- [x] Bundle

This enables to use tools like Litellm without specific bridge with a configuration like this:
```yaml
# config/packages/ai.yaml
ai:
    platform:
        generic:
            litellm:
                base_url: '%env(LITELLM_HOST_URL)%' # e.g. http://localhost:4000
                api_key: '%env(LITELLM_API_KEY)%' # e.g. sk-1234
                model_catalog: 'Symfony\AI\Platform\Bridge\Generic\ModelCatalog' # see below
    agent:
        test:
            platform: 'ai.platform.generic.litellm'
            model: 'mistral-small-latest'
            tools: false

services:
    Symfony\AI\Platform\Bridge\Generic\ModelCatalog:
        $models:
            mistral-small-latest:
                class: 'Symfony\AI\Platform\Bridge\Generic\CompletionsModel'
                capabilities:
                    - !php/const 'Symfony\AI\Platform\Capability::INPUT_MESSAGES'
                    - !php/const 'Symfony\AI\Platform\Capability::OUTPUT_TEXT'
                    - !php/const 'Symfony\AI\Platform\Capability::OUTPUT_STREAMING'
                    - !php/const 'Symfony\AI\Platform\Capability::OUTPUT_STRUCTURED'
                    - !php/const 'Symfony\AI\Platform\Capability::INPUT_IMAGE'
                    - !php/const 'Symfony\AI\Platform\Capability::TOOL_CALLING'
```
image


OskarStark:
I need to double check later on desktop, but this looks really cool and promising, thanks for the hard work, I like the deletion of so many classes while unlocking a lot of new bridges, which use the defacto OpenAI standard 🙋‍♂️

OskarStark:
Thank you Christopher.

PR #1055, opened 2025-12-07
[Platform] Rename MessageBag::prepend to actual use-case withSystemMessage

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | 
| License       | MIT

The actual use case is to mess with the system prompt - let's make it explicit by renaming the method.


OskarStark (approved):

OskarStark:
Thank you Christopher.

PR #1062, opened 2025-12-07
[Platform][LM Studio] Decouple from Open AI & introduce concrete ModelCatalog

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | 
| License       | MIT

So those example didn't work no more, but I at least fixed the examples with requiring concrete models and adding a ModelCatalog for that - needs a follow up for more flexibility tho.


OskarStark:
Thank you Christopher.

PR #1068, opened 2025-12-07
[Platform] Respect SSE comments in streaming mode

lochmueller:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | 
| License       | MIT

I have same problems with OpenRouter streaming mode, because openrouter send a `: OPENROUTER PROCESSING` to keep the SSE connection `open` (details on https://openrouter.ai/docs/api/reference/streaming#additional-information ). The current integration of the RawHttpResult do not handle that. As described in the OpenRouter documentation,  the handling of SSE comments based on the Spec https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation is not handled in the right way. We have to drop lines starting with a `:` because these are comments and cannot be handled via json_decode.

Fixed the integration and add a test incl. a comment.

### DIFF
--- a/src/platform/src/Result/RawHttpResult.php
+++ b/src/platform/src/Result/RawHttpResult.php
@@ -55,6 +55,11 @@ final class RawHttpResult implements RawResultInterface
                     continue;
                 }
 
+                if (str_starts_with(trim($delta), ':')) {
+                    // SEE comments via Spec https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation
+                    continue;
+                }
+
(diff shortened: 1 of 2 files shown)


lochmueller:
Hmm... looks like the problem is not so easy to handle. The Symfony\Component\HttpClient\Chunk\ServerSentEvent should handle comments in the right way, and no comments should be part of the iteration. 
I have to check, why I got errors/exceptions in json_decode, because comments hit the `json_decode` line...

lochmueller:
Will open a fresh one

PR #1076, opened 2025-12-08
[Store][ChromaDB] Add support for `codewithkyrian/chromadb-php` 1.0

xprojects-de:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| License       | MIT

support latest version v1.0 from codewithkyrian/chromadb-php to support latest V2 API of ChromaDb

PR #1081, opened 2025-12-08
[Store] Use `codewithkyrian/chromadb-php` v1.0 and drop 0.x

OskarStark:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | Replaces #1076
| License       | MIT

- Update codewithkyrian/chromadb-php requirement from ^0.2.1|^0.3|^0.4 to ^1.0
- Update namespaces for v1.0 BC breaks:
  - QueryItemsResponse moved to Codewithkyrian\ChromaDB\Responses
  - CollectionResource replaced with Collection in Codewithkyrian\ChromaDB\Models
- Remove queryImages parameter from query() call (removed in v1.0)
- Add queryTexts option to query() method for text-based searches

### DIFF
--- a/src/store/composer.json
+++ b/src/store/composer.json
@@ -48,7 +48,7 @@
         "symfony/uid": "^7.3|^8.0"
     },
     "require-dev": {
-        "codewithkyrian/chromadb-php": "^0.2.1|^0.3|^0.4",
+        "codewithkyrian/chromadb-php": "^1.0",
         "doctrine/dbal": "^3.3|^4.0",
         "mongodb/mongodb": "^1.21|^2.0",
         "phpstan/phpstan": "^2.0",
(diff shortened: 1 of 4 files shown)


xprojects-de:
nice! Like it!

rvmourik:
Could it be that this change causes an error when trying to index the stores, I just upgraded to the latest versions and changed the config. 

image

The set-up is running with docker in combination with symfony cli, before upgrading the indexing worked

chr-hertel:
@rvmourik needed to fix that in the demo as well: https://github.com/symfony/ai/pull/1086/files

not sure if we need to change something here - i think it's just a breaking change we inherit from the package, see https://github.com/CodeWithKyrian/chromadb-php/releases/tag/1.0.0

rvmourik:
> @rvmourik needed to fix that in the demo as well: https://github.com/symfony/ai/pull/1086/files
> 
> not sure if we need to change something here - i think it's just a breaking change we inherit from the package, see https://github.com/CodeWithKyrian/chromadb-php/releases/tag/1.0.0

@chr-hertel thanks for the quick reply, that works indeed.

Offtopic, loved your talk at Symfony Con in Amsterdam

chr-hertel:
Happy to hear that @rvmourik & thanks for the feedback!

Issue #1075, opened 2025-12-08
[AI Bundle] Unresolved dependencies


xprojects-de:
The AI ​​bundle configuration has dependencies on, for example, `use Symfony\AI\Store\Bridge\Postgres\Distance as PostgresDistance` and `use Symfony\AI\Store\Bridge\Redis\Distance`, which are now standalone packages and may not be installed.

```
The enum class "Symfony\AI\Store\Bridge\Postgres\Distance" does not exist i  
  n /Users/.../vendor/symfony/ai-bundle/src/../co  
  nfig/options.php (which is being imported from "/Users/ .../vendor/symfony/ai-bundle/src/AiBundle.php").
```

https://github.com/symfony/ai/blob/5dd24d0ea68f9065e374085f1cd09d28ff531d42/src/ai-bundle/config/options.php#L21

OskarStark:
Should be fixed, right?

Huluti:
@OskarStark Strangely on my side I still have the issue.
That's strange because the new deps are well in my composer.lock... 
I've tried to remove entirely the cache folder to be sure.

<img width="622" height="881" alt="Image" src="https://github.com/user-attachments/assets/a8675490-a1fa-46a7-8f89-763e8352525f" />

xprojects-de:
- Root composer.json requires symfony/ai-redis-store @dev -> satisfiable by symfony/ai-redis-store[dev-main, 0.x-dev (alias of dev-main)].
     - symfony/ai-redis-store[dev-main, 0.x-dev] require ext-redis * -> it is missing from your system. Install or enable PHP's redis extension.

I think this approach requires that all extensions are always installed. Building the cache in the live system would cause problems.

Huluti:
On my side I have the required extensions because if I require manyally the two deps (postgre and redis) then I don't have any remaining errors

VincentLanglet:
Might be worth to reopen this issue @OskarStark ?

I still get the error
```
In FileLoader.php line 177:
!!
!!    The enum class "Symfony\AI\Store\Bridge\Postgres\Distance" does not exist i
!!    n /app/vendor/symfony/ai-bundle/src/../config/options.php (which is being i
!!    mported from "/app/vendor/symfony/ai-bundle/src/AiBundle.php").
!!
!!
!!  In EnumNodeDefinition.php line 52:
!!
!!    The enum class "Symfony\AI\Store\Bridge\Postgres\Distance" does not exist.
!!
```

[...]

PR #1082, opened 2025-12-08
[AI Bundle] Fix unresolved Distance enum dependencies in configuration

OskarStark:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fixes https://github.com/symfony/ai/issues/1075
| License       | MIT

The options.php file was importing Distance enums from optional store bridge packages (postgres-store and redis-store) unconditionally. This caused errors when these packages weren't installed since PHP tries to resolve imports at parse time.

The fix uses string values for the distance configuration instead of enum references, and converts them to enums in AiBundle.php only after verifying the package is installed via willBeAvailable().

### DIFF
--- a/src/ai-bundle/config/options.php
+++ b/src/ai-bundle/config/options.php
@@ -18,8 +18,6 @@ use Symfony\AI\Platform\Bridge\OpenAi\PlatformFactory;
 use Symfony\AI\Platform\Capability;
 use Symfony\AI\Platform\Model;
 use Symfony\AI\Platform\PlatformInterface;
-use Symfony\AI\Store\Bridge\Postgres\Distance as PostgresDistance;
-use Symfony\AI\Store\Bridge\Redis\Distance;
 use Symfony\AI\Store\Document\VectorizerInterface;
 use Symfony\AI\Store\StoreInterface;
 use Symfony\Component\Config\Definition\Exception\InvalidConfigurationException;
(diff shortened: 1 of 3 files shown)


xprojects-de:
Why do we still need the Require-Dev packages?
EDIT: OK used in the AiBundle. It´s ok

OskarStark:
Yes for the tests

src/ai-bundle/config/options.php
```diff
@@ -794,8 +792,8 @@
                                 ->end()
                                 ->enumNode('distance')
                                     ->info('Distance metric to use for vector similarity search')
-                                    ->enumFqcn(PostgresDistance::class)
-                                    ->defaultValue(PostgresDistance::L2)
+                                    ->values(['cosine', 'inner_product', 'l1', 'l2'])
```
VincentLanglet, line 795: The only issue I see here is the fact that if a new case is added/removed in PostgresDistance, this config won't be up-to-date with it.
It will require to update the values (and add a conflict in the composer json ?). 

But I don't see better solution for now. Unless removing the values restrictions and doing the value validation
- either in the AiBundle.php
- either in a callback here (but I'm not sure if it will avoid the unresolved enum...)

I wonder how it's dealed in other symfony packages, this should not be the first enum in an optional dependency isn't it ?

OskarStark: PR welcome if you find something. To unblock, I merge this one in the current state.

PR #1128, opened 2025-12-12
[Platform] Add docs for message templates

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | yes
| Issues        | 
| License       | MIT

Follows #1017 - cc @wachterjohannes makes sense?

### DIFF
--- a/docs/bundles/ai-bundle.rst
+++ b/docs/bundles/ai-bundle.rst
@@ -423,6 +423,14 @@ Then configure the prompt with translation enabled:
 The system prompt text will be automatically translated using the configured translator service.
 If no translation domain is specified, the default domain will be used.
 
+Message Template Support
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+The Platform's feature for using message templates is set up by the bundle, and conditionally also registers the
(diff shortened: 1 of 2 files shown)


OskarStark:
I think we should add an example how you can configure templates in ai bundle config

wachterjohannes:
@OskarStark in the ai-bundle you can just use it. but documenting how that looks like would make sense!

@chr-hertel should we document how the listener has to be configured when no ai-bundle is used? or link to the exanple?

OskarStark:
Yes let's be explicit and explain it in the docs

OskarStark:
Why closing ?

chr-hertel:
oh, removed the branch on my fork because i thought it's merged ... :D

chr-hertel:
> I think we should add an example how you can configure templates in ai bundle config

@OskarStark currently there's nothing additional provided in the bundle - i added a section tho that the feature is just available.

> should we document how the listener has to be configured when no ai-bundle is used?

@wachterjohannes there's a "Setup" section in the platform part

PR #1130, opened 2025-12-13
[Platform] Remove unused code in Vector constructor and add missing tests

camilleislasse:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no  
| Docs?         | no 
| License       | MIT

 The `Vector` constructor had duplicate validation code. The second check (line 36-38) could never be reached because the first check (line 28-30) already handles the same condition.

  Given [`$dimensions` is typed `?int`](https://github.com/symfony/ai/blob/main/src/platform/src/Vector/Vector.php#L26):
  - `null !== $dimensions` = "dimensions is not null"
  - `\is_int($dimensions)` = "dimensions is an int"

  When `$dimensions` is not null, it is an int. So both conditions are equivalent, making the second check unused code.

  This PR:

  - Removes the unreachable duplicate check
  - Adds missing test coverage for dimension validation

### DIFF
--- a/src/platform/src/Vector/Vector.php
+++ b/src/platform/src/Vector/Vector.php
@@ -33,10 +33,6 @@ final class Vector implements VectorInterface
             throw new InvalidArgumentException('Vector must have at least one dimension.');
         }
 
-        if (\is_int($dimensions) && \count($data) !== $dimensions) {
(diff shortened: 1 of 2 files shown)


chr-hertel:
Good catch, thanks @camilleislasse.

PR #1109, opened 2025-12-10
[Elasticsearch] Add bridge

llupa:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | 
| License       | MIT

PR #1134, opened 2025-12-15
[CI] Wait for Docker services to be healthy before running tests

llupa:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  |no 
| Docs?         |
| Issues        | 
| License       | MIT

Adding a wait step in integration-tests workflow that checks all Docker Compose services are healthy before running store commands. In this PR #1109 I had the issue that OpenSearch integrations tests would sometimes pass sometimes not. 

The step waits up to 5 minutes, checking every 5 seconds, for all services to report either "healthy" status or have no healthcheck defined.

I did as many checks as I could on my local macos machine for all services in compose.yaml and at least it works on my machine, but I am not a bash expert, so do not hold back with your reviews 😄

### DIFF
--- a/.github/workflows/integration-tests.yaml
+++ b/.github/workflows/integration-tests.yaml
@@ -44,6 +44,23 @@ jobs:
               working-directory: examples
               run: docker compose up -d
 
+            - name: Wait for services to be healthy
+              working-directory: examples
+              run: |
+                  echo "Waiting for all services to be healthy..."
+                  for i in {1..60}; do
+                      if docker compose ps --format json | jq -s -e 'all(.[]; .Health == "healthy" or .Health == "")' > /dev/null 2>&1; then
+                          echo "All services are healthy!"
+                          exit 0
+                      fi
+                      echo "Attempt $i/60: Some services are still starting..."
+                      docker compose ps
+                      sleep 5
+                  done
+                  echo "Timeout waiting for services to be healthy"
+                  docker compose ps
+                  exit 1
+
             - name: Configure environment
               run: |
                   echo COLUMNS=120 >> $GITHUB_ENV


.github/workflows/integration-tests.yaml
```diff
@@ -44,6 +44,23 @@ jobs:
               working-directory: examples
               run: docker compose up -d
 
+            - name: Wait for services to be healthy
```
OskarStark, line 47: @Guikingone if this works, we can get rid of the 10sec sleep?

Guikingone: Yes.

I'm not sure it will solve the `Opensearch`, when launched locally, it can be considered as "healthy" and not be fully operational, to test on this one to ensure that we're not missing it.

llupa: @Guikingone I am looking into OpenSearch and ElasticSearch locally in more detail. You are right that this does not remove the OS issue for now, but from my local work, it was a necessary step to rule out that the container was up an running before trying out the tests.

chr-hertel:
had the same issue and went lazy on it since you're on it ... thanks already! :)
see #1144

chr-hertel:
Tried to restart the failing tests a couple of times ... at least it doesn't solve them being flaky here due to 403 - or is that a new issue now?

llupa:
@chr-hertel  I am also working on: https://github.com/symfony/ai/pull/1137 but this week have been busy with work. 

It is a combination of both services not being ready in time and OpenSearch and ElasticSearch docker network / initial setup. In the other PR, sometimes if timed right it works, and if not either fails. 

I am trying different ways to make this as reliable / consistent as possible. Maybe we can close this PR and I move this step in the other one. 

But I have this issue living rent-free in my head for days now.

chr-hertel:
Just wanted to understand, easy - thanks for the explanation! Will think about it as well ...

chr-hertel:
Should we close here and in #1137 or do you still want to work on this @llupa?

llupa:
hm, good ping. Let me try this week @chr-hertel if I cannot find anything I will close it myself.

OskarStark:
Lets close here, it is working right now.

PR #1138, opened 2025-12-15
[Store] Extract some code from Indexer to a dedicated class + Introduce Ingester

### LINKED ISSUE #875: [Store] Indexer is nice, but hard to use directly

lyrixx:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | Fix #997, fix #875
| License       | MIT

### DIFF
--- a/demo/AGENTS.md
+++ b/demo/AGENTS.md
@@ -62,7 +62,7 @@ symfony console mcp:server
 - **Agents**: blog, stream, youtube, wikipedia, audio
 - **Platform**: OpenAI integration
 - **Store**: ChromaDB vector store
-- **Indexer**: Text embedding model
+- **Ingester**: Text embedding model
 
 ### Chat Pattern
 - `Chat` class: Message flow and session management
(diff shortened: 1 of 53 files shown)


chr-hertel:
Thanks for kicking off the next round - we're now here, right?

image

we're stacking quite some integration concepts onto each other from architecture point of view - it's not easy to grasp the difference from `Ingester` and `Indexer` - especially for me that term is connected to ETL/ELT, but load and transform are part of Indexer. but that's me maybe.

and we're keeping flaws like that sources state in the Ingester now.

---

compared to
image

I missed to catch up on your arguments in #997, but will create a sketch PR tho.

lyrixx:
wahoo, the diagrams are cool 👏🏼 

> t's not easy to grasp the difference from `Ingester` and `Indexer` - especially for me that term is connected to ETL/ELT, but load and transform are part of Indexer. but that's me maybe.

I don't know, for me it's clear. But I'm biased since I partially built it 😁
I could rename `Ingester` to `Etl` to make it clearer

Anyway, we must have something to index (fileter, transform, vectorize, store) documents. I don't need the loader thing, since data are coming in real time from RabbitMQ.

lyrixx:
So, what's the plan with this PR? Should I rebase, or we close it?

lyrixx:
Too outdated. Let's close it

PR #1145, opened 2025-12-15
[Chat] Fix MongoDB identifier context in MessageNormalizer denormalize

camilleislasse:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | -
| License       | MIT

The `MessageNormalizer::denormalize()` method was ignoring the `identifier` context option, always using hardcoded `'id'` key to restore the UUID.

This caused issues with MongoDB store which uses `_id` as the identifier (MongoDB's primary key convention). When saving messages, the store correctly passed `context: ['identifier' => '_id']` to `normalize()`, but `load()` called `denormalize()` without context, and even if it did, the method ignored it.

**Changes:**
- `MessageNormalizer::denormalize()` now uses `$context['identifier'] ?? 'id'` to read the UUID
- `MongoDb\MessageStore::load()` now passes the `identifier` context to `denormalize()`
- Fixed the MongoDB test to simulate real MongoDB behavior (data with `_id` key)
- Added test for custom identifier roundtrip

### DIFF
--- a/src/chat/src/Bridge/MongoDb/MessageStore.php
+++ b/src/chat/src/Bridge/MongoDb/MessageStore.php
@@ -79,7 +79,9 @@ final class MessageStore implements ManagedStoreInterface, MessageStoreInterface
         ]);
 
         return new MessageBag(...array_map(
-            fn (array $message): MessageInterface => $this->serializer->denormalize($message, MessageInterface::class),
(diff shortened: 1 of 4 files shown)


OskarStark:
Thank you @camilleislasse.

PR #1149, opened 2025-12-15
[Store] Rework state handling of indexer

### LINKED ISSUE #875: [Store] Indexer is nice, but hard to use directly

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| License       | MIT

This is not done, but rather to sketch the idea

- [ ] decision between #997, #1138 and #1149
- [ ] decision on Filter & Transformer vs. Transformer only
- [ ] bundle rework
- [ ] tests
- [ ] docs

This solves:

* State in Indexer
* Indexing from state with `DocumentCollection`
* Enables factory for bundle to keep sources in indexer config (to be proven)
* Opens up for more complex use cases like loading various types in one go, e.g. from one folder

Addressing Concerns from #997:
* "We don't want to support everything. In my humble opinion, it lacks genericness."

  Users could actually quite easy implement their own source objects and loaders to throw them into the indexer - including custom entities or specific URLs, etc.

* "Then, to me, Rss, Url, or Source are data objects, not stateless services. The naming is confusing."
  That's even the idea, yes - they are state objects to be handed over to the indexing process instead of being part of the constructor which lacks flexibility - (why we had the `withSources`)

* "I'm not a fan of the Strategy Design Pattern" + "We are fighting this pattern so much in symfony/serializer to avoid performance issues!"

  I would argue that the performance challenge in serializer comes with object nesting and therefore the runtime complexity grows in a different way since we're mostly dealing with collections here - might be different with folders, but that's a different story - the slowest part of the indexing will be the vectorizing/http calls anyhow

* "This class does too much. It should be split"

  Not sure i follow this reasoning since the amount of operations in total is the same, see sketch in #1138. "indexing" is our high level operation and the indexer itself acts as integrator of multiple operation services. this refactoring makes it a bit more visible now with shifting some loading logic.

cc @lyrixx @OskarStark 

there are still quite some hypothesis and assumption in those line - but i think it helps to follow the idea


examples/indexer/index-file-loader.php
```diff
@@ -23,21 +24,22 @@
...
         new TextSplitTransformer(chunkSize: 500, overlap: 100),
     ],
 );
 
-$indexer->index();
+$sources = TextFileLoader::createSource([
+    dirname(__DIR__, 2).'/fixtures/movies/gladiator.md',
+    dirname(__DIR__, 2).'/fixtures/movies/inception.md',
+    dirname(__DIR__, 2).'/fixtures/movies/jurassic-park.md',
+]);
```
chr-hertel, line 40: could also be like this:
```suggestion
$sources = [
    new TextFile(dirname(__DIR__, 2).'/fixtures/movies/gladiator.md'),
    new TextFile(dirname(__DIR__, 2).'/fixtures/movies/inception.md'),
    new TextFile(dirname(__DIR__, 2).'/fixtures/movies/jurassic-park.md'),
];
```

* assumption is that factory will help with bundle config
* validation could move from factory to constructor

src/store/src/Document/Loader.php
```diff
@@ -0,0 +1,52 @@
...
+     */
+    private readonly array $sourceTypeLoaders;
+
+    /**
+     * @param iterable<SourceTypeLoaderInterface> $sourceTypeLoaders
+     */
+    public function __construct(
+        iterable $sourceTypeLoaders,
+    ) {
+        $this->sourceTypeLoaders = $sourceTypeLoaders instanceof \Traversable ? iterator_to_array($sourceTypeLoaders) : $sourceTypeLoaders;
```
lyrixx: Why do you break lazyness ? I saw theses bits in other place, and I don't get it
It's okay to store iterable in a class: https://3v4l.org/d6h65

src/store/src/Indexer.php
```diff
@@ -13,9 +13,9 @@
 
 use Psr\Log\LoggerInterface;
 use Psr\Log\NullLogger;
-use Symfony\AI\Store\Document\EmbeddableDocumentInterface;
 use Symfony\AI\Store\Document\FilterInterface;
-use Symfony\AI\Store\Document\LoaderInterface;
```
lyrixx, line 18: This interface is not used anymore, it should be removed, shouldn't?

src/store/src/Indexer.php
```diff
@@ -26,57 +26,37 @@
...
-        $this->sources = null === $source ? [] : (array) $source;
-    }
-
-    public function withSource(string|array $source): self
-    {
-        return new self($this->loader, $this->vectorizer, $this->store, $source, $this->filters, $this->transformers, $this->logger);
     }
 
-    public function index(array $options = []): void
+    public function index(iterable|SourceInterface $sources, array $options = []): void
```
lyrixx, line 42: so now, we must pass something when indexing? We loose the ability to run :
```bash
bin/console ai.store.index indexer_name
```
to index everything.


Or I missed something?

src/store/src/Document/Loader.php
```diff
@@ -0,0 +1,52 @@
...
+
+    /**
+     * @return iterable<EmbeddableDocumentInterface>
+     */
+    public function load(iterable $sources): iterable
+    {
+        foreach ($sources as $source) {
+            $loaderFound = false;
+            foreach ($this->sourceTypeLoaders as $sourceTypeLoader) {
+                if ($sourceTypeLoader->supports($source)) {
```
lyrixx: This is the part I don't like. 
We already have a strong coupling between a model, and its implementation.
We could use the DIC (service locator) to solve this.

```php
$this->sourceTypeLoaders->get($source::class)->load($source);
```

To solve this we have a map like
```
[
    source class => loader,
    source class 2 => loader 2
]
```

I [blogged about](https://jolicode.com/blog/how-taggedlocator-can-help-you-design-better-symfony-application) this pattern a while ago

It's quite simple and very performant.

src/store/src/Document/SourceLoaderInterface.php
lyrixx, line 1: `SourceLoaderInterface` is enough to me

src/store/src/Document/Loader.php
chr-hertel: yes, you're totally right ofc

src/store/src/Indexer.php
chr-hertel: not sure i get the point - it was renamed to `SourceTypeLoaderInterface`, but is not needed here anymore

src/store/src/Indexer.php
chr-hertel: yup, that's open and needs to be tackled in the bundle :+1: will follow

src/store/src/Document/Loader.php
chr-hertel: yup, especially without any additional argument or dynamic the `supports(SourceInterface $source)` is over the top - so is the dependency to the entire DIC component for that pattern.

will adopt to something slim like:
```php
/**
 * @return class-string<SourceInterface>
 */
public static function supportedSource(): string;
```

so we could have a map in the loader that makes accessing the corresponding loader also quite efficient

examples/indexer/index-file-loader.php
```diff
@@ -23,21 +24,22 @@
 $store = new InMemoryStore();
 $vectorizer = new Vectorizer($platform, 'text-embedding-3-small');
 $indexer = new Indexer(
-    loader: new TextFileLoader(),
+    loader: new Loader([TextFileLoader::supportedSource() => new TextFileLoader()]),
```
chr-hertel, line 27: could also be
```php
new Loader([TextFile::class => new TextFileLoader()]),
```
or we shift that part into the `Loader` implementation

src/store/src/Indexer.php
lyrixx: Oups, I missed the renaming. Sorry

chr-hertel:
Closed in favor of #1464 - only addressing the state concern. we can make it more complex later if we need those multiple source types on one indexer

PR #1170, opened 2025-12-17
[Agent][Brave] Make field `description` optional

daFish:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Docs?         | no 
| Issues        | not applicable 
| License       | MIT

The `Brave` tool fails when the Brave Search API returns results without a `description` field. According to the [Brave API documentation](https://api-dashboard.search.brave.com/app/documentation/web-search/responses), the `description` field is **optional**, but the current implementation assumes it's always present.

**Error**: `Undefined array key "description"` wrapped in `ToolExecutionException`

**User-visible error**: `"An error occurred while executing tool \"brave_search\"."`

### DIFF
--- a/src/agent/src/Bridge/Brave/Brave.php
+++ b/src/agent/src/Bridge/Brave/Brave.php
@@ -70,12 +70,16 @@ final class Brave implements HasSourcesInterface
 
         foreach ($results as $result) {
             $this->addSource(
-                new Source($result['title'] ?? '', $result['url'] ?? '', $result['description'] ?? '')
+                new Source($result['title'], $result['url'], $result['description'] ?? '')
             );
         }
 
         return array_map(static function (array $result) {
-            return ['title' => $result['title'], 'description' => $result['description'], 'url' => $result['url']];
+            return [
+                'title' => $result['title'],
+                'description' => $result['description'] ?? '',
+                'url' => $result['url'],
+            ];
         }, $data['web']['results'] ?? []);
     }
 }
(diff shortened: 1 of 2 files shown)


src/agent/src/Bridge/Brave/Brave.php
```diff
@@ -75,7 +75,11 @@ public function __invoke(
         }
 
         return array_map(static function (array $result) {
-            return ['title' => $result['title'], 'description' => $result['description'], 'url' => $result['url']];
+            return [
+                'title' => $result['title'] ?? '',
+                'description' => $result['description'] ?? '',
+                'url' => $result['url'] ?? '',
```
OskarStark: All of them are optional? If yes I think it does not make sense to return an array where all values can be an empty string.

OskarStark:
In the PR title you talk about making it nullable, but you set an empty string

daFish:
@OskarStark Right, I have updated the title.

src/agent/src/Bridge/Brave/Brave.php
OskarStark: https://api-dashboard.search.brave.com/app/documentation/web-search/responses#Result

based on this, title and url are always set:
CleanShot 2025-12-17 at 11 00 03@2x

daFish: You are absolutely right (SCNR). I should have RTFM more carefully. 😓

OskarStark (approved):

OskarStark:
Thank you @daFish.

PR #1177, opened 2025-12-17
[AI Bundle] Fix PHPDoc type for collected platform call data

camilleislasse:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | -
| License       | MIT

This PR fixes the PHPDoc type for `DataCollector::getPlatformCalls()`.

## Problem

The `PlatformCallData` type (defined in `TraceablePlatform`) declares `result: DeferredResult`. However, after `DataCollector::lateCollect()` processes the data through `awaitCallResults()`, the `result` field is transformed to `string|iterable|object|null` and a `metadata` field is added.

This causes PHPStan errors when testing code that accesses the transformed data (e.g., `assertNull($dataCollector->getPlatformCalls()[0]['result'])`).

## Solution

Introduce a new PHPDoc type `CollectedPlatformCallData` that accurately describes the data returned by `getPlatformCalls()` after transformation:

```php
@phpstan-type CollectedPlatformCallData array{
    model: string,
    input: array<mixed>|string|object,
    options: array<string, mixed>,
    result: string|iterable<mixed>|object|null,
    metadata: Metadata,
}
```

### DIFF
--- a/src/ai-bundle/src/Profiler/DataCollector.php
+++ b/src/ai-bundle/src/Profiler/DataCollector.php
@@ -25,6 +25,14 @@ use Symfony\Component\HttpKernel\DataCollector\LateDataCollectorInterface;
  * @phpstan-import-type PlatformCallData from TraceablePlatform
  * @phpstan-import-type MessageStoreData from TraceableMessageStore
  * @phpstan-import-type ChatData from TraceableChat
+ *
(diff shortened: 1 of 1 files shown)


OskarStark:
Thank you Camille.

PR #1179, opened 2025-12-17
[Chat][Doctrine] Drop Doctrine DBAL 3 support for `DoctrineDbalMessageStore`

OskarStark:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | --
| License       | MIT

### DIFF
--- a/src/chat/composer.json
+++ b/src/chat/composer.json
@@ -30,7 +30,7 @@
     },
     "require-dev": {
         "ext-redis": "*",
-        "doctrine/dbal": "^3.3|^4.0",
+        "doctrine/dbal": "^4.0",
         "mongodb/mongodb": "^1.21|^2.0",
         "phpstan/phpstan": "^2.1",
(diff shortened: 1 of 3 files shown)


src/chat/composer.json
```diff
@@ -30,7 +30,7 @@
     },
     "require-dev": {
         "ext-redis": "*",
-        "doctrine/dbal": "^3.3|^4.0",
+        "doctrine/dbal": "^4.0",
```
nicolas-grekas, line 33: I'm not sure that's a good idea. We're going to recommend DBAL v3 for the time being in Symfony.
See https://github.com/symfony/symfony/issues/62481 and related
What's the motivation for the bump? We tend to favor wide ranges in the Symfony project, unless reasons.

OskarStark: Just cleanup, so let's keep v3 support

PR #1183, opened 2025-12-17
[Platform][Scaleway] Add support for Responses API

daFish:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | --
| License       | MIT

Thanks to the work by @paulinevos I am adding support for the Responses API from Scaleway. I also added 2 new examples to show use the responses endpoint - one with tool usage.

### DIFF
--- a/src/platform/src/Bridge/Scaleway/CHANGELOG.md
+++ b/src/platform/src/Bridge/Scaleway/CHANGELOG.md
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+0.9
+---
+
+ * Add Responses API support via the Open Responses bridge
+
 0.8
 ---
 
--- a/deptrac.yaml
+++ b/deptrac.yaml
@@ -515,6 +515,7 @@ deptrac:
     ScalewayPlatform:
       - PlatformComponent
       - GenericPlatform
+      - OpenResponsesPlatform
     TransformersPhpPlatform:
       - PlatformComponent
     VertexAiPlatform:
--- a/examples/scaleway/responses-toolcall.php
+++ b/examples/scaleway/responses-toolcall.php
@@ -1,0 +1,37 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use Symfony\AI\Agent\Agent;
+use Symfony\AI\Agent\Toolbox\AgentProcessor;
+use Symfony\AI\Agent\Toolbox\Toolbox;
+use Symfony\AI\Agent\Toolbox\ToolFactory\MemoryToolFactory;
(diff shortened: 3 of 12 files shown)


src/platform/src/Bridge/Scaleway/CHANGELOG.md
```diff
@@ -5,3 +5,4 @@ CHANGELOG
 ---
 
  * Add the bridge
+ * Add support for Responses API
```
OskarStark:
```suggestion
```
not needed for the first release

src/platform/CHANGELOG.md
```diff
@@ -68,3 +68,4 @@ CHANGELOG
  * Allow beta feature flags to be passed into Anthropic model options
  * Add Ollama streaming output support
  * Add multimodal embedding support for Voyage AI
+ * Use Responses API for Scaleway platform when using gpt-oss-120b
```
OskarStark: please revert this line

chr-hertel:
Hi @daFish, what's the benefit of introducing this? feels a bit odd to have both, a lot of duplicated code and only fixed for one model

so ideally we should decide for one API here to ease the implementation

daFish:
> Hi @daFish, what's the benefit of introducing this? feels a bit odd to have both, a lot of duplicated code and only fixed for one model
> 
> so ideally we should decide for one API here to ease the implementation

Scaleway currently supports both but to use `gpt-oss-120b` with tools you need to use the responses API.

From the Scaleway documentation:
> Most supported Generative API models can be used with both Chat Completions and Responses API. For the gpt-oss-120b model, use of the Responses API is recommended, as it will allow you to access all of its features, especially tool-calling.

As the responses API is currently in beta, I'd opt to support both for the time being. Once it is GA then it is the best option to move to responses API only.

chr-hertel:
@daFish yup, sounds reasonable - even if that brings in another dimension of coupling/duplicating code :see_no_evil: 

could you please split into two different `ModelClient` implementations please? you could do the `'gpt-oss-120b` === $model->getName()` check already in the `supports` method and maybe share stuff in an abstract class if needed.

Thanks already!

chr-hertel:
Hi @daFish, nowadays we have an Open Responses bridge that we could just rely on here - basically scaleway adopting it as dependency and building on top - still up for working on that topic, continuing in that direction?

daFish:
Hi @chr-hertel, yes, I was waiting for some more progress on this to avoid unnecessary work. Will start working on that direction.

daFish:
Fabbot failure is a false-positive.

chr-hertel:
@daFish Thanks for the update - not that familiar with Scaleway, do we actually need to keep that generic bridge here or could we completely rely on the Open Responses API instead?

In that case you could trim down the bridge I guess :)

---

Note to myself: currently open responses is in really early shape at scaleway - def keep completions for now.

chr-hertel (approved):
Rebased and reverted some smaller bits.

chr-hertel:
Thank you @daFish.

Issue #1184, opened 2025-12-17
[Platform] RawHttpResult::getDataStream() fails when HTTP chunks split JSON objects


nsrosenqvist:
_Disclaimer: I asked copilot to write up this bug report as it implemented a workaround_

**Description:**

When using the Gemini bridge with streaming enabled, `RawHttpResult::getDataStream()` fails with `JsonException` ("Control character error" or "Syntax error") because it assumes each HTTP chunk contains complete, parseable JSON.

In practice, the HTTP transport layer can split the response at arbitrary byte boundaries. The Gemini API returns responses as JSON arrays (`[{...}, {...}]`), and these can be split mid-object across multiple HTTP chunks.

**Steps to Reproduce:**

1. Configure Gemini platform with streaming enabled
2. Make a chat request that triggers a tool call (these tend to have larger responses with base64 signatures)
3. The response is split across multiple HTTP chunks

**Expected Behavior:**

The stream should buffer incoming data and only yield complete JSON objects.

**Actual Behavior:**

The stream attempts to parse each HTTP chunk as complete JSON, throwing `JsonException` on partial data:

```
JsonException: Control character error, possibly incorrectly encoded
```

**Debug Output:**

The raw chunks received show the split:
- Chunk 1: `[{"candidates": [...partial JSON with base64 signature...`
- Chunk 2: `...base64 continuation...`
- Chunk 3: `...more base64...`
- Chunk 4: `]`

**Root Cause:**

In RawHttpResult.php:

```php
public function getDataStream(): iterable
{
    foreach ((new EventSourceHttpClient())->stream($this->response) as $chunk) {
        // ...
        $jsonDelta = $chunk instanceof ServerSentEvent ? $chunk->getData() : $chunk->getContent();
        
        // This assumes $jsonDelta is complete JSON, but HTTP chunks can split anywhere
        $deltas = explode(",\r\n", $jsonDelta);
        
        foreach ($deltas as $delta) {
            yield json_decode($delta, true, flags: \JSON_THROW_ON_ERROR); // Fails on partial JSON
        }
    }
}
```

**Suggested Fix:**

Buffer the incoming data and only parse when a complete JSON array/object is detected:

```php
public function getDataStream(): iterable
{
    $buffer = '';
    
    foreach ((new EventSourceHttpClient())->stream($this->response) as $chunk) {
        if ($chunk->isFirst() || $chunk->isLast()) {
            continue;
        }
        
[...]

PR #1191, opened 2025-12-18
[Platform] Streamline `RawHttpResult` for streaming

lochmueller:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | 
| License       | MIT

Fix:
- Skip SSE comments in processing
- Skip SSE errors in processing
- Use getArrayData (JSON convert) in case of real SSE in RawHttpResult (Use JSON decode process from the ServerSentEvent)
- Keep the manual json decode process of single data chunks (incl. the "Single items" handling of Google)
- Fix also problems, if Provider like OpenRouter send comments like `: OPENROUTER PROCESSING` https://openrouter.ai/docs/api/reference/streaming#additional-information

### DIFF
--- a/src/platform/src/Result/RawHttpResult.php
+++ b/src/platform/src/Result/RawHttpResult.php
@@ -11,6 +11,7 @@
 
 namespace Symfony\AI\Platform\Result;
 
+use Symfony\Component\HttpClient\Chunk\DataChunk;
 use Symfony\Component\HttpClient\Chunk\ServerSentEvent;
 use Symfony\Component\HttpClient\EventSourceHttpClient;
 use Symfony\Contracts\HttpClient\ResponseInterface;
@@ -33,23 +34,39 @@ final class RawHttpResult implements RawResultInterface
     public function getDataStream(): iterable
     {
         foreach ((new EventSourceHttpClient())->stream($this->response) as $chunk) {
-            // Handle only complete events (no need for handle DataChunk`s)
-            if (!$chunk instanceof ServerSentEvent) {
-                continue;
-            }
-
-            // Do not handle: Init, Terminate, Errors, Comments and openAI specific termination via DONE
+            // Do not handle: First, Last, Error, Comments and API Specific "[DONE]"
             if (
                 $chunk->isFirst()
                 || $chunk->isLast()
                 || null !== $chunk->getError()
                 || str_starts_with($chunk->getContent(), ':')
-                || '[DONE]' === $chunk->getData()
+                || ($chunk instanceof ServerSentEvent && '[DONE]' === $chunk->getData())
             ) {
                 continue;
             }
 
-            yield $chunk->getArrayData();
+            if ($chunk instanceof ServerSentEvent) {
+                // handle complete SSE
+                yield $chunk->getArrayData();
+            } elseif ($chunk instanceof DataChunk) {
+                // Handle single delta arrays
+                $jsonDelta = $chunk->getContent();
+
+                // Remove leading/trailing brackets
+                if (str_starts_with($jsonDelta, '[') || str_starts_with($jsonDelta, ',')) {
+                    $jsonDelta = substr($jsonDelta, 1);
+                }
+                if (str_ends_with($jsonDelta, ']')) {
+                    $jsonDelta = substr($jsonDelta, 0, -1);
+                }
+
+                $deltas = explode(",\r\n", $jsonDelta);
+                $deltas = array_map(function ($string) { return trim($string); }, $deltas);
+                $deltas = array_filter($deltas, function ($item) { return '' !== $item; });
+                foreach ($deltas as $delta) {
+                    yield json_decode($delta, true, flags: \JSON_THROW_ON_ERROR);
+                }
+            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


OskarStark:
Can you please rebase?

lochmueller:
> Can you please rebase?

done. The workflow errors are not related to my changes...

OskarStark:
Lets merge https://github.com/symfony/ai/pull/1204 first, rebase and adjust the testcase

OskarStark:
I merged my PR, can you please update the test? thanks

src/platform/src/Result/RawHttpResult.php
```diff
@@ -33,30 +33,23 @@ public function getData(): array
...
-            // Split in case of multiple JSON objects
-            $deltas = explode(",\r\n", $jsonDelta);
-
-            foreach ($deltas as $delta) {
-                if ('' === trim($delta)) {
-                    continue;
-                }
-
-                yield json_decode($delta, true, flags: \JSON_THROW_ON_ERROR);
-            }
```
chr-hertel, line 59: what's the reason for removing this?
there is no replacement for this part or do i miss something?

chr-hertel: background for having this was gemini api streaming parts of a larger json document - which cannot be decoded right away - testing with `examples/gemini/stream.php` might help

lochmueller: I will check this. The "delta"split is part of the SSE integration and is handled via the symfony/http-client (I think so): https://github.com/symfony/http-client/blob/4f420c441088fc3364647565701f34ace5ba0d5f/EventSourceHttpClient.php#L136 Only complete SSE events create ServerSentEvent objects. But I will check the specification, symfony/http-client and gemini API again. I will share my information in the next days...

lochmueller: ... and the json_decord part is included in the `$chunk->getArrayData()` call. I will check this too...

chr-hertel: okay, so we might need to patch there if there's an issue - fair enough :+1:

lochmueller: @chr-hertel 
double checked. You are right... google really send a "greater JSON structure" with multiple chunks. To work on this items, we still need the "extract the array manually"-stuff. I add the SerevrSendEvent-based encoding "in frontend" of the custom integration and keep the error & comment checks.

OskarStark:
Can you cover this by tests?

lochmueller:
Hey @OskarStark 
yes... but this is still a problem: https://github.com/symfony/ai/issues/1184
Gemini send incomplete chunks, that we "can use" in streaming mode... but there should be some kind of fallback, that we buffer the information if the JSON decode do not work. I will add that and try to add some examples for different SSE event structures, especially for Gemini.
Regards, Tim

Others have also problems with the API specific behavior of google: https://github.com/googleapis/python-genai/issues/1162
Notes: https://developer.chrome.com/docs/ai/streaming?hl=de

chr-hertel:
Please have a look as well at #754 and [my latest comment there](https://github.com/symfony/ai/pull/754#issuecomment-3716302316) - not in conflict I think, but related

src/platform/src/Result/RawHttpResult.php
```diff
@@ -33,29 +34,38 @@ public function getData(): array
...
+                // Remove leading/trailing brackets
+                if (str_starts_with($jsonDelta, '[') || str_starts_with($jsonDelta, ',')) {
+                    $jsonDelta = substr($jsonDelta, 1);
+                }
+                if (str_ends_with($jsonDelta, ']')) {
+                    $jsonDelta = substr($jsonDelta, 0, -1);
                 }
 
-                yield json_decode($delta, true, flags: \JSON_THROW_ON_ERROR);
+                $deltas = explode(",\r\n", $jsonDelta);
```
mhueni-conteo, line 63: Exploding at any comma followed by a newline could potentially break up valid JSON objects. It didn't work for me at least.

Assuming new object chunks are always starting with `,` (as Gemini seems to be doing), this snippet worked for me:

```php
// initialize outside loop
$jsonBuffer = '';
...
} elseif ($chunk instanceof DataChunk) {
    $part = trim($chunk->getContent());
    
    if (str_starts_with($part, '[') || str_starts_with($part, ',')) {
        // beginning of object array or next object, remove delimiter
        $jsonBuffer = substr($part, 1);
    } else {
        // part or end of current object
        $jsonBuffer .= $part;
    }
    
    if (str_ends_with($jsonBuffer, ']')) {
        // end of object array, remove delimiter
        $jsonBuffer = substr($jsonBuffer, 0, -1);
    }
    
    if (str_ends_with($jsonBuffer, '}')) {
        // wrap buffered parts in array again, it could be more than one object
        $objects = json_decode('['.$jsonBuffer.']', true, flags: \JSON_THROW_ON_ERROR);
        foreach ($objects as $object) {
            yield $object;
        }
        $jsonBuffer = '';
    }
}
```

lochmueller:
Close. I think this is a better solution perhaps: https://github.com/symfony/ai/pull/1988

PR #1199, opened 2025-12-18
[AI Bundle][Profiler] avoid serializing closures for files and images

uerka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | 
| License       | MIT

Currently when agent call includes images/files profiler throws exception 'Serialization of 'Closure' is not allowed' and therefore cannot store profile. 
Added __serialize() methods to File class and also displaying of files/images in profiler.


OskarStark:
Thank you @uerka for taking care of this regression.

PR #1203, opened 2025-12-18
[AI Bundle] Remove dependency from AI bundle to OpenAI Bridge

### LINKED ISSUE #1202: [AI Bundle] Implicit dependency on OpenAI PlatformFactory constants
`symfony/ai-bundle` has an implicit hard dependency on `symfony/ai-open-ai-platform`.

In [ai-bundle/config/options.php](https://github.com/symfony/ai/blob/c131b97fed47cb52b35abe2a043aae1ed7741e6f/src/ai-bundle/config/options.php), the bundle imports `Symfony\AI\Platform\Bridge\OpenAi\PlatformFactory` and uses  it's constants `PlatformFactory::REGION_EU` / `PlatformFactory::REGION_US` in the config validation for `ai.platform.openai.region`.

However, `symfony/ai-open-ai-platform` is not listed under `require` in `symfony/ai-bundle`'s `composer.json` (it is only present under `require-dev`). This means installing/booting the bundle without the OpenAI bridge can fail during container compilation, even if the application does not intend to use OpenAI.

## Steps to reproduce

1. Create a new Symfony app (e.g. `symfony/skeleton`).
2. Install the bundle without the OpenAI bridge

## Expected behavior

The AI Bundle should be installable and bootable without requiring `symfony/ai-open-ai-platform`.

## Actual behavior

Container compilation fails because the configuration tree references `PlatformFactory::REGION_EU` and `PlatformFactory::REGION_US`, which require the OpenAI bridge class to be present, even when OpenAI is not used.

## Suggested solutions

- Replace `PlatformFactory::REGION_EU` / `PlatformFactory::REGION_US` with plain string values (`"EU"`, `"US"`) in the configuration validation.
- Add `symfony/ai-open-ai-platform` to the `require` section  of `symfony/ai-bundle`

lochmueller:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #1202
| License       | MIT

Remove the constants of the open AI Bridge in the AI Bundle configuration to remove the hard dependency. I think it is the better solution instead of adding the bridge to the composer.json.

### DIFF
--- a/src/ai-bundle/config/options.php
+++ b/src/ai-bundle/config/options.php
@@ -14,7 +14,6 @@ namespace Symfony\Component\Config\Definition\Configurator;
 use Codewithkyrian\ChromaDB\Client as ChromaDbClient;
 use MongoDB\Client as MongoDbClient;
 use Probots\Pinecone\Client as PineconeClient;
-use Symfony\AI\Platform\Bridge\OpenAi\PlatformFactory;
 use Symfony\AI\Platform\Capability;
 use Symfony\AI\Platform\Model;
 use Symfony\AI\Platform\PlatformInterface;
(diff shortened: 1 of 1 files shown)


OskarStark:
Thanks, that was already on my list.
Can we use an enumNode here?

OskarStark:
Lets keep it like this with the custom validation message

OskarStark:
Thank you Tim.

lochmueller:
I also think about moving the constants to the AI Bundle or create an enumeration. But the message is really clear... Thanks for merging.

PR #1139, opened 2025-12-15
[Platform][VertexAI] Add support for API key authentication

nicodemuz:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | need to be updated
| License       | MIT

Vertex AI supports multiple methods of authentication. This PR adds support for API keys.

PR #1206, opened 2025-12-18
[AI Bundle][Platform] Allow configure Bedrock platform via bundle config

uerka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| Docs?         | no 
| Issues        | 
| License       | MIT

Currently it seems not possible to define bedrock as a platform. 
Added configuration for this platform with support of multiple instances (for example per AWS region).  Added tests for model clients. Changed Bedrock's PlatformFactory to have nullable BedrockRuntimeClient 

So now the AWS bedrock can be configured together with other platforms as follows:

```yaml
ai:
    platform:
        bedrock:
            default: ~
            us:
                bedrock_runtime_client: 'async_aws.client.bedrock_runtime_us'
    agent:
        default:
            platform: 'ai.platform.bedrock_default'
            model:
                name: 'claude-sonnet-4-5-20250929'
 ``` 

Also fixed some smaller bugs for Nova model (exception of sending model name).

### DIFF
--- a/src/platform/src/Bridge/Bedrock/Tests/Nova/NovaModelClientTest.php
+++ b/src/platform/src/Bridge/Bedrock/Tests/Nova/NovaModelClientTest.php
@@ -99,7 +99,7 @@ final class NovaModelClientTest extends TestCase
         $this->modelClient = new NovaModelClient($this->bedrockClient);
 
         $options = [
-            'tools' => ['Tool']
+            'tools' => ['Tool'],
         ];
 
         $response = $this->modelClient->request($this->model, ['message' => 'test'], $options);
@@ -126,7 +126,7 @@ final class NovaModelClientTest extends TestCase
         $this->modelClient = new NovaModelClient($this->bedrockClient);
 
         $options = [
-            'temperature' => 0.35
+            'temperature' => 0.35,
         ];
 
         $response = $this->modelClient->request($this->model, ['message' => 'test'], $options);
@@ -153,7 +153,7 @@ final class NovaModelClientTest extends TestCase
         $this->modelClient = new NovaModelClient($this->bedrockClient);
 
         $options = [
-            'max_tokens' => 1000
+            'max_tokens' => 1000,
         ];
 
         $response = $this->modelClient->request($this->model, ['message' => 'test'], $options);
@@ -181,7 +181,7 @@ final class NovaModelClientTest extends TestCase
 
         $options = [
             'max_tokens' => 1000,
-            'temperature' => 0.35
+            'temperature' => 0.35,
         ];
 
         $response = $this->modelClient->request($this->model, ['message' => 'test'], $options);

### CODE REVIEW

**Comment by uerka** [comment]
It seems like to run tests successfully for ai-bundle link call would be needed (at least it helped for me locally). Is it possible or should i search for another approach? 
Also not sure i understand why demo tests are failing - as far as i see there is nothing related bedrock.

could someone help me out with those ?

PS big thanks for quick merge of previous issue, it was my first one contribution, so pardon me for silly questions))

File: src/ai-bundle/src/AiBundle.php [hunk 1]
```diff
@@ -408,6 +409,31 @@ private function processPlatformConfig(string $type, array $platform, ContainerB
             return;
         }
 
+        if ('bedrock' === $type) {
+            if (!ContainerBuilder::willBeAvailable('symfony/ai-bedrock-platform', BedrockFactory::class, ['symfony/ai-bundle'])) {
+                throw new RuntimeException('Bedrock platform configuration requires "symfony/ai-bedrock-platform" package. Try running "composer require symfony/ai-bedrock-platform".');
+            }
```
chr-hertel: needs a check for `[] === $type['bedrock']` or to be moved into `foreach`

see #1214 for similar

chr-hertel (changes requested):
Thanks @uerka for picking up here :+1:

docs/bundles/ai-bundle.rst
```diff
@@ -100,6 +105,10 @@ Advanced Example with Multiple Agents
                 platform: 'ai.platform.eleven_labs'
                 model: 'text-to-speech'
                 tools: false
+            nova:
+                platform: 'ai.platform.bedrock_default
```
OskarStark:
```suggestion
                platform: 'ai.platform.bedrock_default‚
```

uerka: guess it was about missing quote mark, fixed)

nicodemuz:
I noticed your tests are failing cause of similar reasons as in my PR: https://github.com/symfony/ai/pull/1139

Any idea how to fix these type of issues? Do we need to add new method parameters as last to avoid breaking BC?

src/platform/src/Bridge/Bedrock/PlatformFactory.php
```diff
@@ -33,7 +33,7 @@
 final class PlatformFactory
 {
     public static function create(
-        BedrockRuntimeClient $bedrockRuntimeClient = new BedrockRuntimeClient(),
+        ?BedrockRuntimeClient $bedrockRuntimeClient = null,
```
chr-hertel, line 36: is this change needed?

uerka: tbh - not sure about this one.
When its nullable we can pass null as first argument in [this line](https://github.com/symfony/ai/pull/1206/files#diff-f2828d6c9ecf4603ca1695c4958952ae24994a6e697e0a6e5b57f7f1c336ee47R424) that feels like a better idea than creating BedrockRuntimeClient in the config.
i moved creating client later in [code](https://github.com/symfony/ai/pull/1206/files#diff-9d90025703ce4f7a3c795b186355435b8b0925de16859c0b4d52ab2cb88d5d51R45), so i dont really see the problem of having it nullable from now - except of failing tests atm. 

i dont mind to have it as required argument and always require it for configuring bedrock platform, but seems like creating client on the fly should work also fine (as soon as creds can be resolved through env vars). 

would appreciate your input on this

chr-hertel: understood - let's keep it like that, but go with explicit null check below:
```php
if (null === $bedrockRuntimeClient) {
```

src/ai-bundle/src/AiBundle.php
```diff
@@ -408,6 +409,31 @@ private function processPlatformConfig(string $type, array $platform, ContainerB
             return;
         }
 
+        if ('bedrock' === $type) {
+            foreach ($platform as $name => $config) {
+                if (!ContainerBuilder::willBeAvailable('symfony/ai-bedrock-platform', BedrockFactory::class, ['symfony/ai-bundle'])) {
+                    throw new RuntimeException('Bedrock platform configuration requires "symfony/ai-bedrock-platform" package. Try running "composer require symfony/ai-bedrock-platform".');
+                }
+
+                $platformId = 'ai.platform.bedrock_'.$name;
```
OskarStark: @chr-hertel should we go with
```suggestion
                $platformId = 'ai.platform.bedrock.'.$name;
```
?

src/platform/src/Bridge/Bedrock/Anthropic/ClaudeModelClient.php
```diff
@@ -60,31 +55,6 @@ public function request(Model $model, array|string $payload, array $options = []
         return new RawBedrockResult($this->bedrockRuntimeClient->invokeModel(new InvokeModelRequest($request)));
     }
 
-    public function convert(InvokeModelResponse $bedrockResponse): ToolCallResult|TextResult
```
OskarStark: why removing this?

src/platform/src/Bridge/Bedrock/Tests/Anthropic/ClaudeModelClientTest.php
```diff
@@ -0,0 +1,132 @@
...
+        $this->assertInstanceOf(RawBedrockResult::class, $response);
+    }
+
+    public function testSetsAnthropicVersion()
+    {
+        $this->bedrockClient->expects($this->once())
+            ->method('invokeModel')
+            ->with($this->callback(function ($arg) {
+                $this->assertInstanceOf(InvokeModelRequest::class, $arg);
+                $this->assertEquals('application/json', $arg->getContentType());
```
OskarStark: assertSame where possible please

src/platform/src/Bridge/Bedrock/ModelCatalog.php
```diff
@@ -288,6 +288,26 @@ public function __construct(array $additionalModels = [])
...
+            'claude-sonnet-4-5-20250929' => [
+                'class' => Claude::class,
+                'capabilities' => [
+                    Capability::INPUT_MESSAGES,
+                    Capability::INPUT_IMAGE,
+                    Capability::OUTPUT_TEXT,
+                    Capability::OUTPUT_STREAMING,
+                    Capability::TOOL_CALLING,
+                ],
+            ],
```
OskarStark: Please revert, I extracted it to:
* https://github.com/symfony/ai/pull/1225

src/ai-bundle/src/AiBundle.php
chr-hertel: yes, would be more consistent - see azure

src/platform/src/Bridge/Bedrock/Tests/Anthropic/ClaudeModelClientTest.php
nicodemuz: There's a rector rule that could automate (or at least detect) this:
https://getrector.com/rule-detail/assert-equals-to-same-rector

OskarStark: I created an issue

https://github.com/symfony/ai/issues/1228

src/ai-bundle/src/AiBundle.php
uerka: then i guess we might want to change [here](https://github.com/symfony/ai/blob/main/src/ai-bundle/src/AiBundle.php#L179) to avoid platform named "ai.platform.bedrock.eu" to become "ai.traceable_platform.eu" that could collide with azure/generic or other platforms. should we go with that change or change azure/generic as atm it seems should handle same strange traceable platform name conversion?

src/platform/src/Bridge/Bedrock/Anthropic/ClaudeModelClient.php
uerka: because conversion atm handled with [ResultConverter](https://github.com/symfony/ai/blob/main/src/platform/src/Bridge/Bedrock/Anthropic/ClaudeResultConverter.php), so this code seems not supposed to be called like that.

src/ai-bundle/src/AiBundle.php
OskarStark: I agree, we should change that to `ai.traceable_platform.bedrock.eu`, same for azure, it is more consistent, isn't it @chr-hertel ?

But thats subject to change in a follow up PR, lets merge it like this

src/platform/src/Bridge/Bedrock/Anthropic/ClaudeModelClient.php
OskarStark: Extracted to:
* https://github.com/symfony/ai/pull/1229

src/ai-bundle/src/AiBundle.php
OskarStark: I crafted a PR:
* https://github.com/symfony/ai/pull/1230

src/platform/src/Bridge/Bedrock/Anthropic/ClaudeModelClient.php
uerka: as soon as its other PR now - excluded commit with this change from current branch

src/ai-bundle/src/AiBundle.php
uerka: changed naming to follow azure/generic examples

chr-hertel (changes requested):
Pipeline failure looks related - could you please give it a look @uerka?

OskarStark (approved):

chr-hertel:
Thank you @uerka.

chr-hertel:
Patched the changelog while merging - manual close, changes are integrated with https://github.com/symfony/ai/commit/e376edee2d373633fff81443c32e0d8455bfb03d

PR #1213, opened 2025-12-19
 Remove message store bridge path repositories

OskarStark:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | --
| License       | MIT

The message store bridge packages now exist as proper packages, so the path repository entries are no longer needed.


chr-hertel:
Thank you @OskarStark.

PR #1224, opened 2025-12-20
 Test store bridge pipeline

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | 
| License       | MIT

might help with #867 

not sure @OskarStark if there side effects, but in my understanding we need this?


OskarStark:
Yes, you are right, this is needed

OskarStark:
Thank you @chr-hertel.

PR #1238, opened 2025-12-21
[Mate] Couple Mate name and version + add url & icon to server

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | Fix #... 
| License       | MIT

image


src/mate/src/Command/ServeCommand.php
```diff
@@ -76,7 +78,13 @@ protected function execute(InputInterface $input, OutputInterface $output): int
         );
 
         $server = Server::builder()
-            ->setServerInfo('ai-mate', '0.1.0', 'AI development assistant MCP server')
+            ->setServerInfo(
+                App::NAME,
+                App::VERSION,
+                'AI development assistant MCP server',
+                $this->getIcon(),
+                'https://symfony.com/doc/current/ai/components/mate.html',
```
chr-hertel, line 86: @OskarStark that's the correct URL, right? do you know when rebuild of production docs will be triggered usually?

OskarStark: Url is correct. Don't know when exactly

Issue #1228, opened 2025-12-20
Rector


OskarStark:
There's a rector rule that could automate (or at least detect) this:
https://getrector.com/rule-detail/assert-equals-to-same-rector

_Originally posted by @nicodemuz in https://github.com/symfony/ai/pull/1206#discussion_r2637021442_

VincentLanglet:
Hi @OskarStark, this is also supported by the phpstan extension of phpunit
https://github.com/phpstan/phpstan-phpunit

Since phpstan and phpunit is installed in every package I recommend to add this one.
It adds multiple feature, and also allow to remove the need to add typehint to dataprovider.
See https://github.com/symfony/ai/pull/1241 to see the impact.

chr-hertel:
Closed for now - maybe we'll find another use-case for rector later on

PR #1241, opened 2025-12-21
 Use `phpstan/phpstan-phpunit`

VincentLanglet:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Docs?         | no 
| Issues        | Fix #... 
| License       | MIT

Could be considered enough to close the issue #1228 since it will check the assertEquals


src/platform/tests/ContractTest.php
```diff
@@ -43,12 +47,6 @@ public function testCreateRequestPayload(Model $model, array|string|object $inpu
         $this->assertSame($expected, $actual);
     }
 
-    /**
-     * @return iterable<string, array{
-     *     input: array|string|object,
-     *     expected: array<string, mixed>|string
-     * }>
-     */
```
OskarStark, line 51: We can keep them when we have them, right?

VincentLanglet:
Also, is there a reason to ignore
```
message: "#^Method .*::test.*\\(\\) has no return type specified\\.$#"
```
rather than adding `void` return type? since you require recent version of phpunit and php.

VincentLanglet:
The failure https://github.com/symfony/ai/actions/runs/20429872316/job/58698047458?pr=1241 seems random so I think this is ready to be reviewed @OskarStark :)

OskarStark:
> rather than adding void return type? since you require recent version of phpunit and php.

Yes, Symfony convention and enforced by Fabbot

src/agent/src/Bridge/Wikipedia/phpstan.dist.neon
```diff
@@ -13,10 +14,15 @@ parameters:
...
-            identifier: missingType.iterableValue
-            path: Tests/*
+            message: '#^Call to( static)? method PHPUnit\\Framework\\Assert::.* will always evaluate to true\.$#'
             reportUnmatched: false
         -
             identifier: 'symfonyAi.forbidNativeException'
             path: Tests/*
             reportUnmatched: false
+
+services:
```
OskarStark, line 24: Can we move this to `- ../../../../../.phpstan/extension.neon` to avoid the duplication or is this not doable?

VincentLanglet: We can move more to this extension.neon.

Do you prefer https://github.com/symfony/ai/pull/1246 ?

I just need a decision about what I should do for the Examples folder since it currently does not have phpunit as dependency. Should I just require phpunit on the example even if it's unused ? Or stop using extension.neon in the example folder ? Or exposing two different extension.neon files ?

OskarStark: Let's keep it for now like this and merge this

OskarStark:
Thank you Vincent.

PR #1257, opened 2025-12-22
[Store] Adds platform options to index method

aszenz:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | Pass platform options
| License       | MIT

Pass platform options via indexer

### DIFF
--- a/src/store/src/Indexer.php
+++ b/src/store/src/Indexer.php
@@ -87,13 +87,13 @@ class Indexer implements IndexerInterface
             ++$counter;
 
             if ($chunkSize === \count($chunk)) {
-                $this->store->add(...$this->vectorizer->vectorize($chunk));
+                $this->store->add(...$this->vectorizer->vectorize($chunk, $options['platform_options'] ?? []));
                 $chunk = [];
             }
         }
 
         if ([] !== $chunk) {
-            $this->store->add(...$this->vectorizer->vectorize($chunk));
+            $this->store->add(...$this->vectorizer->vectorize($chunk, $options['platform_options'] ?? []));
         }
 
         $this->logger->debug('Document processing completed', ['total_documents' => $counter]);
--- a/src/store/src/IndexerInterface.php
+++ b/src/store/src/IndexerInterface.php
@@ -21,7 +21,7 @@ interface IndexerInterface
     /**
      * Process sources through the complete document pipeline: load → transform → vectorize → store.
      *
-     * @param array{chunk_size?: int} $options Processing options
+     * @param array{chunk_size?: int, platform_options?: array<string, mixed>} $options Processing options
      */
     public function index(array $options = []): void;
 


src/store/src/Indexer.php
```diff
@@ -87,13 +87,13 @@ public function index(array $options = []): void
             ++$counter;
 
             if ($chunkSize === \count($chunk)) {
-                $this->store->add(...$this->vectorizer->vectorize($chunk));
+                $this->store->add(...$this->vectorizer->vectorize($chunk, $platformOptions));
```
chr-hertel: would be
```suggestion
                $this->store->add(...$this->vectorizer->vectorize($chunk, $options['platform_options'] ?? []));
```

src/store/src/IndexerInterface.php
```diff
@@ -21,7 +21,7 @@ interface IndexerInterface
     /**
      * Process sources through the complete document pipeline: load → transform → vectorize → store.
      *
-     * @param array{chunk_size?: int} $options Processing options
+     * @param array{chunk_size?: int, platform_options?: array<string, mixed>} $options Processing options
```
aszenz, line 24: Unrelated to this pr, but i find the term `chunk_size` here confusing, it's actually batch size, batching vectors and pushing them at once to the underlying store add method.

Is batching is even needed in the indexer, I think batching is best part of the store implementation, since it has the most context

aszenz: Ok I see, it actually batches creating vectors not storing them

chr-hertel: for me "chunk" and "batch" are mostly interchangeable or do i miss something?
AFAIK we're using the term chunk quite consistently in this codebase

aszenz: The key difference:

Chunking = splitting one large thing into smaller pieces like chunking big text
Batching = grouping multiple things together like commits/queries/api calls

Batching is a widely used term for this purpose

aszenz: Also is there a way to document each array option in PHP docblocks?

chr-hertel: so when i hand over one source to the indexer and it gets split into multiple smaller pieces, that's chunking? we're good than i guess :)
batching would be rather the internal perspective of the service than, but the `chunk_size` is defined as part of an argument.

> Also is there a way to document array options in PHP docs?

you already did, didn't you? what are you missing - the values?
i usually look up stuff here: https://phpstan.org/writing-php-code/phpdoc-types

aszenz: > so when i hand over one source to the indexer and it gets split into multiple smaller pieces, that's chunking? we're good than i guess :)

the split part is done by the transformers not the indexer though and it's not affected by chunking argument which actually makes this confusing, i initially thought it's about transforming the source

>  you already did, didn't you? what are you missing - the values?

I documented the type, but how do i add a description for it, just like it can be done for parameters

chr-hertel: AFAIK that's not possible with phpdoc right away - not sure
i'm aware of [this one in http-client](https://github.com/symfony/symfony/blob/8.1/src/Symfony/Contracts/HttpClient/HttpClientInterface.php), but that's not needed here - we're good :+1:

chr-hertel:
Thank you @aszenz.

PR #1260, opened 2025-12-23
[Mate] Use compiled container

Nyholm:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | 
| License       | MIT

To make sure we resolve environment variable references, and to not pass around a container builder, I suggest the ContainerFactory should return a `ContainerInterface` instead. 

~~My only issue is for the Logger used instead the `ContainerFactory` =). I suggest to write to stderr~~

I added a `_build.logger` that is used only before we we compile the container. The definition is removed before we compile. Because we use environment variables for the `_build.logger` (see below) (and not container parameters), we can take the logger from the container before the container resolve the environment variables. 

This also allows the users to configure the `LoggerInterface::class` however they want. (with or without env vars in the container)

https://github.com/symfony/ai/blob/v0.1.0/src/mate/src/default.config.php#L17-L23

### DIFF
--- a/src/mate/src/App.php
+++ b/src/mate/src/App.php
@@ -19,7 +19,7 @@ use Symfony\AI\Mate\Command\ServeCommand;
 use Symfony\AI\Mate\Exception\UnsupportedVersionException;
 use Symfony\Component\Console\Application;
 use Symfony\Component\Console\Command\Command;
-use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\ContainerInterface;
 
 /**
  * @author Johannes Wachter <johannes@sulu.io>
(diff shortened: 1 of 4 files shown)


wachterjohannes:
@Nyholm the configureable logger makes totally sense  because when used in a agent environment stderr is gone forever ...
we dould also set the service instance built from the env vars would be a solution. in all the other services compile has already replaced the env vars - so maybe thats the solution

Nyholm:
The PR is updated and the CI is super green

wachterjohannes:
@Nyholm the idea is to have a "temporary" logger for the time before the container is compiled? Why dont we use that logger also later on? do we have any positive aspects when we remove that and add another one?

Nyholm:
You cannot configure that logger with env vars in the container. 

Also, if we dump the container to file and read the cached file (which we might in the future to optimize), then this logger won't be available.

wachterjohannes:
@Nyholm yeah thats right for optimizing it makes absolutly sense

PR #1261, opened 2025-12-23
[Mate] Allow user to configure protocol version

Nyholm:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        |
| License       | MIT

Not all AI application supports all versions. Jetbrains AI supports `2025-03-26` (that is why it is default). This PR will allow the user to configure the version themselves.

### DIFF
--- a/src/mate/src/Command/ServeCommand.php
+++ b/src/mate/src/Command/ServeCommand.php
@@ -12,6 +12,7 @@
 namespace Symfony\AI\Mate\Command;
 
 use Mcp\Capability\Discovery\Discoverer;
+use Mcp\Schema\Enum\ProtocolVersion;
 use Mcp\Schema\Icon;
 use Mcp\Server;
 use Mcp\Server\Session\FileSessionStore;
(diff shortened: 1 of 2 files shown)


src/mate/src/Command/ServeCommand.php
```diff
@@ -83,6 +84,7 @@ protected function execute(InputInterface $input, OutputInterface $output): int
         );
 
         $server = Server::builder()
+            ->setProtocolVersion(ProtocolVersion::tryFrom($this->container->getParameter('mate.mcp_protocol_version')) ?? ProtocolVersion::V2025_03_26)
```
chr-hertel: if `getParameter` returns null, we might run in a deprecation here
https://3v4l.org/OPKGQ

Nyholm: Thank you. The PR is updated.

OskarStark:
Thank you @Nyholm.

PR #1270, opened 2025-12-23
[Agent] Supporting iterable as vector store return in memory

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #... 
| License       | MIT

We missed it with #1006


OskarStark:
Thank you @chr-hertel.

PR #1271, opened 2025-12-23
 Adopt Multi Agent example after Responses API

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | yes
| Issues        | Fix #... 
| License       | MIT

we missed that with #871


OskarStark:
Thank you @chr-hertel.

Issue #1265, opened 2025-12-23
[Platform] Exposing clients to the container


Guikingone:
Hi 👋🏻 

Small suggestion for future developments regarding the `Platform`, for now, clients are not exposed through the container as they're built using the `*PlatformFactory` (example via `Ollama`): 

```php
<?php

/*
 * This file is part of the Symfony package.
 *
 * (c) Fabien Potencier <fabien@symfony.com>
 *
 * For the full copyright and license information, please view the LICENSE
 * file that was distributed with this source code.
 */

namespace Symfony\AI\Platform\Bridge\Ollama;

use Psr\EventDispatcher\EventDispatcherInterface;
use Symfony\AI\Platform\Bridge\Ollama\Contract\OllamaContract;
use Symfony\AI\Platform\Contract;
use Symfony\AI\Platform\ModelCatalog\ModelCatalogInterface;
use Symfony\AI\Platform\Platform;
use Symfony\Component\HttpClient\EventSourceHttpClient;
use Symfony\Contracts\HttpClient\HttpClientInterface;

/**
 * @author Christopher Hertel <mail@christopher-hertel.de>
 */
final class PlatformFactory
{
    public static function create(
        string $hostUrl = 'http://localhost:11434',
        ?HttpClientInterface $httpClient = null,
        ModelCatalogInterface $modelCatalog = new ModelCatalog(),
        ?Contract $contract = null,
        ?EventDispatcherInterface $eventDispatcher = null,
    ): Platform {
        $httpClient = $httpClient instanceof EventSourceHttpClient ? $httpClient : new EventSourceHttpClient($httpClient);

        return new Platform(
            [new OllamaClient($httpClient, $hostUrl)],
            [new OllamaResultConverter()],
            $modelCatalog,
            $contract ?? OllamaContract::create(),
            $eventDispatcher,
        );
    }
}
```

Once in the bundle, only the platform can be accessed, this approach lock the capacity to interact with the client (for decoration purposes mostly, can be useful for #943 and more) and even if the `Platform` is a good entry point, sometimes, we might need to interact with the client rather than the platform (which is a bit "too late" in the call chain).

Could clients be exposed in the container, maybe via a new definition during the platform building phase?

PR #1273, opened 2025-12-23
[Platform] `ElevenLabs` definitions rework

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | Part of #1265 
| License       | MIT

Hi 👋🏻 

Here's a POC for #1265, I choose EL as it's the easiest to refactor and the easiest to configure, this PR allows to remove the factory usage, add missing services / tags and rework the injection process, every class is now a service and can be decorated / lazy-loaded.


src/ai-bundle/config/services.php
```diff
@@ -102,7 +105,7 @@
         ->set('ai.platform.model_catalog.deepseek', DeepSeekModelCatalog::class)
         ->set('ai.platform.model_catalog.dockermodelrunner', DockerModelRunnerModelCatalog::class)
         ->set('ai.platform.model_catalog.elevenlabs', ElevenLabsModelCatalog::class)
-            ->lazy(true)
+            ->lazy()
```
OskarStark, line 107: why this change?

Guikingone: `true` is the default value, not needed:

```
final public function lazy(bool|string $lazy = true): static
```

chr-hertel (changes requested):
Not completely sold on the need for decoration, but let's isolate the two things here:
1. rework on bundle, skipping the factory
2. removing the `PlatformFactory`

For the second one I would be quite reluctant because that is quite handy when not using the bundle, what - at least for me - is the case regularly in prototypes not requiring the entire framework.

chr-hertel:
@Guikingone want to continue on the first one here or not needed anymore?

> 1. rework on bundle, skipping the factory

Guikingone:
Hi @chr-hertel 👋🏻 

Not needed anymore, this PR was built to ease #943 but the PR is ready without impacting services, I'll close this one.

PR #1276, opened 2025-12-23
 Bump dependencies for release v0.1

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Docs?         | no 
| Issues        | 
| License       | MIT

image

(don't get confused with "current 0.1.0" and "new 0.1.0" - it reads the value from one class constant, that we shouldn't touch manually anymore from now on and which is already on 0.1.0)


fabpot (approved):

fabpot:
Thank you @chr-hertel.

PR #1278, opened 2025-12-23
 Fix default scan directory in mate init command

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | none
| License       | MIT

## Summary

Update the default scan directory from 'mate' to 'mate/src' to correctly point to the source directory structure used in mate projects.

### DIFF
--- a/src/mate/src/Command/InitCommand.php
+++ b/src/mate/src/Command/InitCommand.php
@@ -161,7 +161,7 @@ class InitCommand extends Command
 
         if (!isset($composerJson['extra']['ai-mate'])) {
             $composerJson['extra']['ai-mate'] = [
-                'scan-dirs' => ['mate'],
+                'scan-dirs' => ['mate/src'],
                 'includes' => ['config.php'],
             ];
             $modified = true;


wachterjohannes:
@chr-hertel i found a small issue in the init command

src/mate/src/Command/InitCommand.php
```diff
@@ -161,7 +161,7 @@ private function updateComposerJson(): array
 
         if (!isset($composerJson['extra']['ai-mate'])) {
             $composerJson['extra']['ai-mate'] = [
-                'scan-dirs' => ['mate'],
+                'scan-dirs' => ['mate/src'],
                 'includes' => ['config.php'],
```
chr-hertel, line 165: config file is relative to `mate/` but `scan-dirs` not? :thinking:

wachterjohannes: yeah that seems a bit strange - should we also change this?

chr-hertel:
Thank you @wachterjohannes.

PR #1284, opened 2025-12-24
[Chat] Fix `Cloudflare` on pagination while creating/dropping namespaces

### LINKED ISSUE #1282: [Chat] The implementation of the Cloudflare chat store is broken for accounts with lots of KV namespaces
The implementation of the Cloudflare message store for the chat tries to guess the id of the KV namespace to use by identifying it based on its human-readable title (which is indeed unique in Cloudflare).
This [guessing logic](https://github.com/symfony/ai/blob/506638447e18a3ef84b2c32bd01d085fa1fb2a8a/src/chat/src/Bridge/Cloudflare/MessageStore.php#L136) calls the API endpoint to list namespaces. But it does not account for the fact that this API endpoints returns a *paginated* result: https://developers.cloudflare.com/api/resources/kv/subresources/namespaces/methods/list/

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #1282 
| License       | MIT

### DIFF
--- a/src/chat/src/Bridge/Cloudflare/MessageStore.php
+++ b/src/chat/src/Bridge/Cloudflare/MessageStore.php
@@ -11,7 +11,6 @@
 
 namespace Symfony\AI\Chat\Bridge\Cloudflare;
 
-use Symfony\AI\Chat\Exception\InvalidArgumentException;
 use Symfony\AI\Chat\ManagedStoreInterface;
 use Symfony\AI\Chat\MessageNormalizer;
 use Symfony\AI\Chat\MessageStoreInterface;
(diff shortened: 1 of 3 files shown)


src/chat/src/Bridge/Cloudflare/MessageStore.php
```diff
@@ -43,23 +43,38 @@ public function __construct(
     ) {
     }
 
+    /**
+     * @param array{
+     *     page?: int,
```
stof: I don't think `page` should be exposed as a public option. You should rather use a private method to load the appropriate page of the KV namespaces.

This private method could even be common with the logic of `retrieveCurrentNamespaceUuid` to get the current namespace Uuid (the *only* part of the logic you don't want is the exception when there is no namespace).

src/chat/src/Bridge/Cloudflare/MessageStore.php
```diff
@@ -133,17 +148,21 @@ private function request(string $method, string $endpoint, array $payload = []):
...
+        }
 
         $filteredNamespaces = array_filter(
             $namespaces['result'],
             fn (array $payload): bool => $payload['title'] === $this->namespace,
         );
 
-        if (0 === \count($filteredNamespaces)) {
-            throw new InvalidArgumentException('No namespace found.');
+        if (0 === \count($filteredNamespaces) && $page !== $namespaces['result_info']['total_pages']) {
```
stof, line 150: you still need to handle the case where no namespace is found at all while some KV namespaces with _other_ titles are defined. Checking the `total_count` is not equivalent to the old check on the _filtered_ list.

OskarStark:
Ready to merge?

Guikingone:
Except if @stof see something that I missed, on my side, yes 👍🏻

OskarStark:
Thank you @Guikingone.

PR #1290, opened 2025-12-26
[Platform][Ollama] Fix request options for client

### LINKED ISSUE #1289: [AI Bundle][Agent][Ollama] Ollama /api/chat: model options cannot be passed correctly
## Description

While integrating Ollama via `Symfony\AI\Platform\Bridge\Ollama`, I ran into an issue where model generation options cannot be sent in the format expected by Ollama `/api/chat`.

Ollama expects generation parameters (e.g. `think`, `num_predict`, `temperature`, `num_ctx`) inside the JSON field `options`, but the current bridge implementation makes this difficult or impossible to configure correctly via Symfony AI configuration.

### Current behavior

`OllamaClient::doCompletionRequest()` builds the request like this:

`$json = array_merge($options, $payload);`

As a result:
- All `model.options` from Symfony AI config are merged top-level into the JSON body.
- Ollama `/api/chat` ignores generation parameters unless they are nested under `"options": { ... }`.
- Attempting to configure nested options in model.options (e.g. `options: { think: false }`) breaks container compilation due to `%5B...%` being interpreted as DI parameters (`http_build_query()` in config normalizer).
- This makes it impossible to express the correct Ollama request shape via configuration alone.

### Expected behavior

AI bundle should allow configuring Ollama models so that the outgoing `/api/chat` request can be shaped as:
```json
{
  "stream": false,
  "options": {
    "think": false,
    "num_predict": 64,
    "temperature": 0.2,
    "num_ctx": 1024
  },
  "messages": [...]
}
```
without hacks or client decoration.

###  Reproduction
```yaml
ai:
  agent:
    ollama_debugger:
      platform: 'ai.platform.ollama'
      model:
        class: Symfony\AI\Platform\Bridge\Ollama\Ollama
        name: 'qwen3:0.6b'
        options:
          think: false
          num_predict: 64
          temperature: 0.2
```
Results in outgoing JSON roughly like:

```json
{
  "think": false,
  "num_predict": 64,
  "temperature": 0.2,
  "messages": [...]
}
```

Attempting to configure with nested options:

```
options:
  options:
    think: false
```

breaks container compilation due to `%5Bthink%5D` being treated as a Symfony parameter reference.

###
Why this matters

- Ollama models appear extremely slow or behave unexpectedly because generation options are silently ignored.
- Users currently need custom client decorators or environment hacks to get correct behavior.
- The issue is subtle and hard to diagnose without deep knowledge of Ollama’s API.

sonnymilton:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Docs?         | no 
| Issues        | Fix #1289 
| License       | MIT

Ollama expects model runtime parameters (e.g. `temperature`, `num_ctx`, `num_predict`, etc.) to be sent inside a nested `options` object. This PR adjusts the Ollama bridge so that:

- documented **top-level** request fields remain on the root for:
  - `/api/chat`: `stream`, `format`, `keep_alive`, `tools`, `think`, `logprobs`, `top_logprobs`
  - `/api/embed`: `truncate`, `keep_alive`, `dimensions`
- all other (non-top-level) parameters are moved into `options`

This makes Symfony AI requests compatible with Ollama’s request schema and prevents runtime params from being ignored.

## Changes

- Add top-level key allow-lists for chat and embed requests
- Normalize request options:
  - keep known top-level keys on root
  - merge all remaining keys into `options`
- Add tests verifying:
  - non-top-level params are nested under `options`
  - top-level keys stay on the root
  - structured output `format` stays top-level for chat
  - embedding request options are normalized as expected

## Ollama API docs
`/api/chat` - https://docs.ollama.com/api/chat
`/api/embed` - https://docs.ollama.com/api/embed

### DIFF
--- a/src/platform/src/Bridge/Ollama/OllamaClient.php
+++ b/src/platform/src/Bridge/Ollama/OllamaClient.php
@@ -24,6 +24,22 @@ use Symfony\Contracts\HttpClient\HttpClientInterface;
  */
 final class OllamaClient implements ModelClientInterface
 {
+    private const CHAT_TOP_LEVEL_KEYS = [
+        'stream',
+        'format',
+        'keep_alive',
+        'tools',
+        'think',
+        'logprobs',
+        'top_logprobs',
+    ];
+
+    private const EMBED_TOP_LEVEL_KEYS = [
+        'truncate',
+        'keep_alive',
+        'dimensions',
+    ];
+
     public function __construct(
         private readonly HttpClientInterface $httpClient,
         private readonly string $hostUrl,
@@ -58,6 +74,8 @@ final class OllamaClient implements ModelClientInterface
             unset($options[PlatformSubscriber::RESPONSE_FORMAT]);
         }
 
+        $options = $this->normalizeOllamaOptions($options, self::CHAT_TOP_LEVEL_KEYS);
+
         return new RawHttpResult($this->httpClient->request('POST', \sprintf('%s/api/chat', $this->hostUrl), [
             'headers' => ['Content-Type' => 'application/json'],
             'json' => array_merge($options, $payload),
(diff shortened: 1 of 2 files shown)


chr-hertel:
Hey @sonnymilton, thanks for looking at this, didn't realize so far that there's that nested options thingy ...

So goal is to keep the `$options` array in the `invoke` call a flat array and do the nesting for the user, right?
I agree since this unifies for example how `temperature` is handed over when switching for example between openai and ollama

sonnymilton:
@chr-hertel 
Yes. The idea is to keep the contract as a flat `$options` array.
Then each provider’s client is responsible for shaping the outgoing request to match its API (e.g. the **Gemini** client already does this by wrapping options into the appropriate structure and lifting special fields to the top level).

chr-hertel:
Thank you @sonnymilton.

PR #1297, opened 2025-12-27
[Mate] Remove $_ENV['MATE_ROOT_DIR']

Nyholm:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | 
| License       | MIT

We dont need to set `$_ENV['MATE_ROOT_DIR']`. We pass the `$rootDir` to the `ContainerFactory`. That class is also setting the `mate.root_dir` parameter.

### DIFF
--- a/src/mate/src/default.config.php
+++ b/src/mate/src/default.config.php
@@ -23,7 +23,6 @@ return static function (ContainerConfigurator $container): void {
         : false;
 
     $container->parameters()
-        ->set('mate.root_dir', '%env(MATE_ROOT_DIR)%')
         ->set('mate.cache_dir', sys_get_temp_dir().'/mate')
         ->set('mate.env_file', null)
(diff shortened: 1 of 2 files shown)


src/mate/src/default.config.php
```diff
@@ -23,7 +23,6 @@
         : false;
 
     $container->parameters()
-        ->set('mate.root_dir', '%env(MATE_ROOT_DIR)%')
```
wachterjohannes, line 26: dont we need that param?

wachterjohannes (approved):
if we dont need it just merge it

src/mate/src/default.config.php
Nyholm: It is being set in the ContainerFactory

PR #1313, opened 2025-12-29
[Platform] Add Whisper verbose support

### LINKED ISSUE #1311: [Platform][Bug] Whisper verbose option not usable
Whisper provides a verbose response format, which would require a Platform call like this:
```php
$result = $this->platform->invoke('whisper-1', Audio::fromFile($audioPath), [
    'response_format' => 'verbose_json',
]);
```

Which is currently not possible, since the PlatformSubscriber of the StructuredOutput feature valides if a model has the capability for that when the `response_format` option is set.

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | Fix #1311 
| License       | MIT

image

- [x] Tests

### DIFF
--- a/examples/azure/audio-transcript.php
+++ b/examples/azure/audio-transcript.php
@@ -25,4 +25,6 @@ $file = Audio::fromFile(dirname(__DIR__, 2).'/fixtures/audio.mp3');
 
 $result = $platform->invoke('whisper-1', $file);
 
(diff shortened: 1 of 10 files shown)


OskarStark:
Love it ❤️

chr-hertel:
someone had a good idea - thanks for that :D

Issue #1323, opened 2025-12-30
[Platform] VertexAI does not support HttpClient configuration


koenkivits:
The documentation currently states that you can [override the HTTP client service for every platform](https://symfony.com/doc/current/ai/bundles/ai-bundle.html#http-client-configuration). It doesn't work for VertexAI, though:

```yaml
ai:
    platform:
        vertexai:
            http_client: my_http_client
```

... will result in the following exception:

> Unrecognized option "http_client" under "ai.platform.vertexai". Available options are "location", "project_id".

This is because the VertexAI platform uses its own HTTP client, rather than a reference to another service: https://github.com/symfony/ai/blob/6b5f7bd980133385054c00471d9a3f4822a0d319/src/ai-bundle/src/AiBundle.php#L686-L691

---

I guess the best way to solve this would be to wrap the existing `http_client` service in something that passes along the appropriate credentials for every call, rather than having a dedicated client?

(I'm open to creating a PR to fix this, by the way, but I figured I'd check in with an issue first)

OskarStark:
PR welcome, thank you 🙏

koenkivits:
@OskarStark just checking back in, and I see a PR from me is no longer needed. Nice work! 😄

OskarStark:
😉

PR #1329, opened 2025-12-30
[AI Bundle][VertexAI] Add `http_client` option configuration 

OskarStark:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | Fix https://github.com/symfony/ai/issues/1323
| License       | MIT


src/ai-bundle/composer.json
```diff
@@ -89,7 +89,7 @@
         "symfony/ai-surreal-db-store": "^0.2",
         "symfony/ai-transformers-php-platform": "^0.1",
         "symfony/ai-typesense-store": "^0.2",
-        "symfony/ai-vertex-ai-platform": "^0.1",
+        "symfony/ai-vertex-ai-platform": "^0.2",
```
chr-hertel, line 92: It's a bundle only change - bump shouldn't be required here :thinking:

OskarStark: It is, the new option will otherwise not work in the underlying bridge code

PR #1330, opened 2025-12-30
[Platform][Gemini][VertexAI] Fix nullable parameter type handling in `ToolNormalizer`

### LINKED ISSUE #1127: Gemini ToolNormalizer doesn't handle nullable parameter types correctly
_Disclaimer: I debugged this issue with copilot and asked it to write up the report_

### Describe Your Problem 🎯

**The problem:**
The `ToolNormalizer` for Gemini produces JSON Schema with array-style nullable types (`"type": ["string", "null"]`) which Gemini's API does not accept. Gemini expects the format `"type": "string", "nullable": true` instead.

**Expected behavior:**
Tool parameter schemas with nullable types should be formatted as:

```json
{
  "optional": {
    "type": "string",
    "nullable": true
  }
}
```

**Actual behavior:**
The schema is produced with array-style types:

```json
{
  "optional": {
    "type": ["string", "null"]
  }
}
```

This causes Gemini to return an error:
```
Invalid JSON payload received. Unknown name "type" at 'tools[0].function_declarations[X].parameters.properties[Y].value': Proto field is not repeating, cannot start list.
```

**Steps to reproduce:**

1. Create a tool with nullable parameters:

```php
#[AsTool(
    name: 'example_tool',
    description: 'An example tool with nullable parameters',
)]
final class ExampleTool
{
    public function __invoke(
        string $required,
        ?string $optional = null,
    ): string {
        return 'result';
    }
}
```

2. Register the tool with an Agent using the Gemini platform
3. Send a chat message that triggers tool schema serialization

### Provide Detailed Information 📋

**Component version:** symfony/ai-platform (latest/main)

**Root cause:**
The `JsonSchema\Factory` class in `src/Contract/JsonSchema/Factory.php` produces the array-style type format (around line 104):

```php
if ($type->isNullable()) {
    if (!isset($schema['anyOf'])) {
        $schema['type'] = [$schema['type'], 'null'];
    }
}
```

The `ToolNormalizer` for Gemini (`src/Bridge/Gemini/Contract/ToolNormalizer.php`) only removes `additionalProperties` but doesn't convert the array-style types to Gemini's expected format.

**Suggested fix:**
The Gemini `ToolNormalizer` should recursively transform the schema to convert `"type": ["string", "null"]` to `"type": "string", "nullable": true`.

**Reference:** [Gemini API Schema documentation](https://ai.google.dev/api/caching#Schema) shows `nullable` as a boolean property on Schema objects.

OskarStark:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #1127
| License       | MIT

Convert array-style nullable types like `["string", "null"]` to Gemini's required format `{"type": "string", "nullable": true}`.

This fixes the "Proto field is not repeating, cannot start list" error when using nullable parameters with Gemini and VertexAI.

### DIFF
--- a/src/platform/src/Bridge/Gemini/Contract/ToolNormalizer.php
+++ b/src/platform/src/Bridge/Gemini/Contract/ToolNormalizer.php
@@ -38,7 +38,7 @@ final class ToolNormalizer extends ModelContractNormalizer
         return [
             'description' => $data->getDescription(),
             'name' => $data->getName(),
-            'parameters' => $data->getParameters() ? $this->removeAdditionalProperties($data->getParameters()) : null,
+            'parameters' => $data->getParameters() ? $this->normalizeSchema($data->getParameters()) : null,
         ];
     }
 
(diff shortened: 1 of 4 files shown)


chr-hertel:
Thank you @OskarStark.

PR #1333, opened 2026-01-01
[Mate] Add error handling for PID file creation in ServeCommand

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        |  none
| License       | MIT

## Summary

- Check return value of file_put_contents() when creating PID file
- Log warning if PID file creation fails instead of silently failing
- Add file_exists() check before cleanup in finally block
- Use error suppression operator to prevent PHP warnings


chr-hertel:
Thank you @wachterjohannes.

PR #1334, opened 2026-01-01
[Mate] Add file size protection to prevent memory issues in LogReader

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | none
| License       | MIT

## Summary

- Add 10MB file size threshold check before loading file into memory
- Implement tailFromLargeFile() streaming method for files >10MB
- Use sliding buffer approach instead of loading entire file
- Keep buffer at 2x requested lines to account for filtered entries
- Ensure proper resource cleanup with try-finally block

### DIFF
--- a/src/mate/src/Bridge/Monolog/Service/LogReader.php
+++ b/src/mate/src/Bridge/Monolog/Service/LogReader.php
@@ -173,29 +173,7 @@ final class LogReader
             return [];
         }
 
-        $entries = [];
-        $allLines = file($file, \FILE_IGNORE_NEW_LINES | \FILE_SKIP_EMPTY_LINES);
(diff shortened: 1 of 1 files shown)


chr-hertel:
I wonder if tailing should just be the default - no size checks, only one scalable implementation - WDYT?

wachterjohannes:
> I wonder if tailing should just be the default - no size checks, only one scalable implementation - WDYT?

good point will change that

wachterjohannes:
@chr-hertel its done

chr-hertel:
Thank you @wachterjohannes.

PR #1338, opened 2026-01-03
[MCP Bundle] MCP oauth discovery

camilleislasse:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | -
| License       | MIT

## Summary

Adds OAuth 2.0 support for MCP servers following the [MCP Authorization specification](https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization).

### OAuth Discovery (RFC 9728)

- `OAuthMetadataController` serves `/.well-known/oauth-protected-resource` endpoint
- `OAuthUnauthorizedListener` adds `WWW-Authenticate` header to 401/403 responses
- Configurable via `mcp.oauth.enabled` and `mcp.oauth.authorization_servers`

### OAuth Scope Checking (RFC 6750)

- `#[RequireScope]` attribute to protect tools/prompts/resources by OAuth scope
- `ScopeChecker` validates scopes before executing MCP requests
- `ScopeExtractorInterface` with `RoleScopeExtractor` default (compatible with `league/oauth2-server-bundle`)
- Returns HTTP 403 with `WWW-Authenticate` header containing `error="insufficient_scope"`

### Usage

```php
use Symfony\AI\McpBundle\Security\Attribute\RequireScope;

#[McpTool(name: 'delete-user')]
#[RequireScope('admin')]
public function deleteUser(int $userId): string
{
    // Only accessible with 'admin' scope
}
```

### Configuration

```yaml
mcp:
    oauth:
        enabled: true
        authorization_servers:
            - 'https://auth.example.com'
```

### Scope Extraction

Scopes are extracted via `ScopeExtractorInterface`. The default `RoleScopeExtractor` maps Symfony roles to scopes:

```
ROLE_OAUTH2_READ  → scope "read"
ROLE_OAUTH2_ADMIN → scope "admin"
```

This is compatible with `league/oauth2-server-bundle`. You can register your own extractor (e.g., for JWT claims) by implementing `ScopeExtractorInterface`.

> **Note:** The bundle provides OAuth discovery and scope checking only. To protect your MCP endpoint, configure Symfony's security layer with an OAuth firewall.


chr-hertel:
Hi @camilleislasse, oauth is an important thing to get in place, true, but isn't this duplicating efforts of https://github.com/modelcontextprotocol/php-sdk/pull/194?

camilleislasse:
> Hi @camilleislasse, oauth is an important thing to get in place, true, but isn't this duplicating efforts of [modelcontextprotocol/php-sdk#194](https://github.com/modelcontextprotocol/php-sdk/pull/194)?

Oh, I wasn't aware of that PR. Let me look into it in more detail and I'll get back to you
 
 From what I understand:

  Overlap:
  - /.well-known/oauth-protected-resource endpoint
  - WWW-Authenticate headers

  Difference:
  - SDK validates tokens itself (OAuth2HttpTransport + JwtTokenAuthenticator), independently of Symfony Security
  - My PR delegates token validation to Symfony Security and adds #[RequireScope] for per-tool scope checking

  The #[RequireScope] feature could also be added to the SDK (e.g., a scope param on #[McpTool]).

Spomky:
Hi,

In my opinion, this feature should not be part of the MCP Bundle. 
Symfony already provides a Security component and bundle that offer fully-featured OAuth2/OpenID Connect resource protection.

If a specific requirement is missing (e.g. protecting a resource based on specific scopes) the Security component is the most appropriate place to handle this logic.
I would prefer to see an improvement in `symfony/symfony` rather than duplicating security concepts and logic within bundles.

PR #1343, opened 2026-01-05
[Mate] CS fixes

lochmueller:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | 
| License       | MIT

Just execute PHP-CS-Fixer...

### DIFF
--- a/src/mate/resources/mate/config.php
+++ b/src/mate/resources/mate/config.php
@@ -1,6 +1,14 @@
 <?php
 
-// User's service configuration file
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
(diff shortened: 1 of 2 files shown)


OskarStark:
Thank you @lochmueller.

Nyholm:
These files should be excluded from the CS.

OskarStark:
You are right

chr-hertel:
They are actually excluded :thinking:

lochmueller:
@chr-hertel 
Thanks for checking. Found the problem... executing `./vendor/bin/php-cs-fixer fix src` ignore the exclude list. Next time, I run only `./vendor/bin/php-cs-fixer fix` on root level.
Regards, Tim

PR #1345, opened 2026-01-06
[Platform][OpenRouter] Add streaming and structure output capabilities

lochmueller:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | 
| License       | MIT

Openrouter support streaming options for every model. Capability is added to every model.
Furthermore, the dynamic catalog check the structure output capability and map it to the internal flag.

### DIFF
--- a/examples/openrouter/structured-output-math.php
+++ b/examples/openrouter/structured-output-math.php
@@ -1,0 +1,32 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use Symfony\AI\Platform\Bridge\OpenRouter\PlatformFactory;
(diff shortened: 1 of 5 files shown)


chr-hertel (changes requested):
Nice, thanks @lochmueller!

Can you please add an example similar to `examples/openai/structured-output-math.php` and add support for structured output in the changelog of the bridge for 0.2?
i think streaming worked before, right?

lochmueller:
@chr-hertel 
Example is added. You are right... streaming works before, because there is no check, if the model support streaming - neither before the request, nor after the response (converter).
Regards,
Tim

chr-hertel:
> @chr-hertel Example is added.

@lochmueller Please push :)

OskarStark:
Ask Claude to SSH into his machine and push it on your own 😂

lochmueller:
Here we go ;) Next time I share my SSH key ;)

chr-hertel:
Thank you @lochmueller.

PR #1350, opened 2026-01-06
[AI Bundle] Only throw if a store is actually configured

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #... 
| License       | MIT

I know we don't like extra vars, but this helps with readability i think


OskarStark:
Thank you @chr-hertel.

PR #1356, opened 2026-01-07
[Chat] Fixes on `DoctrineDbalMessageStore`

### LINKED ISSUE #1283: [Chat] the DBAL message store does not use the DBAL API properly.
DBAL has 2 methods to execute DB queries:
- `executeQuery` for queries returning a result set
- `executeStatement` for queries returning only a number of affected rows.

The `PrimaryReadReplicaConnection` of DBAL supports seamlessly switching between the primary and a read-only replica (to reduce the load on the primary) based on some assumptions:
- using a transaction forces using the primary (as it cannot know what the transaction will contain, and it cannot switch in the middle of the transaction as transactions are stateful)
- `executeStatement` forces using the primary (as this is used to run `UPDATE` and `DELETE` queries)
- `executeQuery` keeps the current target, and connects to the replica by default if not connected yet, as this is typically used for `SELECT` queries (some special rarely used queries also need to use it, but require forcing the usage of the primary)

The DBAL store wraps all executions in a DBAL transactions, which is totally unnecessary (running a single query is *already* atomic) and prevents using the read replica for read queries.
The DBAL store also uses `executeQuery` for write queries, which will cause issues with the `PrimaryReadReplicaConnection` if it is not wrapped in a transaction anymore.

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #1283 
| License       | MIT

### DIFF
--- a/src/chat/src/Bridge/Doctrine/DoctrineDbalMessageStore.php
+++ b/src/chat/src/Bridge/Doctrine/DoctrineDbalMessageStore.php
@@ -11,21 +11,21 @@
 
 namespace Symfony\AI\Chat\Bridge\Doctrine;
 
-use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\Connection as DBALConnection;
 use Doctrine\DBAL\Platforms\OraclePlatform;
-use Doctrine\DBAL\Result;
 use Doctrine\DBAL\Schema\Name\Identifier;
 use Doctrine\DBAL\Schema\Name\UnqualifiedName;
 use Doctrine\DBAL\Schema\PrimaryKeyConstraint;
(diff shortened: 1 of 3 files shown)


src/chat/src/Bridge/Doctrine/phpunit.xml.dist
```diff
@@ -8,6 +8,7 @@
          failOnDeprecation="true"
          failOnRisky="true"
          failOnWarning="true"
+         executionOrder="random"
```
OskarStark, line 11: please create a PR and add this to all phpunit.xml.dist files

OskarStark:
Thank you @Guikingone.

PR #1358, opened 2026-01-07
[Platform] Move stream up in convert function, in front of $response access

lochmueller:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | 
| License       | MIT

By design, I suggest checking the stream output always first in the Result converter. Reason: If the response is really a AsyncResponse, then even a "getStatusCode" on the response trigger the initialize process of the async handling. I suggest handling the stream in the first line of the convert will avoid this problem.

Any thought related to this?

### DIFF
--- a/src/platform/src/Bridge/Anthropic/ResultConverter.php
+++ b/src/platform/src/Bridge/Anthropic/ResultConverter.php
@@ -35,16 +35,16 @@ class ResultConverter implements ResultConverterInterface
 
     public function convert(RawHttpResult|RawResultInterface $result, array $options = []): ResultInterface
     {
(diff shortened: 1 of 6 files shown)


chr-hertel:
what is the bug you're acting on here?
not sure i can follow, since the the call to the result converter itself is delay until we access the results content - but this might be a misunderstanding

lochmueller:
Hey @chr-hertel 
cannot reproduce the related issue anymore. I will work on the StreamResult the next days. I I can reproduce the issue, I will recreate the PR. Thanks,
TIm

PR #1372, opened 2026-01-09
 [Mate] Fix mate logger failing when log directory doesn't exist

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | N/A
| License       | MIT

### Summary

The mate logger was failing to start when the configured log file path included a non-existing directory, because file_put_contents() does not automatically create parent directories.

Changes:
- Add ensureDirectoryExists() method to Logger service
- Automatically creates parent directories with 0755 permissions
- Throws clear exception if directory creation fails due to permissions
- Add test coverage for automatic directory creation
- Add test for permission-based directory creation failures
- Update existing test to expect directory creation error message

### DIFF
--- a/src/mate/src/Service/Logger.php
+++ b/src/mate/src/Service/Logger.php
@@ -47,6 +47,8 @@ class Logger extends AbstractLogger
         );
 
         if ($this->fileLogEnabled || !\defined('STDERR')) {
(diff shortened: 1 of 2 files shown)


chr-hertel:
Thank you @wachterjohannes.

PR #1374, opened 2026-01-10
 Limit root package dependencies

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | 
| License       | MIT

Not sure why we need this wide range - it's not a published package, no external incoming dependency. Pipeline will tell ...

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -8,10 +8,10 @@
     "require-dev": {
         "php": ">=8.2",
         "deptrac/deptrac": "^4.2",
-        "php-cs-fixer/shim": "^3.91.2",
-        "symfony/console": "^5.4|^6.4|^7.4|^8.0",
-        "symfony/filesystem": "^5.4|^6.4|^7.4|^8.0",
-        "symfony/finder": "^5.4|^6.4|^7.4|^8.0"
+        "php-cs-fixer/shim": "^3.92",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -8,10 +8,10 @@
...
         "php": ">=8.2",
         "deptrac/deptrac": "^4.2",
-        "php-cs-fixer/shim": "^3.91.2",
-        "symfony/console": "^5.4|^6.4|^7.4|^8.0",
-        "symfony/filesystem": "^5.4|^6.4|^7.4|^8.0",
-        "symfony/finder": "^5.4|^6.4|^7.4|^8.0"
+        "php-cs-fixer/shim": "^3.92",
+        "symfony/console": "^7.4",
+        "symfony/filesystem": "^7.4",
+        "symfony/finder": "^7.4"
```
smnandre, line 14: Is 8.0 really problematic ?  😐

OskarStark: No, its more about 5.4 and 6.4

smnandre: I mean, do you think we can keep 8 ? 😊

OskarStark: I think yes

PR #1375, opened 2026-01-12
[Evaluator] Add component

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | --
| License       | MIT

This PR (as a POC for now) introduce the `Evaluator` component, this component is mainly used to evaluate and add a score to output from platforms / agents, I moved it to a new directory as a mirror of the `Validator` component, the goal is to allows to add new scorers without impacting the platforms (even if they can define their own scorers).

The PR is built around the following concepts:

- An `Evaluator` that receive `Scorers` and compute the final score
- `ScorerInterface` implementations that defines score using `score` method (could be improved, I'm not locked on the name) 
- An `AbstractScorer` used to define a reason (mostly used for "LLM as judge" scorers) for a score
- An implementation on `AiBundle` via the Profiler and a subscriber (no configuration for now).

### DIFF
--- a/splitsh.json
+++ b/splitsh.json
@@ -30,6 +30,9 @@
         "ai-pogocache-message-store": "src/chat/src/Bridge/Pogocache",
         "ai-redis-message-store": "src/chat/src/Bridge/Redis",
         "ai-surreal-db-message-store": "src/chat/src/Bridge/SurrealDb",
+        "ai-evaluator": {
+            "prefixes": [{ "from": "src/evaluator", "to": "", "excludes": ["src/Bridge"] }]
+        },
         "ai-mate": {
             "prefixes": [{ "from": "src/mate", "to": "", "excludes": ["src/Bridge", "composer-plugin"] }]
         },
--- a/examples/composer.json
+++ b/examples/composer.json
@@ -33,6 +33,7 @@
         "symfony/ai-doctrine-message-store": "^0.11",
         "symfony/ai-elasticsearch-store": "^0.11",
         "symfony/ai-eleven-labs-platform": "^0.11",
+        "symfony/ai-evaluator": "^0.11",
         "symfony/ai-failover-platform": "^0.11",
         "symfony/ai-gemini-platform": "^0.11",
         "symfony/ai-generic-platform": "^0.11",
--- a/examples/evaluator/evaluator-output-end-with.php
+++ b/examples/evaluator/evaluator-output-end-with.php
@@ -1,0 +1,36 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use Symfony\AI\Evaluator\Evaluator;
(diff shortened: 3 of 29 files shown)


splitsh.json
```diff
@@ -28,6 +28,9 @@
         "ai-pogocache-message-store": "src/chat/src/Bridge/Pogocache",
         "ai-redis-message-store": "src/chat/src/Bridge/Redis",
         "ai-surreal-db-message-store": "src/chat/src/Bridge/SurrealDb",
+        "ai-evaluator": {
+            "prefixes": [{ "from": "src/evaluator", "to": "", "excludes": ["src/Bridge"] }]
```
OskarStark, line 34:
```suggestion
            "prefixes": [{ "from": "src/evaluator", "to": "" }]
```
not sure, as we don't have a bridge folder

Guikingone: Oops, wrong copy/paste 😅

OskarStark:
Does this need to be a dedicated component? If not, where would it fit best? Platform or Agent?

Guikingone:
> Does this need to be a dedicated component?

Open to debate and I don't have a strong opinion on this one, the first version was a sub-directory of `Platform` but as PR's tend to stay "simple" and once complexity is added, we tend to move it as "bridge", I preferred to move it as a new component to keep clean the `Platform` one.

If it needs to be moved to an existing one, I would say `Platform`, we don't depends on `Agent`.

aszenz:
> > Does this need to be a dedicated component?
> 
> Open to debate and I don't have a strong opinion on this one, the first version was a sub-directory of `Platform` but as PR's tend to stay "simple" and once complexity is added, we tend to move it as "bridge", I preferred to move it as a new component to keep clean the `Platform` one.
> 
> If it needs to be moved to an existing one, I would say `Platform`, we don't depends on `Agent`.

Can it be integrated into PHPUnit, i think it kind of provides testing utilities for the platform component so it should ideally be in platform component itself

Guikingone:
> Can it be integrated into PHPUnit

Like `Mailer` / `Messenger` assertions? Of course 🙂

chr-hertel:
First of all: yes we need this! i wonder if it's too early tho ...

To me this is rather a standalone component, yes, but the integration aspect with PHPUnit is interesting, but not sure we want to do this - i might want to use evaluations not only in my test suite :thinking: 

For me it's too early since i feel like putting something on top of agent/chat without them being stable enough - and i would rather focus on that first - that shouldn't keep you from exploring the field here ofc. it's just another thing in parallel and i'm a bit worried about too many moving parts tbh.

what's your blueprint for tackling this? do you use some kind of reference?
i recall discussing this briefly with @tgalopin in Amsterdam - was it dspy or pydantic? or sth else?

Guikingone:
Hi @chr-hertel 👋🏻 

References mostly comes from `Agno` and `Mastra`, they both implement a "scorer / evaluator" component, my first draft was built around the `Validator` component but I wasn't convinced about the API and the real interest of using it as a "core logic handler".

Dspy is not about evaluating the input/output but about building it *for* the model / *by* the model rather than sending it as an "incomplete prompt".

PR #1385, opened 2026-01-13
[Platform][OpenRouter] Add Cached HTTP client example for OpenRouter ModelApiCatalog

lochmueller:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | 
| License       | MIT

Small example to use the Symfony CachingHttpClient to fetch the model List via ModelApiCatalog (GET https://openrouter.ai/api/v1/models ). I already contacted the OpenRouter guys, to drop the "Cache-Control: private" from the response of the endpoint, if there is no Token attached to the request, to allow caching based on HTTP defaults.

### DIFF
--- a/examples/openrouter/cached-api-model-catalog.php
+++ b/examples/openrouter/cached-api-model-catalog.php
@@ -1,0 +1,43 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use Symfony\AI\Platform\Bridge\OpenRouter\ModelApiCatalog;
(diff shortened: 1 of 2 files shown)


examples/openrouter/cached-api-model-catalog.php
```diff
@@ -0,0 +1,45 @@
...
+$cache = new FilesystemTagAwareAdapter(
+    namespace: 'model_catalog',
+    defaultLifetime: ONE_WEEK,
+    directory: dirname(__DIR__).'/var/',
+);
+
+$cachedHttpClient = new CachingHttpClient(
+    client: http_client(),
+    cache: $cache,
+    maxTtl: ONE_WEEK
```
OskarStark: No need for a constant I guess, let's add // one week here as comment

chr-hertel:
Thank you @lochmueller.

PR #1389, opened 2026-01-13
[Store] Add support for "remove" operation

### LINKED ISSUE #1373: [Store] Add Support to "delete" Operation
Hello, sometimes you need to remove vectors from the store, particularly when working with library documentation. Every time the documentation changes, I need to update the store with new vectors and remove the old ones.
Actually, that's the last operation I need to fully remove my custom Pinecone integration in favor of Symfony AI-store
Let me know if you agree with my proposal, and I can help with a PR

MolloKhan:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | Fix #1373 
| License       | MIT

Added `remove()` method to the `StoreInterface` and added real implementation to the Pinecone bridge

### DIFF
--- a/src/store/CHANGELOG.md
+++ b/src/store/CHANGELOG.md
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+0.4
+---
+
+ * Add `StoreInterface::remove()` method
+
 0.3
 ---
 
--- a/src/store/src/Bridge/AzureSearch/SearchStore.php
+++ b/src/store/src/Bridge/AzureSearch/SearchStore.php
@@ -15,6 +15,7 @@ use Symfony\AI\Platform\Vector\NullVector;
 use Symfony\AI\Platform\Vector\Vector;
 use Symfony\AI\Store\Document\Metadata;
 use Symfony\AI\Store\Document\VectorDocument;
+use Symfony\AI\Store\Exception\LogicException;
 use Symfony\AI\Store\StoreInterface;
 use Symfony\Contracts\HttpClient\HttpClientInterface;
 
@@ -45,6 +46,11 @@ final class SearchStore implements StoreInterface
         $this->request('index', [
             'value' => array_map([$this, 'convertToIndexableArray'], $documents),
         ]);
+    }
+
+    public function remove(string|array $ids, array $options = []): void
+    {
+        throw new LogicException('Method not implemented yet.');
     }
 
     public function query(Vector $vector, array $options = []): iterable
--- a/src/store/src/StoreInterface.php
+++ b/src/store/src/StoreInterface.php
@@ -25,6 +25,12 @@ interface StoreInterface
     public function add(VectorDocument|array $documents): void;
 
     /**
+     * @param string|array<string> $ids
+     * @param array<string, mixed> $options
+     */
+    public function remove(string|array $ids, array $options = []): void;
+
+    /**
      * @param array<string, mixed> $options
      *
      * @return iterable<VectorDocument>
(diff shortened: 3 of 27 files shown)


src/store/src/Bridge/AzureSearch/SearchStore.php
```diff
@@ -48,6 +48,11 @@ public function add(VectorDocument|array $documents): void
         ]);
     }
 
+    public function remove(VectorDocument|array $documents, array $options = []): void
+    {
+        // TODO: Implement remove() method.
```
MolloKhan: Is this ok, or should it throw an exception?

OskarStark: I think we should either implement all or have a dedicated interface for now and only implement it where we have a proper remove method instead place.

WDYT @chr-hertel ?

MolloKhan: Implementing it on all bridges would be a good amount of work. It would be nice if more people from the community could help with it. I'm not a fan of adding an interface only for the `remove()` method, but it's probably a good temporary solution

src/store/src/StoreInterface.php
```diff
@@ -24,6 +24,12 @@ interface StoreInterface
      */
     public function add(VectorDocument|array $documents): void;
 
+    /**
+     * @param VectorDocument|VectorDocument[] $documents
+     * @param array<string, mixed>            $options
+     */
+    public function remove(VectorDocument|array $documents, array $options = []): void;
```
chr-hertel: is it really a `VectorDocument` here? we only need the ID, right?

src/store/src/Bridge/AzureSearch/SearchStore.php
chr-hertel: can we have a `Good first issue` umbrella issue?

MolloKhan: Sure, as I mentioned in #1373 whenever I need to update the documentation of a library, I delete all the related vectors first
Another use-case is when you store the comments of your users, they can be edited or deleted, so in any case, you need to delete vectors
Basically, this applies whenever the stored data is not relevant anymore.

src/store/src/StoreInterface.php
MolloKhan: Yes, it only needs the ids. I just thought it would be easier for the end-user to send the `VectorDocument` object because that's what you get when calling `query()`. But that's just my point of view, I'm open to suggestions

chr-hertel: i think i'd favor the ID only or are there downsides to it?

MolloKhan: Nop, I don't think there's a downside. I did it just for convenience

MolloKhan: Alright, done :)

MolloKhan:
Sounds good to me

MolloKhan:
This one is ready

src/store/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ CHANGELOG
 0.3
 ---
 
+ * Add `StoreInterface::remove()` method
```
OskarStark, line 7: 0.4 in the meantime

MolloKhan: Ha! this is moving too fast

chr-hertel:
Thank you @MolloKhan.

PR #1392, opened 2026-01-14
[Mate] Add support for extension exclusion in Mate

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | N/A
| License       | MIT

## Summary

Add support for 'extension: false' flag in extra.ai-mate configuration to exclude packages from being discovered as extensions. This is useful for vendor packages that use Mate for internal development tooling but should not be detected as user-facing extensions.

### DIFF
--- a/demo/composer.json
+++ b/demo/composer.json
@@ -103,6 +103,7 @@
             "require": "8.0.*"
         },
         "ai-mate": {
+            "extension": false,
             "scan-dirs": [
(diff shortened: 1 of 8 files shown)


chr-hertel:
Thank you @wachterjohannes.

PR #1397, opened 2026-01-15
[Store] Allow `id` to be `int|string|Uuid` for `VectorDocument` and `TextDocument`

### LINKED ISSUE #857: [Store] Allow VectorDocument ID to be a string

MolloKhan:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no 
| Issues        | Fix #857
| License       | MIT

This PR is a duplicate of #867 but with a clean commit history

### DIFF
--- a/src/store/src/Bridge/MongoDb/Store.php
+++ b/src/store/src/Bridge/MongoDb/Store.php
@@ -11,7 +11,6 @@
 
 namespace Symfony\AI\Store\Bridge\MongoDb;
 
-use MongoDB\BSON\Binary;
 use MongoDB\Client;
 use MongoDB\Collection;
 use MongoDB\Driver\Exception\CommandException;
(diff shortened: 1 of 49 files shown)


src/store/src/Bridge/MongoDb/Store.php
OskarStark, line 1: @GromNaN are you fine with this change?

chr-hertel (approved):

src/store/src/Bridge/MongoDb/Store.php
chr-hertel: please address if needed as follow up, maybe detecting UUID strings to opt-in to binary?

chr-hertel:
Thank you @MolloKhan.

PR #1400, opened 2026-01-15
[Platform][ElevenLabs] Support `voice_settings` options

laszlo-karpati:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | 
| License       | MIT

### This PR adds support for passing voice_settings options to the ElevenLabs Text-to-Speech API.

Previously, the agent only forwarded the required text and model_id fields when performing a TTS request. ElevenLabs also supports optional voice_settings (e.g. stability, similarity boost, style, etc.), but there was no way to provide them through the agent.

This change conditionally includes voice_settings in the request payload when the option is provided, while preserving the existing behavior when it is not.

### Changes

Add optional voice_settings to the JSON payload of the Text-to-Speech request

Keep backward compatibility by only sending voice_settings when explicitly defined

### Example
```php
$result = $this->platform->invoke('eleven_flash_v2_5', ['text' => $text], [
    'voice' => 'my-voice-id',
    'voice_settings' => [
        'stability' => 0.5,
        'use_speaker_boost' => 1,
        'similarity_boost' => 0.7,
        'style' => 0.2,
        'speed' => 1.2,
    ],
]);
```

### DIFF
--- a/src/platform/src/Bridge/ElevenLabs/ElevenLabsClient.php
+++ b/src/platform/src/Bridge/ElevenLabs/ElevenLabsClient.php
@@ -93,6 +93,8 @@ final class ElevenLabsClient implements ModelClientInterface
             ? \sprintf('%s/text-to-speech/%s/stream', $this->hostUrl, $voice)
             : \sprintf('%s/text-to-speech/%s', $this->hostUrl, $voice);
 
+        unset($options['voice'], $options['stream']);
+
         return new RawHttpResult($this->httpClient->request('POST', $url, [
             'headers' => [
                 'xi-api-key' => $this->apiKey,
@@ -100,6 +102,7 @@ final class ElevenLabsClient implements ModelClientInterface
             'json' => [
                 'text' => $payload['text'],
                 'model_id' => $model->getName(),
+                ...$options,
             ],
         ]));
     }
--- a/src/platform/src/Bridge/ElevenLabs/Tests/ElevenLabsClientTest.php
+++ b/src/platform/src/Bridge/ElevenLabs/Tests/ElevenLabsClientTest.php
@@ -229,4 +229,51 @@ final class ElevenLabsClientTest extends TestCase
         $this->assertInstanceOf(RawHttpResult::class, $result);
         $this->assertSame(1, $httpClient->getRequestsCount());
     }
+
+    public function testClientCanPerformTextToSpeechRequestWithExtraApiOptions()
+    {
+        $payload = Audio::fromFile(\dirname(__DIR__, 6).'/fixtures/audio.mp3');
+
+        $httpClient = new MockHttpClient(function (string $method, string $url, array $options) use ($payload) {
+            $this->assertSame('POST', $method);
+            $this->assertArrayHasKey('body', $options);
+            $body = json_decode($options['body'], true);
+            $this->assertArrayHasKey('voice_settings', $body);
+            $this->assertArrayNotHasKey('voice', $body);
+            $this->assertArrayNotHasKey('stream', $body);
+            $this->assertSame([
+                'stability' => 0.5,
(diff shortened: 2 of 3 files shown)


OskarStark:
Please add a test case and use the same table for the PR header like done in other PRs, thanks

laszlo-karpati:
> Please add a test case and use the same table for the PR header like done in other PRs, thanks

Hi @OskarStark ! I’ve added a test case for `voice_options` and updated the PR header to use the same request table as in the other PRs.

src/platform/src/Bridge/ElevenLabs/ElevenLabsClient.php
```diff
@@ -100,6 +100,7 @@ private function doTextToSpeechRequest(Model $model, array|string $payload, arra
             'json' => [
                 'text' => $payload['text'],
                 'model_id' => $model->getName(),
+                ...(isset($options['voice_settings']) ? ['voice_settings' => $options['voice_settings']] : []),
```
chr-hertel: we could make this even a bit easier and potentially open for more with just:
```suggestion
                ...$options,
```

chr-hertel (changes requested):
Thanks @laszlo-karpati, please see that one comment - should be a quick one :)

src/platform/src/Bridge/ElevenLabs/Tests/ElevenLabsClientTest.php
```diff
@@ -229,4 +229,50 @@ public function testClientCanPerformTextToSpeechRequestAsStreamVoiceKeyIsProvide
...
+                    'style' => 0.2,
+                    'speed' => 1.2,
+                ],
+            ]
+        );
+
+        $this->assertSame(1, $httpClient->getRequestsCount());
+    }
+
 }
```
chr-hertel, line 279: line break too much - see fabbot
```suggestion
    }
}
```

src/platform/src/Bridge/ElevenLabs/ElevenLabsClient.php
laszlo-karpati: hi! 
I don’t think spreading `...$options` into the JSON body is safe here. The `$options` array already contains request-level configuration (e.g. `voice` and `stream`) that are used to build the URL or control behavior, not part of the API payload itself.

If we spread `...$options`, keys like voice would end up in the JSON body even though the API expects them in the URL.

chr-hertel: yes, `voice` and `stream` should go into that payload - that's a good catch!

however, the approach you chose doesn't scale well over time - we missed to bring in options to the request body in the beginning, that's why you needed to patch it for `voice_settings`, but there are way more options, like `seed`, `apply_text_normalization`, and [more](https://elevenlabs.io/docs/api-reference/text-to-speech/convert#request).

we would end up with could like
```php
...(isset($options['voice_settings']) ? ['voice_settings' => $options['voice_settings']] : []),
...(isset($options['seed']) ? ['seed' => $options['seed']] : []),
...(isset($options['apply_text_normalization']) ? ['apply_text_normalization' => $options['apply_text_normalization']] : []),
```
and that would req patching whenever they add one option to their api

the solution that we use in different places is simply to unset the unwanted options, in this case `voice` and `stream`

laszlo-karpati: I’ve updated the code to unset `voice` and `stream` before spreading `...$options` into the payload.
Tests have been updated accordingly.

OskarStark (approved):

chr-hertel (approved):

chr-hertel:
Thank you @laszlo-karpati.

PR #1403, opened 2026-01-15
 Use generic PlatformFactory for generic based factories

lochmueller:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | 
| License       | MIT

- Used central PlatformFactory of the generic package in the "simple packages" that based on the generic
- Drop `Contract::create()` fallback (Platform create also the default contract itself as fallback)


chr-hertel (approved):
Thanks @lochmueller - removed the classification as bug

chr-hertel:
Thank you @lochmueller.

PR #1410, opened 2026-01-16
[Platform] `ElevenLabsApiCatalog` fix on key / headers

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | --
| License       | MIT

* Looks like the headers is now `xi-api-key`
* Small fix on the key used to enable `STT`


src/platform/src/Bridge/ElevenLabs/CHANGELOG.md
```diff
@@ -5,3 +5,9 @@ CHANGELOG
 ---
 
  * Add the bridge
+
+0.3
```
OskarStark: only features deserve an entry here

chr-hertel:
Thank you @Guikingone.

PR #1412, opened 2026-01-16
[Mate] Add mcp:tools:call command for tool execution via JSON input

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | N/A
| License       | MIT

## Summary

This PR adds a new `mcp:tools:call` command that enables executing MCP tools directly from the command line with JSON input parameters. This complements the existing `mcp:tools:list` and `mcp:tools:inspect` commands, providing a complete CLI workflow for tool discovery, inspection, and execution.

## Features

- Execute MCP tools with JSON parameter input
- Automatic JSON schema validation against tool input schemas
- Two output formats:
  - **Pretty format** (default): Human-readable output with clear headers, descriptions, and formatted results
  - **JSON format**: Machine-parseable output for scripting and automation
- Comprehensive error handling for invalid tools, malformed JSON, and schema validation failures
- Clear execution headers showing tool name, description, and extension

### Execute tool with parameters
```bash
$ vendor/bin/mate mcp:tools:call monolog-search '{"term": "about"}'

Executing Tool: monolog-search
==============================

 Search log entries by text term with optional level, channel, environment, and date filters
 Extension: symfony/ai-monolog-mate-extension

Result
------

 {"datetime":"2026-01-16T18:28:19+00:00","channel":"console","level":"CRITICAL","message":"Error thrown while running command \"about --profiler\". Message: \"The \"--profiler\" option does not exist.\"","context":{...},"source_file":"dev.log","line_number":21}
 {"datetime":"2026-01-16T18:28:19+00:00","channel":"console","level":"DEBUG","message":"Command \"about --profiler\" exited with code \"1\"","context":{...},"source_file":"dev.log","line_number":22}
```

### Execute tool with empty parameters

```bash
$ vendor/bin/mate mcp:tools:call php-version "{}"

Executing Tool: php-version
===========================

 Get the version of PHP
 Extension: symfony/ai-mate

Result
------

 8.4.15

JSON output format

$ vendor/bin/mate mcp:tools:call php-version '{}' --format=json
"8.4.15"
```

## Implementation Details

- Reuses the same capability collection system as mcp:tools:list and mcp:tools:inspect
- Properly handles schema default values for optional parameters
- Resolves tool handlers from the Symfony DI container
- Supports both ClassName::method and ClassName (for __invoke) handler formats
- Includes comprehensive test coverage (5 test cases)

### DIFF
--- a/src/mate/src/Command/ToolsCallCommand.php
+++ b/src/mate/src/Command/ToolsCallCommand.php
@@ -1,0 +1,224 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Mate\Command;
+
+use Mcp\Capability\Discovery\Discoverer;
+use Mcp\Capability\Registry;
+use Mcp\Capability\Registry\ReferenceHandler;
+use Mcp\Exception\ToolNotFoundException;
+use Mcp\Schema\Request\CallToolRequest;
+use Psr\Log\LoggerInterface;
+use Symfony\AI\Mate\Command\Session\CliSession;
+use Symfony\AI\Mate\Discovery\FilteredDiscoveryLoader;
+use Symfony\Component\Console\Attribute\AsCommand;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Execute MCP tools via JSON input.
+ *
+ * @author Johannes Wachter <johannes@sulu.io>
+ */
+#[AsCommand('mcp:tools:call', 'Execute MCP tools via JSON input')]
+class ToolsCallCommand extends Command
+{
+    private Registry $registry;
+    private ReferenceHandler $referenceHandler;
+
+    public function __construct(
+        LoggerInterface $logger,
+        private ContainerInterface $container,
+    ) {
+        parent::__construct(self::getDefaultName());
+
+        $rootDir = $container->getParameter('mate.root_dir');
+        \assert(\is_string($rootDir));
+
+        $extensions = $this->container->getParameter('mate.extensions') ?? [];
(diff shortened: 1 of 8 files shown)


demo/composer.json
```diff
@@ -48,6 +48,7 @@
         "twig/twig": "^3.22"
     },
     "require-dev": {
+        "mcp/sdk": "0.3.0 as 0.2.3",
```
wachterjohannes: this is currently needed to run the newest versions of mate (requires 0.3)

chr-hertel: if this is only for development, i would rather not add this

it get's split into https://github.com/symfony/ai-demo as well - which currently only runs on ^0.2 versions of mate - so that issue is not relevant there i guess

demo/config/reference.php
chr-hertel, line 1: unrelated - please revert here

src/mate/src/Command/ToolsCallCommand.php
```diff
@@ -0,0 +1,357 @@
...
+    private function renderPretty(mixed $result, SymfonyStyle $io): void
+    {
+        if (\is_array($result)) {
+            if ($this->isAssociativeArray($result)) {
+                $io->definitionList(...array_map(fn ($key, $value) => [$key => $this->formatValue($value)], array_keys($result), $result));
+            } else {
+                foreach ($result as $item) {
+                    $io->text($this->formatValue($item));
+                }
+            }
```
chr-hertel: that custom `isAssociativeArray` is similar but inverted to [array_is_list](https://php.net/array_is_list)?

```suggestion
            if (array_is_list($result)) {
                foreach ($result as $item) {
                    $io->text($this->formatValue($item));
                }
            } else {
                $io->definitionList(...array_map(fn ($key, $value) => [$key => $this->formatValue($value)], array_keys($result), $result));
            }
```

chr-hertel:
Not sure about this - from a theory point of view. You're skipping almost the entire SDK infrastructure and features for tools - which raises a couple of issues for tools that:

* rely on `RequestContext` as argument
* interact with the client-side for sampling or elicitations

I would recommend to go through the entire stack there.

demo/composer.json
wachterjohannes: yeah it only happens when we run link - so i will remove it

wachterjohannes:
@chr-hertel tried it a different way by calling the ReferenceHandler directly - what do you think?

chr-hertel:
Thank you @wachterjohannes.

wachterjohannes:
> Potential follow-up could be to pull `Server` one level higher from `ServeCommand` into `App` to reuse it for `ToolsCallCommand` - but hard compromise since it's "only" debugging, right?

@chr-hertel yeah that sounds like a good idea. will tackle that in a follow up

PR #1416, opened 2026-01-18
[Agent] ToolCallArgumentResolver use and implement interface

chriskapp:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Docs?         | no
| Issues        | no
| License       | MIT

Hi, I'am currently trying to build a toolbox where the tools are defined in a database table and not as dedicated PHP class with attributes. Therefor I need to provide a custom `ToolFactory` and `ToolCallArgumentResolver`. Currently the `ToolCallArgumentResolver` has no interface but there is an interface `ToolCallArgumentResolverInterface` which is not used, this PR implements this interface to the `ToolCallArgumentResolver` so that it is possible to provide a custom implementation to the `Toolbox`. Let me know if I'am missing anything here, and thanks for this project!

### DIFF
--- a/src/agent/src/Toolbox/ToolCallArgumentResolver.php
+++ b/src/agent/src/Toolbox/ToolCallArgumentResolver.php
@@ -25,7 +25,7 @@ use Symfony\Component\TypeInfo\TypeResolver\TypeResolver;
 /**
  * @author Valtteri R <valtzu@gmail.com>
  */
-final class ToolCallArgumentResolver
+final class ToolCallArgumentResolver implements ToolCallArgumentResolverInterface
 {
     private readonly TypeResolver $typeResolver;
 
--- a/src/agent/src/Toolbox/ToolCallArgumentResolverInterface.php
+++ b/src/agent/src/Toolbox/ToolCallArgumentResolverInterface.php
@@ -11,6 +11,7 @@
 
 namespace Symfony\AI\Agent\Toolbox;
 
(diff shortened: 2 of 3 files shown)


chr-hertel (approved):
Good catch, thanks!

Was wondering about that use-case already - do you need to bend a lot?
Can you use the built-in `Toolbox` and how do you populate the `ExecutionReference`?

chr-hertel:
Thank you @chriskapp.

chriskapp:
@chr-hertel nice, from the architecture it felt intuitive, I could also use the existing `Toolbox` and I have created a single `DatabaseTool` which I pass as tool to the standard `Toolbox` and then in the factory I go through each row and return a tool with always the same `new ExecutionReference(DatabaseTool::class, 'invoke')`, now with the custom `DatabaseToolCallArgumentResolver` I can simply forward the `ToolCall $toolCall` to my `invoke` method and then I can use the name and arguments to get the tool from the database and invoke it. At least this is my plan :) I'am currently trying to migrate from a different mcp library, but it looks like this should work.

I am not sure how many users have also such a use case, but we could rename the `ToolCallArgumentResolver` to `ReflectionToolCallArgumentResolver` and add a second i.e. `SimpleToolCallArgumentResolver` (there is probably a better name) which only passes `$metadata` and `$toolCall` as arguments, maybe this could be useful for other users.

PR #1422, opened 2026-01-20
[Platform][OpenRouter] Add the new audio output capabilities

lochmueller:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | 
| Issues        | 
| License       | MIT

- Fix the API based catalog to avoid the exception for the output capability
- Add the new models to the static catalog
- Add an example for speech to text

### DIFF
--- a/examples/openrouter/speech-to-text.php
+++ b/examples/openrouter/speech-to-text.php
@@ -1,0 +1,28 @@
+<?php
+
+/*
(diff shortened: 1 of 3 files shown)


OskarStark:
Thank you Tim, I tested the example

PR #1429, opened 2026-01-23
[Agent] Support limiting iteration loop when tool calling

### LINKED ISSUE #1376: [Agent] Support limiting iteration loop when tool calling
LLMs can spiral into loops of repeated tool calls, it would be nice to have a fixed set of *tries* that we pass to the agent, and if it can't end before that we throw an exception

lochmueller:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | Fix #1376
| License       | MIT

Count the iterations in the AgentProcessor to limit the tool calls in the iteration. Default value is 50. Add information to CHANGELOG.md and UPGRADE.md

### DIFF
--- a/src/agent/CHANGELOG.md
+++ b/src/agent/CHANGELOG.md
@@ -7,6 +7,7 @@ CHANGELOG
  * [BC BREAK] Drop toolboxes `StreamResult` in favor of `StreamListener` on top of Platform's `StreamResult`
  * [BC BREAK] Rename `SourceMap` to `SourceCollection`, its methods from `getSources()` to `all()` and `addSource()` to `add()`
  * [BC BREAK] Third Argument of `ToolResult::__construct()` now expects `SourceCollection` instead of `array<int, Source>`
+ * Add `maxToolCalls` parameter to `AgentProcessor` to limit tool calling iterations and prevent infinite loops
 
 0.2
 ---
--- a/src/agent/src/Toolbox/AgentProcessor.php
+++ b/src/agent/src/Toolbox/AgentProcessor.php
@@ -13,6 +13,7 @@ namespace Symfony\AI\Agent\Toolbox;
 
 use Symfony\AI\Agent\AgentAwareInterface;
 use Symfony\AI\Agent\AgentAwareTrait;
+use Symfony\AI\Agent\Exception\MaxIterationsExceededException;
 use Symfony\AI\Agent\Input;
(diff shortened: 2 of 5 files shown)


src/agent/CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ CHANGELOG
  * [BC BREAK] Drop toolboxes `StreamResult` in favor of `StreamListener` on top of Platform's `StreamResult`
  * [BC BREAK] Rename `SourceMap` to `SourceCollection`, its methods from `getSources()` to `all()` and `addSource()` to `add()`
  * [BC BREAK] Third Argument of `ToolResult::__construct()` now expects `SourceCollection` instead of `array<int, Source>`
+ * Add `maxIterations` parameter to `AgentProcessor` to limit tool calling iterations and prevent infinite loops
+ * [BC BREAK] `AgentProcessor` now limits tool calling iterations to 50 by default
```
OskarStark: I would not add this as a BC break, as an inifinite loop is also not helpful and breaking the application

lochmueller: Set the default to \PHP_INT_MAX, and drop the BC information and the UPGRADE instructions as well.

src/agent/src/Toolbox/AgentProcessor.php
```diff
@@ -53,6 +54,7 @@ public function __construct(
         private readonly ?EventDispatcherInterface $eventDispatcher = null,
         private readonly bool $keepToolMessages = false,
         private readonly bool $includeSources = false,
+        private readonly int $maxIterations = \PHP_INT_MAX,
```
chr-hertel: i'd prefer something more explicit, like `maxToolCalls` - "iterations" doesn't state the subject of _what_ - would also remove that `PHP_INT_MAX` - looks like a explicit decision, but there's no reason from tech point of view for that, is there?
```suggestion
        private readonly ?int $maxToolCalls = null,
```

chr-hertel:
potential follow up to overwrite with option on `$agent->call($input, ['max_tool_calls' => 123])`

src/agent/src/Toolbox/AgentProcessor.php
lochmueller: done.

chr-hertel:
Thank you @lochmueller.

PR #1446, opened 2026-01-23
[Platform] Add object support in `AssistantMessage` content

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | N/A
| License       | MIT

### Summary

After the [`ContentInterface` rework](https://github.com/symfony/ai/commit/c531f337a3bb6a2579b2c443469ccfbf0dc9e36a) of `AssistantMessage`, the original "accept objects directly" idea no longer fit the model — every part of an assistant message is now a typed content block. This rewrite keeps the spirit of the original PR (let applications hold a typed payload in their message bag without metadata workarounds) by introducing it as a first-class content block.

Adds `Message\Content\Json`, a `ContentInterface` that wraps a `JsonSerializable`, `Stringable`, or any object. `Message::ofAssistant()` now also accepts arbitrary objects and wraps them automatically. When the message is replayed to a provider, every bridge `AssistantMessageNormalizer` (Anthropic, Bedrock Nova, Gemini, OpenResponses, VertexAI, AiMlApi, and the base normalizer) serializes the wrapped value to a JSON string and emits it as text content — `Stringable`-only via `__toString()`, otherwise via `json_encode`.

```php
$recipe = $agent->call($messages, ['response_format' => Recipe::class])->getContent();
$messages->add(Message::ofAssistant($recipe)); // wrapped as Json content

// later, in the UI
foreach ($message->getContent() as $part) {
    if ($part instanceof Json && $part->getObject() instanceof Recipe) {
        return $part->getObject();
    }
}
```

The recipe demo (`demo/src/Recipe/Chat.php`) now uses this path instead of the `metadata['recipe']` workaround.

### Notes for reviewers
- Single squashed commit rebased onto current `main`.
- Tests added for `Json`, `Message::ofAssistant`, the base normalizer, and each updated bridge normalizer.
- No BC break: existing usages keep working; this is purely additive.


src/platform/src/Contract/Normalizer/Message/AssistantMessageNormalizer.php
```diff
@@ -46,8 +46,14 @@ public function normalize(mixed $data, ?string $format = null, array $context =
...
         ];
 
-        if (null !== $data->getContent()) {
-            $array['content'] = $data->getContent();
+        $content = $data->getContent();
+        if (null !== $content) {
+            $array['content'] = match (true) {
+                \is_string($content) => $content,
+                $content instanceof \Stringable => (string) $content,
+                $content instanceof \JsonSerializable => json_encode($content, \JSON_THROW_ON_ERROR),
```
wachterjohannes: @chr-hertel should we directly do Toon encoding here?

chr-hertel: follow up i'd say

wachterjohannes: talked on slack: we will postpone toon integration here for now

wachterjohannes:
@chr-hertel now its ready :)

src/platform/src/Contract/Normalizer/Message/AssistantMessageNormalizer.php
```diff
@@ -46,8 +46,14 @@ public function normalize(mixed $data, ?string $format = null, array $context =
...
+                default => json_encode($this->normalizer->normalize($content, $format, $context), \JSON_THROW_ON_ERROR),
+            };
```
chr-hertel: i wonder if this could be just like this:
```suggestion
            $array['content'] = json_encode($this->normalizer->normalize($content, $format, $context), \JSON_THROW_ON_ERROR);
```

at least `JsonSerializable` is supported by normalizer

wachterjohannes: yeah should not be an issue - the only thing which changes is `$string = 'test message";` will be serialized to `$array['content'] = '"test message"';`

but in this case i think it should not be an issue - will update that

wachterjohannes:
@chr-hertel fixed your review

src/platform/src/Contract/Normalizer/Message/AssistantMessageNormalizer.php
```diff
@@ -46,8 +46,9 @@ public function normalize(mixed $data, ?string $format = null, array $context =
             'role' => $data->getRole()->value,
         ];
 
-        if (null !== $data->getContent()) {
-            $array['content'] = $data->getContent();
+        $content = $data->getContent();
+        if (null !== $content) {
+            $array['content'] = json_encode($this->normalizer->normalize($content, $format, $context), \JSON_THROW_ON_ERROR);
```
chr-hertel: Let's use the `SerializerAwareTrait` instead and directly use that instead of plain `json_encode`

chr-hertel (changes requested):
Small thing, other than that didn't give it a test run yet, but looks good - thanks! :)

wachterjohannes:
@chr-hertel fixed

OskarStark:
Copilot basically says "No" 😄

wachterjohannes:
@OskarStark @chr-hertel to keep BC we have to adapt also the bridges - what do you think?

wachterjohannes:
@chr-hertel I've reworked this following your feedback. After the ContentInterface rework of AssistantMessage, the original "accept objects directly" approach no longer fit, so this is now introduced as a first-class typed content block instead — keeping the original intent (let apps carry a typed payload in their message bag without metadata workarounds).

PR #1454, opened 2026-01-25
[Store][MariaDB] Fix for invalid parameter number

michielvaneerd:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | No issue
| License       | MIT

When storing the documents in the MariaDB Store Bridge, the 'metadata' and 'vector' parameter names are bound 2 times, which isn't allowed and results in a 'SQLSTATE[HY093]: Invalid parameter number on PDO Statement' error. This fixes it.

### DIFF
--- a/src/store/src/Bridge/MariaDb/Store.php
+++ b/src/store/src/Bridge/MariaDb/Store.php
@@ -115,7 +115,7 @@ final class Store implements ManagedStoreInterface, StoreInterface
                 <<<'SQL'
                     INSERT INTO %1$s (id, metadata, `%2$s`)
                     VALUES (:id, :metadata, VEC_FromText(:vector))
-                    ON DUPLICATE KEY UPDATE metadata = :metadata, `%2$s` = VEC_FromText(:vector)
+                    ON DUPLICATE KEY UPDATE metadata = :metadata2, `%2$s` = VEC_FromText(:vector2)
                     SQL,
                 $this->tableName,
                 $this->vectorFieldName,
@@ -126,6 +126,8 @@ final class Store implements ManagedStoreInterface, StoreInterface
             $statement->bindValue(':id', $document->getId());
             $statement->bindValue(':metadata', json_encode($document->getMetadata()->getArrayCopy()));
             $statement->bindValue(':vector', json_encode($document->getVector()->getData()));
+            $statement->bindValue(':metadata2', json_encode($document->getMetadata()->getArrayCopy()));
+            $statement->bindValue(':vector2', json_encode($document->getVector()->getData()));
 
             $statement->execute();
         }


valtzu:
1. Are you using emulated prepared statements or not? (`PDO::ATTR_EMULATE_PREPARES`/`PDO::MYSQL_ATTR_DIRECT_QUERY`)
2. Which MariaDB version you're using?

This store is covered by integration tests so it's definitely working with some version & configuration.

For the fix, instead of double-binding and double-encoding (and decoding on DB side) it could also be fixed within SQL which should have better performance:

```sql
INSERT INTO %1$s (id, metadata, `%2$s`)
VALUES (:id, @metadata := :metadata, @vector := VEC_FromText(:vector))
ON DUPLICATE KEY UPDATE metadata = @metadata, `%2$s` = @vector
```

michielvaneerd:
I am using PHP8.3 and MariaDB 11.8.5.

I think in PHP7 this wasn't allowed as well, but it didn't raise a fatal error, but since PHP8 it does.

And this is the code that raised the error:

```php
$docs = [new TextDocument(Uuid::v4(), 'Some content', new Metadata(['chunk_num' => 1]))];
$platform = PlatformFactory::create(config('myapp.mistral_api_key'));
$store = Store::fromPdo(DB::connection()->getPdo(), 'rag_chunks');
$store->setup(['dimensions' => 1024]);
$vectorizer = new Vectorizer($platform, 'mistral-embed', logger());
$indexer = new Indexer(new InMemoryLoader($docs), $vectorizer, $store, logger: logger());
$indexer->index($docs); // SQLSTATE[HY093] error
```

valtzu:
I did some testing and indeed the issue only occurs when **not** using emulated prepares. With default config there's no issue.

```php
$pdo = new PDO('mysql:host=db', 'root');
$pdo->setAttribute(PDO::ATTR_EMULATE_PREPARES, false); // comment out this line or set to true and the error is gone
$stmt = $pdo->prepare("SELECT :a, :a");
$stmt->execute(['a' => 1]); // throws SQLSTATE[HY093]: Invalid parameter number
```

chr-hertel:
Thanks for reporting & fixing @michielvaneerd - and double-checking @valtzu!

Since this fix is neutral for setups with `$pdo->setAttribute(PDO::ATTR_EMULATE_PREPARES, true);` we can safely merge it, right?
At least I think there's a reason for having that as a default setting

valtzu:
I'd say it's safe to merge, though as mentioned, will probably have some performance impact. If you are indexing millions of documents the impact may be noticeable, otherwise probably insignificant.

Let's merge :+1:

OskarStark (approved):

chr-hertel:
Thank you @michielvaneerd.

PR #1456, opened 2026-01-25
[AI Bundle] Remove cache result normalizer if package is not available

### LINKED ISSUE #1453: missing dependency symfony/ai-cache-platform
I installed ^0.3 and got:

```
In DefinitionErrorExceptionPass.php line 48:
                                                                                                                                  
  [Symfony\Component\DependencyInjection\Exception\RuntimeException]                                                              
  Invalid service "ai.platform.cache.result_normalizer": class "Symfony\AI\Platform\Bridge\Cache\ResultNormalizer" does not exis  
  t.                                                                                                                              
                                                                                                                                  

```

Fixed by installing 

```bash
composer require symfony/ai-cache-platform
```

So I imagine that should be a dependency somewhere.  Happy to add it and submit a PR, but where?

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #1453
| License       | MIT

We need to remove it, when package is not available.

### DIFF
--- a/src/ai-bundle/src/AiBundle.php
+++ b/src/ai-bundle/src/AiBundle.php
@@ -538,6 +538,9 @@ final class AiBundle extends AbstractBundle
 
             return;
         }
+        if (!ContainerBuilder::willBeAvailable('symfony/ai-cache-platform', CachePlatform::class, ['symfony/ai-bundle'])) {
+            $container->removeDefinition('ai.platform.cache.result_normalizer');
+        }
 
         if ('cartesia' === $type) {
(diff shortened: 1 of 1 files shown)


Guikingone:
Hi @chr-hertel 👋🏻 

I would have inverted the condition by registering the normalizer in the condition related to the `cache` platform (this remove the lines from `services.php`) in `processPlatform...` method 🤔

chr-hertel:
yea this is not super intuitive, but mostly how we do it right now - for now it's more consistent, but will look at this later again to optimize a bit how we handle that bridge specific stuff - similar to #1448

PR #1459, opened 2026-01-26
 Condition for message stores packages moved into the loop

### LINKED ISSUE #1458: Unable to upgrade to 0.3 due to required message stores
My current configuration describes only the use of the session `message_store`

```yaml
ai:
    agent:
        default:
            platform: ...
            model: ...
            prompt:
              text: ...
              include_tools: true
            tools:
              - ...
    message_store:
      session:
        default:
          identifier: 'ai.message_store.session'
    chat:
      default:
        agent: 'ai.agent.default'
        message_store: 'ai.message_store.session.default'
```

When upgrading to 0.3 I am required to install all available types of `message_store` because every item in the config is processed by `processMessageStoreConfig`

Image

Every type that is not installed yields a `throw new RuntimeException('... message store configuration requires....` Specifically Mongo-DB is a problem because it requires `ext-mongodb`.

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #1458 
| License       | MIT

### DIFF
--- a/src/ai-bundle/src/AiBundle.php
+++ b/src/ai-bundle/src/AiBundle.php
@@ -1944,11 +1944,11 @@ final class AiBundle extends AbstractBundle
     private function processMessageStoreConfig(string $type, array $messageStores, ContainerBuilder $container): void
     {
         if ('cache' === $type) {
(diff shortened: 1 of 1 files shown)


chr-hertel:
Thank you @Guikingone.

PR #1464, opened 2026-01-27
[Store] Make `Indexer` stateless by moving source to `index()` parameter

### LINKED ISSUE #875: [Store] Indexer is nice, but hard to use directly

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| Docs?         | no
| License       | MIT

## Summary

- Make `Indexer` stateless by removing `$sources` property and `$source` constructor parameter
- Add `$source` parameter to `index()` method instead of storing it in the constructor
- Remove `withSource()` method from `IndexerInterface`
- Add `ConfiguredIndexer` decorator for bundle configuration backward compatibility

## Motivation

The `Indexer` was a stateful service due to the `$sources` member being stored in the constructor. Stateful services are generally a bad practice in Symfony applications. The source is only used as input for the `LoaderInterface`, which already accepts it as a parameter.

## Changes

**Store Component:**
- `IndexerInterface`: Removed `withSource()`, added `$source` parameter to `index()`
- `Indexer`: Removed `$sources` property and constructor parameter, now stateless
- `ConfiguredIndexer`: New decorator that holds default source for bundle config
- `IndexCommand`: Passes source directly to `index()` instead of `withSource()`

**AI Bundle:**
- When `source` is configured in YAML, wraps indexer with `ConfiguredIndexer`
- Inner indexer registered as `ai.indexer.<name>.inner` when source is present

## Test plan

- [x] All existing `IndexerTest` tests updated and passing
- [x] New `ConfiguredIndexerTest` with tests for the decorator
- [x] All `AiBundleTest` indexer tests updated and passing
- [x] PHP-CS-Fixer applied

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,3 +1,39 @@
+UPGRADE FROM 0.3 to 0.4
+=======================
+
+AI Bundle
+---------
+
+ * An indexer configured with a `source`, now wraps the indexer with a `ConfiguredIndexer` decorator. This is
+   transparent - the configured source is still used by default, but can be overridden by passing a source to `index()`.
+
+Store
+-----
+
+ * The `Indexer` class is now stateless. The `$source` constructor parameter and `withSource()` method have been removed.
+   Source is now passed directly to the `index()` method:
+
+   ```diff
+   -$indexer = new Indexer($loader, $vectorizer, $store, '/path/to/source');
+   -$indexer->index();
+   +$indexer = new Indexer($loader, $vectorizer, $store);
+   +$indexer->index('/path/to/source');
+   ```
+
+ * The `IndexerInterface::withSource()` method has been removed. Use the `$source` parameter of `index()` instead:
+
+   ```diff
+   -$indexer->withSource('/new/source')->index();
+   +$indexer->index('/new/source');
+   ```
+
+ * The `Indexer` constructor parameter order has changed. Filters and transformers have shifted positions:
+
+   ```diff
+   -new Indexer($loader, $vectorizer, $store, $source, $filters, $transformers, $logger);
+   +new Indexer($loader, $vectorizer, $store, $filters, $transformers, $logger);
+   ```
+
 UPGRADE FROM 0.2 to 0.3
 =======================
 
--- a/src/store/CHANGELOG.md
+++ b/src/store/CHANGELOG.md
@@ -5,6 +5,8 @@ CHANGELOG
 ---
 
  * Add `StoreInterface::remove()` method
+ * [BC BREAK] Make `Indexer` stateless - removed `$source` constructor parameter and `withSource()` method
+ * Add `ConfiguredIndexer` decorator for pre-configuring default sources
 
 0.3
 ---
--- a/src/store/src/ConfiguredIndexer.php
+++ b/src/store/src/ConfiguredIndexer.php
@@ -1,0 +1,37 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Store;
+
+/**
+ * Decorator that wraps an IndexerInterface with a pre-configured default source.
+ *
+ * This is useful for bundle configuration where the source is defined in YAML/PHP config
+ * but can still be overridden at runtime by passing a source to index().
+ *
+ * @author Christopher Hertel <mail@christopher-hertel.de>
+ */
+final class ConfiguredIndexer implements IndexerInterface
+{
+    /**
+     * @param string|array<string>|null $defaultSource Default source to use when none is provided to index()
+     */
+    public function __construct(
+        private IndexerInterface $indexer,
+        private string|array|null $defaultSource = null,
+    ) {
+    }
+
+    public function index(string|array|null $source = null, array $options = []): void
+    {
+        $this->indexer->index($source ?? $this->defaultSource, $options);
+    }
+}
--- a/src/store/tests/ConfiguredIndexerTest.php
+++ b/src/store/tests/ConfiguredIndexerTest.php
@@ -1,0 +1,123 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Store\Tests;
+
+use PHPUnit\Framework\TestCase;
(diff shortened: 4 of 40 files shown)


chr-hertel:
Hey @lyrixx, this is closer to your PRs again and also less intrusive than my previous ones.
Do you see your concerns addressed?

lyrixx:
Hello,

I'm gonna test your PR this morning and I'll report feedback. Thanks !

OskarStark:
Needs a rebase

src/store/src/ConfiguredIndexer.php
```diff
@@ -0,0 +1,37 @@
...
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Store;
+
+/**
+ * Decorator that wraps an IndexerInterface with a pre-configured default source.
```
OskarStark, line 15: Smart!

src/store/tests/ConfiguredIndexerTest.php
```diff
@@ -0,0 +1,123 @@
...
+use Symfony\AI\Store\Document\TextDocument;
+use Symfony\AI\Store\Document\Vectorizer;
+use Symfony\AI\Store\Indexer;
+use Symfony\AI\Store\Tests\Double\PlatformTestHandler;
+use Symfony\AI\Store\Tests\Double\TestStore;
+use Symfony\Component\Uid\Uuid;
+
+final class ConfiguredIndexerTest extends TestCase
+{
+    public function testIndexUsesDefaultSourceWhenNoneProvided()
```
OskarStark, line 29:
```suggestion
    public function testIndexerUsesDefaultSourceWhenNoneProvided()
```

Same for the others?

src/store/CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 CHANGELOG
 =========
 
+0.4
+---
+
+ * [BC BREAK] Make `Indexer` stateless - `$source` constructor parameter and `withSource()` method removed
```
OskarStark:
```suggestion
 * [BC BREAK] Make `Indexer` stateless - removed `$source` constructor parameter and `withSource()` method
```

UPGRADE.md
```diff
@@ -1,3 +1,37 @@
...
+   ```
+
+ * The `Indexer` constructor parameter order has changed. Filters and transformers have shifted positions:
+
+   ```diff
+   -new Indexer($loader, $vectorizer, $store, $source, $filters, $transformers, $logger);
+   +new Indexer($loader, $vectorizer, $store, $filters, $transformers, $logger);
+   ```
+
+ * For bundle users: If you configured a `source` in your indexer configuration, the bundle now wraps the indexer
```
OskarStark:
```suggestion
AI Bundle
---------

 * For bundle users: If you configured a `source` in your indexer configuration, the bundle now wraps the indexer
```

and move it above the `Store` headline in this file?

lyrixx:
There is still an issue here.

**TL;DR: I want to index a document**

---

My current code is:
```php
private function indexCrawlUrl(string $id, string $crawlId, string $url, string $body): void
{
    $document = new TextDocument(
        new Uuid($id),
        $body,
        new Metadata([
            'url' => $url,
            'crawl_id' => $crawlId,
        ]),
    );
    $this->indexer->index([$document]); // I need to pass an array, but ideally, It would be better to pass the document directly 
}
```

With the current architecture, it's not possible.


1. A loader is mandatory in the bundle configuration. I thought I could use an `InMemoryLoader`, but it has not be built for that
2. So I built a passthru loader : 
    ```php
    <?php
    
    namespace App\Loader;
    
    use Symfony\AI\Store\Document\EmbeddableDocumentInterface;
    use Symfony\AI\Store\Document\LoaderInterface;
    
    final class PassthruLoaderLoader implements LoaderInterface
    {
        /**
         * @param EmbeddableDocumentInterface[] $documents
         */
        // ⚠⚠⚠⚠⚠ I HAD TO REMOVE THE TYPE HINT on $documents ⚠⚠⚠⚠⚠
        public function load($documents = null, array $options = []): iterable
        {
            if (is_iterable($documents)) {
                yield from $documents;
            } else {
                yield $documents;
            }
        }
    }
    ```
    This code is a bit weird to me :/ 
3. Then, I got the following error
    ```
      "detail": "Symfony\\AI\\Store\\Indexer::loadSource(): Argument #1 ($source) must be of type ?string, Symfony\\AI\\Store\\Document\\TextDocument given, called in \/app\/server\/ai\/src\/store\/src\/Indexer.php on line 50",
    ```
    I fixed it by removing the hint here too

---

This PR is better than the previous version, but it does not fix my initial issue.

To recall, I need to index a **document** (`final class TextDocument implements EmbeddableDocumentInterface`).

chr-hertel:
okay, true, let's come back to the overloading idea for making the loading part optional then:
1. you can call `$indexer->index($sources)` with being `string|string[]` of references used by a loader
2. you can call `$indexer->index($documents)` with TextDocument|TextDocument[]` 

but making sure that we don't have to have an awkward loader
follow up PR to this one tho. sorry for the extra loops, but decoupling those two makes it easier i guess.

src/store/tests/ConfiguredIndexerTest.php
chr-hertel: refers to the method name `index` to me

OskarStark:
Lets merge it in this state, its a good progress in the right direction!

OskarStark:
It took time, but here we go, this is in now. Thank you very much Christopher.

lyrixx:
> 2\. you can call `$indexer->index($documents)` with TextDocument|TextDocument[]`
> but making sure that we don't have to have an awkward loader

Sorry, But I can't make it work! 

1. I'm forced to have a loader, right? https://github.com/symfony/ai/blob/8c35eb037b48634226f76aafb4749ad288011692/src/store/src/Indexer.php#L33
2. $document (without `s`) is passed to [`private function loadSource(?string $source): array`](https://github.com/symfony/ai/blob/8c35eb037b48634226f76aafb4749ad288011692/src/store/src/Indexer.php#L90C5-L90C56). `TextDocument` is not a `string`
3. InderxerInterace::index() does'n allow Object: https://github.com/symfony/ai/blob/8c35eb037b48634226f76aafb4749ad288011692/src/store/src/IndexerInterface.php#L24

lyrixx:
I made minor adjustment to this pr: https://github.com/symfony/ai/pull/1517

(but my main concerns are not fixed yet)

PR #1516, opened 2026-01-29
[Store] Make loader optional in Indexer when passing documents directly

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | -
| License       | MIT

## Summary

Allow callers to pass `TextDocument` or `TextDocument[]` directly to `Indexer::index()`, bypassing the loader. This is useful when documents are already available in memory and don't need to be loaded from an external source.

### Changes

- Reorder `Indexer` constructor: `(vectorizer, store, loader?, filters, transformers, logger)`
- Make `$loader` parameter optional (nullable)
- Accept `EmbeddableDocumentInterface|array<EmbeddableDocumentInterface>` in `index()` method
- Throw `RuntimeException` when using source strings without a configured loader
- Add example demonstrating direct document indexing

### Usage

```php
// Without loader - passing documents directly
$indexer = new Indexer($vectorizer, $store);
$indexer->index($textDocument);
$indexer->index([$document1, $document2]);

// With loader (existing behavior, using named parameters)
$indexer = new Indexer(
    vectorizer: $vectorizer,
    store: $store,
    loader: $loader,
);
$indexer->index('path/to/file.txt');
```

## Test plan

- [x] All existing Indexer tests pass
- [x] New tests for direct document indexing
- [x] New tests for error handling when loader is missing
- [x] PHP syntax validation on all examples

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/src/store/CHANGELOG.md
+++ b/src/store/CHANGELOG.md
@@ -7,6 +7,8 @@ CHANGELOG
  * Add `StoreInterface::remove()` method
  * [BC BREAK] Make `Indexer` stateless - removed `$source` constructor parameter and `withSource()` method
  * Add `ConfiguredIndexer` decorator for pre-configuring default sources
+ * [BC BREAK] Reorder `Indexer` constructor parameters and make `$loader` optional
+ * Add support for passing documents directly to `Indexer::index()` without requiring a loader
 
 0.3
 ---
--- a/src/store/src/Indexer.php
+++ b/src/store/src/Indexer.php
@@ -18,6 +18,7 @@ use Symfony\AI\Store\Document\FilterInterface;
 use Symfony\AI\Store\Document\LoaderInterface;
 use Symfony\AI\Store\Document\TransformerInterface;
 use Symfony\AI\Store\Document\VectorizerInterface;
+use Symfony\AI\Store\Exception\RuntimeException;
 
 /**
  * @author Christopher Hertel <mail@christopher-hertel.de>
@@ -30,28 +31,21 @@ class Indexer implements IndexerInterface
      * @param TransformerInterface[] $transformers Transformers to mutate documents after filtering (chunking, cleaning, etc.)
      */
     public function __construct(
-        private LoaderInterface $loader,
         private VectorizerInterface $vectorizer,
         private StoreInterface $store,
+        private ?LoaderInterface $loader = null,
         private array $filters = [],
         private array $transformers = [],
         private LoggerInterface $logger = new NullLogger(),
     ) {
     }
 
-    public function index(string|array|null $source = null, array $options = []): void
+    public function index(string|array|EmbeddableDocumentInterface|null $source = null, array $options = []): void
     {
-        $sources = null === $source ? [null] : (array) $source;
-
-        $this->logger->debug('Starting document processing', ['sources' => $sources, 'options' => $options]);
-
-        $documents = [];
-        foreach ($sources as $singleSource) {
-            $documents = array_merge($documents, $this->loadSource($singleSource));
-        }
+        $documents = $this->resolveDocuments($source);
 
         if ([] === $documents) {
-            $this->logger->debug('No documents to process', ['sources' => $sources]);
+            $this->logger->debug('No documents to process');
 
             return;
         }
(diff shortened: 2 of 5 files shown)


src/store/src/Indexer.php
```diff
@@ -30,28 +31,21 @@ class Indexer implements IndexerInterface
...
         private StoreInterface $store,
+        private ?LoaderInterface $loader = null,
         private array $filters = [],
         private array $transformers = [],
         private LoggerInterface $logger = new NullLogger(),
     ) {
     }
 
-    public function index(string|array|null $source = null, array $options = []): void
+    public function index(string|array|EmbeddableDocumentInterface|null $source = null, array $options = []): void
```
chr-hertel, line 43: let's rename to `$input`

```suggestion
    public function index(string|array|EmbeddableDocumentInterface|null $input = null, array $options = []): void
```

src/store/src/Indexer.php
```diff
@@ -30,28 +31,21 @@ class Indexer implements IndexerInterface
      * @param TransformerInterface[] $transformers Transformers to mutate documents after filtering (chunking, cleaning, etc.)
      */
     public function __construct(
-        private LoaderInterface $loader,
         private VectorizerInterface $vectorizer,
         private StoreInterface $store,
+        private ?LoaderInterface $loader = null,
```
OskarStark, line 36: needs to be added to UPGRADE.md

chr-hertel:
@lyrixx just dropped my previous comment to claude as prompt, but haven't really reviewed this here yet - but this would try to address the issue you're facing

src/store/src/Indexer.php
```diff
@@ -84,11 +78,99 @@ public function index(string|array|null $source = null, array $options = []): vo
...
+     *
+     * @param array<string|null> $sources
+     *
+     * @return EmbeddableDocumentInterface[]
+     */
+    private function loadFromSources(array $sources): array
+    {
+        if (null === $this->loader) {
+            throw new RuntimeException('A loader is required when indexing from sources. Either provide a loader in the constructor or pass documents directly.');
+        }
```
chr-hertel, line 155: should be shifted into the `loadSource(...)` instead of the `assert(...)`

src/store/CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ CHANGELOG
  * Add `StoreInterface::remove()` method
  * [BC BREAK] Make `Indexer` stateless - removed `$source` constructor parameter and `withSource()` method
  * Add `ConfiguredIndexer` decorator for pre-configuring default sources
+ * [BC BREAK] Reorder `Indexer` constructor parameters and make `$loader` optional
```
lyrixx, line 10: Both BC break could be squash together, couldn't it?

src/store/src/Indexer.php
```diff
@@ -84,11 +78,99 @@ public function index(string|array|null $source = null, array $options = []): vo
...
     }
 
+    /**
+     * Resolve the source parameter into an array of documents.
+     *
+     * @param string|array<string>|EmbeddableDocumentInterface|array<EmbeddableDocumentInterface>|null $source
+     *
+     * @return EmbeddableDocumentInterface[]
+     */
+    private function resolveDocuments(string|array|EmbeddableDocumentInterface|null $source): array
```
lyrixx, line 88: It yould be better to use iterator everywhere

src/store/src/Indexer.php
```diff
@@ -84,11 +78,99 @@ public function index(string|array|null $source = null, array $options = []): vo
...
+
+                return [];
+            }
+
+            $firstElement = reset($source);
+            if ($firstElement instanceof EmbeddableDocumentInterface) {
+                $this->logger->debug('Processing document array directly', ['count' => \count($source)]);
+
+                return $this->filterDocuments($source);
+            }
```
lyrixx, line 111: I don't like this kind of code. 

Passing many kind of type to a method is a code smell. Having two methods in the Indexer would be much cleaner. I know you already register `index()`+`loadThenIndex()`, but I think we could find something else, better than this kind of magic.

src/store/src/Indexer.php
```diff
@@ -84,11 +78,99 @@ public function index(string|array|null $source = null, array $options = []): vo
...
+
+            // Array contains strings - will be processed by the loader below
+            // We know $firstElement is string here since it's not EmbeddableDocumentInterface
+            \assert(\is_string($firstElement));
```
lyrixx, line 115: We tend to not use `assert()`  function in Symfony

src/store/src/Indexer.php
chr-hertel: yup, need that rebase after merging #1517

src/store/src/Indexer.php
chr-hertel: yea, i feel you. just "input" also feels like too wiggly - so let's split the use cases.

so let's figure out the naming:
a) `$indexer->fromSource($sources, $options)` & `$indexer->fromDocuments($documents, $options)`
b) `$indexer->loadThenIndex($sources, $options)` & `$indexer->index($documents, $options)`
c) `$indexer->loadAndIndex($sources, $options)` & `$indexer->index($documents, $options)`
d) `$indexer->indexSources($sources, $options)` & `$indexer->indexDocuments($documents, $options)`

... and more? :thinking:

lyrixx: Hmm, actually, I like A & D (more, than B & C)! Good proposal.

So I would vote for D>A>B>C

chr-hertel:
Closing in favor of #1526

PR #1517, opened 2026-01-29
[Store] Allow iterable of source in Indexer::index() + doesn't break lazyness

lyrixx:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        |
| License       | MIT

### DIFF
--- a/src/store/src/ConfiguredIndexer.php
+++ b/src/store/src/ConfiguredIndexer.php
@@ -30,7 +30,7 @@ final class ConfiguredIndexer implements IndexerInterface
     ) {
     }
 
-    public function index(string|array|null $source = null, array $options = []): void
+    public function index(string|iterable|null $source = null, array $options = []): void
     {
         $this->indexer->index($source ?? $this->defaultSource, $options);
     }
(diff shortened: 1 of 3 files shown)


chr-hertel:
yup, hunting down something fishy similar to this - kinda easy to mess that up while chaining all those services. no idea to test that properly tho

lyrixx:
It's already tested, right? I mean, I did broke the tests while coding that!

chr-hertel:
yea, i rather meant the concern that users could easily break that generator behavior that relies on every service involved to respect that.

if there's only one transformer that messes up, it's mostly broken for the entire indexing process and that's not easy to catch by design - or i fail to see how

lyrixx:
I'm not sure to get your point.

I edited my code to throw an exception in a Transformer, and I got this exception.

image


It's quite clean IMHO. We already have such iterator chaining / processing in Symfony or other lib and it's common IMHO

chr-hertel (approved):

chr-hertel:
Thank you @lyrixx.

PR #1523, opened 2026-01-29
[Mate] Refactor command registration to use service container

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | yes
| Issues        | N/A
| License       | MIT

## Summary

Refactored the Mate application to use dependency injection container with services for command registration instead of manual instantiation. This improves modularity and makes the application more maintainable.

### DIFF
--- a/src/mate/src/App.php
+++ b/src/mate/src/App.php
@@ -13,7 +13,6 @@ namespace Symfony\AI\Mate;
 
 use Mcp\Server\Transport\Stdio\RunnerControl;
 use Mcp\Server\Transport\Stdio\RunnerState;
-use Psr\Log\LoggerInterface;
 use Symfony\AI\Mate\Command\ClearCacheCommand;
 use Symfony\AI\Mate\Command\DebugCapabilitiesCommand;
 use Symfony\AI\Mate\Command\DebugExtensionsCommand;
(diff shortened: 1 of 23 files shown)


src/mate/src/default.config.php
```diff
@@ -49,5 +73,81 @@
...
             ->arg('$debugEnabled', '%mate.debug_enabled%')
             ->alias(Logger::class, LoggerInterface::class)
+
+        // Container service for commands that need it
+        ->alias(ContainerInterface::class, 'service_container')
+
+        // Register discovery services
+        ->set(Discoverer::class)
+            ->autowire()
+            ->public()
```
chr-hertel: `autowire()` is already default and looks like `public()` should be too - you wouldn't have to repeat yourself

src/mate/src/App.php
```diff
@@ -40,27 +29,20 @@ final class App
...
     public static function build(ContainerInterface $container): Application
     {
-        $logger = $container->get(LoggerInterface::class);
-        \assert($logger instanceof LoggerInterface);
+        $application = new Application(self::NAME, self::VERSION);
 
-        $rootDir = $container->getParameter('mate.root_dir');
-        \assert(\is_string($rootDir));
+        // Discover all tagged commands from the container
+        if (!method_exists($container, 'findTaggedServiceIds')) {
```
chr-hertel: why do we need `method_exists`? 😬

chr-hertel:
Thank you @wachterjohannes.

PR #1535, opened 2026-02-01
[Platform] Read discriminator mapping from serializer metadata

valtzu:
| Q             | A
| ------------- | ---
| Bug fix?      | yes (kinda)
| New feature?  | no
| Docs?         | no
| Issues        | Related to #559
| License       | MIT

The discriminator mapping introduced in #585 only supports mapping via attributes – I found it quite unexpected behavior. Let's use serializer metadata instead?

### DIFF
--- a/src/ai-bundle/config/services.php
+++ b/src/ai-bundle/config/services.php
@@ -156,6 +156,7 @@ return static function (ContainerConfigurator $container): void {
             ->args([
                 service('ai.platform.json_schema.description_parser'),
                 service('type_info.resolver')->nullOnInvalid(),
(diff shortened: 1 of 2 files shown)


OskarStark:
Sorry for asking, but is this already covered by a test? I mean, can I trust the CI here?  😄

valtzu:
> is this already covered by a test

Yes I would say so, e.g. https://github.com/symfony/ai/blob/main/src/platform/tests/Contract/JsonSchema/FactoryTest.php#L224-L276

chr-hertel:
Thank you @valtzu.

PR #1549, opened 2026-02-04
[Agent] Add context compression strategies

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | 
| License       | MIT

PR #1553, opened 2026-02-05
[Agent] improv: Better defaults, preserve tool msgs

### LINKED ISSUE #1402: [Agent] Keep tool messages by default in agent processor
Defaults matter and `keepToolMessages = false` in AgentProcessor right now, this causes the LLM to forget what tool calls it made before and their results during subsequent requests.

Keeping tool messages and results is important since it enables the LLM to be fully aware of the things it has tried and potentially failed at in later messages.

Of course the downside is increase in context size, but shrinking the message bag (context) should be done by a separate service.

aszenz:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Docs?         | yes 
| Issues        | Fix #1402  
| License       | MIT

Still not sure if we should change the default or remove this option altogether and move it to compression feature in #1549

### DIFF
--- a/docs/components/agent.rst
+++ b/docs/components/agent.rst
@@ -405,12 +405,12 @@ If you need to react more granularly to the lifecycle of individual tool calls, 
         // Let the client know, that the tool $event->toolCall->name failed with the exception: $event->exception
     });
 
-Keeping Tool Messages
-~~~~~~~~~~~~~~~~~~~~~
+Excluding Tool Messages from MessageBag
+---------------------------------------
 
-Sometimes you might wish to keep the tool messages (:class:`Symfony\\AI\\Platform\\Message\\AssistantMessage` containing the ``toolCalls`` and :class:`Symfony\\AI\\Platform\\Message\\ToolCallMessage`
-containing the result) in the context. Enable the ``keepToolMessages`` flag of the toolbox' :class:`Symfony\\AI\\Agent\\Toolbox\\AgentProcessor`
-to ensure those messages will be added to your :class:`Symfony\\AI\\Platform\\Message\\MessageBag`::
+Sometimes you might wish to exclude the tool messages (:class:`Symfony\\AI\\Platform\\Message\\AssistantMessage` containing the ``toolCalls`` and :class:`Symfony\\AI\\Platform\\Message\\ToolCallMessage`
+containing the result) in the context. Enable the ``excludeToolMessages`` flag of the toolbox' :class:`Symfony\\AI\\Agent\\Toolbox\\AgentProcessor`
+to ensure those messages will be removed from your :class:`Symfony\\AI\\Platform\\Message\\MessageBag`::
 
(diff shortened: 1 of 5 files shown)


chr-hertel:
Let's keep it here, but should we rename it to `excludeToolMessage: false`?
we're not hiding them, they're gone - and feels more consistent to `includeSources`

src/agent/src/Toolbox/AgentProcessor.php
```diff
@@ -50,7 +50,7 @@ public function __construct(
         private readonly ToolboxInterface $toolbox,
         private readonly ToolResultConverter $resultConverter = new ToolResultConverter(),
         private readonly ?EventDispatcherInterface $eventDispatcher = null,
-        private readonly bool $keepToolMessages = false,
+        private readonly bool $keepToolMessages = true,
```
chr-hertel: meant this as well:

```suggestion
        private readonly bool $excludeToolMessage = false,
```

needs a invert of the logic below

chr-hertel (approved):
Thanks @aszenz - quite a change when it comes to token consumption, but the right default - i hope :D

I think we should still add a note in the `UPGRADE.md` - good to be merged after that

docs/components/agent.rst
```diff
@@ -405,12 +405,12 @@ If you need to react more granularly to the lifecycle of individual tool calls,
         // Let the client know, that the tool $event->toolCall->name failed with the exception: $event->exception
     });
 
-Keeping Tool Messages
-~~~~~~~~~~~~~~~~~~~~~
+Excluding Tool Messages from :class:`Symfony\\AI\\Platform\\Message\\MessageBag`
+~~~~~~~~~~~~~~~~~~~~------------------------------------------------------------
```
OskarStark:
```suggestion
Excluding Tool Messages from MessageBag
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

src/agent/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ CHANGELOG
 
  * [BC BREAK] Rename `Symfony\AI\Agent\Toolbox\Tool\Agent` to `Symfony\AI\Agent\Toolbox\Tool\Subagent`
  * Add `MetaDataAwareTrait` to `MockResponse`, the metadata will also be set on the returned `TextResult` when calling the `toResult` function
+ * [BC BREAK] Refactor Agent Processor `keepToolMessages` to `excludeToolMessages` and change default behaviour to preserve tool messages.
```
```suggestion
 * [BC BREAK] Change AgentProcessor `keepToolMessages` to `excludeToolMessages` and default behaviour to preserve tool messages
```

chr-hertel:
Looks like the rebase brought in some unrelated commits => conflicts

chr-hertel:
Thank you @aszenz.

PR #1557, opened 2026-02-06
[Platform] Add support for DiscriminatorMap on single polymorphic interface parameters

### LINKED ISSUE #1159: [Platform] Support Discriminator Map when generation json schema for interfaces arguments
When declaring a tool which takes as argument an interface with a DiscriminatorMap, generated json schema only considers interface properties and doesn't use `oneOff` to describe all possible implementations.

Right now discriminator map is only supported within array of polymorphic types, see https://github.com/symfony/ai/pull/585

## Tool Code

```php
#[AsTool(
    name: 'navigator',
    description: <<<TEXT
            Navigate to various resources
        TEXT,
    method: 'navigate'
)]
final class ToolNavigator implements ToolInterface
{
    public function __construct(
        private UrlGeneratorInterface $urlGenerator,
    ) {}

    public function toolName(): string
    {
        return 'navigator';
    }

    public function navigate(Filterable $filter): string
    {
        return match(true) {
            $filter instanceof OrderFilter => $this->urlGenerator->generate(
                'order_searchresult',
                [...$filter->toUrlParameter()],
            ),
            $filter instanceof PurchaseContractFilter => $this->urlGenerator->generate(
                'contract_searchresult',
                [...$filter->toUrlParameter()],
            ),
        };
    }
}
```

## Interface

```php
<?php

#[DiscriminatorMap(
    typeProperty: 'type',
    mapping: [
        'order' => OrderFilter::class,
        'purchase_contract' => PurchaseContractFilter::class,
    ]
)]
interface Filterable
{
    public string $type {get;}
}
```

## Implementations

```php

<?php
final class OrderFilter implements Filterable
{
    public function __construct(
        #[With(const: 'order')]
        public string $type = 'order',
        public ?string $number = null,
        public ?string $userresponsible = null,
        public ?\DateTimeImmutable $departureDate = null,
    ) {}
}
```

```php

<?php
final class PurchaseContractFilter implements Filterable
{
    public function __construct(
        #[With(const: 'purchase_contract')]
        public string $type = 'purchase_contract',
        public ?string $contractNumber = null,
        public ?string $subsidiary = null,
    ) {}
}
```

**Note**: I tried using union types instead of interface for the tool, the json schema is correctly generated then but since all properties are optional in the implemenatation classes symfony serializer always deserializes tool call results into the first class `OrderFilter` by creating an empty object instead of actually matching properties.

## Code patch

```patch
From f6f37e0d8b51aea166b5ebe78aedf60cbac166c6 Mon Sep 17 00:00:00 2001
From: asrar <aszenz@gmail.com>
Date: Tue, 16 Dec 2025 21:15:13 +0100
Subject: [PATCH] Implement DiscriminatorMap handling for polymorphism

Added handling for polymorphic interfaces using DiscriminatorMap attribute in the JSON schema factory.
---
 src/platform/src/Contract/JsonSchema/Factory.php | 13 +++++++++++++
 1 file changed, 13 insertions(+)

diff --git a/src/platform/src/Contract/JsonSchema/Factory.php b/src/platform/src/Contract/JsonSchema/Factory.php
index 776537e4fe..dafd845986 100644
--- a/src/platform/src/Contract/JsonSchema/Factory.php
+++ b/src/platform/src/Contract/JsonSchema/Factory.php
@@ -229,6 +229,19 @@ private function getTypeSchema(Type $type): array
                 if (\in_array($className, ['DateTime', 'DateTimeImmutable', 'DateTimeInterface'], true)) {
                     return ['type' => 'string', 'format' => 'date-time'];
                 } else {
+                    // Check for the DiscriminatorMap attribute to handle polymorphic interfaces
+                    $discriminatorMapping = $this->findDiscriminatorMapping($className);
+                    if ($discriminatorMapping) {
+                        $discriminators = [];
+                        foreach ($discriminatorMapping as $_ => $discriminator) {
+                            $discriminators[] = $this->buildProperties($discriminator);
+                        }
+
+                        return [
+                            'type' => 'object',
+                            'oneOf' => $discriminators,
+                        ];
+                    }
                     // Recursively build the schema for an object type
                     return $this->buildProperties($className) ?? ['type' => 'object'];
                 }

```

aszenz:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | Fix #1159
| License       | MIT

This PR extends JSON schema generation to support `DiscriminatorMap` on single interface parameters, not just arrays. The implementation generates a `oneOf` schema with all discriminator implementations, allowing LLMs to properly understand and use polymorphic tool parameters.

See `examples/anthropic/toolcall-polymorphic-interface.php` for a working example.

### DIFF
--- a/examples/agent/toolcall-polymorphic-interface.php
+++ b/examples/agent/toolcall-polymorphic-interface.php
@@ -1,0 +1,85 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use Symfony\AI\Agent\Agent;
+use Symfony\AI\Agent\Toolbox\AgentProcessor;
+use Symfony\AI\Agent\Toolbox\Attribute\AsTool;
+use Symfony\AI\Agent\Toolbox\Toolbox;
+use Symfony\AI\Platform\Bridge\Anthropic\PlatformFactory as AnthropicPlatformFactory;
+use Symfony\AI\Platform\Bridge\Gemini\PlatformFactory as GeminiPlatformFactory;
+use Symfony\AI\Platform\Bridge\OpenAi\PlatformFactory as OpenAiPlatformFactory;
+use Symfony\AI\Platform\Message\Message;
+use Symfony\AI\Platform\Message\MessageBag;
+use Symfony\AI\Platform\Tests\Fixtures\StructuredOutput\PolymorphicType\Filterable;
+use Symfony\AI\Platform\Tests\Fixtures\StructuredOutput\PolymorphicType\OrderFilter;
+use Symfony\AI\Platform\Tests\Fixtures\StructuredOutput\PolymorphicType\PurchaseContractFilter;
+
+require_once dirname(__DIR__).'/bootstrap.php';
+
+$platforms = [
+    'claude-sonnet-4-5-20250929' => AnthropicPlatformFactory::create(env('ANTHROPIC_API_KEY'), httpClient: http_client()),
+    'gpt-5-mini' => OpenAiPlatformFactory::create(env('OPENAI_API_KEY'), httpClient: http_client()),
+    'gemini-2.5-flash' => GeminiPlatformFactory::create(env('GEMINI_API_KEY'), httpClient: http_client()),
+];
+
+foreach ($platforms as $model => $platform) {
+    $toolbox = new Toolbox(tools: [new SearchTool()]);
+    $processor = new AgentProcessor(toolbox: $toolbox);
+    $agent = new Agent($platform, $model, [$processor], [$processor]);
+
+    echo "\n=== Testing with model: $model ===\n\n";
+
+    $systemMsg = Message::forSystem('You are a helpful search assistant. Help users search for different resources.');
+    $userMsg = 'Search for order number ORD-12345 with user responsible John Doe';
+    $messages = new MessageBag(
+        $systemMsg,
+        Message::ofUser($userMsg)
+    );
+
+    $result = $agent->call($messages);
+
+    echo $userMsg." : \n".$result->getContent().\PHP_EOL.\PHP_EOL;
+
+    $userMsg = 'Search for purchase contract number PC-67890 with subsidiary Acme Corp';
+    $messages = new MessageBag(
+        $systemMsg,
+        Message::ofUser($userMsg),
+    );
+
+    $result = $agent->call($messages);
+
+    echo $userMsg." : \n".$result->getContent().\PHP_EOL;
+}
+
+// Tool that uses polymorphic interface as parameter
+#[AsTool('search', 'Search various resources based on filter criteria', method: 'search')]
+final class SearchTool
+{
+    /**
+     * @param Filterable $filter The filter to use for search (order or purchase_contract)
+     */
+    public function search(Filterable $filter): string
+    {
+        return match (true) {
+            $filter instanceof OrderFilter => sprintf(
+                'Searching for order: %s (User: %s) Found 1 result.',
+                $filter->number ?? 'N/A',
+                $filter->userResponsible ?? 'N/A'
+            ),
+            $filter instanceof PurchaseContractFilter => sprintf(
+                'Searching for purchase contract: %s (Subsidiary: %s) Found 1 result.',
+                $filter->contractNumber ?? 'N/A',
+                $filter->subsidiary ?? 'N/A'
+            ),
+            default => 'Unknown filter type',
+        };
+    }
+}
--- a/src/platform/CHANGELOG.md
+++ b/src/platform/CHANGELOG.md
@@ -10,6 +10,7 @@ CHANGELOG
  * [BC BREAK] Replace variadic constructor parameters with array parameters in `VectorResult`, `ToolCallResult`, `RerankingResult`, `ToolCallComplete`, and `ImageResult` (OpenAI DallE bridge)
  * [BC BREAK] Rename `#[With]` attribute to `#[Schema]` and `WithAttributeDescriber` to `SchemaAttributeDescriber`
  * Add `ref` property to `#[Schema]` attribute to allow providing schema as file
+ * Add support for `DiscriminatorMap` and `DateTimeImmutable` for structured output and tool calls
 
 0.7
 ---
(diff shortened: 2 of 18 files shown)


chr-hertel (changes requested):
Please just use the test classes in the examples, so we don't need to repeat those classes there - that's alright since we don't ship the examples standalone.

did you test with OpenAI or other providers? the change has impact to all platform bridges basically

aszenz:
> Please just use the test classes in the examples, so we don't need to repeat those classes there - that's alright since we don't ship the examples standalone.

Done
 
> did you test with OpenAI or other providers? the change has impact to all platform bridges basically

Revamped the example to test multiple providers and indeed it surfaced new issues:

1. oneOff was not working on anthropic, so switched to anyOf
2. gemini doesn't support const, so switched to enum in the example

During testing I found more issues:

### gemini-3-flash-preview failed to work with the error:

```
Error "400" - "INVALID_ARGUMENT": "Function call is missing a thought_signature in functionCall parts. This is required for tools to work correctly, and missing thought_signature may lead to degraded model performance. Additional data, function call `default_api:search` , position 2. Please refer to https://ai.google.dev/gemini-api/docs/thought-signatures for more details.".
```

### JsonSchema Factory marks optional class fields as required

This makes models like Gemini not work when user provides only one search term. 

The code here: 

https://github.com/symfony/ai/blob/main/src/platform/src/Contract/JsonSchema/Factory.php#L125-L127

Only checks for reflection parameter not property

examples/agent/toolcall-polymorphic-interface.php
```diff
@@ -0,0 +1,109 @@
...
+];
+
+foreach ($platforms as $model => $platform) {
+    $searchTool = new SearchTool();
+    $classMetadataFactory = new ClassMetadataFactory(new AttributeLoader());
+    $discriminator = new ClassDiscriminatorFromClassMetadata($classMetadataFactory);
+
+    $toolArgNormalizers = [
+        new DateTimeNormalizer(),
+        new ObjectNormalizer(
```
aszenz: should we include discriminator resolver in the default tool argument normalizer, would make sense since it's also included for structured outputs here:

https://github.com/symfony/ai/blob/2a2b6be79bf80b992e5b78a539b3f27e9d75e30c/src/platform/src/StructuredOutput/Serializer.php#L41-L48

src/platform/src/Contract/JsonSchema/Factory.php
```diff
@@ -233,6 +233,19 @@ private function getTypeSchema(Type $type): array
...
+                // Check for the DiscriminatorMap attribute to handle polymorphic interfaces
+                $discriminatorMapping = $this->classMetadataFactory->getMetadataFor($className)->getClassDiscriminatorMapping()?->getTypesMapping();
+                if ($discriminatorMapping) {
+                    $discriminators = [];
+                    foreach ($discriminatorMapping as $_ => $discriminator) {
+                        $discriminators[] = $this->buildProperties($discriminator);
+                    }
+
+                    return [
+                        'anyOf' => $discriminators,
```
aszenz: this should have been `oneOff` but that didn't work with anthropics and gemini models

aszenz: Execution of tool "search" failed with error: Type property "type" not found for the abstract object "Symfony\AI\Platform\Tests\Fixtures\StructuredOutput\PolymorphicType\Filterable".

src/platform/tests/Fixtures/StructuredOutput/PolymorphicType/OrderFilter.php
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+namespace Symfony\AI\Platform\Tests\Fixtures\StructuredOutput\PolymorphicType;
+
+use Symfony\AI\Platform\Contract\JsonSchema\Attribute\With;
+
+final class OrderFilter implements Filterable
+{
+    public function __construct(
+        #[With(enum: ['order'])]
```
aszenz: this should have simply been `const: 'order'`, but gemini doesn't support const

src/platform/CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+0.4
+---
+
+ * Add support for `DiscriminatorMap` on single polymorphic interface parameters in JSON schema generation
```
aszenz: won't work with default tool argument resolver though since it lacks discriminator

src/platform/src/Contract/JsonSchema/Factory.php
aszenz: anyOf is also used in list of discriminators above, see:

https://github.com/symfony/ai/blob/2a2b6be79bf80b992e5b78a539b3f27e9d75e30c/src/platform/src/Contract/JsonSchema/Factory.php#L197-L211

examples/agent/toolcall-polymorphic-interface.php
chr-hertel: Yes, I think users would also expect consistency in the feature set of both parts

src/platform/src/Contract/JsonSchema/Factory.php
chr-hertel: can you please add a comment documenting that, so we can follow that later without checking the PR

chr-hertel:
heads up, there is #1568 going on in parallel - we'll merge that one first I assume, and you can rebase.

examples/agent/toolcall-polymorphic-interface.php
aszenz: added php doc block, enum and property type support too to keep it consistent with structured output.

also added examples that showcase those features across platforms

aszenz:
> heads up, there is #1568 going on in parallel - we'll merge that one first I assume, and you can rebase.

That change seems big, maybe it makes more sense to merge this smaller improvement first

chr-hertel:
Merged #1683, rebase unlocked :+1:

aszenz:
@chr-hertel Can you review this again

aszenz:
@chr-hertel Ping , is this still needed, then i can resolve conflicts

chr-hertel:
Yes, sorry, i was also having a look the other day and got into thinking regarding the examples - i get why you did it with the `foreach` but it's rather not the style like we do it atm - didn't have a clever idea yet - examples are going quite crazy, but also they don't cost a lot

aszenz:
> Yes, sorry, i was also having a look the other day and got into thinking regarding the examples - i get why you did it with the `foreach` but it's rather not the style like we do it atm - didn't have a clever idea yet - examples are going quite crazy, but also they don't cost a lot

I think many of the examples are serving more as test cases. Hence I used foreach

chr-hertel (approved):

chr-hertel:
Thank you @aszenz.

PR #1560, opened 2026-02-06
[Agent][Chat] Add `metadata` to `AssistantMessage`

### LINKED ISSUE #1558: [Chat] Metadata in returned AssistantMessage is always empty
I am using the [symfony/ai-chat](https://github.com/symfony/ai-chat) package (v0.3.0) to create a interactive chatbot. I have configured the package to use an agent (which contains some custom tools) along with the following [bridge for session-based message storage](https://github.com/symfony/ai-session-message-store). An issue I notice however is that the metadata on the returned AssistantMessage always seems to be empty.

As you can see in the code below from the [Chat component](https://github.com/symfony/ai-chat/blob/main/src/Chat.php), only the content of the result gets added to the AssisantMessage instance. I would expect the submit function of this component to take the commented line which I added into account so the metadata is properly set on the AssistantMessage.

```php
{

    $messages->add($message);

    // $assistantMessage->getMetadata()->add('sources', $result->getMetadata()->get('sources', []));

    $this->store->save($messages);

    return $assistantMessage;
}
```

If there was a deliberate choice on not to include the metadata when using the Chat component, I would really appreciate any feedback!

janssensglenn:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #1558
| License       | MIT

This PR adds the sources of a TextResult to the metadata of the AssistantMessage within the [Chat](https://github.com/symfony/ai-chat) component.

### DIFF
--- a/src/agent/CHANGELOG.md
+++ b/src/agent/CHANGELOG.md
@@ -5,6 +5,7 @@ CHANGELOG
 ---
 
  * [BC BREAK] Rename `Symfony\AI\Agent\Toolbox\Tool\Agent` to `Symfony\AI\Agent\Toolbox\Tool\Subagent`
+ * Add `MetaDataAwareTrait` to `MockResponse`, the metadata will also be set on the returned `TextResult` when calling the `toResult` function
 
 0.3
 ---
--- a/src/chat/CHANGELOG.md
+++ b/src/chat/CHANGELOG.md
@@ -5,3 +5,4 @@ CHANGELOG
 ---
 
  * Add the component
+ * Add `metadata` from `TextResult` to `AssistantMessage`
--- a/src/chat/tests/ChatTest.php
+++ b/src/chat/tests/ChatTest.php
@@ -11,25 +11,24 @@
 
 namespace Symfony\AI\Chat\Tests;
 
-use PHPUnit\Framework\MockObject\MockObject;
 use PHPUnit\Framework\TestCase;
-use Symfony\AI\Agent\AgentInterface;
+use Symfony\AI\Agent\MockAgent;
+use Symfony\AI\Agent\MockResponse;
 use Symfony\AI\Chat\Chat;
 use Symfony\AI\Chat\InMemory\Store as InMemoryStore;
 use Symfony\AI\Platform\Message\AssistantMessage;
 use Symfony\AI\Platform\Message\Message;
 use Symfony\AI\Platform\Message\MessageBag;
-use Symfony\AI\Platform\Result\TextResult;
 
 final class ChatTest extends TestCase
 {
-    private AgentInterface&MockObject $agent;
+    private MockAgent $agent;
     private InMemoryStore $store;
     private Chat $chat;
 
     protected function setUp(): void
     {
-        $this->agent = $this->createMock(AgentInterface::class);
+        $this->agent = new MockAgent();
         $this->store = new InMemoryStore();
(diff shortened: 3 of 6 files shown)


src/chat/tests/ChatTest.php
Guikingone: While you're adding the metadata, let's benefit from this small refactoring to update the code and use `MockAgent`: 

```php
$agent = new MockAgent([
    'Hello, how are you?' => 'I am doing well, thank you!',
]);

$chat = self::createChat($agent);
$result = $chat->submit($userMessage);
```

Easier than using the mocks from PHPUnit.

janssensglenn: @Guikingone  Good idea! I do notice however that the MockAgent makes it unable for me to test that the sources are included in the returned AssistantMessage, since these cannot be added to the MockResponse. Is it perhaps an idea to add the Metadata object as an optional argument in the constructor?

    final class MockResponse
    {

        public function __construct(
            private readonly string $content = '',
        ) {
        }

        public function toResult(): ResultInterface
        {
            return new TextResult($this->content);
        }

        public function getContent(): string
        {
            return $this->content;
        }

        /**
         * Create a MockResponse from a string.
         */
        public static function create(string $content): self
        {
            return new self($content);
        }
    }

Guikingone: We can also use the `MetadataAwareTrait` in `MockResponse`, would be "more compliant" than changing the API of `MockResponse` IMO, @OskarStark any thoughts?

janssensglenn: @Guikingone I updated the test class to use the MockAgent now, thanks for the suggestion. The MockAgent however will call the getResult() function of the MockResponse and return it's value. To make sure the metadata was included here as well I tried doing something like the screenshot below, but I do not know if this is the route we want to take. What are your thoughts?

Screenshot 2026-02-06 at 21 46 56

OskarStark:
Please add entries in changelog file for chat and agent

src/agent/CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ CHANGELOG
 ---
 
  * [BC BREAK] Rename `Symfony\AI\Agent\Toolbox\Tool\Agent` to `Symfony\AI\Agent\Toolbox\Tool\Subagent`
+ * [BC BREAK] Added `MetaDataAwareTrait` to `MockResponse`, the metadata will also be set on the returned `TextResult` when calling the `toResult` function
```
```suggestion
 * Add `MetaDataAwareTrait` to `MockResponse`, the metadata will also be set on the returned `TextResult` when calling the `toResult` function
```
this is not a BC break to me

src/chat/CHANGELOG.md
```diff
@@ -5,3 +5,4 @@ CHANGELOG
 ---
 
  * Add the component
+ * Add metadata from TextResult to AssistantMessage
```
OskarStark:
```suggestion
 * Add `metadata` from `TextResult` to `AssistantMessage`
```

janssensglenn:
@OskarStark Thanks for the feedback, I have updated both CHANGELOG.md files now.

src/chat/tests/ChatTest.php
Guikingone: Sorry for the delay, I'd say that as `MockResponse` is only used in tests, adding this extra line is not an issue 🤔

chr-hertel:
Thank you @janssensglenn.

PR #1563, opened 2026-02-07
[Store] Introduce integration tests for stores

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | 
| License       | MIT

Want to tackle a few things here
* `examples/commands/store.php` is not an example, but integration tests
  back then we went for the synergy about the docker setup, but with that `remove` PR it gets a bit more important for tests
* store tests are by far the slowest in our pipeline since we bring up all the container together and run it sequentially
* bring in a store conformance test with a base integration test

Initially built by Claude

### DIFF
--- a/src/store/src/Test/AbstractStoreIntegrationTestCase.php
+++ b/src/store/src/Test/AbstractStoreIntegrationTestCase.php
@@ -1,0 +1,124 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Store\Test;
+
+use PHPUnit\Framework\Attributes\Depends;
+use PHPUnit\Framework\TestCase;
+use Symfony\AI\Platform\Vector\Vector;
+use Symfony\AI\Store\Document\Metadata;
+use Symfony\AI\Store\Document\VectorDocument;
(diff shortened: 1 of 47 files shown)


src/store/src/Test/AbstractStoreIntegrationTestCase.php
```diff
@@ -0,0 +1,124 @@
...
+        $this->assertSame('third document', $found->getMetadata()['name']);
+    }
+
+    #[Depends('testQueryDocuments')]
+    public function testRemoveDocuments()
+    {
+        try {
+            self::$store->remove(self::DOCUMENT_ID_3);
+        } catch (LogicException) {
+            $this->markTestSkipped('Store does not support remove().');
```
wachterjohannes, line 86: We should introduce a specific exception - not that we hide here real exceptions

src/store/src/Test/AbstractStoreIntegrationTestCase.php
```diff
@@ -0,0 +1,124 @@
...
+        foreach ($results as $result) {
+            $this->assertNotSame(self::DOCUMENT_ID_3, $result->getId());
+        }
+    }
+
+    #[Depends('testRemoveDocuments')]
+    public function testDropStore()
+    {
+        if (!self::$store instanceof ManagedStoreInterface) {
+            $this->markTestSkipped('Store does not implement ManagedStoreInterface.');
```
wachterjohannes, line 102: hmm did not realy like the Skipped for all the unsupported features - but for now its OK

PR #1564, opened 2026-02-07
[Agent] Add support for `thoughtSignature` in result conversion and normalization

### LINKED ISSUE #1539: [Platform][Gemini] Support thought (reasoning) in tool calls to prevent 400 INVALID_ARGUMENT
**Description**
When using Google Gemini "Thinking" models (like `gemini-3-pro-preview`) with Tool Calling enabled, the API requires that the reasoning generated by the model (the `thought` part) be sent back in the message history when a tool call is performed.

Currently, `symfony/ai-gemini-platform` does not extract the `thought` field from the response, and subsequently does not include it when normalizing the `AssistantMessage` for the next request. This results in the following error from the Gemini API:

```
Error "400" - "INVALID_ARGUMENT": "Function call is missing a thought_signature in functionCall parts. 
This is required for tools to work correctly, and missing thought_signature may lead to degraded model performance. 
Additional data, function call default_api:similarity_search , position 2. 
Please refer to https://ai.google.dev/gemini-api/docs/thought-signatures for more details."
```

**How to reproduce**
1. Configure an agent with a Gemini Thinking model (e.g., `gemini-3-pro-preview`).
2. Give the agent access to a tool.
3. Ask a question that triggers a tool call.
4. The API returns a 400 error because the `thought` part is missing in the message history sent by the bridge.

**Possible Solution**
According to the [Google Gemini documentation](https://ai.google.dev/gemini-api/docs/thought-signatures), the `thought_signature` part must be preserved and sent back as the first part of the message containing the `functionCall`.

1. **`AssistantMessage` evolution**: Add a way to store the thinking/reasoning content in the `AssistantMessage` (perhaps via metadata or a specific property).
2. **`ResultConverter`**: Update `Symfony\AI\Platform\Bridge\Gemini\Gemini\ResultConverter` to extract the `thought` field from the candidate parts.
3. **`AssistantMessageNormalizer`**: Update `Symfony\AI\Platform\Bridge\Gemini\Contract\AssistantMessageNormalizer` to prepend the `thought` part before the `functionCall` part in the request payload.

I can try to implement the feature myself. I would just like some validation beforehand on how to do it.

ehibes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | Fix #1539 
| License       | MIT


OskarStark:
Thanks for contributing. This is a new feature to me. Can you please update the corresponding CHANGELOG file? Thanks

src/agent/tests/Toolbox/AgentProcessorTest.php
```diff
@@ -413,4 +413,40 @@ public function testCustomMaxIterationsLimitAllowsConfiguredIterations()
...
+    public function testMetadataGetsPropagatedFromToolCallResultToAssistantMessage()
+    {
+        $toolCall = new ToolCall('id1', 'tool1', ['arg1' => 'value1']);
+        $toolbox = $this->createMock(ToolboxInterface::class);
+        $toolbox
+            ->expects($this->once())
+            ->method('execute')
+            ->willReturn(new ToolResult($toolCall, 'Test response'));
+
+        // On prépare un ToolCallResult avec une métadonnée 'thought'
```
OskarStark: English please

fabpot:
Related to #1635?

ehibes:
> Related to #1635?

It seems so, except that in the case of Gemini, we just get a `thoughtSignature` when we call tools.

chr-hertel:
I think it would be great to adopt the `ThinkingContent` here, but it also happens on non-streaming responses for tool calls?

ehibes:
I’m not sure whether my custom agent calls Gemini-3 in streaming or non-streaming mode.
ThinkingSignature is required for Gemini-3, but not for Gemini-2.5.
I can also try implementing ThinkingContent.

ehibes:
> I think it would be great to adopt the `ThinkingContent` here, but it also happens on non-streaming responses for tool calls?

@chr-hertel I can confirm that ThinkingSignature is also returned in non-streaming responses for tool calls. In that case, returning a single ResultInterface becomes difficult, since two results are actually needed.
Maybe introducing a MultipartResultInterface at the platform level could be a solution.

chr-hertel:
i think we have multiple options here, my favorite one currently is to have the `ThinkingContent` as optional part of the `ToolCall`, cc @fabpot WDYT?

it can be yielded prior to the ToolCall when streaming, but still be part of that object later on and in case on non-streaming responses.

valtzu:
[The docs](https://ai.google.dev/gemini-api/docs/thought-signatures) mention that `thoughtSignature` can be included in any part, therefore I'm not sure if making it only a part of `ToolCall` is sufficient / correct place:

> the API may return a thoughtSignature field within the content parts of the response (e.g., `text` or `functionCall` parts).

[Though it's not currently enforced to be provided for non-function-call parts](https://ai.google.dev/gemini-api/docs/thought-signatures#non-function-call):

> The API does not strictly enforce validation. You won't receive a blocking error if you omit them, though **performance may degrade.**

---

 For example, when using [combination of server tools & function calls](https://ai.google.dev/gemini-api/docs/tool-combination), `thoughtSignature` is included in all parts and the platform expects to get those all back as-is:

<details>
  <summary>Request</summary>

```json
{
    "contents": [
        {
            "role": "user",
            "parts": [
                {
                    "text": "What was the 12 month Euribor rate a week ago based on https://www.euribor-rates.eu/en/current-euribor-rates/4/euribor-rate-12-months/"
                }
            ]
        }
    ],
    "tools": [
        {
            "url_context": {}
        },
        {
            "functionDeclarations": [
                {
                    "name": "get_current_time",
                    "description": "Get current date and time."
                }
            ]
        }
    ],
    "toolConfig": {
        "includeServerSideToolInvocations": true
    }
}
```
</details>
<details>
<summary>Response</summary>

```json5
{
  "candidates": [
    {
      "content": {
        "parts": [
          {
            "thoughtSignature": "...",
            "toolCall": {
              "toolType": "URL_CONTEXT",
              "args": {
                "urls": [
                  "https://www.euribor-rates.eu/en/current-euribor-rates/4/euribor-rate-12-months/"
                ]
              },
              "id": "sbhhfjsp"
            }
          },
          {
            "thoughtSignature": "...",
            "toolResponse": {
              "toolType": "URL_CONTEXT",
              "response": {
                "url_metadata": [
                  {
                    "retrieved_url": "https://www.euribor-rates.eu/en/current-euribor-rates/4/euribor-rate-12-months/",
                    "url_retrieval_status": "URL_RETRIEVAL_STATUS_SUCCESS"
                  }
                ]
              },
              "id": "sbhhfjsp"
            }
          },
          {
            "text": "Based on the provided URL, a week ago was Sunday, March 15, 2026. Since Euribor rates are not published on weekends, there is no rate for that exact date. \n\nHowever, the rates for the closest surrounding trading days were:\n* **Friday, March 13, 2026:** 2.522%\n* **Monday, March 16, 2026:** 2.540%",
            "thoughtSignature": "..."
          }
        ],
        "role": "model"
      },
      // ...grounding etc metadata
    }
  ],
  // ...token count etc
}
```

</details>

It's not possible to use combination of server & client tools unless you have `toolConfig.includeServerSideToolInvocations` set to `true`, and once you do that there's multiple parts with signatures that are enforced to be included in subsequent requests. Only the `text` part signature can be omitted, though it does seem the response takes longer when done so.

Also worth noting that currently `Gemini` implementation in this repo does not support these server-side calls (`toolCall`/`toolResponse`) at all – just wanted to jump in here so that implementing it later would not be too difficult.

chr-hertel:
I guess we're done here with #1853 (and some follow ups), feel free to reopen if i missed something. Thanks for raising the topic - we changed a lot in the end to improve this holistically.

PR #1568, opened 2026-02-08
[Platform] Make JSON schema generation extensible

### LINKED ISSUE #1534: [Platform] Make json schema generation extendable
### Feature request

It'd be nice to have an optional `symfony/validator` integration that would transform Symfony validation constraints to JSON schema (for tool calls / structured output) + it would also check those with validator before invoking the tool. To make all this possible I think the json schema generation needs to be decoupled from `With` attribute and do something similar to https://github.com/nelmio/NelmioApiDocBundle/tree/5.x/src/ModelDescriber/Annotations.

In addition to using validation constraints, it'd still be possible to use `#[With]`, the same way as you can use `#[OA\Property]` with nelmio api doc bundle in case it's not possible to infer the whole schema from validation constraints alone.

valtzu:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | Fix #1534
| License       | MIT

Rearrange schema generation logic to allow extending it.

After this it will be possible to add optional `ValidatorConstraintDescriber` that amends the schema based on Symfony Validator metadata.


chr-hertel:
> I would really like if "normal" Serializer behavior worked here, e.g. getters and setters. If this support is planned, the proposed interface here would not work.

like sitting on to of the metadata collected by serializer? not sure how that would work :thinking:

valtzu:
Thanks for taking a look, but tbh I'm not sure if we should merge this 😄

It is clear what I want to achieve, but how-to part not so clear

If we marked the interface as internal, then maybe more mergeable as its less set-in-stone

chr-hertel:
well, i'm not sure about some things about it as well, but i think it's not a step in the wrong direction.

i wonder about
* can we have some kind of data model instead of `&array`
* is reflector maybe to concrete already?

but still, i feel like it's better than before - are you worried about BC promise here? then `@internal` could help, yes

OskarStark:
If it can ease things, let's make it internal

valtzu:
Ok, after some thought marking it as `@internal` would mean the PR title would not hold. If you think the changes here are fine, let's merge as-is and improve later if needed?

One caveat with the current solution is that it's not possible to alter the "outer" schema from a `DescriberInterface` implementation: this means that e.g. with `Assert\NotNull` constraint it won't be possible to add the property to `required`. Though currently all properties are marked as required so in that sense it's a non-issue.

---

> can we have some kind of data model instead of &array

I do agree a proper `JsonSchema` DTO instead of just phpstan type could be nice. This class could possibly be the same as the attribute (currently named `#[With]`). Or maybe we could just use an already tested package like https://github.com/zircote/swagger-php though I do undestand json schema & openapi are not the same thing (but close).

> is reflector maybe to concrete already?

Maybe, and interface with union type is not the cleanest.

chr-hertel:
so we'd have two approaches here:
1. use case driven, bottom up, from what we want to do
  => if this PR solves a problem for what you are facing right now, let's do it
2. zoom out to abstraction more and tackle the theory basically
  => not on my priority list, but happy to contribute ideas & direction

for 2, i would think we'd need to go into this
image

what could be an evolution of this, with the question if we have to do it ourselves or if we could sit on top of something doing the work - ofc with schema inference that's already the feature, pulling information in that are defined with Symfony Validator constraints already as example. should always be possible to do it standalone tho - we wouldn't want people to install Symfony Validator to describer their tools or response format - so there needs to be that agnostic layer inbetween - and if it's only json schema data model, i think we're fine maintaining that - at least i feel like that kind of code is rather cheap to maintain.

in the end it's a matter of priority tho - i don't have the pain or use-case right now - the only thing I might tackle atm is to extend enums for coming from a dynamic data source, e.g. DB

valtzu:
> if this PR solves a problem for what you are facing right now, let's do it

To evaluate this, I created a follow-up PR for that Validator stuff: https://github.com/symfony/ai/pull/1606

While I'm not super happy with the Describer abstraction it does seem to do the job.

src/platform/CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+0.4
```
chr-hertel, line 9: needs a bump now:

```suggestion
0.6
```

src/platform/src/Contract/JsonSchema/Describer/TypeInfoDescriber.php
```diff
@@ -0,0 +1,190 @@
...
+                        'anyOf' => $discriminators,
+                    ];
+                }
+
+                if (\in_array($className, ['DateTime', 'DateTimeImmutable', 'DateTimeInterface'], true)) {
+                    return ['type' => 'string', 'format' => 'date-time'];
+                }
+
+                // Recursively build the schema for an object type
+                return $this->factory->buildProperties($className) ?? ['type' => 'object'];
```
chr-hertel, line 131: Any blockers for moving this into a `SerializerMetadataDescriber`?

src/platform/src/Contract/JsonSchema/Describer/TypeInfoDescriber.php
```diff
@@ -0,0 +1,173 @@
...
+
+            case $type->isIdentifiedBy(TypeIdentifier::NULL):
+                return ['type' => 'null'];
+            case $type->isIdentifiedBy(TypeIdentifier::STRING):
+            default:
+                // Fallback to string for any unhandled types
```
valtzu, line 137: For me fallback to string does not make any sense – if no type available, then schema should not limit the type at all (i.e. no schema / empty schema). But I did not want to change the existing logic regarding this.

src/platform/src/Contract/JsonSchema/Describer/TypeInfoDescriber.php
valtzu: Assuming you meant the discriminator part, no blockers, done.

valtzu:
> 2. zoom out to abstraction more and tackle the theory basically
>    => not on my priority list, but happy to contribute ideas & direction
> 
> for 2, i would think we'd need to go into this image


Something like this is now implemented in https://github.com/symfony/ai/pull/1683 (replacing this)

chr-hertel:
Closed in favor of #1683

PR #1569, opened 2026-02-08
[AI Bundle] Fix ResultNormalizer parse-time import causing fatal error

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | N/A
| License       | MIT

## Summary

When the symfony/ai-cache-platform bridge package is not installed, the parse-time import of ResultNormalizer in services.php causes a fatal error before the container build phase can remove it.

This fix moves the service registration from parse-time configuration (services.php) to runtime conditional registration (AiBundle.php), using a string class name to avoid class loading at parse time.

The service is now only registered when ContainerBuilder::willBeAvailable() confirms the cache platform bridge is present, following the same pattern used for other optional bridge integrations.

### DIFF
--- a/src/ai-bundle/src/AiBundle.php
+++ b/src/ai-bundle/src/AiBundle.php
@@ -56,6 +56,7 @@ use Symfony\AI\Platform\Bridge\Anthropic\PlatformFactory as AnthropicPlatformFac
 use Symfony\AI\Platform\Bridge\Azure\OpenAi\PlatformFactory as AzureOpenAiPlatformFactory;
 use Symfony\AI\Platform\Bridge\Bedrock\PlatformFactory as BedrockFactory;
 use Symfony\AI\Platform\Bridge\Cache\CachePlatform;
+use Symfony\AI\Platform\Bridge\Cache\ResultNormalizer;
 use Symfony\AI\Platform\Bridge\Cartesia\PlatformFactory as CartesiaPlatformFactory;
 use Symfony\AI\Platform\Bridge\Cerebras\PlatformFactory as CerebrasPlatformFactory;
 use Symfony\AI\Platform\Bridge\Decart\PlatformFactory as DecartPlatformFactory;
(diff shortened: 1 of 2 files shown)


src/ai-bundle/src/AiBundle.php
```diff
@@ -541,8 +541,13 @@ private function processPlatformConfig(string $type, array $platform, ContainerB
 
             return;
         }
-        if (!ContainerBuilder::willBeAvailable('symfony/ai-cache-platform', CachePlatform::class, ['symfony/ai-bundle'])) {
-            $container->removeDefinition('ai.platform.cache.result_normalizer');
+
+        if (ContainerBuilder::willBeAvailable('symfony/ai-cache-platform', CachePlatform::class, ['symfony/ai-bundle'])) {
+            $container->register('ai.platform.cache.result_normalizer', 'Symfony\\AI\\Platform\\Bridge\\Cache\\ResultNormalizer')
```
OskarStark: Using ::class shouldn't be a problem here, as all other classes from other bridges are also not autoloaded using parse time, right?

wachterjohannes: @OskarStark absolutly right - have fixed it

wachterjohannes:
@OskarStark the failing testcase seems something different - in my opinion this PR is ready

OskarStark:
What do you mean? All checks passed 🤔

wachterjohannes:
> What do you mean? All checks passed 🤔

image

A check is failing 😂

OskarStark:
Oh, it's unrelated. Weird, it was green, or maybe caching issue

chr-hertel (approved):
somehow i feel like we patched that before 😂 🙈

chr-hertel:
Thank you @wachterjohannes.

PR #1570, opened 2026-02-08
[Store] Add query abstraction with filter support

### LINKED ISSUE #1562: [Store] Open `Store::query(Vector)` method for different query types
Currently the entire `Store` infrastructure is really coupled to `Document` and `Vector` - which made sense when thinking of "classic RAG", but we see more thing emerging right now:

* Hybrid queries, like #783 
* Vectorization internalized, like ChromaDB, also leading to plain text as input
* Additional filters

On top we aim to build up a high-level service called `Retriever`, that supports various RAG scenarios:
Image

with **pre query** extensions like "language normalization" or "multi-query" and **post query** extensions like user-land reranking.

while discussing with @wachterjohannes and looking again at @ahmed-bhs PR, there was the idea of introducing some kind of `QueryInterface` or `Query` to open up the first parameter of `Store::query()` method.

**1. QueryInterface**

`Store::query(QueryInterface $query, array $options = [])`

could work with `VectorQuery`, `TextQuery` and `HybridQuery`, potentially combined with `StoreInterface` implementations signalling what they support - or just throwing on invalid arguments

**2. Query**

`Store::query(Query $query, array $options = [])`

could work with `Query::fromVector()` and `Query::fromText()` or `Query->addVector()` and `Query->addText()` - making hybrid implicit

the difference is not that huge tho, we should just decide
* do we want to go there?
* do we want to be graceful? like you can add text, but a store might just ignore it
* do we shift filter options into the query as well?

cc @Guikingone @OskarStark @wachterjohannes @ahmed-bhs

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | Fix #1562
| License       | MIT

## Description

This PR implements a query abstraction pattern for the Store component, enabling support for multiple query types (Vector, Text, Hybrid) with an integrated filter system. This addresses RFC #1562 and provides a flexible foundation for
advanced retrieval use cases.

### Query Abstraction
- **QueryInterface**: Base interface for all query types with `getType()` and `getFilter()` methods
- **QueryType Enum**: Defines Vector, Text, and Hybrid query types
- **Query Implementations**:
- `VectorQuery`: Classic vector similarity search (supported by all stores)
- `TextQuery`: Full-text search or internal vectorization (ChromaDB, Postgres, Cache)
- `HybridQuery`: Combined vector + text search with configurable `semanticRatio` parameter (Postgres, ChromaDB, Cache)

### Filter System
- **FilterInterface**: Base interface for query filters with `getType()` and `toArray()` methods
- **EqualFilter**: Equality filter for metadata filtering (e.g., filtering by locale, category, status)
- Filters are integrated directly into query objects

### Store Updates

**Postgres Store** - Extended to support all three query types:
- Vector search using pgvector extension
- Full-text search using PostgreSQL's native FTS
- Hybrid search with weighted scoring combining vector similarity and text relevance
- EqualFilter support via JSONB operators

**ChromaDB Store** - Extended to support all three query types:
- Vector and text queries via ChromaDB's native API
- Hybrid query implementation with result merging
- EqualFilter support via ChromaDB's filtering syntax

**Pinecone, Redis, Qdrant Stores** - Vector only with filter support:
- Vector similarity search
- EqualFilter support via respective backend filter syntaxes

**Cache Store** - All query types with filter support:
- In-memory implementations for all query types
- EqualFilter support via metadata filtering

### API Changes
- **StoreInterface**: Updated with new `query(QueryInterface $query, array $options = [])` signature
- **Capability Signaling**: New `supports(string $queryClass): bool` method to check store capabilities
- **Retriever**: Updated to use query abstraction with automatic capability detection
- **Exception**: New `UnsupportedQueryTypeException` thrown for unsupported query types

### Store::query() Signature Change

**Before:**
```php
$store->query($vector, ['limit' => 10]);
```

**After:**
```php
use Symfony\AI\Store\Query\VectorQuery;

$store->query(new VectorQuery($vector), ['limit' => 10]);
```

All existing code needs to wrap Vector objects in VectorQuery constructor.

### Vector Query with Filter
```php
use Symfony\AI\Store\Query\VectorQuery;
use Symfony\AI\Store\Query\Filter\EqualFilter;

// Search for documents in English locale
$query = new VectorQuery($vector, new EqualFilter('locale', 'en'));
$results = $store->query($query, ['limit' => 10]);
```

### Text Query
```php
use Symfony\AI\Store\Query\TextQuery;

// Full-text search (Postgres, ChromaDB, Cache)
$query = new TextQuery('space adventure', new EqualFilter('category', 'scifi'));
$results = $store->query($query, ['limit' => 5]);
```

### Hybrid Query
```php
use Symfony\AI\Store\Query\HybridQuery;

// Combine vector and text with 70% vector weight
$query = new HybridQuery(
    $vector,
    'space adventure',
    0.7,  // semanticRatio (vector weight)
    new EqualFilter('status', 'published')
);
$results = $store->query($query);
```

### Capability Checking
```php
// Check store support before using query type
if ($store->supports(HybridQuery::class)) {
   $query = new HybridQuery($vector, $text);
} else {
    $query = new VectorQuery($vector);
}
```

### DIFF
--- a/src/store/src/Bridge/ChromaDb/Store.php
+++ b/src/store/src/Bridge/ChromaDb/Store.php
@@ -17,7 +17,11 @@ use Symfony\AI\Platform\Vector\Vector;
 use Symfony\AI\Store\Document\Metadata;
 use Symfony\AI\Store\Document\VectorDocument;
 use Symfony\AI\Store\Exception\RuntimeException;
+use Symfony\AI\Store\Exception\UnsupportedQueryTypeException;
 use Symfony\AI\Store\ManagedStoreInterface;
+use Symfony\AI\Store\Query\QueryInterface;
+use Symfony\AI\Store\Query\TextQuery;
+use Symfony\AI\Store\Query\VectorQuery;
 use Symfony\AI\Store\StoreInterface;
 
 /**
@@ -88,30 +92,95 @@ final class Store implements ManagedStoreInterface, StoreInterface
         $collection->delete(ids: $ids);
     }
 
-    /**
-     * @param array{where?: array<string, string>, whereDocument?: array<string, mixed>, include?: array<string>, queryTexts?: array<string>} $options
-     */
-    public function query(Vector $vector, array $options = []): iterable
+    public function supports(string $queryClass): bool
     {
-        $include = null;
-        if ([] !== ($options['include'] ?? [])) {
-            $include = array_values(
-                array_unique(
-                    array_merge(['embeddings', 'metadatas', 'distances'], $options['include'])
-                )
-            );
+        return \in_array($queryClass, [
+            VectorQuery::class,
+            TextQuery::class,
+        ], true);
+    }
+
+    /**
+     * @param array{where?: array<string, string>, whereDocument?: array<string, mixed>, include?: array<string>, limit?: positive-int} $options
+     */
+    public function query(QueryInterface $query, array $options = []): iterable
+    {
+        if (!$this->supports($query::class)) {
+            throw new UnsupportedQueryTypeException($query::class, $this);
         }
+
+        return match (true) {
+            $query instanceof VectorQuery => $this->queryVector($query, $options),
+            $query instanceof TextQuery => $this->queryText($query, $options),
+            default => throw new UnsupportedQueryTypeException($query::class, $this),
+        };
+    }
+
+    /**
+     * @param array{where?: array<string, string>, whereDocument?: array<string, mixed>, include?: array<string>, limit?: positive-int} $options
+     *
+     * @return iterable<VectorDocument>
+     */
+    private function queryVector(VectorQuery $query, array $options): iterable
+    {
+        $include = $this->buildInclude($options);
 
         $collection = $this->client->getOrCreateCollection($this->collectionName);
         $queryResponse = $collection->query(
-            queryEmbeddings: [$vector->getData()],
-            queryTexts: $options['queryTexts'] ?? null,
-            nResults: 4,
+            queryEmbeddings: [$query->getVector()->getData()],
+            nResults: $options['limit'] ?? 4,
             where: $options['where'] ?? null,
             whereDocument: $options['whereDocument'] ?? null,
             include: $include,
         );
 
+        yield from $this->transformResponse($queryResponse);
+    }
+
+    /**
+     * @param array{where?: array<string, string>, whereDocument?: array<string, mixed>, include?: array<string>, limit?: positive-int} $options
+     *
+     * @return iterable<VectorDocument>
+     */
+    private function queryText(TextQuery $query, array $options): iterable
+    {
+        $include = $this->buildInclude($options);
+
+        $collection = $this->client->getOrCreateCollection($this->collectionName);
+        $queryResponse = $collection->query(
+            queryTexts: $query->getTexts(),
+            nResults: $options['limit'] ?? 4,
+            where: $options['where'] ?? null,
+            whereDocument: $options['whereDocument'] ?? null,
+            include: $include,
+        );
+
+        yield from $this->transformResponse($queryResponse);
+    }
+
+    /**
+     * @param array<string, mixed> $options
+     *
+     * @return array<string>|null
+     */
+    private function buildInclude(array $options): ?array
+    {
+        if ([] === ($options['include'] ?? [])) {
+            return null;
+        }
+
+        return array_values(
+            array_unique(
+                array_merge(['embeddings', 'metadatas', 'distances'], $options['include'])
+            )
+        );
+    }
+
+    /**
+     * @return iterable<VectorDocument>
+     */
+    private function transformResponse(object $queryResponse): iterable
+    {
         $metaCount = \count($queryResponse->metadatas[0]);
 
         for ($i = 0; $i < $metaCount; ++$i) {
--- a/src/store/src/Bridge/ChromaDb/Tests/IntegrationTest.php
+++ b/src/store/src/Bridge/ChromaDb/Tests/IntegrationTest.php
@@ -23,6 +23,11 @@ use Symfony\AI\Store\Test\AbstractStoreIntegrationTestCase;
 #[Group('integration')]
 final class IntegrationTest extends AbstractStoreIntegrationTestCase
 {
+    public function testQueryDocumentsWithTextQuery()
+    {
+        $this->markTestSkipped('ChromaDb TextQuery requires an embedding function to be configured.');
+    }
+
     protected static function createStore(): StoreInterface
     {
         $client = ChromaDB::factory()
--- a/src/store/src/Query/HybridQuery.php
+++ b/src/store/src/Query/HybridQuery.php
@@ -1,0 +1,67 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Store\Query;
+
+use Symfony\AI\Platform\Vector\Vector;
+use Symfony\AI\Store\Exception\InvalidArgumentException;
+
+/**
+ * Combined vector + text search query for hybrid retrieval.
+ *
+ * Hybrid queries combine semantic similarity (vector search) with keyword
+ * matching (full-text search) to provide more accurate and relevant results.
+ *
+ * @author Johannes Wachter <johannes@sulu.io>
+ */
+final class HybridQuery implements QueryInterface
+{
+    /**
+     * @param string|array<string> $text Single search text or array of texts (combined with OR logic)
+     */
+    public function __construct(
+        private readonly Vector $vector,
+        private readonly string|array $text,
+        private readonly float $semanticRatio = 0.5,
+    ) {
+        if ($semanticRatio < 0.0 || $semanticRatio > 1.0) {
+            throw new InvalidArgumentException(\sprintf('Semantic ratio must be between 0.0 and 1.0, got %.2f', $semanticRatio));
+        }
+    }
+
+    public function getVector(): Vector
+    {
+        return $this->vector;
+    }
+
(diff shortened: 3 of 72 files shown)


src/store/src/Retriever.php
```diff
@@ -49,4 +53,29 @@ public function retrieve(string $query, array $options = []): iterable
 
         $this->logger->debug('Document retrieval completed', ['retrieved_count' => $count]);
     }
+
+    private function createQuery(string $query, array $options): QueryInterface
```
wachterjohannes, line 62: This is the most important part of this PR - it selects the right type of query

src/store/src/Query/QueryInterface.php
```diff
@@ -0,0 +1,29 @@
...
+ * Queries represent search intent and are passed to Store::query() along with
+ * execution options (limit, etc.) in the $options array.
+ *
+ * @author Johannes Wachter <johannes@sulu.io>
+ */
+interface QueryInterface
+{
+    public function getType(): QueryType;
+
+   public function getFilter(): ?FilterInterface;
```
wachterjohannes: not sure how the filter should look like and how we could validate them in the store. @chr-hertel for equal its easy but for more complex filters it could be hard to implement that in the stores

wachterjohannes:
for all who take a look at this PR this is a POC as a discussion base

src/store/src/Query/TextQuery.php
```diff
@@ -0,0 +1,48 @@
...
+
+    public function getFilter(): ?FilterInterface
+    {
+        return $this->filter;
+    }
+
+    public function getType(): QueryType
+    {
+        return QueryType::Text;
+    }
```
chr-hertel: we only use that at exceptions right now? let's drop it then or do i miss something?

src/store/src/Retriever.php
```diff
@@ -49,4 +53,29 @@ public function retrieve(string $query, array $options = []): iterable
...
+        if (!$this->store->supports(VectorQuery::class)) {
+            $this->logger->debug('Store does not support vector queries, falling back to TextQuery');
+
+            return new TextQuery($query);
+        }
+
+        if ($this->store->supports(HybridQuery::class)) {
+            $this->logger->debug('Store supports hybrid queries, using HybridQuery with semantic ratio', ['semanticRatio' => $options['semanticRatio'] ?? 0.5]);
+
+            return new HybridQuery($this->vectorizer->vectorize($query), $query, $options['semanticRatio'] ?? 0.5);
```
chr-hertel, line 79: is the assumption fair to use `$query` for both parts?

src/store/src/Query/QueryType.php
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+namespace Symfony\AI\Store\Query;
+
+/**
+ * Defines the types of queries supported by the Store component.
+ *
+ * @author Johannes Wachter <johannes@sulu.io>
+ */
+enum QueryType: string
```
chr-hertel: not sure we need that, do we?

src/store/src/Query/QueryInterface.php
chr-hertel: let's go incremental and skip the filter part for the first PR - we can slice that, right?

src/store/src/Query/HybridQuery.php
```diff
@@ -0,0 +1,67 @@
...
+final class HybridQuery implements QueryInterface
+{
+    public function __construct(
+        private readonly Vector $vector,
+        private readonly string $text,
+        private readonly float $semanticRatio = 0.5,
+        private readonly ?FilterInterface $filter = null,
+    ) {
+        if ($semanticRatio < 0.0 || $semanticRatio > 1.0) {
+            throw new \InvalidArgumentException(\sprintf('Semantic ratio must be between 0.0 and 1.0, got %.2f', $semanticRatio));
```
chr-hertel: use `Symfony\AI\Store\Exception\InvalidArgumentException` instead of native one

src/store/src/Retriever.php
wachterjohannes: do you mean:

* the default semanticRatio?  - what would you use as a default? Or would you only do it when the option is set?
* using the query for both Text and Vector? - how would we get the text for each other?

src/store/src/Query/QueryType.php
wachterjohannes: will remove it

src/store/src/Query/HybridQuery.php
wachterjohannes: will do that

src/store/src/Query/TextQuery.php
wachterjohannes: will remove it

wachterjohannes:
@chr-hertel I’ll remove the filter for now and address it in a separate issue. However, a lightweight filter implementation across all stores would be valuable. Starting with basic equality filters should cover the primary use case and work for most, if not all, stores. More complex filtering can be handled at the application level, where it’s already store-specific and can leverage those advantages.

src/store/src/Retriever.php
chr-hertel: donno :shrug: :joy: 

sometimes i just ask questions to make us aware of some decisions - i think it's fair to couple it - don't have a better idea for now

src/store/src/Bridge/ChromaDb/Store.php
```diff
@@ -67,30 +71,95 @@ public function remove(string|array $ids, array $options = []): void
...
+    public function supports(string $queryClass): bool
+    {
+        return \in_array($queryClass, [
+            VectorQuery::class,
+            TextQuery::class,
+        ], true);
+    }
+
     /**
-     * @param array{where?: array<string, string>, whereDocument?: array<string, mixed>, include?: array<string>, queryTexts?: array<string>} $options
```
wachterjohannes, line 92: @chr-hertel this is a BC break `queryTexts`  was able to handle multiple texts the new implementation is only able to query one text

src/store/src/Bridge/ChromaDb/Tests/IntegrationTest.php
```diff
@@ -23,6 +23,11 @@
 #[Group('integration')]
 final class IntegrationTest extends AbstractStoreIntegrationTestCase
 {
+    public function testQueryDocumentsWithTextQuery()
+    {
+        $this->markTestSkipped('ChromaDb TextQuery requires an embedding function to be configured.');
```
wachterjohannes, line 28: @chr-hertel the ChromaDb - meant to work before but never worked with the text-queries before.

i have done a research with Claude-Code:

```
ChromaDB Text Query Requirements

When using queryTexts parameter (which is what the Symfony AI Store's TextQuery does), an embedding function must be configured on the collection. This is because:

1. Text needs to be converted to vectors: ChromaDB performs similarity search using vector embeddings, so query text must first be embedded
2. The collection's embedding function does this: When you query with text, ChromaDB uses the collection's configured embedding function to convert your query text into a vector
3. Without an embedding function, it fails: If no embedding function is set, there's no way to convert the text query into a searchable vector, resulting in the "Call to a member function generate() on null" error

Relevant Documentation

Official ChromaDB Documentation:

- https://docs.trychroma.com/docs/embeddings/embedding-functions - Main documentation about how embedding functions work in ChromaDB
- https://docs.trychroma.com/guides - General ChromaDB usage patterns
- https://cookbook.chromadb.dev/embeddings/bring-your-own-embeddings/ - ChromaDB Cookbook for custom embeddings

PHP Client (codewithkyrian/chromadb-php):

- https://github.com/CodeWithKyrian/chromadb-php - Main PHP client repository with examples
- https://github.com/CodeWithKyrian/chromadb-php/blob/main/README.md - Setup and usage examples

The Issue in Our Case

The Symfony AI Store's ChromaDB bridge creates collections without an embedding function:
$this->client->createCollection($this->collectionName);

When TextQuery calls the collection's query() method with queryTexts parameter, ChromaDB tries to use $embeddingFunction->generate() but since none is configured, it's null, causing the error.

Solution: Collections need to be created with an embedding function to support text-based queries.

Sources:
- https://docs.trychroma.com/docs/embeddings/embedding-functions
- https://github.com/CodeWithKyrian/chromadb-php
- https://docs.trychroma.com/guides
```

wachterjohannes: created an issue for that: https://github.com/symfony/ai/issues/1601

wachterjohannes:
@chr-hertel final in my opinion - the only issue is the chromadb stuff - but created the issue already

src/store/src/Bridge/ChromaDb/Tests/IntegrationTest.php
chr-hertel: thanks for that! :+1:

src/store/src/StoreInterface.php
```diff
@@ -34,6 +35,13 @@ public function remove(string|array $ids, array $options = []): void;
...
+     *
+     * @throws UnsupportedQueryTypeException if query type not supported
+     */
+    public function query(QueryInterface $query, array $options = []): iterable;
+
+    /**
+     * @param class-string<QueryInterface> $queryClass The query class to check
      */
-    public function query(Vector $vector, array $options = []): iterable;
+    public function supports(string $queryClass): bool;
```
chr-hertel, line 46: let's add the `supports` when we need it - currently it's not used and you've been super explicit in the implementations with those exceptions

chr-hertel (changes requested):

src/store/src/StoreInterface.php
wachterjohannes: it is use in the Retriever to find out what the store is able to query for https://github.com/symfony/ai/pull/1570/files#diff-4eb38714aa9287457a8f9c404ad19af65af7f87dceafb329d134513ab5dc7805R70

chr-hertel: you are correct of course! :pray:

chr-hertel:
Thank you @wachterjohannes.

PR #1578, opened 2026-02-10
[Agent] Add `HasSourcesTrait` to `Subagent`

### LINKED ISSUE #1574: [Agent] No metadata available when using agent as subagent
I have an agent setup with some custom tools. In these tools I make use of the HasSourcesInterface to add some sources to my returned result. When I directly interact with this agent I can access these sources in my frontend just fine, but when I use this agent as a subagent (agent-as-a-tool) the metadata gets lost. 

When looking at the [Symfony\AI\Agent\Toolbox\Tool\Subagent class](https://github.com/symfony/ai-agent/blob/main/src/Toolbox/Tool/Subagent.php) this makes complete sense, however it does make me wonder if there is a way to preserve the metadata (or at least the sources) when using an agent as a tool. If this is not possible, should I perhaps think about ditching the agent-as-a-tool approach?

Image

janssensglenn:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | Fix #1574 
| License       | MIT

This PR adds the sources of the TextResult to the [Subagent](https://github.com/symfony/ai-agent/blob/main/src/Toolbox/Tool/Subagent.php) class. This way the sources do not get lost when using an agent as a tool.

### DIFF
--- a/src/agent/CHANGELOG.md
+++ b/src/agent/CHANGELOG.md
@@ -6,6 +6,7 @@ CHANGELOG
 
  * [BC BREAK] Rename `Symfony\AI\Agent\Toolbox\Tool\Agent` to `Symfony\AI\Agent\Toolbox\Tool\Subagent`
  * Add `MetaDataAwareTrait` to `MockResponse`, the metadata will also be set on the returned `TextResult` when calling the `toResult` function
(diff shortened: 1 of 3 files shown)


OskarStark (commented):
Please add a testcase, thanks

chr-hertel:
Thank you @janssensglenn.

PR #1582, opened 2026-02-11
[Store][Weaviate] Add collection existence check before creation

rensjanszen:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | 
| License       | MIT

Added a collection existence check so multiple setups with the same collection name don't crash. Used the schema list API since the GET call was case-sensitive. 
Saw that some of the other bridges already had this functionality.

Skip collection creation if it already exists, using case-insensitive matching via the schema API.

### DIFF
--- a/src/store/src/Bridge/Weaviate/Store.php
+++ b/src/store/src/Bridge/Weaviate/Store.php
@@ -40,6 +40,10 @@ final class Store implements ManagedStoreInterface, StoreInterface
             throw new InvalidArgumentException('No supported options.');
         }
 
+        if ($this->collectionExists()) {
+            return;
+        }
+
         $this->request('POST', 'v1/schema', [
             'class' => $this->collection,
         ]);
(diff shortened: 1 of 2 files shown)


src/store/src/Bridge/Weaviate/Store.php
```diff
@@ -148,4 +152,24 @@ private function convertToVectorDocument(array $data): VectorDocument
 
         return new VectorDocument($id, $vector, new Metadata(json_decode($data['_metadata'], true)));
     }
+
+    /**
+     * Helper function that checks if the collection exists, using the schema API since the GET is case-sensitive.
```
Guikingone: IMHO, the comment is not needed, the function name is clear enough

Guikingone:
@rensjanszen Can you add an extra test to ensure that this case is covered?

rensjanszen:
@Guikingone sure

rensjanszen:
> @rensjanszen Can you add an extra test to ensure that this case is covered?

@Guikingone Does this new test case suffice?

src/store/src/Bridge/Weaviate/Store.php
rensjanszen: Felt like i needed to specify why I couldn't just do a GET and check if 200 or 404

Guikingone: Open to debate, IMHO, the name is enough to explain what the method is doing (the code is clean so that's another good point IMHO to remove the comment), each store has their own implementation / API / weird endpoints, no need to justify the usage of this one here IMO 🙂

Guikingone:
Yes, good for me 👍🏻

chr-hertel (approved):
Not sure we have that kind of gracefulness with others, but +1 from me - thanks!

chr-hertel:
Thank you @rensjanszen.

src/store/src/Bridge/Weaviate/Store.php
chr-hertel: agree, removed on merge

Issue #1301, opened 2025-12-28
Road to Symfony AI v1.0


chr-hertel:
Of course, we want Symfony AI to become a stable and reliable set of packages, complying to known Symfony standards, like the BC promise. This will only be achieved with the release of v1.0.

Reaching that level of reliability is important for adoption and using Symfony AI in production without patching your code all the time.

When going with `symfony/ai` instead of integration into `symfony/symfony` we decoupled releases deliberately since we're still figuring out what we need to build. And as long, as we are aware of flaws, that make BC breaks necessary, it would create extra effort to provide a BC layer - slowing is down.

However, we want to create transparency, where we are on this road, for users to follow along.
This issue tracks high-level items, that need to be resolved until we can tag a v1.0 of Symfony AI.
When all items are tackled, a feature freeze around 4 weeks will kick in, to stabilize the codebase.

* _This uses the [MoSCoW method](https://en.wikipedia.org/wiki/MoSCoW_method) for priotization. Items are up for discussion._
* _Items are labeled as `MUST` if impact to component API is expected._

## Open to resolve
 
* **Platform Component**
  * `MUST` Solve dynamic model catalogs decoupling package lifecycle of model lifecycle
  * `MUST` Better stream handling for exposing thinking/reasoning as well
  * `MUST` Better handling of multi-modal responses (generated speech, images or video)
  * `SHOULD` Meaningful adoption of capabilities, validation and routing in ModelCatalog/Platform
  * `SHOULD` Remote tool support
  * `COULD` Streamlined Error Handling
  * `COULD` Streamlined JSON to data conversion
* **Agent Component**
  * `MUST` Redesigned API to tailored Agent API instead of being a slim Platform decorator
  * `MUST` Introduce traces to come out-of-the-box instead of custom 
  * `SHOULD` Integration with workflow component
  * `SHOULD` Integrate hand-off into one universal Agent
  * `SHOULD` Sources/traces support across built-in tools
  * `SHOULD` Message template support for `SimilaritySearch` (and others?)
  * `COULD` MCP support, agent acting as
* **Store Component**
  * `MUST` Universal storage component supporting all use-cases instead of components/bridges per use-case
  * `MUST` Introduce retrieval extension points for features like multi-query or reranking
  * `MUST` Work out-of-the-box with entities/repositories - unified store vs. additional store
  * `MUST` `TBD` Extract Retrieval into separate component
* **Chat Component**
  * `MUST` Handle multi-user conversations
 * **AI Bundle**
[...]

PR #1583, opened 2026-02-11
[Chat] Add support for streaming

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | Rework of #928 - Part of #1301
| License       | MIT

Solve the streaming aspect of `Chat` by using the `StreamResult` listeners, `TraceableChat` use a `return` to prevent waiting for the `Generator` to be consumed before storing informations.

### DIFF
--- a/src/chat/tests/ChatTest.php
+++ b/src/chat/tests/ChatTest.php
@@ -19,6 +19,7 @@ use Symfony\AI\Chat\InMemory\Store as InMemoryStore;
 use Symfony\AI\Platform\Message\AssistantMessage;
 use Symfony\AI\Platform\Message\Message;
 use Symfony\AI\Platform\Message\MessageBag;
+use Symfony\AI\Platform\Result\StreamResult;
 
 final class ChatTest extends TestCase
 {
@@ -35,13 +36,13 @@ final class ChatTest extends TestCase
 
     public function testItInitiatesChatByClearingAndSavingMessages()
     {
-        $messages = $this->createMock(MessageBag::class);
+        $agent = new MockAgent();
 
-        $this->chat->initiate($messages);
+        $chat = new Chat($agent, new InMemoryStore());
+        $chat->initiate(new MessageBag());
 
+        $agent->assertNotCalled();
         $this->assertCount(0, $this->store->load());
-
-        $this->agent->assertNotCalled();
     }
 
     public function testItSubmitsUserMessageAndReturnsAssistantMessage()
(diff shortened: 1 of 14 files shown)


Guikingone:
@marco-jouwweb if you want to take a look 🙂

src/platform/src/Message/AssistantMessage.php
marco-jouwweb, line 21: Nitpick: could it be possible to make the class itself `readonly`, instead of each separate constructor parameter?

src/chat/tests/ChatTest.php
```diff
@@ -108,4 +111,71 @@ public function testItHandlesEmptyMessageStore()
...
+                    yield 'Yes, ';
+                    yield 'I can!';
+                })());
+            }
+
+            public function getName(): string
+            {
+                return 'mock-stream';
+            }
+        };
```
marco-jouwweb: Somewhere else in this PR, you used `MockAgent`. Shouldn't we use that here as well?
(I expect this is related to the `MockAgent` not supporting streaming right now, but I feel like this'd be a good reason to update it instead of working around it :-))

marco-jouwweb:
> @marco-jouwweb if you want to take a look 🙂

I did a quick check, looks very nice & clean! I can review more in-depth somewhere next week and do some testing (I don't have the focus right now). If you haven't merged by then, of course :-)

Thanks for the cool feature!

marco-jouwweb (approved):
I checked it out, I think it looks very clean. Good job!
I left 2 minor comments. One other thing I was still wondering: 

I remember streaming being incompatible with session conversation storage, due to the nature of HTTP/session lifecycles & streaming trying to update after headers are sent. This might not be the most important thing right now, but I think it would be nice to throw early if streaming is combined with session store. Fast failure will save a lot of people from a painful debugging session. Or we could at least explicitly document this limitation somewhere, so people won't be surprised. :-)

P.S. I purely looked at code, I didn't actively test the branch.

src/platform/src/Message/AssistantMessage.php
Guikingone: Project convention, Symfony applies the same rule 🙂

src/chat/tests/ChatTest.php
Guikingone: Fixed, thanks for the catch 😅

src/chat/composer.json
chr-hertel, line 1: please revert and use `link` instead

OskarStark (approved):

OskarStark:
Thank you @Guikingone.

PR #1590, opened 2026-02-13
[Store] Update Symfony dev dependencies in Store

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | 
| License       | MIT

### DIFF
--- a/src/store/composer.json
+++ b/src/store/composer.json
@@ -53,9 +53,9 @@
         "phpstan/phpstan-phpunit": "^2.0",
         "phpstan/phpstan-strict-rules": "^2.0",
         "phpunit/phpunit": "^11.5.46",
-        "symfony/console": "^6.4|^7.1",
-        "symfony/dependency-injection": "^6.4|^7.1",
-        "symfony/dom-crawler": "^6.4|^7.1"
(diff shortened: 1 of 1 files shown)


src/store/composer.json
```diff
@@ -53,9 +53,7 @@
         "phpstan/phpstan-phpunit": "^2.0",
         "phpstan/phpstan-strict-rules": "^2.0",
         "phpunit/phpunit": "^11.5.46",
-        "symfony/console": "^6.4|^7.1",
-        "symfony/dependency-injection": "^6.4|^7.1",
-        "symfony/dom-crawler": "^6.4|^7.1"
+        "symfony/console": "^7.3|^8.0"
```
GromNaN: It's also updating the version constraint.

chr-hertel: i know - that's the point actually

OskarStark:
Thank you @chr-hertel.

src/store/composer.json
chr-hertel: PR title was misleading - didn't get in the beginning why we have them at all, but that was wrong :D

PR #1452, opened 2026-01-24
[Platform] Add Load Balanced Platform

clegginabox:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| License       | MIT

This PR is a working proof-of-concept to gather feedback on the approach before I flesh it out further, add tests and documentation.

Adds a `LoadBalancedPlatform` that distributes requests across multiple platforms based on their available capacity.

## Use Case
I have multiple API keys across different providers (OpenAI, Anthropic, Bedrock) with varying rate limits and concurrency constraints. I want to fan out requests and have them distributed automatically whilst respecting each platform's limits.

More context: https://github.com/symfony/ai/issues/41#issuecomment-3794465762

## Design
Follows the same pattern as `FailoverPlatform`, but for capacity distribution rather than error recovery.

- `CapacityProvider` - abstracts capacity checking (rate limits, concurrency, or both)
- `PlatformCapacity` - pairs a platform with its capacity provider
- `PlatformSelectionStrategy`- determines platform ordering (random, round-robin, etc.)

## Open Questions

Model handling across providers
Currently the `$model` parameter is passed directly to whichever platform is selected. This works when balancing multiple keys for the same provider, but not when mixing providers (OpenAI/Anthropic/Bedrock) since model names differ.

Possible approaches:
1. Model-per-platform - PlatformCapacity specifies the model to use, request-level $model is ignored or used as a capability hint
2. Model mapping - translate a generic model name to provider-specific models
3. Document as a constraint - all configured platforms must support the requested model

Leaning towards option 1 for flexibility, but interested in feedback.

## Example usage 

```php
use Redis;
use Symfony\AI\Platform\Bridge\Anthropic\PlatformFactory as AnthropicPlatformFactory;
use Symfony\AI\Platform\Bridge\LoadBalanced\LoadBalancedPlatform;
use Symfony\AI\Platform\Bridge\LoadBalanced\PlatformCapacity;
use Symfony\AI\Platform\Bridge\LoadBalanced\Provider\CompositeCapacityProvider;
use Symfony\AI\Platform\Bridge\LoadBalanced\Provider\NoLimitCapacityProvider;
use Symfony\AI\Platform\Bridge\LoadBalanced\Provider\RateLimitedCapacityProvider;
use Symfony\AI\Platform\Bridge\LoadBalanced\Strategy\RandomStrategy;
[...]

Issue #1519, opened 2026-01-29
[Platform] ResultEvent is dispatched before the result is in


chr-hertel:
Due to the async setup of `DeferredResult` the Platform dispatches the `ResultEvent` before the result is actually available - that's not really helpful since you would expect that event only when the result is available.

Should we add just a third event like `ResultConvertedEvent` or just shift the current one to later point in time?

PR #1596, opened 2026-02-14
[Platform] Add onResolved callback to `DeferredResult`

clegginabox:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes/no
| License       | MIT

Linked to: https://github.com/symfony/ai/pull/1452

Adds `onResolved()` to `DeferredResult` - triggered when a stream is fully consumed/a request is complete. 

Provides a hook to release capacity for the load balanced platform.

### DIFF
--- a/src/platform/src/Result/DeferredResult.php
+++ b/src/platform/src/Result/DeferredResult.php
@@ -14,9 +14,9 @@ namespace Symfony\AI\Platform\Result;
 use Symfony\AI\Platform\Exception\ExceptionInterface;
 use Symfony\AI\Platform\Exception\UnexpectedResultTypeException;
 use Symfony\AI\Platform\Metadata\MetadataAwareTrait;
+use Symfony\AI\Platform\Result\Stream\CallbackStreamListener;
 use Symfony\AI\Platform\ResultConverterInterface;
 use Symfony\AI\Platform\TokenUsage\StreamListener;
-use Symfony\AI\Platform\TokenUsage\TokenUsage;
 use Symfony\AI\Platform\Vector\Vector;
 
 /**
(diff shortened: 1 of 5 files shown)


chr-hertel:
somehow this reminds me of #1519 - would it also be possible to solve your issue with an event that is dispatched if the result (no matter if streamed or not) is resolved?

clegginabox:
> somehow this reminds me of #1519 - would it also be possible to solve your issue with an event that is dispatched if the result (no matter if streamed or not) is resolved?

I don't think so.

I've just had a quick look - it would couple my bridge to the event dispatcher? I would need a way to know which event corresponds to which platform so I release the correct token. I also need to release on failures - would it make sense to use a single event for success and failure?

clegginabox:
If my understanding of https://github.com/symfony/ai/issues/1519 is correct - with this PR you could do something like

```php
  $result = $this->convertResult($model, $this->doInvoke($model, $payload, $options), $options);                                                                                                                                                                                                                            
                  
  $result->onResolved(function () use ($model, $result, $options, $input): void {
      $this->eventDispatcher?->dispatch(new ResultEvent($model, $result, $options, $input));
  });

  return $result;
  ```

**Comment by clegginabox** [comment]
could maybe also be used to fix - https://github.com/symfony/ai/issues/1391?

**Comment by chr-hertel** [comment]
I'm still wondering if we can use the EventDispatcher right away with `ResultConversionEvent` and `ResultErrorEvent` with:
```php
// In DeferredResult.php
    public function __construct(
        private readonly ResultConverterInterface $resultConverter,
        private readonly RawResultInterface $rawResult,
        private readonly \Closure $onConversion,
        private readonly \Closure $onError,
    ) {
    }
```

```php
// Platform.php
return new DeferredResult($resultConverter, $result, $options, $this->dispatcher->dispatch(...), $this->dispatcher->dispatch(...));
```

or similar code

clegginabox:
superseded by #2222

PR #1602, opened 2026-02-14
[Mate] Add RegistryProvider service to eliminate Registry duplication

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | -
| License       | MIT

## Summary

This PR introduces a `RegistryProvider` service to centralize Registry creation and eliminate code duplication.

**Problem:**
Both `ServeCommand` and `ToolsCallCommand` were independently creating and populating `Registry` instances with the same
 logic:
```php
$this->registry = new Registry(null, $logger);
$loader->load($this->registry);
```

#### Solution:
- Add RegistryProvider service that lazily creates and populates a shared Registry instance
- Update ServeCommand to use RegistryProvider::getRegistry() instead of addLoader()
- Update ToolsCallCommand to depend on RegistryProvider instead of creating its own Registry
- Commands now depend on RegistryInterface rather than the concrete Registry class

#### Benefits:
- Single source of truth for Registry creation
- Lazy initialization avoids unnecessary capability discovery during container compilation
- Better separation of concerns - commands don't need to know how to populate the Registry
- Improved testability through interface dependency

### DIFF
--- a/src/mate/src/Command/ServeCommand.php
+++ b/src/mate/src/Command/ServeCommand.php
@@ -19,7 +19,7 @@ use Mcp\Server\Transport\StdioTransport;
 use Psr\Log\LoggerInterface;
 use Symfony\AI\Mate\Agent\AgentInstructionsAggregator;
 use Symfony\AI\Mate\App;
-use Symfony\AI\Mate\Discovery\FilteredDiscoveryLoader;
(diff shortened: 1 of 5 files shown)


chr-hertel:
Thank you @wachterjohannes.

PR #1609, opened 2026-02-16
[Platform][Ollama] Improve support for `ScopingHttpClient`

### LINKED ISSUE #899: [Platform] OllamaClient does not really respect the HttpClient configuration
> [!NOTE]
> The issue title is not perfect, but I cannot find something better

---

I have the following configuration
```yaml
ai:
    platform:
        ollama:
            http_client: http_client.ollama

framework:
    http_client:
        scoped_clients:
            http_client.ollama:
                base_uri: '%env(OLLAMA_ENDPOINT)%'
```

```
OLLAMA_ENDPOINT=http://user:password@1.2.3.4
```

And I got the following exception:

>   #message: "Failed to connect to 127.0.0.1 port 11434 after 0 ms: Couldn't connect to server for "http://127.0.0.1:11434/api/show"."

🤔 This is strange, this is not the IP I configured.

But, after reading the [code](https://github.com/symfony/ai/blob/main/src/platform/src/Bridge/Ollama/OllamaClient.php#L39), it's kinda legit. We always use the `host` configured.

IMHO, if an HTTP client is provided (because I need custom option, and I want a dedicated HTTP client for profiling), we must not seed the `base_uri` (At DIC level).

WDYT?

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| BC BREAK | Yes
| Issues        | Fix #899 
| License       | MIT

* Improve `Ollama` platform to support `ScopingHttpClient`
* Rename the `host_url` configuration key to `endpoint`
* Add support for `api_key` for `Ollama.com` Cloud endpoint


chr-hertel:
Thank you @Guikingone.

Issue #1053, opened 2025-12-07
[Platform][Gemini] Response with multiple content parts ends up with an error


chr-hertel:
While working on #1054 I ran into the issue that those image models might generate a response with two content parts: 1 text part and one inlineData part - which is not handled well currently and makes that added example flaky - especially with `gemini-2.5-flash-image`.

Use the example `gemini/image-editing.php` for reproduction and this method is to fix:
`Symfony\AI\Platform\Bridge\Gemini\Gemini\ResultConverter::convertChoice`, see TODO comment

NicolasGraph:
Thanks @chr-hertel for introducing gemini-2.5-flash-image and gemini-3-pro-image-preview models!
I can't see a way to return multiparts responses without returning multiple appropriate results

microHoffman:
I got this issue even when I received multiple content parts with only `text`:

```
      "content": {
        "parts": [
          {
            "text": "\nThe Content...."
          },
          {
            "text": "Another content..."
          },
          {
            "text": "```json\n{\n  ... \n```"
          }
        ],
        "role": "model"
      },
```

Is there a reason why in this case we just do not concatenate the results? or e.g. return array?

NicolasGraph:
@microHoffman, concatenating the result as proposed by @TNAJanssen in [#1046](https://github.com/symfony/ai/pull/1046) is not enough because multiple content parts can be of different types, as decribed by @chr-hertel. I think it is also not as simple as returning an array because of mutliple chained methods, but something like a `ResultBag` should maybe be introduced

PR #1610, opened 2026-02-16
[Platform][Gemini] Skip streaming candidates without content parts

stijnkuppens:
| Q             | A
  | ------------- | ---
  | Bug fix?      | yes
  | New feature?  | no
  | Docs?         | no
  | Issues        | Related #1053
  | License       | MIT

  The `convertStream()` method passes all candidates to `convertChoice()`,
  which directly accesses `$choice['content']['parts']`. During streaming,
  Gemini can send chunks without `content.parts` (e.g. the final chunk
  with only `finishReason`), causing:

  > Warning: Undefined array key "parts"

  The non-streaming path already guards against this on line 57. This PR
  adds the same guard to the streaming path by filtering out candidates
  that don't have `content.parts` set.

### DIFF
--- a/src/platform/src/Bridge/Gemini/Gemini/ResultConverter.php
+++ b/src/platform/src/Bridge/Gemini/Gemini/ResultConverter.php
@@ -75,7 +75,7 @@ final class ResultConverter implements ResultConverterInterface
     private function convertStream(RawResultInterface $result): \Generator
     {
         foreach ($result->getDataStream() as $data) {
-            $choices = array_map($this->convertChoice(...), $data['candidates'] ?? []);
+            $choices = array_values(array_filter(array_map($this->convertChoice(...), $data['candidates'] ?? [])));
 
             if (!$choices) {
                 continue;
@@ -93,7 +93,7 @@ final class ResultConverter implements ResultConverterInterface
     /**
      * @param array{
      *     finishReason?: string,
-     *     content: array{
+     *     content?: array{
      *         parts: array{
      *             functionCall?: array{
      *                 id: string,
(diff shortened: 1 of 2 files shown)


src/platform/src/Bridge/Gemini/Gemini/ResultConverter.php
```diff
@@ -75,7 +75,11 @@ public function getTokenUsageExtractor(): TokenUsageExtractor
     private function convertStream(RawResultInterface $result): \Generator
     {
         foreach ($result->getDataStream() as $data) {
-            $choices = array_map($this->convertChoice(...), $data['candidates'] ?? []);
+            $candidates = array_filter(
+                $data['candidates'] ?? [],
+                static fn (array $candidate): bool => isset($candidate['content']['parts']),
+            );
+            $choices = array_map($this->convertChoice(...), $candidates);
```
chr-hertel: i think it would be great to move the `isset` closer to the data access:

```suggestion
            $choices = array_filter(array_map($this->convertChoice(...), $candidates));
```

in combination with:
```php
    private function convertChoice(array $choice): ToolCallResult|TextResult|BinaryResult|null
    {
        if (!isset($choice['content']['parts']) {
            return null;
        }

        $contentParts = $choice['content']['parts'];
```

stijnkuppens: @chr-hertel Feedback is applied! Just another safety done with array_values restoring array keys for a possible edge case where $choices[0] could still fail if that value would have no parts.

chr-hertel:
Thank you @stijnkuppens.

PR #1611, opened 2026-02-16
[Store] Allow to retrieve documents without vectors

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | 
| License       | MIT

* Introduce a `include_vectors` option in stores to retrieve documents + vectors at runtime
* By default, vectors are not retrieved when using `Store::query()`

### DIFF
--- a/src/agent/src/Bridge/SimilaritySearch/SimilaritySearch.php
+++ b/src/agent/src/Bridge/SimilaritySearch/SimilaritySearch.php
@@ -39,7 +39,7 @@ final class SimilaritySearch
      */
     public function __invoke(string $searchTerm): string
     {
-        $vector = $this->vectorizer->vectorize($searchTerm, ['include_vectors' => true]);
+        $vector = $this->vectorizer->vectorize($searchTerm);
         $this->usedDocuments = iterator_to_array($this->store->query(new VectorQuery($vector)));
 
         if ([] === $this->usedDocuments) {
--- a/src/agent/src/Bridge/SimilaritySearch/SimilaritySearch.php
+++ b/src/agent/src/Bridge/SimilaritySearch/SimilaritySearch.php
@@ -39,7 +39,7 @@ final class SimilaritySearch
      */
     public function __invoke(string $searchTerm): string
     {
-        $vector = $this->vectorizer->vectorize($searchTerm);
+        $vector = $this->vectorizer->vectorize($searchTerm, ['include_vectors' => true]);
         $this->usedDocuments = iterator_to_array($this->store->query(new VectorQuery($vector)));
 
         if ([] === $this->usedDocuments) {
--- a/src/store/src/Bridge/Cache/Store.php
+++ b/src/store/src/Bridge/Cache/Store.php
@@ -12,6 +12,7 @@
 namespace Symfony\AI\Store\Bridge\Cache;
 
 use Psr\Cache\CacheItemPoolInterface;
+use Symfony\AI\Platform\Vector\NullVector;
 use Symfony\AI\Platform\Vector\Vector;
 use Symfony\AI\Store\Distance\DistanceCalculator;
 use Symfony\AI\Store\Document\Metadata;
@@ -113,7 +114,8 @@ final class Store implements ManagedStoreInterface, StoreInterface
     /**
      * @param array{
      *     maxItems?: positive-int,
-     *     filter?: callable(VectorDocument): bool
+     *     filter?: callable(VectorDocument): bool,
+     *     include_vectors?: bool,
      * } $options If maxItems is provided, only the top N results will be returned.
      *            If filter is provided, only documents matching the filter will be considered.
      */
@@ -150,7 +152,7 @@ final class Store implements ManagedStoreInterface, StoreInterface
 
         $vectorDocuments = array_map(static fn (array $document): VectorDocument => new VectorDocument(
             id: $document['id'],
-            vector: new Vector($document['vector']),
+            vector: $options['include_vectors'] ?? false ? new Vector($document['vector']) : new NullVector(),
             metadata: new Metadata($document['metadata']),
         ), $documents);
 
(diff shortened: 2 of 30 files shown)


chr-hertel:
i need some more explanations here tbh - what are we addressing/solving?

Guikingone:
> i need some more explanations here tbh - what are we addressing/solving?

Mainly a performance issue (related to RAM handling of embeddings), currently, when a store is retrieving a set of `VectorDocument`, vectors are retrieved per-document then stored in memory even if you're not accessing them (as `Vector` is not lazy-loaded, once vectors are sent into the class, the array is allocated right away), this could lead to a spike on RAM usage (for `1536` dimensions and around 1000 documents - as returned by OpenAI - the spike can be around `10Mo`) and as the number of documents grows, the RAM usage follow. 

Problem is, most of the time, when dealing with stores, you want to retrieve the document without querying the vectors (as the store already handle them), actually, you're dealing with both the document and the vectors in the current process.

The PR introduce a `LazyVector` (mostly for BC compatibility as the `store` package is targeting `>= 8.2`) that acts as a non-initialized wrapper until the "vectors-related" methods are called (for filtering and so on), in `>= 8.4`, we use the `lazyGhost` to trigger the same behavior while using the "dev-friendly" API from PHP.

My first implementation was focused around the idea of not querying the vectors but not all stores allows to retrieve only the document or document+vectors so handling it in the "public API" looked like a better approach.

Here's the output of `rag/cache.php` with this approach:

Capture d’écran 2026-02-16 à 21 04 36

The best approach would be to adapt converters (pushed in a recent commit) for all platforms that supports embeddings then on stores (as implemented here) to "close the loop".

OskarStark:
Why not making a Vector then lazy by design?

Guikingone:
> Why not making a Vector then lazy by design?

Could be something to explore, I wanted to implement "a workaround" without breaking the current internal API but handling the problem at the `Vector` level would be easier for everyone (and can lead to the removal of the ugly `LazyVector` 😄 )

chr-hertel:
just asked claude to write some benchmarking about your PR and got this:
https://gist.github.com/chr-hertel/679658a100097f947ebc15f05c49d6f8

not saying this vibe-coded script is really worth it - especially since it is just _one_ script, but in general it would be great to decide stuff like this data driven.

"lazy" is just not intuitive for me here, since the data is around already anyways :thinking: 

BUT, there's one question in general: do we need the vector on results anyways? i never used it until now, so maybe a waste of memory. but i never had the case of working with 1000 documents on read as well :D

Guikingone:
> BUT, there's one question in general: do we need the vector on results anyways? i never used it until now, so maybe a waste of memory. but i never had the case of working with 1000 documents on read as well :D

Could be something to discuss, the only usage that I'm seeing is in `SimilaritySearch`, apart from that, never used them on my side (I want to retrieve the document, not the vectors 😄 ), maybe something to explore when using a `with_vectors` option on stores?

Might be cleaner than a "lazy-loaded-like" vector at the end, not returning it is always better than "hiding it" 😅

Guikingone:
Ok, just pushed a refactoring, now, vectors are not loaded into the `VectorDocument` by default, the only way to load them is to use the `with_vectors` option when using `Store::query()` (`SimilaritySearch` is the only one that use it for now), this way, we can retrieve a `VectorDocument` with a `NullVector` (that throws exceptions by default).

I also updated the CHANGELOG + UPGRADE.

src/platform/src/Bridge/Cache/ResultNormalizer.php
```diff
@@ -66,7 +67,7 @@ public function normalize(mixed $data, ?string $format = null, array $context =
                     $data->getContent(),
                 ),
                 VectorResult::class => array_map(
-                    static fn (Vector $vector): array => [
+                    static fn (VectorInterface $vector): array => [
```
OskarStark: these changes can be moved to an extra PR to keep focus, thanks

src/store/src/Bridge/Cache/Tests/StoreTest.php
```diff
@@ -63,7 +69,9 @@ public function testStoreCanSearchUsingCosineDistance()
             new VectorDocument(Uuid::v4(), new Vector([0.3, 0.7, 0.1])),
         ]);
 
-        $result = iterator_to_array($store->query(new VectorQuery(new Vector([0.0, 0.1, 0.6]))));
+        $result = iterator_to_array($store->query(new VectorQuery(new Vector([0.0, 0.1, 0.6])), [
+            'with_vectors' => true,
```
OskarStark:
```suggestion
            'include_vectors' => true,
```
Sounds better IMHO

src/agent/src/Bridge/SimilaritySearch/SimilaritySearch.php
```diff
@@ -39,7 +39,7 @@ public function __construct(
      */
     public function __invoke(string $searchTerm): string
     {
-        $vector = $this->vectorizer->vectorize($searchTerm);
+        $vector = $this->vectorizer->vectorize($searchTerm, ['include_vectors' => true]);
```
chr-hertel: this is def unintuitive - the purpose of the `Vectorizer` is to return a vector - i feel like the option shouldn't be needed here

Guikingone: We can invert the condition in stores, by default, we're returning vectors and if `include_vectors => false` is passed, we're not returning them, it would solve the logic issue here and keep the API "as it" 🙂

OskarStark:
Rebase unlocked after #1628

src/store/src/Bridge/Cache/Store.php
```diff
@@ -150,7 +152,7 @@ private function queryVector(VectorQuery $query, array $options): iterable
 
         $vectorDocuments = array_map(static fn (array $document): VectorDocument => new VectorDocument(
             id: $document['id'],
-            vector: new Vector($document['vector']),
+            vector: $options['include_vectors'] ?? true ? new Vector($document['vector']) : new NullVector(),
             metadata: new Metadata($document['metadata']),
         ), $documents);
```
OskarStark: As this PR is about performance, the vector is now still available in `$document['vector']`, do we need to unset this variable or so to free the memory? otherwise we could keep the code as is and close this PR, no?

chr-hertel:
Please reopen if you want to follow up - in that case please provide some evidence/numbers.

PR #1618, opened 2026-02-17
[Platform][Anthropic] Fix `AssistantMessageNormalizer` to support tools properly

fabpot:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Docs?         | no 
| Issues        | 
| License       | MIT

When Claude prefixes a tool call with a sentence, the `text` block must be preserved in the serialized content array alongside the `tool_use` block. Previously a ternary chose either `text` OR `tool_use`, silently dropping the text.

See https://platform.claude.com/docs/en/agents-and-tools/tool-use/implement-tool-use#model-responses-with-tools


OskarStark:
Thanks for fixing this bug Fabien.

PR #1622, opened 2026-02-17
[Platform][ModelsDev] Add `ModelResolver`

fabpot:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Docs?         | no 
| Issues        | 
| License       | MIT

A small addition to let users use "shortcuts" for the model names they want to use.
We are now sorting models as well so that when using just "anthropic" as a model name, we get back the latest available model.

### DIFF
--- a/src/platform/src/Bridge/ModelsDev/ModelResolver.php
+++ b/src/platform/src/Bridge/ModelsDev/ModelResolver.php
@@ -1,0 +1,130 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Platform\Bridge\ModelsDev;
+
+use Symfony\AI\Platform\Exception\InvalidArgumentException;
+
+/**
+ * Resolves a model specification string to a (provider, model ID) pair.
+ *
+ * Accepted input formats (resolved in that order):
+ *
+ *   * "provider::model": Explicit form like "anthropic::claude-opus-4-5"
+ *
+ *   * "provider": Provider-only, returns the first non-deprecated model listed
+ *     in that provider's catalog.
+ *
+ *   * "model": Model ID, scans all provider catalogs to find the owning
+ *     provider, checking canonical providers first over aggregator/proxy
+ *     providers being preferred.
+ *
+ * @author Fabien Potencier <fabien@symfony.com>
+ */
+final class ModelResolver
+{
+    /**
+     * @var list<string>
+     */
+    private const CANONICAL_PROVIDERS = [
+        'anthropic', 'openai', 'google', 'mistral', 'xai',
+        'deepseek', 'groq', 'cerebras', 'cohere',
+    ];
+
+    private readonly ProviderRegistry $registry;
+
+    public function __construct(
+        ?ProviderRegistry $registry = null,
+        ?string $dataPath = null,
+    ) {
+        $this->registry = $registry ?? new ProviderRegistry($dataPath);
+    }
+
+    /**
+     * Resolves a model specification to a (provider, modelId) pair.
+     *
+     * @return array{provider: string, model_id: string}
+     *
+     * @throws InvalidArgumentException when the provider or model cannot be determined
+     */
+    public function resolve(string $modelSpec): array
+    {
+        // "provider::model"
+        if (str_contains($modelSpec, '::')) {
+            [$provider, $modelId] = explode('::', $modelSpec, 2);
+
+            if ('' === $provider) {
+                throw new InvalidArgumentException(\sprintf('Invalid model specification "%s": provider part is empty.', $modelSpec));
+            }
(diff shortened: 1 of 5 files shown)


src/platform/src/Bridge/ModelsDev/ModelResolver.php
```diff
@@ -0,0 +1,145 @@
...
+     * Resolves a model specification to a (provider, modelId) pair.
+     *
+     * @return array{provider: string, model_id: string}
+     *
+     * @throws InvalidArgumentException when the provider or model cannot be determined
+     */
+    public function resolve(string $modelSpec): array
+    {
+        // "provider:model"
+        if (str_contains($modelSpec, ':')) {
```
chr-hertel: if we'd want to scale that into a more universal feature, that might conflict with other conventions, like the parameters at Ollama:
Image

maybe just `::`, like `huggingface::Qwen/Qwen3.5-397B-A17B`

fabpot: Done

src/platform/src/Bridge/ModelsDev/ModelResolver.php
```diff
@@ -0,0 +1,145 @@
...
+        }
+
+        throw new InvalidArgumentException(\sprintf('Cannot determine provider for model "%s"; Use "provider:model" format to specify it explicitly.', $modelId));
+    }
+
+    private function getCatalog(string $provider): ?ModelCatalog
+    {
+        try {
+            return new ModelCatalog($provider, $this->dataPath);
+        } catch (\Throwable) {
```
chr-hertel: it's only `InvalidArgumentException` that is really to be expected here if i get it right

src/platform/src/Bridge/ModelsDev/ModelResolver.php
```diff
@@ -0,0 +1,145 @@
...
+                continue; // already checked
+            }
+
+            $catalog = $this->getCatalog($providerId);
+            if (null !== $catalog && isset($catalog->getModels()[$modelId])) {
+                return $providerId;
+            }
+        }
+
+        throw new InvalidArgumentException(\sprintf('Cannot determine provider for model "%s"; Use "provider:model" format to specify it explicitly.', $modelId));
```
chr-hertel:
```suggestion
        throw new InvalidArgumentException(\sprintf('Cannot determine provider for model "%s"; Use "provider::model" format to specify it explicitly.', $modelId));
```

src/platform/src/Bridge/ModelsDev/ModelResolver.php
```diff
@@ -0,0 +1,145 @@
...
+
+    /**
+     * Returns the first non-deprecated model ID from a provider's catalog.
+     *
+     * @throws InvalidArgumentException when no models are found for the provider
+     */
+    private function resolveFirstModel(string $provider): string
+    {
+        if (null === $catalog = $this->getCatalog($provider)) {
+            throw new InvalidArgumentException(\sprintf('No model catalog found for provider "%s"; use "provider:model" format.', $provider));
```
chr-hertel:
```suggestion
            throw new InvalidArgumentException(\sprintf('No model catalog found for provider "%s"; use "provider::model" format.', $provider));
```

src/platform/src/Bridge/ModelsDev/ModelResolver.php
```diff
@@ -0,0 +1,145 @@
...
+        }
+
+        if ([] === $models = array_keys($catalog->getModels())) {
+            throw new InvalidArgumentException(\sprintf('No models found for provider "%s"; use "provider:model" format.', $provider));
```
chr-hertel:
```suggestion
            throw new InvalidArgumentException(\sprintf('No models found for provider "%s"; use "provider::model" format.', $provider));
```

src/platform/src/Bridge/ModelsDev/ModelResolver.php
chr-hertel, line 1: what do you think about shifting the `getCatalog` to the `ProviderRegistry`?
feels more related and might ease parts of the implementation

fabpot: Good idea

src/platform/src/Bridge/ModelsDev/composer.json
```diff
@@ -30,7 +30,10 @@
         "phpstan/phpstan-strict-rules": "^2.0",
         "phpunit/phpunit": "^11.5.53",
         "symfony/ai-anthropic-platform": "^0.3",
-        "symfony/ai-gemini-platform": "^0.3"
+        "symfony/ai-bedrock-platform": "^0.3",
+        "symfony/ai-gemini-platform": "^0.3",
+        "symfony/ai-meta-platform": "^0.3",
```
OskarStark, line 35: As the build-packages.php magic foo script does not change path repositories for transitive depoendencies, the fix is to add symfony/ai-meta-platform here too

OskarStark:
I manually squashed the commits

OskarStark:
Thank you Fabien.

PR #1624, opened 2026-02-17
[Chat][Bridge] DoctrineDBALMessageStore - adding support for databases with other existing tables

xprojects-de:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | 
| License       | MIT

If the store is applied to a database that already contains other tables, the schema will not be defined correctly. The existing tables will be defined using a CREATE statement, which will result in an error. The pull request respects the existing structure, including the existing tables, and cleanly adds the new table.

### DIFF
--- a/src/chat/src/Bridge/Doctrine/DoctrineDbalMessageStore.php
+++ b/src/chat/src/Bridge/Doctrine/DoctrineDbalMessageStore.php
@@ -13,6 +13,7 @@ namespace Symfony\AI\Chat\Bridge\Doctrine;
 
 use Doctrine\DBAL\Connection as DBALConnection;
 use Doctrine\DBAL\Platforms\OraclePlatform;
+use Doctrine\DBAL\Schema\ComparatorConfig;
 use Doctrine\DBAL\Schema\Name\Identifier;
 use Doctrine\DBAL\Schema\Name\UnqualifiedName;
 use Doctrine\DBAL\Schema\PrimaryKeyConstraint;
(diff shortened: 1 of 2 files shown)


Guikingone:
Hi @xprojects-de 👋🏻 

Could you rebase against `main` and add tests for this update? Thanks again for the fix 🙂

xprojects-de:
@Guikingone rebased and tests added.
Let me know if you'd like me to do more.

xprojects-de:
"Fabbot" is difficult to fix :-)

OskarStark:
This is a new feature, please add a line in the changelog, thanks

chr-hertel:
Thanks @xprojects-de for working on a smoother integration here! Pretty much needed indeed!

What's your outer context here? E.g. using the ORM in Symfony? In that caser, how does the `doctrine:schema:update` react to this? does this require to configure a schema filter? in that case we should document that as well i feel.

xprojects-de:
Thanks for the feedback @chr-hertel 

I think we're still a long way from seamless integration of schema migration within the entire application. For example, the store would need to integrate schema listeners and respond to them.

Currently, everything is done directly on the connection, which is certainly not good.

As a first step, I would fix the bug so that existing databases can be used as a store, and then, as a next step, aim for complete schema integration using...

e.g. 

```php 
#[AsDoctrineListener(event: 'postGenerateSchema')]
class DoctrineSchemaListener
{
    /**
     * @param GenerateSchemaEventArgs $event
     * @return void
     * @throws TypesException
     */
    public function __invoke(GenerateSchemaEventArgs $event): void
    {
        $table = $event->getSchema()->createTable(...);
        $table->addColumn(...);

    }

}
```

chr-hertel:
Thank you @xprojects-de.

PR #1633, opened 2026-02-18
[Store][Qdrant] Support for `ScopingHttpClient` improved

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | --
| BC BREAK | Yes
| License       | MIT


examples/rag/qdrant.php
```diff
@@ -23,17 +23,17 @@
...
-    http_client(),
-    env('QDRANT_HOST'),
-    env('QDRANT_SERVICE_API_KEY'),
-    'movies',
-);
+$store = new Store(ScopingHttpClient::forBaseUri(http_client(), env('QDRANT_HOST'), [
+    'headers' => [
+        'api-key' => env('QDRANT_SERVICE_API_KEY'),
+    ],
+]), 'movies');
```
OskarStark: I would keep the example simple and avoid using ScopingHttpClient here, you can add another one if you want

src/store/src/Bridge/Qdrant/CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 CHANGELOG
 =========
 
+0.4
```
OskarStark: 0.5

src/store/src/Bridge/Qdrant/CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
...
+---
+
+ * [BC BREAK] The `endpoint` parameter for `Store` has been removed
```
OskarStark:
```suggestion
 * [BC BREAK] The `endpointUrl` parameter for `Store` has been removed
```

examples/rag/qdrant.php
Guikingone: Yes, agree, an other option would be to introduce a `StoreFactory::create` like platforms, this would abstract the creation of the store and remove the `new Store(...)`, WDYT?

OskarStark:
Thank you @Guikingone.

PR #1635, opened 2026-02-18
[Platform] Add thinking support

fabpot:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Docs?         | no 
| Issues        | n/a
| License       | MIT


src/platform/src/Message/AssistantMessage.php
```diff
@@ -29,6 +29,8 @@ final class AssistantMessage implements MessageInterface
     public function __construct(
         private ?string $content = null,
         private ?array $toolCalls = null,
+        private ?string $thinkingContent = null,
+        private ?string $thinkingSignature = null,
```
chr-hertel, line 33: does the caller side have the `ThinkingContent` instance at hand?
might be easier to have:

```suggestion
        private ?ThinkingContent $thinkingContent = null,
```

chr-hertel (approved):
Tested with anthropic - works nicely thanks! :)

only thing i noted was that thinking is always one blob since the stream is buffered, but i get it and don't see a better way 👍

chr-hertel:
Thank you @fabpot.

PR #1641, opened 2026-02-19
[Store][MongoDB] Add `setup_options` to store config

OskarStark:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | -
| License       | MIT

The MongoDB store's `setup()` method accepts a `fields` option to add extra Atlas Search index fields (e.g. filter fields for metadata). However, there was no way to configure these via `ai.yaml` — users had to write a custom command to call `setup()` with the options manually.

This PR adds a `setup_options` configuration node to the MongoDB store config, with a `fields` entry, so users can configure extra index fields directly in `ai.yaml` and rely on the built-in `ai:store:setup` command:

```yaml
ai:
  store:
    mongodb:
      my_store:
        index_name: vector_index
        setup_options:
          fields:
            - { path: 'metadata.store_code', type: 'filter' }
```

Then simply run:

```
php bin/console ai:store:setup ai.store.mongodb.my_store
```

A `variableNode` is used for `fields` because each Atlas Search field object's structure varies by `type` (`filter` fields have `path`/`type`; `vector` fields add `numDimensions`/`similarity`).

The PR also fixes a missing wire-up in `AiBundle::processStoreConfig()` — the MongoDB block was building the store definition but never collecting `setup_options` into `$setupStoresOptions`, meaning `SetupStoreCommand` would always call `$store->setup([])` regardless of config.


src/ai-bundle/CHANGELOG.md
```diff
@@ -4,7 +4,7 @@ CHANGELOG
 0.5
 ---
 
- * [BC BREAK] The `host_url` configuration key for `ElevenLabs` has been renamed `endpoint`
+ * Add `setup_options` configuration for MongoDB store to pass extra Atlas Search index fields to `ai:store:setup`
```
OskarStark:
```suggestion
 * Add `setup_options` configuration for MongoDB store to pass extra fields to `ai:store:setup`
```

PR #1644, opened 2026-02-19
[Platform][Anthropic] Add support for prompt caching

fabpot:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Docs?         | no 
| Issues        | 
| License       | MIT

~~`CacheRetentionInputProcessor` is part of the agent component as this is where it should be. It does nothing if used for a non-Anthropic model. Most of the logic is then in the Anthropic platform bridge.~~


src/agent/src/InputProcessor/CacheRetentionInputProcessor.php
```diff
@@ -0,0 +1,59 @@
...
+ * via {@see \Symfony\AI\Platform\TokenUsage\TokenUsageInterface::getCachedTokens()}.
+ * Using this processor with a non-Anthropic platform is a no-op: the option
+ * is stripped by each platform's ModelClient before the request is sent.
+ *
+ * @author Fabien Potencier <fabien@symfony.com>
+ */
+final readonly class CacheRetentionInputProcessor implements InputProcessorInterface
+{
+    /**
+     * @param 'none'|'short'|'long' $cacheRetention
```
fabpot: Using strings here as we have the same in platform and I didn't want to share an enum between the 2 components.

src/platform/src/Bridge/Anthropic/Contract/AnthropicContract.php
```diff
@@ -27,7 +27,7 @@ public static function create(NormalizerInterface ...$normalizer): Contract
             new DocumentUrlNormalizer(),
             new ImageNormalizer(),
             new ImageUrlNormalizer(),
-            new MessageBagNormalizer(),
+            new PromptCacheNormalizer(), // replaces MessageBagNormalizer; also handles cache injection
```
fabpot: I could make it more explicit by using `new PromptCacheNormalizer(new MessageBagNormalizer())` but not sure if this would be better.

src/platform/src/Contract.php
```diff
@@ -71,12 +72,16 @@ public static function create(NormalizerInterface ...$normalizer): self
...
      *
      * @return array<string, mixed>|string
      */
-    final public function createRequestPayload(Model $model, object|array|string $input): string|array
+    final public function createRequestPayload(Model $model, object|array|string $input, array $options = []): string|array
     {
-        return $this->normalizer->normalize($input, context: [self::CONTEXT_MODEL => $model]);
+        return $this->normalizer->normalize($input, context: [
+            self::CONTEXT_MODEL => $model,
+            self::CONTEXT_OPTIONS => $options,
```
fabpot, line 83: Maybe there is a better way, tell me if that works for you.

src/agent/tests/InputProcessor/CacheRetentionInputProcessorTest.php
```diff
@@ -0,0 +1,94 @@
...
+
+use PHPUnit\Framework\TestCase;
+use Symfony\AI\Agent\Input;
+use Symfony\AI\Agent\InputProcessor\CacheRetentionInputProcessor;
+use Symfony\AI\Platform\Message\MessageBag;
+
+/**
+ * @author Fabien Potencier <fabien@symfony.com>
+ */
+final class CacheRetentionInputProcessorTest extends TestCase
```
OskarStark: should we test with "foo" as unsupported one?

src/platform/src/Bridge/Anthropic/Contract/PromptCacheNormalizer.php
```diff
@@ -0,0 +1,145 @@
...
+ * Injects Anthropic prompt-caching markers into the normalised message payload.
+ *
+ * Anthropic prompt caching requires a {"cache_control": {"type": "ephemeral"}}
+ * annotation on the *last block* of the *last user message*.  When the
+ * `cacheRetention` option is present in the invocation options (propagated via
+ * {@see Contract::CONTEXT_OPTIONS}), this normalizer wraps the standard
+ * {@see MessageBagNormalizer} output and splices in the required annotation.
+ *
+ * Supported `cacheRetention` values
+ * ----------------------------------
```
OskarStark:
```suggestion
 * ---------------------------------
```
nitpick

src/agent/src/InputProcessor/CacheRetentionInputProcessor.php
```diff
@@ -0,0 +1,59 @@
...
+ *               the plain ephemeral marker
+ * - `'none'`  – prompt caching disabled for this invocation
+ *
+ * Other platforms
+ * ---------------
+ * OpenAI caches prompt prefixes automatically (no annotation required).
+ * Cached-token counts are already surfaced by the OpenAI TokenUsageExtractor
+ * via {@see \Symfony\AI\Platform\TokenUsage\TokenUsageInterface::getCachedTokens()}.
+ * Using this processor with a non-Anthropic platform is a no-op: the option
+ * is stripped by each platform's ModelClient before the request is sent.
```
chr-hertel: That coupling of high-level and low-level behavior is a bit scary to me - I see two options to isolate this two Anthropic usage more:

1. use options directly when sending requests to Anthropic, like:
    ```php
    $platform->invoke('claude-3-5-sonnet-latest', $messages, ['cacheRetention' => 'short']);
    ```
    (_could be possible already_)

2. add an EventListener for `Symfony\AI\Platform\Event\InvocationEvent` to Anthropic bridge, maybe added via `PlatformFactory` to basically do the same like this InputProcessor.

I guess 1. is not easy in your case - or with failover platform, but 2. could be more elegant and isolated if it works in reality and not only in my head.

WDYT?

src/platform/src/Contract.php
chr-hertel: works for me :)

src/agent/src/InputProcessor/CacheRetentionInputProcessor.php
chr-hertel: or 3: handle it in the Anthropic ModelClient :thinking:

fabpot: You're right. I've reworked this PR to move the feature directly to the ModelClient instead.

UPGRADE.md
```diff
@@ -7,6 +7,8 @@ Platform
  * The `hostUrl` parameter for `ElevenLabsClient` has been removed
  * The `host` parameter for `ElevenLabsApiCatalog` has been removed
  * The `hostUrl` parameter for `PlatformFactory::create()` in `ElevenLabs` has been renamed to `endpoint`
+ * `Symfony\AI\Platform\TokenUsage\TokenUsageInterface` has two new methods:
+   `getCacheCreationTokens()` and `getCacheReadTokens()`
```
chr-hertel: would be for a new section for 0.5 to 0.6

chr-hertel (approved):
good to be merged from my end after that `UPGRADE.md` comment, thanks already!

OskarStark (approved):

UPGRADE.md
fabpot: Fixed

fabpot:
Added support in the AI bundle

chr-hertel:
Thank you @fabpot.

PR #1649, opened 2026-02-19
[Platform][Scaleway] Fix tool call without arguments

guillaume-ro-fr:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Docs?         | no 
| Issues        | None 
| License       | MIT

When calling a tool on Scaleway, the `arguments` array may not be included in the results, which causes problems when converting the result with `ResultConverter`.

This PR fixes the issue by checking that the array exists before retrieving its content.

### DIFF
--- a/ai.symfony.com/templates/base.html.twig
+++ b/ai.symfony.com/templates/base.html.twig
@@ -49,7 +49,7 @@
                                 <a href="https://x.com/symfony" target="_blank" class="footer-meta text-decoration-none">
                                     {{ ux_icon('simple-icons:x', {width: 16, height: 16, class: 'ms-1'}) }}
                                 </a>
-                                <a href="ttps://github.com/symfony/ai" target="_blank" class="footer-meta text-decoration-none">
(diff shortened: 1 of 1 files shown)


OskarStark:
Thanks for fixing this bug, please add tests to avoid further regressions. It would also be helpful to add a script in examples/ folder.

guillaume-ro-fr:
@OskarStark Thanks for your feedback. I added some tests, and I just noticed another issue detected with LiteLLM (using `symfony/ai-generic-plaform`) connected to Scaleway in backend: the list of arguments may be present but empty, which causes a syntax error (since PHP 7.0, an empty string in `json_decode` throws an exception).

The schema: Symfony <--> LiteLLM <--> Scaleway

Should I make another PR to fix the issue in `ai-generic-plaform` (same code as here), or can I change the title and mention both platforms?

OskarStark:
Please create a separate one, thanks.

PR #1651, opened 2026-02-19
feat: Enhance Model Catalog with labels

### LINKED ISSUE #1648: [Platform] Add a human-readable label to models
**Is your feature request related to a problem?**

When building UIs on top of Symfony AI (for example, integrating with Drupal AI), model selection dropdowns and admin screens currently have to display raw model IDs like `gemini-1.5-pro-002` or `claude-3-5-sonnet-20241022`. There is no standardized way to get a human-readable name from the catalog, so every integration has to maintain its own ID-to-label mapping.

**Describe the solution you'd like**

Extend both the `Model` class and `ModelCatalogInterface` so that each model exposes a mandatory human-readable `label`. This label is what UI consumers (Drupal, Symfony apps, etc.) can render in select lists and admin screens.

Because this is still in the `0.x` phase, it is acceptable for this to be a breaking change that requires catalog implementors to update their code.

**Proposed changes to `Model`**

```php
class Model
{
    /**
     * @param non-empty-string     $name
     * @param non-empty-string     $label       Human-readable display name
     * @param Capability[]         $capabilities
     * @param array<string, mixed> $options     The default options for the model usage
     */
    public function __construct(
        private readonly string $name,
        private readonly string $label,
        private readonly array $capabilities = [],
    ) {
        if ('' === trim($name)) {
            throw new InvalidArgumentException('Model name cannot be empty.');
        }

        if ('' === trim($label)) {
            throw new InvalidArgumentException('Model label cannot be empty.');
        }
    }

...

    /**
     * @return non-empty-string
     */
    public function getLabel(): string
    {
        return $this->label;
    }

  ...
}
```

**Proposed changes to `ModelCatalogInterface`**

```php

interface ModelCatalogInterface
{
....

    /**
     * @return array<string, array{
     *     class: string,
     *     label: string,
     *     capabilities: list<Capability>
     * }>
     */
    public function getModels(): array;
}
```

**Usage example**

```php
$model = $catalog->getModel('gemini-1.5-pro-002');

$model->getName();  // "gemini-1.5-pro-002"
$model->getLabel(); // "Gemini 1.5 Pro"
```

**Why make `label` mandatory?**

- Every model exposed to users has some human-readable name that makes sense in a UI.
- Making `label` required keeps catalog implementations consistent across providers.
- It removes the need for null checks and fallbacks in every consumer.

**Describe alternatives you have considered**

- Keeping `Model` as-is and only extending the array returned by `getModels()` with an optional `label`. This still pushes null-handling into every consumer and is less robust.
- Maintaining ID-to-label maps in consumer projects (like [Drupal AI](https://www.drupal.org/project/ai)). This is what we currently do and it tends to drift out of sync and duplicates effort across projects.

**Additional context**

This requirement came up while integrating [Drupal AI](https://www.drupal.org/project/ai) with Symfony AI, where both administrative and user-facing UIs display human-readable model names instead of internal IDs. Having the label defined at the Symfony AI level keeps this concern close to the model metadata and avoids each integration reinventing it. 

Before Symfony AI (Ignore outdated model names :) )

Image

With Symfony AI

Image

mxr576:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | Fix #1648
| License       | MIT

TODO:
- [ ] Changelog entry: should follow https://symfony.com/doc/current/contributing/code/conventions.html#writing-a-changelog-entry
- [ ] Audit Amazee AI 
- [ ] Check if other AI providers were missed by Copilot

### DIFF
--- a/src/platform/tests/ModelCatalog/AbstractModelCatalogTest.php
+++ b/src/platform/tests/ModelCatalog/AbstractModelCatalogTest.php
@@ -166,6 +166,7 @@ final class AbstractModelCatalogTest extends TestCase
                 $this->models = [
                     'test-model' => [
                         'class' => Model::class,
+                        'label' => 'Test Model',
                         'capabilities' => [Capability::INPUT_TEXT],
                     ],
                 ];


OskarStark:
If no label is set and this is fine IMHO when someone provide its own catalog, we should fallback to the models name.

src/platform/src/Bridge/Cartesia/Tests/CartesiaClientTest.php
```diff
@@ -35,7 +35,7 @@ public function testSupportsModel()
         );
 
         $this->assertTrue($client->supports(new Cartesia('sonic-3')));
-        $this->assertFalse($client->supports(new Model('any-model')));
+        $this->assertFalse($client->supports(new Model('any-model', 'Any Model')));
```
OskarStark, line 38: It should never be required to provide a label

mxr576: Thanks @OskarStark  for the early review on this probably, the decision is crucial about whether label should be mandatory or not. Can we discuss PROS and CONS in the related issue? Originally I was thinking about being lenient on it, but probably requiring a label to be always specified is better (the provider package can always decide falling back to the ID if they do not provide a label or could try to generate a label from an ID (see Amazee AI in the PR)).

PR #1652, opened 2026-02-20
[AI Bundle] Make OVH platform configurable

loicsapone:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| License       | MIT

The new OVH platform cannot be used with the Bundle (I forgot to add the necessary configuration in the initial PR...)

### DIFF
--- a/src/ai-bundle/src/AiBundle.php
+++ b/src/ai-bundle/src/AiBundle.php
@@ -75,6 +75,7 @@ use Symfony\AI\Platform\Bridge\Ollama\OllamaApiCatalog;
 use Symfony\AI\Platform\Bridge\Ollama\PlatformFactory as OllamaPlatformFactory;
 use Symfony\AI\Platform\Bridge\OpenAi\PlatformFactory as OpenAiPlatformFactory;
 use Symfony\AI\Platform\Bridge\OpenRouter\PlatformFactory as OpenRouterPlatformFactory;
+use Symfony\AI\Platform\Bridge\Ovh\PlatformFactory as OvhPlatformFactory;
 use Symfony\AI\Platform\Bridge\Perplexity\PlatformFactory as PerplexityPlatformFactory;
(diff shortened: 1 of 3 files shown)


OskarStark:
Nice catch, could you please add a testcase? Thanks

src/ai-bundle/src/AiBundle.php
```diff
@@ -1127,6 +1128,30 @@ private function processPlatformConfig(string $type, array $platform, ContainerB
...
+                    new Reference('http_client', ContainerInterface::NULL_ON_INVALID_REFERENCE),
+                    new Reference('ai.platform.model_catalog.ovh'),
+                    null,
+                    new Reference('event_dispatcher'),
+                ])
+                ->addTag('ai.platform');
+
+            $container->setDefinition($platformId, $definition);
+
+            return;
```
Guikingone, line 1152: Don't forget to handle the situation where a `ScopingHttpClient` is used (see `Ollama` for more informations).

loicsapone: The endpoint is not configurable for OVH (like Scaleway or OpenRouter). ScopedClient is currently only supported for providers with a configurable endpoint.
Shouldn't this be standardized in a separate PR?

Guikingone: The usage of `ScopingHttpClient` is not related to configurable endpoint but to the fact that some developers prefer to use scoped http client than adding configuration in `ai.yaml` but yes, we can handle this situation in a separate PR 🙂

OskarStark:
Please add a line to the AI bundle's changelog, let's merge it as a feature for visibility. Afterwards good to merge. Thanks

src/ai-bundle/src/AiBundle.php
loicsapone: I'll try to do it this weekend 😉

loicsapone:
Changelog updated @OskarStark 🙂

OskarStark:
Thank you @loicsapone.

PR #1655, opened 2026-02-20
[Docs] Review structure

OskarStark:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | yes
| Issues        | --
| License       | MIT

### DIFF
--- a/docs/components/agent/tools.rst
+++ b/docs/components/agent/tools.rst
@@ -1,0 +1,201 @@
+Built-in Tools
+==============
+
+The Agent component ships with a collection of ready-made tool bridges that can be added to any agent.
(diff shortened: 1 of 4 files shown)


docs/components/agent/tools.rst
chr-hertel, line 1: I think it would be great to list the tools with signature and example values - so users understand what they get, e.g. Wikipedia brings two tools:
1. `wikipedia_search(string $query)` returns a list of 10 articles
2. `wikipedia_article(string $title)` returns the text content of an article

and maybe if the tools supports exposing sources.

OskarStark: Did this before, wanted to avoid leaking too much code, but will provide it again, to discuss

chr-hertel:
Thank you @OskarStark.

PR #1660, opened 2026-02-21
[MCP Bundle][Docs] Add discovery configuration for php-sdk example in docs

alenpokos:
Added configuration example for discovery in mcp-bundle. Default behaviour scans entire codebase which is no fun on a bigger project.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | yes
| Issues        | No
| License       | MIT

A simple documentation example update to provide, from my view, a really important configuration to stop mcp sdk from scaning entire codebase without need.

### DIFF
--- a/docs/bundles/mcp-bundle.rst
+++ b/docs/bundles/mcp-bundle.rst
@@ -202,6 +202,10 @@ Configuration
 
             Example contexts: logging, debugging, time-sensitive operations.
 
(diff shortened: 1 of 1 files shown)


chr-hertel:
Thanks @alenpokos, that's a very valid point - i think that shouldn't be the default behavior after all, but we should use the container for that - would provide caching out of the box.

chr-hertel:
Thank you @alenpokos.

PR #1667, opened 2026-02-23
[Store][Vektor] Introduce `StoreFactory`

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | Follow-up of #1633 
| License       | MIT

### DIFF
--- a/src/store/src/Bridge/Vektor/StoreFactory.php
+++ b/src/store/src/Bridge/Vektor/StoreFactory.php
@@ -1,0 +1,30 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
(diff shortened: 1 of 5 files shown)


src/store/src/Bridge/Vektor/StoreFactory.php
```diff
@@ -0,0 +1,30 @@
...
+namespace Symfony\AI\Store\Bridge\Vektor;
+
+use Symfony\AI\Store\ManagedStoreInterface;
+use Symfony\AI\Store\StoreInterface;
+use Symfony\Component\Filesystem\Filesystem;
+
+/**
+ * @author Guillaume Loulier <personal@guillaumeloulier.fr>
+ */
+final class StoreFactory
```
OskarStark, line 21: Why do we need this factory?

Guikingone: The idea is to standardize the creation of stores using factories (like Qdrant) but as this one doesn't require HttpClient, probably not need (much a helper than a requirement)

PR #1670, opened 2026-02-23
[Platform][Anthropic] Fix AssistantMessageNormalizer producing null content for Anthropic API

fabpot:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Docs?         | no 
| Issues        | n/a
| License       | MIT

When an AssistantMessage has no content, no tool calls, and no thinking, the normalizer returns "content": null in the payload. The Anthropic API rejects this with: messages.X.content: Input should be a valid list.

This can happen when the AssistantMessage is constructed with null content, which is a valid state per its constructor signature (?string $content = null).


chr-hertel:
Thank you @fabpot.

PR #1673, opened 2026-02-24
[CI] Auto-discover store bridges with integration tests in CI matrix

OskarStark:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | --
| License       | MIT

Instead of maintaining a hardcoded list of store bridges in the `store-bridges-integration` job, dynamically detect which bridges have integration tests by scanning for the `#[Group('integration')]` attribute. This way a new bridge with integration tests is automatically picked up without needing to update the workflow file.

@chr-hertel I need to check if real all of them got caught, lets run the CI, check and only merge if that's the case ⚠️

### DIFF
--- a/.github/workflows/build-matrix.yaml
+++ b/.github/workflows/build-matrix.yaml
@@ -15,6 +15,9 @@ on:
             store-bridges:
                 description: 'List of store bridges'
                 value: ${{ jobs.matrix.outputs.store-bridges }}
+            store-bridges-integration:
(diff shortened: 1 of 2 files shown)


chr-hertel (approved):
Nailed it 👍

chr-hertel:
Thank you @OskarStark.

PR #1675, opened 2026-02-24
 Fix ai profiler page

### LINKED ISSUE #1674: DataCollector not visible in Web Profiler
Im currently not seeing the symfony ai tab in the profiler. As im not really understanding the inner works here, i got this description from claude.

## Description
The `ai.data_collector` service is registered with the `data_collector` tag but without an `id` attribute:                                                                        

  ```php
  $builder->setDefinition('ai.data_collector', ...)->addTag('data_collector');
```

  Symfony's ProfilerPass uses the id tag attribute as the template key. When absent, it falls
  back to the fully-qualified class name (Symfony\AI\AiBundle\Profiler\DataCollector). However,
  TemplateManager::getNames() later checks $profile->hasCollector($name), which compares against
  the value returned by DataCollector::getName() — which is 'ai'. The mismatch causes the panel
  to be silently excluded from the profiler UI.

  The data is collected and stored correctly, but the "Symfony AI" tab never appears in the
  Web Profiler sidebar.

  Steps to reproduce

  1. Install symfony/ai-bundle with symfony/web-profiler-bundle
  2. Make any request that triggers an AI platform call
  3. Open the Web Profiler for that request
  4. The "Symfony AI" panel is absent from the sidebar

  Expected behavior

  The "Symfony AI" panel appears in the Web Profiler sidebar and shows platform calls,
  tool calls, and message store activity.

  Actual behavior

  The panel is never rendered. The data is collected and stored in the profile, but
  TemplateManager::getNames() skips it because the template key
  (Symfony\AI\AiBundle\Profiler\DataCollector) does not match the collector name (ai).

  Fix

  Add id: 'ai' to the tag registration in AiBundle.php:

  ```php
  // Before
  ->addTag('data_collector')
```

  ```php
  // After
  ->addTag('data_collector', ['id' => 'ai'])
```
---

Claude made this fix and registered it in Kernel.php, and im seeing the panel again in the profiler

```php
//Kernel.php
//...
    protected function build(ContainerBuilder $container): void
    {
        $container->addCompilerPass(new FixAiProfilerTemplatePass());
    }
```

```php
<?php 
//FixAiProfilerTemplatePass.php

declare(strict_types=1);

namespace App\DependencyInjection\Compiler;

use Symfony\AI\AiBundle\Profiler\DataCollector;
use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
use Symfony\Component\DependencyInjection\ContainerBuilder;

/**
 * Workaround for Symfony AI bundle missing `id: ai` in its data_collector tag.
 * ProfilerPass uses the class name as the template key, but TemplateManager
 * needs the collector's getName() value ('ai') to match the profile's stored data.
 *
 * @see https://github.com/symfony/ai-bundle/issues
 */
final class FixAiProfilerTemplatePass implements CompilerPassInterface
{
    public function process(ContainerBuilder $container): void
    {
        if (!$container->hasParameter('data_collector.templates')) {
            return;
        }

        $templates = $container->getParameter('data_collector.templates');

        foreach ($templates as $serviceId => $args) {
            if ($args && ($args[0] ?? null) === DataCollector::class) {
                $templates[$serviceId] = ['ai', $args[1]];
            }
        }

        $container->setParameter('data_collector.templates', $templates);
    }
}
```

pkothmair:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Docs?         | no
| Issues        | Fix #1674
| License       | MIT

Fixed the profiler page

### DIFF
--- a/src/ai-bundle/config/services.php
+++ b/src/ai-bundle/config/services.php
@@ -227,7 +227,7 @@ return static function (ContainerConfigurator $container): void {
                 tagged_iterator('ai.traceable_agent'),
                 tagged_iterator('ai.traceable_store'),
             ])
-            ->tag('data_collector')
+            ->tag('data_collector', ['id' => 'ai'])
 
         // serializer
         ->set('ai.chat.message_bag.normalizer', MessageNormalizer::class)
--- a/src/ai-bundle/tests/DependencyInjection/AiBundleTest.php
+++ b/src/ai-bundle/tests/DependencyInjection/AiBundleTest.php
@@ -143,6 +143,14 @@ class AiBundleTest extends TestCase
         }
     }
 
+    public function testDataCollectorTagIncludesId()
+    {
+        $container = $this->buildContainer($this->getFullConfig());
+        $definition = $container->getDefinition('ai.data_collector');
+        $this->assertTrue($definition->hasTag('data_collector'));
+        $this->assertSame([['id' => 'ai']], $definition->getTag('data_collector'));
+    }
+
     public function testStoreCommandsArentDefinedWithoutStore()
     {
         $container = $this->buildContainer([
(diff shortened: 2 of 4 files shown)


chr-hertel:
Thank you @pkothmair.

PR #1683, opened 2026-02-24
[Platform] Make JSON schema generation extensible (v2)

### LINKED ISSUE #1534: [Platform] Make json schema generation extendable

valtzu:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | todo
| Issues        | Fix #1534
| License       | MIT

An alternative/evolution of #1568. This version supports "normal" symfony serializer stuff, meaning f.e. getter, setters & constructor injection.

```mermaid
classDiagram
    %% 1. Entities & Packages (Represented as Classes for compatibility)
    class TypeInfo_Pkg["Package: symfony/type-info"]
    class PropertyInfo_Pkg["Package: symfony/property-info"]
    class Serializer_Pkg["Package: symfony/serializer"]
    
    class Factory
    class Model {
        +string name
        +reflector
    }
    class Property {
        +string name
        +reflector
    }

    class ModelDescriberInterface {
        <<interface>>
        +describeModel()
    }
    class PropertyDescriberInterface {
        <<interface>>
        +describeProperty()
    }

    class Describer
    class MethodDescriber
    class PropertyInfoDescriber
    class SerializerDescriber
    class TypeInfoDescriber
    class WithAttributeDescriber

    %% 2. Interface Realizations
    ModelDescriberInterface <|.. Describer
    PropertyDescriberInterface <|.. Describer
    ModelDescriberInterface <|.. MethodDescriber
    PropertyDescriberInterface <|.. MethodDescriber
    ModelDescriberInterface <|.. PropertyInfoDescriber
    PropertyDescriberInterface <|.. PropertyInfoDescriber
    ModelDescriberInterface <|.. SerializerDescriber
    ModelDescriberInterface <|.. TypeInfoDescriber
    PropertyDescriberInterface <|.. TypeInfoDescriber
    PropertyDescriberInterface <|.. WithAttributeDescriber

    %% 3. Internal Compositions
    Describer *-- MethodDescriber
    Describer *-- PropertyInfoDescriber
    Describer *-- SerializerDescriber
    Describer *-- TypeInfoDescriber
    Describer *-- WithAttributeDescriber

    %% 4. Orchestration & Usage
    Factory *-- Describer
    Describer ..> Model : uses
    Describer ..> Property : uses

    %% 5. External Mappings
    TypeInfoDescriber --> TypeInfo_Pkg
    SerializerDescriber --> Serializer_Pkg
    PropertyInfoDescriber --> PropertyInfo_Pkg
```

---

I created a new separate PR instead of pushing to the old one in case this is too much.


src/platform/src/Contract/JsonSchema/Model.php
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Platform\Contract\JsonSchema;
+
+final class Model
```
chr-hertel: that naming messed with my mind a bit to be honest 😆 at least for me it would be less confusing with switching to `DataClass` or something different

valtzu: I agree it can be confusing, but it represents a json schema model (or "something to convert into Json Schema Object"). Would `DataClass` be confusing as it represents tool arguments too? Maybe `SchemaSubject`/`SchemaObject` ?

The basic idea here was 
- describe `Model` -> output names of properties + allow editing schema (in case you e.g. wanted to add description of the model)
- describe `Property` -> describe single property of a `Model`

chr-hertel: ah, understood, let's go with `SchemaObject` then

valtzu: Oops I missed your comment :face_with_peeking_eye: :eyes: 

I changed
- `Model` -> `Subject\ObjectSubject`
- `Property` -> `Subject\PropertySubject`
- `ModelDescriberInterface` -> `ObjectDescriberInterface`
- `describeModel` -> `describeObject`
- `$model` parameter to `$subject` in both `ModelDescriberInterface` and `PropertyDescriberInterface`
   - I think it makes sense more now: the `$subject` is described to `&$schema`. I tried that `SchemaObject` but it looked confusing as you could think that you're supposed to write a schema into it. :thinking: 

WDYT?

chr-hertel: i like the consistency with property, yup :+1:

valtzu:
I'm wondering could/should we remove [`StructuredOutput/Serializer`](https://github.com/symfony/ai/blob/6962d6d3026f5f955f4c5f72a0288b99e297b1b2/src/platform/src/StructuredOutput/Serializer.php) now :thinking:

chr-hertel:
> I'm wondering could/should we remove [`StructuredOutput/Serializer`](https://github.com/symfony/ai/blob/6962d6d3026f5f955f4c5f72a0288b99e297b1b2/src/platform/src/StructuredOutput/Serializer.php) now 🤔

there is no immediate interference when merging this one, right?

valtzu:
Nothing should break – not sure if getter/setter would work with structured output though, before wiring the proper serializer.

chr-hertel:
> Nothing should break – not sure if getter/setter would work with structured output though, before wiring the proper serializer.

Let's go with a follow up then, and I'll merge this one! Thanks @valtzu!

PR #1685, opened 2026-02-25
[Platform] Option to mark models as deprecated

alex-ception:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | 
| License       | MIT

Models provided by platforms quickly evolve (renamed/removed, and so on).
We should be able to inform end developers about model being discontinued somehow.

I suggest adding in catalog the following :
```PHP
'test-model' => [
    'class' => Model::class,
    'capabilities' => [Capability::INPUT_TEXT],
    'deprecated' => true,
    // will trigger: "test-model" is deprecated.
],
'gemini-2.0-flash' => [
    'class' => Model::class,
    'capabilities' => [Capability::INPUT_TEXT],
    'deprecated' => 'symfony/ai-gemini-platform::0.5',
    // Will trigger: Since symfony/ai-gemini-platform 0.5: "gemini-2.0-flash" is deprecated.
],
'test-model-no-trigger' => [
    'class' => Model::class,
    'capabilities' => [Capability::INPUT_TEXT],
    'deprecated',
    // Will NOT trigger.
],

```

### DIFF
--- a/src/platform/src/ModelCatalog/AbstractModelCatalog.php
+++ b/src/platform/src/ModelCatalog/AbstractModelCatalog.php
@@ -55,7 +55,7 @@ abstract class AbstractModelCatalog implements ModelCatalogInterface
 
         if (isset($modelConfig['deprecated'])) {
             $deprecation = \is_string($modelConfig['deprecated']) ? $modelConfig['deprecated'] : '';
-            $r = preg_match('#(?<package>[^:]*)::(?<version>.*)#', $deprecation, $packageDestructuration);
+            preg_match('#(?<package>[^:]*)::(?<version>.*)#', $deprecation, $packageDestructuration);
             $package = $packageDestructuration['package'] ?? '';
             $version = $packageDestructuration['version'] ?? '';
 


src/platform/composer.json
```diff
@@ -55,6 +55,7 @@
         "phpstan/phpdoc-parser": "^2.1",
         "psr/log": "^3.0",
         "symfony/clock": "^7.3|^8.0",
+        "symfony/deprecation-contracts": "^3.6",
```
OskarStark: I think we can also support the 2.x version here, can we?

src/platform/src/ModelCatalog/AbstractModelCatalog.php
```diff
@@ -38,7 +38,7 @@ public function getModel(string $modelName): Model
         $options = $parsed['options'];
 
         if (!isset($this->models[$catalogKey])) {
-            throw new ModelNotFoundException(\sprintf('Model "%s" not found in %s.', $actualModelName, static::class));
+            throw new ModelNotFoundException(\sprintf('Model "%s" not found in "%s".', $actualModelName, static::class));
```
OskarStark: Please revert

src/platform/src/ModelCatalog/AbstractModelCatalog.php
```diff
@@ -53,6 +53,16 @@ public function getModel(string $modelName): Model
             throw new InvalidArgumentException(\sprintf('Model class "%s" must extend "%s".', $modelClass, Model::class));
         }
 
+        if (true === isset($modelConfig['deprecated'])) {
+            $deprecationMessage = \sprintf('Using "%s" is deprecated.', $catalogKey);
+
+            if (true === \is_string($modelConfig['deprecated'])) {
```
OskarStark: I would only allow bool here, it does not add any value I think

src/platform/src/ModelCatalog/AbstractModelCatalog.php
```diff
@@ -53,6 +53,16 @@ public function getModel(string $modelName): Model
...
+                $deprecationMessage .= ' '.$modelConfig['deprecated'];
+            }
+
+            trigger_deprecation('', '', $deprecationMessage);
```
OskarStark: Could we somehow get the info about the package and the version here?

src/platform/src/Test/ModelCatalogTestCase.php
```diff
@@ -36,8 +36,12 @@ abstract public static function modelsProvider(): iterable;
      * @param list<Capability>    $expectedCapabilities
      */
     #[DataProvider('modelsProvider')]
-    public function testGetModel(string $modelName, string $expectedClass, array $expectedCapabilities)
+    public function testGetModel(string $modelName, string $expectedClass, array $expectedCapabilities, bool $isDeprecated = false)
     {
+        if (true === $isDeprecated) {
```
OskarStark:
```suggestion
        if ($isDeprecated) {
```

src/platform/src/Test/ModelCatalogTestCase.php
```diff
@@ -36,8 +36,12 @@ abstract public static function modelsProvider(): iterable;
      * @param list<Capability>    $expectedCapabilities
      */
     #[DataProvider('modelsProvider')]
-    public function testGetModel(string $modelName, string $expectedClass, array $expectedCapabilities)
+    public function testGetModel(string $modelName, string $expectedClass, array $expectedCapabilities, bool $isDeprecated = false)
```
OskarStark:
```suggestion
    public function testGetModel(string $modelName, string $expectedClass, array $expectedCapabilities, bool $deprecated = false)
```

src/platform/src/ModelCatalog/AbstractModelCatalog.php
OskarStark: Ah or we could leverage the string with "package::version" to route this information

src/platform/src/ModelCatalog/AbstractModelCatalog.php
alex-ception: I agree but could not find how ?
Any idea ? From the class perspective, we only have access to the bounded namespace triggering this (by solving some reflection and stuff with static ?).

I first thought about hard coding "symfony/ai-platform" and version to "0.5" but It would imply that we do a hard work on every platform bridges.

In an ideal point of view it would be the package name of the platform bridge involved.
Besides, the version is not the real trigger here, it's more part of the lifecycle of the platform, not the version of the package.

If really, I'm opened to any suggestion that can help me through that :)

src/platform/src/ModelCatalog/AbstractModelCatalog.php
alex-ception: So you suggest, for example (would be a real example of Gemini 2.0 flash model shutting down in june 2026) :
`'deprecated' => 'symfony/ai-gemini-platform::0.5'` ? That would solve this : https://github.com/symfony/ai/pull/1685#discussion_r2857573655

src/platform/src/ModelCatalog/AbstractModelCatalog.php
alex-ception: I remember now why I did that :D https://github.com/symfony/ai/actions/runs/22435033581/job/64962572684?pr=1685

src/platform/src/ModelCatalog/AbstractModelCatalog.php
```diff
@@ -53,6 +53,15 @@ public function getModel(string $modelName): Model
             throw new InvalidArgumentException(\sprintf('Model class "%s" must extend "%s".', $modelClass, Model::class));
         }
 
+        if (isset($modelConfig['deprecated'])) {
+            $deprecation = \is_string($modelConfig['deprecated']) ? $modelConfig['deprecated'] : '';
+            $r = preg_match('#(?<package>[^:]*)::(?<version>.*)#', $deprecation, $packageDestructuration);
```
alex-ception: keep $matches to be more human readable ?

src/platform/src/ModelCatalog/AbstractModelCatalog.php
OskarStark: Yes, as an idea

WDYT @chr-hertel ?

chr-hertel: i don't think we should do this. but the issue is bigger than this PR. we need to find a way to decouple the change frequency of model catalogs and the library versions.

Opened #1692 for that ...

alex-ception:
see : https://github.com/symfony/models-dev

PR #1690, opened 2026-02-26
[Platform] Fix OpenAI tool call serialization

fabpot:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Docs?         | no 
| Issues        | 
| License       | MIT

When a tool has no required parameters and is called with no arguments, the empty PHP array was serialized as `[]` instead of `{}`. Also ensure assistant messages always include a `content` key (`null` when there is no text) as required by the OpenAI specification for messages with `tool_calls`.


chr-hertel:
Thank you @fabpot.

PR #1691, opened 2026-02-26
 Fix scientific notation strings being cast to int instead of float in model option parsing

fabpot:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Docs?         | no 
| Issues        | n/a
| License       | MIT


chr-hertel:
Thank you @fabpot.

PR #1704, opened 2026-02-28
[Store] Add platform-level reranking and hybrid retriever

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | -
| License       | MIT

## Description

This PR adds reranking support to the **Platform** abstraction layer and a hybrid retriever to the **Store** component, enabling more accurate document retrieval in RAG pipelines.

### Platform: Reranking Abstraction

Reranking is now a first-class platform capability, following the same pattern as embeddings (`Vectorizer → PlatformInterface → VectorResult`):

- `RerankingResult` / `RerankingEntry` — typed result objects for reranking scores
- `Capability::RERANKING` — new capability enum case
- `DeferredResult::asRerankingResult()` — convenience accessor
- **HuggingFace bridge** — `text-ranking` task support in `ModelClient` and `ResultConverter`, compatible with both the serverless Inference API and self-hosted TEI

### Store: Platform-based Reranker

A single `Reranker` replaces the previous vendor-specific bridge implementations (Cohere, Jina, HuggingFace). It delegates to `PlatformInterface`, so it works with **any provider** that supports reranking:

```php
use Symfony\AI\Store\Reranker\Reranker;

$reranker = new Reranker($platform, 'BAAI/bge-reranker-base?task=text-ranking');
$reranked = $reranker->rerank('query', $documents, topK: 5);
```

### Store: HybridRetriever

A new `HybridRetriever` combines results from two retrievers (e.g. vector and full-text) using Reciprocal Rank Fusion (RRF), with optional cross-encoder reranking:

```php
use Symfony\AI\Store\HybridRetriever;
use Symfony\AI\Store\Reranker\Reranker;

$retriever = new HybridRetriever(
    vectorRetriever: $vectorRetriever,
    textRetriever: $textRetriever,
    reranker: new Reranker($platform, 'BAAI/bge-reranker-base?task=text-ranking'),
    rrfK: 60,
    candidateCount: 20,
    topK: 5,
);

$results = $retriever->retrieve('search query');
```

### Store: Vectorizer text preservation

The `Vectorizer` now automatically stores the document content as `_text` in metadata during vectorization, so text search and reranking have access to the original content without requiring manual metadata setup.

### DIFF
--- a/src/platform/src/Bridge/HuggingFace/CHANGELOG.md
+++ b/src/platform/src/Bridge/HuggingFace/CHANGELOG.md
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+0.7
+---
+
+ * Add text-ranking task support for cross-encoder reranking models
+
 0.1
 ---
 
--- a/src/platform/src/Bridge/HuggingFace/Output/ZeroShotClassificationResult.php
+++ b/src/platform/src/Bridge/HuggingFace/Output/ZeroShotClassificationResult.php
@@ -28,10 +28,18 @@ final class ZeroShotClassificationResult
     }
 
     /**
-     * @param array{labels: array<string>, scores: array<float>, sequence?: string} $data
+     * @param array{labels: array<string>, scores: array<float>, sequence?: string}|list<array{label: string, score: float}> $data
      */
     public static function fromArray(array $data): self
     {
+        // Serverless format: [{label: "refund", score: 0.97}, ...]
+        if (isset($data[0]['label'])) {
+            return new self(
+                array_map(static fn (array $item): string => $item['label'], $data),
+                array_map(static fn (array $item): float => $item['score'], $data),
+            );
+        }
+
         return new self(
             $data['labels'],
             $data['scores'],
--- a/src/platform/src/Reranking/RerankingEntry.php
+++ b/src/platform/src/Reranking/RerankingEntry.php
@@ -1,0 +1,34 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Platform\Reranking;
+
+/**
+ * @author Johannes Wachter <johannes@sulu.io>
+ */
+final class RerankingEntry
+{
+    public function __construct(
+        private readonly int $index,
+        private readonly float $score,
+    ) {
+    }
+
+    public function getIndex(): int
+    {
+        return $this->index;
+    }
+
+    public function getScore(): float
+    {
+        return $this->score;
+    }
+}
--- a/src/platform/src/Result/DeferredResult.php
+++ b/src/platform/src/Result/DeferredResult.php
@@ -14,6 +14,7 @@ namespace Symfony\AI\Platform\Result;
 use Symfony\AI\Platform\Exception\ExceptionInterface;
 use Symfony\AI\Platform\Exception\UnexpectedResultTypeException;
 use Symfony\AI\Platform\Metadata\MetadataAwareTrait;
+use Symfony\AI\Platform\Reranking\RerankingEntry;
 use Symfony\AI\Platform\ResultConverterInterface;
 use Symfony\AI\Platform\TokenUsage\StreamListener;
 use Symfony\AI\Platform\TokenUsage\TokenUsage;
@@ -128,6 +129,16 @@ final class DeferredResult
     public function asVectors(): array
     {
         return $this->as(VectorResult::class)->getContent();
+    }
+
+    /**
+     * @return list<RerankingEntry>
+     *
+     * @throws ExceptionInterface
+     */
+    public function asReranking(): array
+    {
+        return $this->as(RerankingResult::class)->getContent();
     }
 
     /**
--- a/src/store/src/Document/Vectorizer.php
+++ b/src/store/src/Document/Vectorizer.php
@@ -109,7 +109,14 @@ final class Vectorizer implements VectorizerInterface
             throw new RuntimeException('No vector returned for vectorization.');
         }
 
-        return new VectorDocument($document->getId(), $vectors[0], $document->getMetadata());
+        // Preserve the original text in metadata so downstream consumers
+        // (e.g. text search, reranking) can access it via Metadata::getText().
+        $metadata = $document->getMetadata();
+        if (!$metadata->hasText()) {
+            $metadata->setText($document->getContent());
+        }
+
+        return new VectorDocument($document->getId(), $vectors[0], $metadata);
     }
 
     /**
@@ -186,7 +193,14 @@ final class Vectorizer implements VectorizerInterface
 
         $vectorDocuments = [];
         foreach ($documents as $i => $document) {
-            $vectorDocuments[] = new VectorDocument($document->getId(), $vectors[$i], $document->getMetadata());
+            // Preserve the original text in metadata so downstream consumers
+            // (e.g. text search, reranking) can access it via Metadata::getText().
+            $metadata = $document->getMetadata();
+            if (!$metadata->hasText()) {
+                $metadata->setText($document->getContent());
+            }
+
+            $vectorDocuments[] = new VectorDocument($document->getId(), $vectors[$i], $metadata);
         }
 
         $this->logger->info('Vectorization process completed', [
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,3 +1,18 @@
+UPGRADE FROM 0.6 to 0.7
+=======================
+
+Store
+-----
+
+ * The `Retriever` constructor has a new `?EventDispatcherInterface $eventDispatcher` parameter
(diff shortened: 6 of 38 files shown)


src/store/src/Bridge/HuggingFace/Reranker.php
wachterjohannes, line 1: @chr-hertel @OskarStark the implementations of the Reranker are at the absolut wrong place - where would you place it? in the platform? maybe with a Platform Client or a different approach?

chr-hertel: yea, i see where you're coming from. `RerankerInterface` implementations def belong to the Store, bringing in `reranking` as a capability in platform and abstract that a bit away for the implementation here with a dependency to `PlatformInterface` would be nice

wachterjohannes:
@chr-hertel i also fixed the examples for HF - a few are not possible to fix and i wanted to spend to much time about it :)

wachterjohannes:
@chr-hertel in my opinion this is ready :)

examples/huggingface/audio-classification.php
chr-hertel, line 1: that's alright - not all HF examples work all the time because models are not always warm - we don't have to kill the example for that right now. please revert

examples/huggingface/text-generation.php
chr-hertel, line 1: there's already another example for chat completion tho

src/platform/src/Bridge/HuggingFace/Output/ZeroShotClassificationResult.php
```diff
@@ -28,10 +28,18 @@ public function __construct(
     }
 
     /**
-     * @param array{labels: array<string>, scores: array<float>, sequence?: string} $data
+     * @param array{labels: array<string>, scores: array<float>, sequence?: string}|list<array{label: string, score: float}> $data
      */
     public static function fromArray(array $data): self
     {
+        // New HF serverless format: [{label: "refund", score: 0.97}, ...]
```
chr-hertel: Won't be "new" for long
```suggestion
        // Serverless format: [{label: "refund", score: 0.97}, ...]
```

src/store/src/HybridRetriever.php
chr-hertel, line 1: can we somehow encapsulate the search strategy (vector vs. text vs. hybrid) within the store and only have one consolidated retriever for all cases?

if you consider this target picture, we would have to implement the same extension points in all Retriever implementations:
Image

and you implemented a dynamic query logic already before:
https://github.com/symfony/ai/blob/64a8bd7029e577b5908cc96e290c988e430c15c1/src/store/src/Retriever.php#L62-L85

examples/huggingface/text-generation.php
wachterjohannes: should i remove it?

src/platform/src/Bridge/HuggingFace/Output/ZeroShotClassificationResult.php
wachterjohannes: you are absolut right 😂

src/store/src/HybridRetriever.php
wachterjohannes: @chr-hertel I had a look at what it would take. The tricky part is that HybridRetriever does something quite different from what Retriever does currently — it doesn't use the store's native hybrid search at all. Instead it runs two separate retrievals (vector + text), merges them client-side with RRF, and optionally reranks with a cross-encoder. That's the main value for stores that don't support HybridQuery natively.

I could see a few ways to bring this into Retriever:

1. Add an optional RerankerInterface + some kind of strategy concept (e.g. "native" vs "RRF") to Retriever
2. Or keep it as a decorator — HybridRetriever wraps two retrievers, which feels pretty natural composition-wise and keeps each class focused

We could also potentially model this through HybridQuery somehow — like having Retriever detect that the store doesn't support it natively, and then fall back to the dual-retrieval + RRF approach itself. But either way we'd end up introducing some kind of strategy concept into Retriever, which makes its constructor and logic noticeably more complex.

I'm slightly leaning toward keeping them separate since merging would make Retriever's constructor and logic noticeably more complex, but I can totally see the argument for having a single entry point. What's your preference?

examples/huggingface/text-generation.php
chr-hertel: you can just revert the changes here - with hyperbolic i get another error message or does it work for you?

wachterjohannes: at least last week it worked 😂 i will revert it

examples/huggingface/chat-completion.php
chr-hertel, line 1: please revert this provider switch

examples/huggingface/text-generation.php
```diff
@@ -18,6 +19,7 @@
 
 $result = $platform->invoke('katanemo/Arch-Router-1.5B', 'The quick brown fox jumps over the lazy', [
     'task' => Task::TEXT_GENERATION,
+    'provider' => Provider::HYPERBOLIC,
```
chr-hertel: please remove this provider switch

src/platform/src/Result/DeferredResult.php
```diff
@@ -130,6 +131,16 @@ public function asVectors(): array
         return $this->as(VectorResult::class)->getContent();
     }
 
+    /**
+     * @return list<RerankingEntry>
+     *
+     * @throws ExceptionInterface
+     */
+    public function asRerankingResult(): array
```
chr-hertel:
```suggestion
    public function asReranking(): array
```

src/platform/src/Reranking/RerankingEntry.php
```diff
@@ -0,0 +1,24 @@
...
+namespace Symfony\AI\Platform\Reranking;
+
+/**
+ * @author Johannes Wachter <johannes@sulu.io>
+ */
+final class RerankingEntry
+{
+    public function __construct(
+        public readonly int $index,
+        public readonly float $score,
```
chr-hertel: please `private readonly` + getters

src/platform/src/Bridge/HuggingFace/CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+0.6
```
chr-hertel:
```suggestion
0.7
```

src/store/src/Event/PostRetrievalEvent.php
```diff
@@ -0,0 +1,65 @@
...
+use Symfony\Contracts\EventDispatcher\Event;
+
+/**
+ * Event dispatched after documents are retrieved from the store.
+ *
+ * Listeners can modify the documents list, for example to rerank results.
+ *
+ * @author Johannes Wachter <johannes@sulu.io>
+ */
+final class PostRetrievalEvent extends Event
```
chr-hertel: would prefer `PostQueryEvent` (+ `PreQueryEvent` later)

chr-hertel: cc @OskarStark opinion here?

chr-hertel (changes requested):

src/store/src/Event/PostRetrievalEvent.php
wachterjohannes: 😂 https://github.com/symfony/ai/pull/1745

wachterjohannes:
@chr-hertel i just have implemented your review - if @OskarStark means that the old name is better i can revert that! but all of your feedback is implemented now

chr-hertel (approved):

chr-hertel:
Needs a rebase, good to go in afterwards.

wachterjohannes:
@chr-hertel done :)

chr-hertel (approved):

chr-hertel:
We'll need to clean up the event dispatcher dependencies over all packages as a follow up - but that's larger than this PR

chr-hertel:
Thank you @wachterjohannes.

src/store/src/Document/Vectorizer.php
```diff
@@ -109,7 +109,14 @@ private function vectorizeEmbeddableDocument(EmbeddableDocumentInterface $docume
             throw new RuntimeException('No vector returned for vectorization.');
         }
 
-        return new VectorDocument($document->getId(), $vectors[0], $document->getMetadata());
+        // Preserve the original text in metadata so downstream consumers
+        // (e.g. text search, reranking) can access it via Metadata::getText().
+        $metadata = $document->getMetadata();
+        if (!$metadata->hasText()) {
+            $metadata->setText($document->getContent());
+        }
```
chr-hertel, line 117: @wachterjohannes turns out this is a hidden breaking change since it changes the expected schema for Azure Search implementation since it renders the metadata flat into the schema:

https://github.com/symfony/ai/blob/6ff20085eb9b08f853b0886575714aa4caa62118/src/store/src/Bridge/AzureSearch/SearchStore.php#L46-L49

wachterjohannes: @chr-hertel OK that sounds crazy - i think that should be fixed in the AzureSearch as this will break all the time something (Listener or Developer add a new metadata value) - what do you think?

PR #1708, opened 2026-02-28
[AI Bundle] Defer DataCollector to lateCollect

Spomky:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        |
| License       | MIT

`DataCollector::collect()` calls `$this->lateCollect()` directly, defeating the purpose of implementing `LateDataCollectorInterface`. This causes all tool and platform call resolution to happen during the HTTP response instead of after it.

By making `collect()` a no-op, the data collection is deferred to `kernel.terminate` (via `Profiler::saveProfile()`), no longer blocking the HTTP response.

### DIFF
--- a/src/ai-bundle/src/Profiler/DataCollector.php
+++ b/src/ai-bundle/src/Profiler/DataCollector.php
@@ -76,7 +76,7 @@ final class DataCollector extends AbstractDataCollector implements LateDataColle
 
     public function collect(Request $request, Response $response, ?\Throwable $exception = null): void
     {
-        $this->lateCollect();
+        // Deferred to lateCollect() to avoid blocking the response.
     }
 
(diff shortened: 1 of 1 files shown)


valtzu:
As I recall it was done for a reason, otherwise the calls dont show up in the toolbar? For streamed responses this is of course expected, but for non-streamed not

If that's the case, then maybe we can detect the `StreamedResponse` and only then skip early collect?

chr-hertel (approved):

chr-hertel:
@valtzu - you made the change with #173 for the stream example in the demo - it still works here :+1:

chr-hertel:
Thank you @Spomky.

PR #1710, opened 2026-03-01
[Platform][OpenAI] Add error handling for stream conversions

oesteve:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Docs?         | no 
| Issues        |  
| License       | MIT

When an error chunk type is received while using streams, it is ignored, resulting in an unexpected situation where something is not working, but the platform does not return any error. This PR provides handling for an error type chunk by throwing a RuntimeException.


chr-hertel:
Thank you @oesteve.

chr-hertel:
Replaced `DummyRawResult` with existing `InMemoryRawResult` while merging.

PR #1711, opened 2026-03-01
[Platform] Throw InvalidArgumentException when ModelClient receives a string payload

fabpot:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Docs?         | no 
| Issues        | n/a
| License       | MIT

While working on #1644, I realized that passing a string to `ModelClientInterface::request()` was allowed, but most ModelClient implementations use `array_merge($options, $payload)` aithout checking if `$payload` is a string.

### DIFF
--- a/src/platform/src/Bridge/Anthropic/ModelClient.php
+++ b/src/platform/src/Bridge/Anthropic/ModelClient.php
@@ -11,6 +11,7 @@
 
 namespace Symfony\AI\Platform\Bridge\Anthropic;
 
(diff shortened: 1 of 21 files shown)


chr-hertel:
Oh yes, that makes a whole lot of sense :)

fabpot:
Tests are green now

OskarStark (approved):

OskarStark:
Thank you @fabpot.

PR #1712, opened 2026-03-01
[Platform][ModelsDev] Skip specialized bridge check when a custom baseUrl is provided

fabpot:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Docs?         | no 
| Issues        | n/A
| License       | MIT

A custom `baseUrl` signals the user is pointing at an OpenAI-compatible proxy, so the generic bridge is always the right choice even for providers like `google-vertex` that would normally require a non-routable specialized bridge.


OskarStark:
Thank you Fabien.

PR #1713, opened 2026-03-01
[Platform] Add primary catalog delegation and embed-name heuristic to FallbackModelCatalog

fabpot:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Docs?         | no 
| Issues        | n/a
| License       | MIT

`FallbackModelCatalog` now accepts an optional primary `ModelCatalogInterface` (leveraged by the Modeldev bridge). When provided, the primary catalog is tried first; on `ModelNotFoundException`, the catalog falls back to a naming heuristic.

The fallback heuristic now returns `CompletionsModel` or `EmbeddingsModel` (based on whether the model name contains "embed") instead of a plain `Model`. This fixes compatibility with the `Generic` bridge, whose model clients require specific model types to route requests correctly.

### DIFF
--- a/src/platform/src/Bridge/ModelsDev/CHANGELOG.md
+++ b/src/platform/src/Bridge/ModelsDev/CHANGELOG.md
@@ -4,6 +4,7 @@ CHANGELOG
 0.6
 ---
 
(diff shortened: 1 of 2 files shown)


fabpot:
Wrong approach, I will open a simpler/better PR soon.

PR #1714, opened 2026-03-01
[Platform][Generic]  Fix FallbackModelCatalog

fabpot:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Docs?         | no 
| Issues        | n/a
| License       | MIT

Alternative to #1713

The Generic bridge's `Completions\ModelClient` and `Embeddings\ModelClient` use `instanceof CompletionsModel` and `instanceof EmbeddingsModel` to determine whether they support a given model. However, the shared `FallbackModelCatalog` creates plain `Model` instances, which neither `ModelClient` recognizes. This means that any usage of `GenericPlatformFactory::create()` without an explicit `ModelCatalog` is broken, every `invoke()` call fails with 'No ModelClient registered'.

Add a bridge-specific `FallbackModelCatalog` (following the pattern of `OpenResponses\FallbackModelCatalog`) that returns `CompletionsModel` or `EmbeddingsModel` based on a naming convention: model names containing 'embed' are treated as embedding models, everything else as completions models (the heuristic might not be perfect, but I didn't find a better way).


chr-hertel:
Thank you @fabpot.

chr-hertel:
Patched fabbot failure on merge.

PR #1715, opened 2026-03-02
[Platform][Generic] Add token usage handling in stream

### LINKED ISSUE #1700: [Platform] Generic token usage is empty when stream is true
There are two options: either I’m using this incorrectly, or there’s a bug in the code:

when I use (Lite LLM)
`Symfony\AI\Platform\Bridge\Generic\PlatformFactory;`

`
$resp = $platform->invoke('gpt-4o-mini', $messages, [
    'stream' => true, 
]);

dump($resp->getMetadata());
`
is empty array

genosajd:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | Fix #1700 
| License       | MIT

This PR adds support for token usage extraction during streaming responses in the Generic platform bridge.

### DIFF
--- a/examples/generic/stream-token-metadata.php
+++ b/examples/generic/stream-token-metadata.php
@@ -1,0 +1,60 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use Symfony\AI\Agent\Agent;
+use Symfony\AI\Platform\Bridge\Generic\CompletionsModel;
+use Symfony\AI\Platform\Bridge\Generic\ModelCatalog;
+use Symfony\AI\Platform\Bridge\Generic\PlatformFactory;
+use Symfony\AI\Platform\Capability;
+use Symfony\AI\Platform\Message\Message;
+use Symfony\AI\Platform\Message\MessageBag;
+
+require_once dirname(__DIR__).'/bootstrap.php';
+
+$modelCatalog = new ModelCatalog([
+    'mistral-small-latest' => [
+        'class' => CompletionsModel::class,
+        'capabilities' => [
+            Capability::INPUT_MESSAGES,
+            Capability::OUTPUT_TEXT,
+            Capability::OUTPUT_STREAMING,
+            Capability::OUTPUT_STRUCTURED,
+            Capability::INPUT_IMAGE,
+            Capability::TOOL_CALLING,
+        ],
+    ],
+]);
+
+$platform = PlatformFactory::create(
+    env('LITELLM_HOST_URL'),
+    env('LITELLM_API_KEY'),
+    http_client(),
+    $modelCatalog,
+);
+
+$agent = new Agent($platform, 'mistral-small-latest');
+$messages = new MessageBag(
+    Message::forSystem('You are a pirate and you write funny.'),
+    Message::ofUser('What is the Symfony framework?'),
+);
+$result = $agent->call($messages, [
+    'stream' => true,
+]);
+
+foreach ($result->getContent() as $textChunk) {
+    echo $textChunk;
+}
+
+echo \PHP_EOL;
+
+print_token_usage($result->getMetadata()->get('token_usage'));
+
+echo \PHP_EOL;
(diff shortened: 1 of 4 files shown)


chr-hertel (changes requested):
Thanks @genosajd for patching this - doesn't work for me right now tho - can you please add an example similar to this one:
https://github.com/symfony/ai/blob/main/examples/openai/stream-token-metadata.php

And also the generic bridge would need `CHANGELOG` entry for this new feature.

Edit: and code style issue is real - `php vendor/bin/php-cs-fixer fix` in root dir should fix that

genosajd:
> Thanks @genosajd for patching this - doesn't work for me right now tho - can you please add an example similar to this one: https://github.com/symfony/ai/blob/main/examples/openai/stream-token-metadata.php
> 
> And also the generic bridge would need `CHANGELOG` entry for this new feature.
> 
> Edit: and code style issue is real - `php vendor/bin/php-cs-fixer fix` in root dir should fix that

okay, should I add feature to 0.4 in changelog or create 0.5?

chr-hertel:
> > Thanks @genosajd for patching this - doesn't work for me right now tho - can you please add an example similar to this one: https://github.com/symfony/ai/blob/main/examples/openai/stream-token-metadata.php
> > And also the generic bridge would need `CHANGELOG` entry for this new feature.
> > Edit: and code style issue is real - `php vendor/bin/php-cs-fixer fix` in root dir should fix that
> 
> okay, should I add feature to 0.4 in changelog or create 0.5?

To 0.7 please

OskarStark:
Can you please fix the rebase? Thanks

OskarStark (approved):

chr-hertel:
Needs a rebase after #1756

chr-hertel:
Sorry @genosajd, needs to wait for #1759 as well - a lot going on here in parallel

genosajd:
> Sorry @genosajd, needs to wait for #1759 as well - a lot going on here in parallel

no problem :)

ineersa:
@genosajd @chr-hertel 
I think it needs some additional work.
While actual logic in TokenUsage conversion is correct (I used it in https://github.com/ineersa/re-search/blob/main/src/Platform/Generic/Completions/ResultConverter.php#L34) this won't work for generic platforms.

https://developers.openai.com/api/reference/resources/chat

> stream_options: Options for streaming response. Only set this when you set stream: true.
>include_usage: optional boolean -- If set, an additional chunk will be streamed before the data: [DONE] message. The usage field on this chunk shows the token usage statistics for the entire request, and the choices field will always be an empty array.

So for OpenAI compatible APIs you actually need to pass those options.

In theory you could pass them in `ModelCatalog` as model options or via DSN `modelname?stream_options[include_usage]=true`, but there is a caveat.

> Only set this when you set stream: true

So your non-streaming requests will fail with validation errors on API providers which do strict validations.

To fix this I pass it inside `ModelClient` and if it's `stream` set `stream_options`.

Maybe there is a better way, when we set options for request check for stream_options and set them only to stream?

chr-hertel (changes requested):
The example currently does not work for me in combination with litellm:

```bash
docker compose up -d litellm litellm-db
php generic/stream-token-metadata.php
```

results in 
Image

chr-hertel:

- `src/platform/src/Bridge/Generic/CHANGELOG.md`

Thank you!

---

genosajd:
> Hi! Version **0.7** has been released. This PR has changelog/upgrade entries under the `0.7` heading, which is now a released version.
> 
> Please update the following files to place your entries under the next version **0.8**:
> 
> * `src/platform/src/Bridge/Generic/CHANGELOG.md`
> 
> **For `CHANGELOG.md` files**: change the version heading from `0.7` to `0.8` **For `UPGRADE.md`**: place entries under `UPGRADE FROM 0.7 to 0.8`
> 
> Thank you!
> 
> 🤖 This comment was generated automatically by an agent.

Is this PR still necessary if you already implemented:

`if (isset($data['usage'])) {
                yield $this->convertStreamUsage($data['usage']);
            }`

ineersa:
Need to test but https://github.com/symfony/ai/pull/1715#issuecomment-4173420404 is still not there.

chr-hertel:
closed in favor of #2131 - thanks @genosajd and let us know if we're missing something here.

PR #1717, opened 2026-03-02
[Platform][ModelsDev] Resolve models.dev JSON from symfony/models-dev Composer package

fabpot:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Docs?         | no 
| Issues        | n/a
| License       | MIT

DataLoader now resolves the default data path from the `symfony/models-dev` Composer package.

Users can either:

* Pass an explicit JSON file path via `$dataPath`
* Install `symfony/models-dev`


chr-hertel:
Thank you @fabpot.

PR #1719, opened 2026-03-03
[Platform][Azure] Replace Chat Completions client with Responses API client

pkothmair:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| Docs?         | no
| Issues        |
| License       | MIT

## Summary

Azure OpenAI now supports the unified Responses API (`/openai/v1/responses`),
which is the same endpoint the standard OpenAI bridge already uses. This PR
migrates the Azure platform bridge to the Responses API and fixes PDF document
support along the way.

### Changes

**`src/platform/src/Bridge/Azure/`**

- **Remove** `OpenAi/CompletionsModelClient` (Chat Completions, `/chat/completions`)
- **Add** `Responses/ModelClient` — Azure-specific Responses API client that uses
  `api-key` header auth and targets `https://{baseUrl}/openai/v1/responses`
- **Update** `OpenAi/ModelCatalog` — map `Gpt` models to `ResponsesModel` instead
  of `CompletionsModel`, enabling file/PDF input and unified streaming behavior
- **Update** `OpenAi/PlatformFactory` — use the new `Responses\ModelClient` together
  with `OpenAiContract` (which registers `DocumentNormalizer`) and `OpenResponses\ResultConverter`

**`src/platform/src/Bridge/OpenAi/Contract/`**

- **Fix** `DocumentNormalizer::supportsModel()` — replace `$model instanceof Gpt`
  with `$model->supports(Capability::INPUT_PDF)` so the normalizer works for any
  model declaring PDF support, not just `Gpt` (which Azure remaps to `ResponsesModel`)
- **Fix** `OpenAiContract::create()` — register `DocumentNormalizer` so PDF documents
  are correctly serialized when sent via agents

**`deptrac.yaml`**

- Add `OpenResponsesPlatform` to `AzurePlatform` allowed dependencies

### Why

The `Contract` class builds its own isolated `Serializer` with a hardcoded normalizer
list — it does not pick up normalizers from the Symfony DI container. `DocumentNormalizer`
existed but was never registered in `OpenAiContract`, and its `instanceof Gpt` guard
broke on Azure where GPT models are represented as `ResponsesModel`.

The Azure bridge was the only remaining bridge still using Chat Completions. Migrating
to the Responses API aligns it with the OpenAI bridge and unlocks file input support.

## Test plan

- `Responses/ModelClientTest` — validates URL construction, `api-key` header auth,
  `ResponsesModel` support, and correct request format
- Removed `CompletionsModelClientTest` (client deleted)
- Existing `EmbeddingsModelClientTest` and `WhisperModelClientTest` unaffected
- Updated `DocumentNormalizerTest` to assert capability-based model check


.codespellrc
chr-hertel, line 1: Nice one, thanks! 👍

chr-hertel:
Thank you @pkothmair.

chr-hertel:
Fixed missing CHANGELOG entry while merging.

PR #1723, opened 2026-03-04
[Platform] Improve structured output test coverage

valtzu:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| License       | MIT

Related to comment https://github.com/symfony/ai/pull/1683#issuecomment-4000631686 I added a test that makes sure the getters/setters work with `StructuredOutput\Serializer`.

---

We probably should aim to use the normal `serializer` (or maybe add a child definition of it) because currently even basic things like `DateTime`s are not supported – let alone custom types with custom normalizers.
But the comment in `StructuredOutput\Serializer` still partly holds – so I'd say it's good enough for now / good for release.

### DIFF
--- a/src/platform/tests/Fixtures/StructuredOutput/UserWithAccessors.php
+++ b/src/platform/tests/Fixtures/StructuredOutput/UserWithAccessors.php
@@ -1,0 +1,74 @@
+<?php
+
+/*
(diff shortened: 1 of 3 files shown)


chr-hertel:
Thank you @valtzu.

PR #1729, opened 2026-03-05
[Platform][Bedrock] Add match statement to determine bedrock inference profile region

rrigby:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         |  no
| Issues        | N/A
| License       | MIT

The model client was operating under the assumption that all inference profiles use a 2 character region taken from the AWS region .e.g `ap-southeast-2 -> ap`. For US and EU regions this worked fine.

However, this is not the case for Asia Pacific regions and some others, see here for AWS documentation: https://docs.aws.amazon.com/bedrock/latest/userguide/inference-profiles-support.html

This is not the most elegant solution, maybe a trait or utility method somewhere should provide this functionality.


chr-hertel (changes requested):
Hey @rrigby, thanks for solving this issue!

two small things I need from you tho:
1. can you briefly explain the mapping, like `af => eu`
2. please extract into `RegionMapper::map(string $region): string` (+ tests)

Thanks already! :)

chr-hertel (approved):

chr-hertel:
Thank you @rrigby - addressed my comment while merging.

PR #1730, opened 2026-03-05
[MCP Bundle] Use Symfony Service Discovery

### LINKED ISSUE #1744: [MCP Bundle] Long response from Tool
I use **Symfony 7.4 and  "symfony/mcp-bundle": "^0.6.0"**

Using, MCP inspector or mcporter I get long response from tool, about 6-10 seconds.
Profiler shows that main problem is in **controller.get_callable 6177.9 ms / 22 MiB.**

Tool example:
```
<?php

namespace App\Mcp;

class HelloTool
{
    #[McpTool(name: 'hello')]
    public function hello(string $name = 'World'): string
    {
        return "Hello $name from Symfony MCP!";
    }
}
```
Switching to prod env does not resolve the problem. 

I have 4000+ services (php bin/console debug:container | wc -l), maybe bundle scan all services until find needed tool?

Symfony on Debian 13.3 (Docker)

EldoranDev:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | Fix #1744 
| License       | MIT

Hi this MR changes the Discovery of Tools/Resources etc. to use the Symfony Container instead of the FileBased Discovery of `mcp/sdk` following the suggestion from https://github.com/symfony/ai/pull/1660#issuecomment-3938496927

I've implemented this for McpTools right now, to check back with you if that would be an approach you'd agree with or if anything should be done differently.

If you agree with this solution I'd finish it for the other Types as well as adding tests (and update docs etc.).

Thanks for your feedback or suggestions


chr-hertel:
Hey @EldoranDev, thanks for working on this - much appreciated :+1: 

The SDK itself supports setting a container, and the `Mcp\Capability\Registry\ReferenceHandler` is capable of pulling services from that container. so we could potentially use the `Mcp\Capability\Registry\Loader\ArrayLoader` under the hood to avoid rebuilding how we parse tool implementations. or did you check that already? if there's an impediment to that, we should consider patching it in the SDK.

Cheers!

EldoranDev:
> Hey @EldoranDev, thanks for working on this - much appreciated 👍
> 
> The SDK itself supports setting a container, and the `Mcp\Capability\Registry\ReferenceHandler` is capable of pulling services from that container. so we could potentially use the `Mcp\Capability\Registry\Loader\ArrayLoader` under the hood to avoid rebuilding how we parse tool implementations. or did you check that already? if there's an impediment to that, we should consider patching it in the SDK.
> 
> Cheers!

For some reason I thought that the Icon loading for example would break due to serialisation issues and went for a custom loader because of that.

I've switched to Array Loader

chr-hertel:

- `src/platform/src/Bridge/Ollama/CHANGELOG.md`

Thank you!

---

PR #1732, opened 2026-03-05
[Platform][Azure] Pass deployment name to `ResponsesModelClient`

pkothmair:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        |
| License       | MIT

## Summary

`ResponsesModelClient` was not receiving the configured deployment name, so it always sent `$model->getName()` (e.g. `gpt-5`) as the `model` field in the Azure Responses API request body.

Azure OpenAI requires the **deployment name** as the `model` field — not the model family name — so any project where the Azure deployment is not named identically to the OpenAI model would receive a `404 DeploymentNotFound` error.

**Changes:**
- Add `$deployment` parameter to `ResponsesModelClient::__construct()`, placed before `$apiKey` to match the convention of `EmbeddingsModelClient` and `WhisperModelClient`
- Use `$this->deployment ?? $model->getName()` in the request, falling back to the model name when no deployment is configured
- Pass `$deployment` from `PlatformFactory::create()` to `ResponsesModelClient`
- Add tests covering deployment name override and null fallback

### DIFF
--- a/src/platform/src/Bridge/Azure/Responses/ModelClient.php
+++ b/src/platform/src/Bridge/Azure/Responses/ModelClient.php
@@ -31,6 +31,7 @@ final class ModelClient implements ModelClientInterface
         HttpClientInterface $httpClient,
         string $baseUrl,
         #[\SensitiveParameter] private readonly string $apiKey,
+        private readonly ?string $deployment = null,
     ) {
         if ('' === $baseUrl) {
             throw new InvalidArgumentException('The base URL must not be empty.');
@@ -57,7 +58,7 @@ final class ModelClient implements ModelClientInterface
             'headers' => [
                 'api-key' => $this->apiKey,
             ],
-            'json' => array_merge($options, ['model' => $model->getName()], $payload),
+            'json' => array_merge($options, ['model' => $this->deployment ?? $model->getName()], $payload),
         ]));
     }
 }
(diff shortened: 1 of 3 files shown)


pkothmair:
@chr-hertel the pr from yesterday works, but only if the deployment name is the same as the model name. 

This should fix cases, where the deployment name is not the model name.

in theory we could also now allow the `deployment` option in the `ai_azure_platform.yaml` to be optional, when a responses model is used and deployments are named after models. That way every responses model could be handled by the same platform in the config.

Let me know what you think 👍

chr-hertel:
> so any project where the Azure deployment is not named identically to the OpenAI model

so guess my setup :see_no_evil: :laughing:

src/platform/src/Bridge/Azure/Responses/ModelClient.php
```diff
@@ -30,6 +30,7 @@ final class ModelClient implements ModelClientInterface
     public function __construct(
         HttpClientInterface $httpClient,
         string $baseUrl,
+        private readonly ?string $deployment,
         #[\SensitiveParameter] private readonly string $apiKey,
```
chr-hertel: let's switch around please:

```suggestion
        #[\SensitiveParameter] private readonly string $apiKey,
        private readonly ?string $deployment = null,
```

pkothmair:
> so guess my setup 🙈 😆

@chr-hertel yes, mine too. only caught that because of some misconfiguration

OskarStark:
Thank you @pkothmair.

PR #1739, opened 2026-03-09
[Store][Pinecone] Support `includeMetadata` and `includeValues` options

MolloKhan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| Docs?         | yes
| Issues        |  N/A
| License       | MIT

The option `includeValues` is set to `false` by default by the Probots library, but the store sets it to `true` by default
And, there is no way to specify if you don't want to fetch metadata


OskarStark:
Thank you @MolloKhan.

OskarStark:
I fixed the CHANGELOG in #1752

PR #1742, opened 2026-03-10
[Platform][Cache] Fix `MessageBag` handling

### LINKED ISSUE #1734: `CachePlatform` input gets shared when it's a `MessageBag`
When `CachePlatform` is used to cache call output, the input part of the key is discarded when it is of type `MessageBag`. This means the result will be shared for all invocations that have the same `prompt_cache_key` and model. The bag needs to be normalized instead of assumed to be JSON encodable in https://github.com/symfony/ai/blob/9642a27db615c1e8b828f3c348ade83a0dc4d604/src/platform/src/Bridge/Cache/CachePlatform.php#L68

Reproduction code:

```php

$bag = new MessageBag(Message::ofUser('User prompt that gets discarded'));
var_dump(json_encode($bag)); // string(2) "{}¨
```

The MD5 of the input will therefore always be the same ("99914b932bd37a50b983c5e7c90ae93b" to be precise), regardless of input. This effectively shares the output with all other invocations of the model/agent and can be a severe security vulnerability.

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #1734 
| License       | MIT

### DIFF
--- a/src/platform/src/Bridge/Cache/CachePlatform.php
+++ b/src/platform/src/Bridge/Cache/CachePlatform.php
@@ -11,6 +11,8 @@
 
 namespace Symfony\AI\Platform\Bridge\Cache;
 
+use Symfony\AI\Platform\Exception\InvalidArgumentException;
+use Symfony\AI\Platform\Message\MessageBag;
 use Symfony\AI\Platform\ModelCatalog\ModelCatalogInterface;
 use Symfony\AI\Platform\PlainConverter;
 use Symfony\AI\Platform\PlatformInterface;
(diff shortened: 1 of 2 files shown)


OskarStark (approved):

OskarStark:
Thank you Guillaume.

src/platform/src/Bridge/Cache/CachePlatform.php
```diff
@@ -62,10 +64,17 @@ public function invoke(string $model, array|string|object $input, array $options
             return $this->platform->invoke($model, $input, $options);
         }
 
+        $normalizedInput = match (true) {
+            \is_string($input) => md5($input),
+            \is_array($input) => json_encode($input),
+            $input instanceof MessageBag => $input->getId()->toString(),
```
villermen, line 70: @Guikingone @OskarStark Sorry for not responding or suggesting a fix sooner. I've been keeping myself busy.

In your original comment https://github.com/symfony/ai/issues/1734#issuecomment-4012401207, you mention additional challenges in using normalizers. Is this what made you switch to using the `MessageBag` ID?

I don't think this is quite the right solution to the problem. IDs are freshly generated for each new bag, so this will not cache the results of the prompt even when the prompt is identical to a prior one.

I have an additional question about the new case introduced for array inputs. It no longer hashes the input, but uses its JSON-encoded string directly for the cache key. I imagine there must be limits to key length that this can violate. Has that been taken into consideration?

Guikingone: Hi @villermen 👋🏻 

You're right about the missing hash of the input when an array is submitted, I'll push a new PR for this one.

Regarding the message bag, it works if you're using the same bag over and over but you're right if the bag is updated during the process, my main concern about using normalizers is that the `$input` can be updated as you suggested and I wonder is the `CachedPlatform` should be in charge of checking this situation 🤔 

One solution might be to use the latest user input as "cache key" and check if in the new bag, the input still the same, not perfect but as bags are just "bags" ... 🤔

villermen: Continued in https://github.com/symfony/ai/pull/1754#discussion_r2930501392.

PR #1745, opened 2026-03-10
[Store] Add PreQueryEvent for query enhancement

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | -
| License       | MIT

## Summary

Requires: https://github.com/symfony/ai/pull/1704 (merged)

This adds a `PreQueryEvent` dispatched **before vectorization** in the `Retriever`, symmetric to the existing `PostQueryEvent`. Together they give full control over both the input and output of the retrieval pipeline.

### Use cases

- **Query expansion** — enrich vague queries with synonyms or related terms
- **Spelling correction** — fix typos before they get vectorized
- **Query rewriting** — rephrase via LLM for better embedding quality
- **Option adjustment** — dynamically change `semanticRatio` or other options

### Pipeline

```
Retriever::retrieve("search text", $options)
  → dispatch PreQueryEvent($query, $options)
      → listeners may modify query and options
  → createQuery() → store->query()
  → dispatch PostQueryEvent($query, $documents, $options)
  → yield documents
```

### Usage

**Query expansion:**
```php
$dispatcher->addListener(PreQueryEvent::class, function (PreQueryEvent $event) {
    $event->setQuery($event->getQuery() . ' crime family mafia organized crime');
});
```

**Option adjustment:**
```php
$dispatcher->addListener(PreQueryEvent::class, function (PreQueryEvent $event) {
    $event->setOptions(['semanticRatio' => 0.9]);
});
```

### DIFF
--- a/src/store/src/Event/PreQueryEvent.php
+++ b/src/store/src/Event/PreQueryEvent.php
@@ -1,0 +1,60 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Store\Event;
+
+use Symfony\Contracts\EventDispatcher\Event;
+
+/**
+ * Event dispatched before documents are retrieved from the store.
+ *
+ * Listeners can modify the query string and options, for example to expand
+ * the query, correct spelling, inject synonyms, or adjust options like semanticRatio.
+ *
+ * @author Johannes Wachter <johannes@sulu.io>
+ */
+final class PreQueryEvent extends Event
+{
+    /**
+     * @param array<string, mixed> $options
+     */
+    public function __construct(
+        private string $query,
+        private array $options = [],
+    ) {
+    }
+
+    public function getQuery(): string
+    {
+        return $this->query;
+    }
+
+    public function setQuery(string $query): void
+    {
+        $this->query = $query;
+    }
+
+    /**
+     * @return array<string, mixed>
+     */
+    public function getOptions(): array
+    {
+        return $this->options;
+    }
+
+    /**
+     * @param array<string, mixed> $options
+     */
+    public function setOptions(array $options): void
+    {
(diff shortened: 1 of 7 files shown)


examples/rag/hybrid-reranking.php
chr-hertel, line 1: Please split this file in six files in `examples/reranking/` instead

src/store/src/Event/PreQueryEvent.php
```diff
@@ -0,0 +1,96 @@
...
+     */
+    public function setDocuments(iterable $documents): void
+    {
+        $this->documents = $documents;
+    }
+
+    public function hasDocuments(): bool
+    {
+        return null !== $this->documents;
+    }
```
chr-hertel: setting documents via event feels a bit drastic - do you have an example in mind?

my main use case is mostly about normalizing language or later maybe multiplying queries

wachterjohannes: The main usecase for this would be caching. so if you detect that this was done before you can preset the documents to skip the search

chr-hertel: okay, i see. i would recommend to use a decorator for the retriever in this use case over an event listener

wachterjohannes: Ok so we should remove the documents from the even?

chr-hertel: yes, would prefer to reduce the scope for now :+1:

examples/rag/hybrid-reranking.php
wachterjohannes: i reverted this file and added new retriever examples (as they make more sense for this)

src/store/src/Event/PreQueryEvent.php
wachterjohannes: Not totally sure about the decorator vs. listener here. But for now this is not functionally necessary and the decision can be postponed :)

wachterjohannes:
@OskarStark is something missing to merge it?

chr-hertel:
Thank you @wachterjohannes.

PR #1747, opened 2026-03-11
[Platform] Clarify thrown exceptions and normalize exception taxonomy

### LINKED ISSUE #1637: Improve documentation of thrown exceptions
When integrating Symfony AI into another framework, I need a clear and reliable overview of the exceptions that each public method can throw. Right now, that information is either incomplete or too generic to be actionable.

For example, ResultConverter::convert() can throw several specific exceptions. Authentication-related exceptions were introduced in this commit: https://github.com/symfony/ai/commit/56881ba32645edfc279df55b2fb28cd02ab14f99

ModelNotFoundException was added in this commit: https://github.com/symfony/ai/commit/3374b14da44b958bd9c28af709c52b886060d4d6

However, the interface PHPDoc only documents `ExceptionInterface`. That interface extends `\Throwable` and does not provide any meaningful distinction between different failure scenarios. From an integration perspective, this forces developers to either read the implementation in detail or fall back to broad catch (`\Throwable`) blocks, which defeats the purpose of having typed exceptions in the first place.

Having a complete and accurate list of possible exceptions, or at least well-defined base exceptions per domain (for example authentication, transport, model resolution, rate limiting, etc.), would make it much easier to:

- Implement precise error handling in host frameworks
- Map Symfony AI exceptions to framework-specific exceptions
- Avoid overly defensive catch-all blocks
- Rely on static analysis tools like PHPStan to verify that checked exceptions are handled before production

Better exception documentation would significantly improve the developer experience for framework integrators and help ensure more robust integrations overall.

Image

(Source: https://docs.google.com/presentation/d/17gfiWgCewU8F9bFXmKW2qXqvKbAF5cqIoFELfw8b0Cc/mobilepresent#slide=id.g2bfa81ea89f_0_86)

mxr576:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | yes
| Issues        | Fix #1637
| License       | MIT

This PR showcases potential improvements that could be added related to #1637. Code in this PR got inspired by how [Symfony Messenger component ](https://github.com/symfony/messenger/tree/8.1/Exception)categories and exposes exceptions. 

TODO:
-  [ ] Consider moving exception hierarchy documentation to docs

### DIFF
--- a/src/platform/README.md
+++ b/src/platform/README.md
@@ -56,6 +56,57 @@ To use a specific AI platform, install the corresponding bridge package:
 **This repository is a READ-ONLY sub-tree split**. See
 https://github.com/symfony/ai to create issues or submit pull requests.
 
+## Exception strategy
+
+The Platform component defines its own exception taxonomy. Consumers of this component should depend on Platform exceptions rather than on exceptions from lower-level libraries such as HTTP clients, process execution libraries, or provider SDKs.
+
+In other words, the public contract of this component is expressed through exceptions in the `Symfony\AI\Platform\Exception\` namespace. Bridge implementations are expected to normalize provider-specific and transport-specific failures into these Platform exceptions before they cross the `PlatformInterface`, `ModelClientInterface`, or `ResultConverterInterface` boundaries.
+
+### Exception categories
+
+The current exception set is intended to cover several different kinds of failures, including:
+
+- authentication and authorization problems, for example `AuthenticationException`
+- invalid input or request problems:
+  - `InvalidArgumentException` for invalid usage of the Platform API by the caller
+  - `InvalidRequestException` for semantically invalid requests in the Platform/provider contract
+  - `BadRequestException` for generic provider-side request rejection when no narrower Platform exception applies
+- model resolution and support problems, for example `ModelNotFoundException` or `MissingModelSupportException`
+- provider-side runtime limits or policies, for example `RateLimitExceededException`, `ExceedContextSizeException`, or `ContentFilterException`
+- conversion or contract mismatches, for example `UnexpectedResultTypeException`
+- broader component-level failures, through `LogicException`, `RuntimeException`, and `ExceptionInterface`
+
+### Recoverable and unrecoverable failures
+
+When handling failures, consumers will often want to distinguish between:
+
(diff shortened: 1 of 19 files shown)


chr-hertel:
Hi @mxr576, thanks for this draft - I usually filter out drafts, this is how i lost this out of sight, sorry for that!

I think this is a great start - the only thing I wonder is about the interpretation of `Recoverable` vs. `Unrecoverable`.
At first I thought it's great guidance for developers and makes it easier to implement retry mechanisms, but two things here:
* if users want retry I would mostly recommend to configure that on [http client level](https://symfony.com/doc/current/http_client.html#retry-failed-requests)
* the interpretation if recoverable or not, might be due to the use-case, e.g. context windows can be compressed which is some kind of recovery strategy

do you need this underlying differentiation on your end already?

mxr576:
I would generally agree if all LLM connectors in this codebase were HTTP-based, but that's not the case. The codebase already has working examples of non-HTTP transports, such as [ClaudeCode](https://github.com/symfony/ai/tree/main/src/platform/src/Bridge/ClaudeCode) and [Codex](https://github.com/symfony/ai/tree/main/src/platform/src/Bridge/Codex), which communicate via CLI processes rather than HTTP. Configuring retry at the Symfony HTTP client level simply doesn't apply to those bridges. Because of this, any retry or failover strategy that needs to work across all platform bridges has to live at a higher abstraction level.

That said, you're right that no retry or failover mechanism exists at the platform level yet, so this classification is partly forward-looking. I'm comfortable with that tradeoff because retrofitting exception semantics later, once retry logic is actually built, tends to be messier than establishing the contract early. But I recognize that's a judgment call and I'm open to deferring if the project prefers to wait until there's a concrete consumer for this distinction.

On the context window example: fair point. "Recoverable" from the platform's perspective (the operation could be retried or handled differently) is not the same as "recoverable" from the application's perspective (the application knows how to actually handle it). I'd be happy to clarify this in the docblocks, making it explicit that the classification describes the nature of the error at the platform boundary, not a guarantee that the calling code has a viable recovery path.

mxr576:
Have I managed to answer your questions @chr-hertel ? How do you see the feature of this PR, especially in the contrast of other ongoing work? Should I just rebase it and finalize it for review?

mxr576:
@chr-hertel  please allow me a kind reminder here 

Probably this PR also needs an update after #1923.

PR #1748, opened 2026-03-11
[AI Bundle][Profiler] Fix all tools listing duplicates for each agent

DZunke:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | 
| License       | MIT

If one is registering multiple agents the tools listing in the profiler is getting full of duplicates. So, for example, there is a `search_entries` tool and a `entry_detail` tool but you have two different agents and both get all tools, then the "Tools" section in the profiler which should just be a list of all tools available shows each of these tools in double.

I first deduplicated by taking the execution class and method as a unique selling point but had to take also the name into account as, with the names, it is also possible to have multiple tools with the same execution reference but just different names. That popped up as i thought about the subagent tool. 

So hopefully `__name__::__class__::__method__` is unique enough for deduplication in the profiler all tools listing in any case.

### DIFF
--- a/src/ai-bundle/tests/Profiler/DataCollectorTest.php
+++ b/src/ai-bundle/tests/Profiler/DataCollectorTest.php
@@ -14,12 +14,15 @@ namespace Symfony\AI\AiBundle\Tests\Profiler;
 use PHPUnit\Framework\TestCase;
 use Symfony\AI\Agent\AgentInterface;
 use Symfony\AI\Agent\MockAgent;
+use Symfony\AI\Agent\Toolbox\Tool\Subagent;
+use Symfony\AI\Agent\Toolbox\ToolboxInterface;
 use Symfony\AI\AiBundle\Profiler\DataCollector;
 use Symfony\AI\AiBundle\Profiler\TraceableAgent;
 use Symfony\AI\AiBundle\Profiler\TraceableChat;
 use Symfony\AI\AiBundle\Profiler\TraceableMessageStore;
 use Symfony\AI\AiBundle\Profiler\TraceablePlatform;
 use Symfony\AI\AiBundle\Profiler\TraceableStore;
+use Symfony\AI\AiBundle\Profiler\TraceableToolbox;
 use Symfony\AI\Chat\Chat;
 use Symfony\AI\Chat\InMemory\Store as InMemoryStore;
 use Symfony\AI\Platform\Message\Content\Text;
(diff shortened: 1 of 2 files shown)


src/ai-bundle/tests/Profiler/DataCollectorTest.php
```diff
@@ -236,4 +241,76 @@ public function testCollectsDataForStores()
...
+            'Does Something'
+        );
+
+        $tool3 = new Tool(
+            new ExecutionReference('App\Tool\SecondTool', 'second'),
+            'second_tool',
+            'Does Something Else'
+        );
+
+        $toolbox1 = $this->createMock(ToolboxInterface::class);
```
chr-hertel:
```suggestion
        $toolbox1 = $this->createStub(ToolboxInterface::class);
```

src/ai-bundle/tests/Profiler/DataCollectorTest.php
```diff
@@ -236,4 +241,76 @@ public function testCollectsDataForStores()
...
+        $toolbox1->method('getTools')->willReturn([$tool1, $tool3]);
+
+        $toolbox2 = $this->createMock(ToolboxInterface::class);
```
chr-hertel:
```suggestion
        $toolbox2 = $this->createStub(ToolboxInterface::class);
```

src/ai-bundle/tests/Profiler/DataCollectorTest.php
```diff
@@ -236,4 +241,76 @@ public function testCollectsDataForStores()
...
+            'Research Agent'
+        );
+
+        $tool2 = new Tool(
+            new ExecutionReference(Subagent::class, '__invoke'),
+            'writer_agent',
+            'Writer Agent'
+        );
+
+        $toolbox1 = $this->createMock(ToolboxInterface::class);
```
chr-hertel:
```suggestion
        $toolbox1 = $this->createStub(ToolboxInterface::class);
```

src/ai-bundle/tests/Profiler/DataCollectorTest.php
```diff
@@ -236,4 +241,76 @@ public function testCollectsDataForStores()
...
+        $toolbox1->method('getTools')->willReturn([$tool1]);
+
+        $toolbox2 = $this->createMock(ToolboxInterface::class);
```
chr-hertel:
```suggestion
        $toolbox2 = $this->createStub(ToolboxInterface::class);
```

chr-hertel (changes requested):
Thanks Denis, only some nit picks we nowadays care about :D

DZunke:
Thanks for the review! I have changed the mocks to stubs and it was absolutely right, latest wioth future PHPUnit versions. Sorry for missing it! 🙇‍♂️

chr-hertel:
Thank you @DZunke.

PR #1754, opened 2026-03-12
[Cache] Fix support for `MessageBag` and string

### LINKED ISSUE #1734: `CachePlatform` input gets shared when it's a `MessageBag`

### LINKED ISSUE #2192: [Platform][Cache] CachePlatform never hits cache when input is a MessageBag
`CachePlatform` is not capable of returning a cached result for a `MessageBag`. Every invocation produces a cache miss because the normalized input key is derived from the `MessageBag`'s randomly generated UUID rather than its content.

If that's intended and has reasons, please let me know. Perhaps i'm missing something here.

Here's more details on that issue:

  ### Root cause
  
  `CachePlatform::invoke()` builds the cache key using this match:

  ```php
      \is_array($input)  => json_encode($input),
      $input instanceof MessageBag => $input->getId()->toString(), // ← this is the case for me
      ...
  };
```
  
  MessageBag generates a fresh Uuid::v7() in its constructor on every instantiation:

```php
  {
      $this->id = Uuid::v7(); // new random UUID every time
  }
```

  Since a new MessageBag is constructed for each request, getId()->toString() always returns a different value, making the cache key unique on every call — regardless of whether the message content is identical.

  The string and array branches use content-based normalization and work correctly. The MessageBag branch does not.

###  Steps to reproduce
  
  1. Configure a CachePlatform wrapping an inner platform, with a working cache adapter.
  2. Ensure prompt_cache_key is set in the options.
  3. Call platform->invoke() twice with two separate MessageBag instances containing identical messages.
  4. Log requests to the LLM to find out if the request is being sent twice

##  Expected behavior

The second invocation with identical message content returns the cached result without calling the inner platform.

##  Actual behavior
  
The MessageBag always has a different ``$normalizedInput`` (generated uuid) and therefore does not hit the cache.

##  Suggested fix
  
Some kind of normalization of the MessageBag to find out if cache should hit. 

The content serialization needs to handle all ContentInterface implementations (Text, Image, File, etc.) without including any randomly generated identifiers.

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| Docs?         | yes
| Issues        | Fix #1734 - Fix #2192 - Part of #1742
| License       | MIT

`MessageBagCacheKeyGenerator` keys a bag on `MessageBag::getId()`, a random `Uuid::v7()` regenerated on
every construction, so a lookup only ever hits when the very same instance is passed twice. Two bags
carrying the same conversation never share an entry, which defeats the point of the decorator.

The key is now derived from the message content through the platform `Contract` serializer, without a
model bound to the context, so it reflects the whole bag - system prompt and prior turns included - and
stays provider agnostic. A bare `MessageInterface` input gets the same treatment through the new
`MessageCacheKeyGenerator`, and an `array` input is hashed instead of being concatenated raw, which could
produce a key holding PSR-6 reserved characters.

To keep that normalization complete, base `Contract` normalizers are added for `DocumentUrl`, `File`
(covering `Video` and `Document`), `Collection`, `Template`, `Thinking`, `ExecutableCode` and
`CodeExecution`, `AssistantMessageNormalizer` stops dropping the provider specific parts when no model is
bound (a provider request payload keeps ignoring them, so no unknown key reaches the API), and
`SystemMessageNormalizer` stringifies a `Template` content instead of serializing it to an empty object.

The cache layer is hardened so it never breaks the request:

 * a `ToolCallResult` is normalized inline, so it is cacheable with the default serializer, which no
   longer needs a dedicated tool call normalizer to be registered;
 * the signature of `TextResult` and `ToolCall` survives the round-trip, as it already did for
   `ThinkingResult`, so a provider scoped signature (Gemini/Vertex AI `thoughtSignature`) is not lost;
 * a streamed call bypasses the cache instead of throwing on `StreamResult`;
 * an input that cannot be keyed, a result the serializer cannot handle (every result type added since
   0.11 - web search, MCP calls, ... - currently makes the request throw) and a stale or corrupted entry
   all fall back to a live invocation;
 * the constructor `cacheKey`, which was never read, acts as the default namespace, the key segments are
   joined with a `.` to avoid collisions at the namespace/model boundary, and
   `CachePlatform::invalidateTags()` drops entries by model or namespace tag.

### DIFF
--- a/src/platform/src/Bridge/Cache/CachePlatform.php
+++ b/src/platform/src/Bridge/Cache/CachePlatform.php
@@ -44,6 +44,18 @@ use Symfony\Contracts\Cache\ItemInterface;
 final class CachePlatform implements PlatformInterface
 {
     /**
+     * Invocation option holding the cache namespace of the call, it defaults to the constructor
+     * level cache key and opts the call out of caching when set to an empty string.
+     */
+    public const OPTION_CACHE_KEY = 'prompt_cache_key';
+
+    /**
+     * Invocation option holding the lifetime (in seconds) of the cached entry, it defaults to the
+     * constructor level TTL.
+     */
+    public const OPTION_CACHE_TTL = 'prompt_cache_ttl';
+
+    /**
      * @param iterable<CacheKeyGenerator> $cacheKeyGenerators Tried in order to key non-scalar inputs (objects)
      */
     public function __construct(
@@ -60,39 +72,65 @@ final class CachePlatform implements PlatformInterface
         private readonly ?int $cacheTtl = null,
         private iterable $cacheKeyGenerators = [
             new MessageBagCacheKeyGenerator(),
+            new MessageCacheKeyGenerator(),
             new DocumentUrlCacheKeyGenerator(),
             new ImageUrlCacheKeyGenerator(),
             new FileCacheKeyGenerator(),
         ],
+        private readonly InputHasher $inputHasher = new InputHasher(),
     ) {
     }
 
     public function invoke(string|Model $model, array|string|object $input, array $options = []): DeferredResult
     {
-        if (null === $this->cache || !\array_key_exists('prompt_cache_key', $options) || '' === $options['prompt_cache_key']) {
+        $namespace = $options[self::OPTION_CACHE_KEY] ?? $this->cacheKey;
+        $ttl = $options[self::OPTION_CACHE_TTL] ?? $this->cacheTtl;
+
+        // The decorator consumes its own options: they must not reach the decorated platform, whether
+        // the call is cached or bypasses the cache.
+        unset($options[self::OPTION_CACHE_KEY], $options[self::OPTION_CACHE_TTL]);
+
+        if (null === $this->cache || null === $namespace || '' === $namespace) {
+            return $this->platform->invoke($model, $input, $options);
+        }
+
+        if (false !== strpbrk($namespace, '{}()/\\@:')) {
+            throw new InvalidArgumentException(\sprintf('The cache namespace "%s" contains reserved characters ("{}()/\@:") and cannot be used to build a cache key.', $namespace));
+        }
+
+        // A stream is consumed while it is read, caching it would drain it before the caller sees it.
+        if (true === ($options['stream'] ?? false)) {
             return $this->platform->invoke($model, $input, $options);
         }
 
         $modelName = $model instanceof Model ? $model->getName() : $model;
 
-        $normalizedInput = match (true) {
-            \is_string($input) => md5($input),
-            \is_array($input) => json_encode($input),
-            default => $this->generateInputCacheKey($input),
-        };
+        try {
+            $normalizedInput = $this->generateInputCacheKey($input);
+            $normalizedOptions = $this->inputHasher->hash($options);
+        } catch (UncacheableInputException) {
+            // Fail open: an input or an option set that cannot be keyed deterministically is served live.
+            return $this->platform->invoke($model, $input, $options);
+        }
 
-        $cacheKey = (new UnicodeString())->join([
-            $options['prompt_cache_key'] ?? $this->cacheKey,
+        // "." separates the segments: ":" and "/" are reserved by PSR-6 and a delimiter avoids
+        // boundary collisions (namespace "sy" + model "m" versus namespace "s" + model "ym"). The
+        // options are part of the key: the same input answered with a different temperature, tool set
+        // or response format is a different request.
+        $cacheKey = (new UnicodeString('.'))->join([
+            $namespace,
             (new UnicodeString($modelName))->camel(),
             $normalizedInput,
+            $normalizedOptions,
         ]);
 
-        $ttl = $options['prompt_cache_ttl'] ?? $this->cacheTtl;
+        $uncacheableResult = null;
 
-        unset($options['prompt_cache_key'], $options['prompt_cache_ttl']);
-
-        $cached = $this->cache->get($cacheKey, function (ItemInterface $item) use ($model, $modelName, $input, $options, $cacheKey, $ttl): array {
-            $item->tag((new UnicodeString($modelName))->camel());
+        $cached = $this->cache->get($cacheKey, function (ItemInterface $item, bool &$save) use ($model, $modelName, $input, $options, $cacheKey, $namespace, $ttl, &$uncacheableResult): array {
+            $item->tag([
+                (new UnicodeString($modelName))->camel(),
+                'namespace.'.$namespace,
+            ]);
 
             if (null !== $ttl) {
                 $item->expiresAfter($ttl);
@@ -102,8 +140,18 @@ final class CachePlatform implements PlatformInterface
 
             $result = $deferredResult->getResult();
 
+            try {
+                $normalizedResult = $this->serializer->normalize($result);
+            } catch (\Throwable) {
+                // Fail open: a result the serializer cannot handle is served live and not stored.
+                $save = false;
+                $uncacheableResult = $deferredResult;
+
+                return [];
+            }
+
             return [
-                'result' => $this->serializer->normalize($result),
+                'result' => $normalizedResult,
                 'raw_data' => $deferredResult->getRawResult()->getData(),
                 'metadata' => $result->getMetadata()->all(),
                 'cached_at' => $this->clock->now()->getTimestamp(),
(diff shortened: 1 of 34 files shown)


Guikingone:
@villermen If you want to continue the discussion 🙂

src/platform/src/Bridge/Cache/CachePlatform.php
```diff
@@ -66,8 +67,8 @@ public function invoke(string $model, array|string|object $input, array $options
 
         $normalizedInput = match (true) {
             \is_string($input) => md5($input),
-            \is_array($input) => json_encode($input),
-            $input instanceof MessageBag => $input->getId()->toString(),
+            \is_array($input) => md5(json_encode($input)),
+            $input instanceof MessageBag => $input->latestAs(Role::User)->getId()->toRfc4122(),
```
villermen: Continued from https://github.com/symfony/ai/pull/1742/changes#r2925408069

@Guikingone 

> One solution might be to use the latest user input as "cache key" and check if in the new bag, the input still the same, not perfect but as bags are just "bags" ... 🤔

You're referring to using the input, but what I'm seeing is usage of the ID only. IMO, there's an inherent problem in using IDs. The content of the message can change without affecting the ID.

What `CachePlatform` provides is an identical result for identical inputs without invoking the wrapped platform. Since this is not using the input at all, I think it kinda misses the point. 

Why is normalizing the bag with `MessageBagNormalizer` not an option in your eyes? (`$this->serializer->normalize($input)`) Am I missing something that makes this impossible?

villermen (changes requested):

src/platform/src/Bridge/Cache/CachePlatform.php
Guikingone: > Why is normalizing the bag with MessageBagNormalizer not an option in your eyes? `($this->serializer->normalize($input))` Am I missing something that makes this impossible?

The `MessageBag` cannot be normalized without loosing its own identifier, could be updated, not an issue to be honest, It's not done for now as I'm wondering about the serialization of messages.

Guikingone:
@villermen @push-mann I pushed a new version with an `InputHasher` if you want to take a look at it 🙂

src/platform/src/Bridge/Cache/CachePlatform.php
```diff
@@ -56,26 +55,36 @@ classDiscriminatorResolver: new ClassDiscriminatorFromClassMetadata(new ClassMet
...
         private readonly ?string $cacheKey = null,
         private readonly ?int $cacheTtl = null,
+        private readonly InputHasher $inputHasher = new InputHasher(),
     ) {
     }
 
     public function invoke(string|Model $model, array|string|object $input, array $options = []): DeferredResult
     {
-        if (null === $this->cache || !\array_key_exists('prompt_cache_key', $options) || '' === $options['prompt_cache_key']) {
+        $namespace = $options['prompt_cache_key'] ?? $this->cacheKey;
```
push-mann: is it possible to expose this array-key as public const so it's usable from outside when setting the option?

push-mann (approved):

push-mann:
Caching works well now for me 🙂 

@Guikingone I also came across #2160 while investigating. Not sure whether it's relevant here but i think it addresses at least caching for content-inputs.

src/platform/src/Contract.php
villermen, line 1: Unrelated observation: I don't really get this pattern `Contract` uses. It is supposed to be extended, but returns `self`. That means implementing classes will never be instantiated.

@chr-hertel Wouldn't it make more sense to have this class be abstract and provide a method with the default normalizers to (not) use? Something along the lines of...

```php
abstract class Contract
{
    /**
     * @return NormalizerInterface[]
     */
    public static function getDefaultNormalizers(): array
    {
        // ...
    }
}
```

src/platform/src/Contract/Normalizer/Message/AssistantMessageNormalizer.php
```diff
@@ -56,6 +57,8 @@ public function normalize(mixed $data, ?string $format = null, array $context =
                 $reasoning .= $part->getContent();
             } elseif ($part instanceof ToolCall) {
                 $toolCalls[] = $part;
+            } else {
+                $additional[] = $this->normalizer->normalize($part, $format, $context);
```
villermen: Suggestion: This feels like a straight up improvement over typechecking against all potential part types. I don't know how this project does TODO management, but I would consider deprecating the dedicated "tool_calls" and "content" keys in favor of just a "parts" key with everything in it in this case. Would also save you from having to call it "additional" =)

src/platform/src/Contract/Normalizer/Message/AssistantMessageNormalizer.php
```diff
@@ -72,6 +75,10 @@ public function normalize(mixed $data, ?string $format = null, array $context =
             $array['reasoning_content'] = $reasoning;
         }
 
+        if ([] !== $additional) {
+            $array['content_parts'] = $additional;
```
villermen, line 86: This PR does not add a test for this new key. I suggest adding it in the `Contract` variety of `AssistantMessageNormalizerTest`.

villermen (commented):
Had to read in a bit deeper than before when I just reported the issue I observed without really using the library itself 😅 My earlier suggestion was based on the assumption this normalizing logic was already in place, so making it available is a solid improvement in my eyes. Thanks for taking the time, and good work!

I've found nothing serious in my review, but ideally would like to see a test on the new main normalizer before merge.

push-mann:
@Guikingone are there any news on how generating cache-keys for `MessageBag`s will work going forward? This draft provides a working solution. But there now have been some changes in the CachePlatform. There is a new `MessageBagCacheKeyGenerator`. In there, we still have the non-working behavior, using the `MeesageBag`s uuid as CacheKey, which is always being regenerated on construction.

Just wanted to kindly ask whether this draft is still on the roadmap.

Guikingone:
Hi @push-mann 👋🏻 

Yes, it's planned to fix it using this PR, don't have any merge time, need to rebase the branch and adapt the code, can't promise anything for this week but in the next few days, for sure.

Guikingone:
Rebased on top of `main` and adapted to the `CacheKeyGenerator` strategy merged in the meantime — thanks @push-mann for pointing out that `MessageBagCacheKeyGenerator` had reintroduced the identifier-based key. The content-based key now lives in that generator instead of replacing the strategy, and `MessageCacheKeyGenerator` covers a bare `MessageInterface` input.

Two points I would like a maintainer's call on.

**1. `content_parts` is emitted only when no model is bound to the context.**

@villermen, this is the one change I made to what you reviewed. Emitting the key unconditionally is no longer safe: since this PR was written, 8 new `ContentInterface` classes landed (`WebSearch`, `McpCall`, `McpListTools`, `McpApprovalRequest`, `ComputerCall`, `LocalShellCall`, `FileSearch`, `CustomToolCall`) and `Message::ofAssistant()` puts them inside assistant messages built from real provider responses. The OpenAI chat-completions bridge has no `AssistantMessageNormalizer` of its own, so the base one's output *is* its request body — an unconditional `content_parts` would both throw on those 8 types and add an unknown key to a payload OpenAI rejects.

Gating on `!isset($context[Contract::CONTEXT_MODEL])` keeps your improvement exactly where it matters (the content-hashing path, which is the only caller of the new `Contract::normalize()`) and leaves provider request payloads untouched. It reuses the discriminator the codebase already has: `ModelContractNormalizer` already treats "a model is bound" as "this is a provider request". A bag holding one of the 8 un-normalizable parts throws inside the hasher, which the cache layer turns into a bypass — never a wrong hit, just an uncached call.

**2. The `cacheKey` constructor argument becoming effective is a visible behaviour change.**

It was dead code: `invoke()` returned early unless `prompt_cache_key` was passed, so the argument did nothing. Reading it matches what the bundle already documents (`cache_key`: *"if not set, the current platform name will be used, the `prompt_cache_key` can be set during platform call to override this value"*) and makes the documented example in `docs/components/platform.rst` actually cache.

But it flips every bundle-configured cache platform from "caches nothing unless you opt in per call" to "caches everything", which a deployer cannot predict from the old code. I have documented it in `UPGRADE.md` accordingly — **which means this PR needs the `BC Break` label, and `Validate Changelog & Upgrade Entries` will stay red until someone with triage rights adds it.** If you would rather treat it as a plain bug fix, say so and I will drop the `UPGRADE.md` entry and keep it in the bridge CHANGELOG only.

**Also fixed while in there**, three defects the review of the rebase surfaced:

- the two cache options leaked into the decorated platform on the bypass paths, while the cached path stripped them;
- the invocation options were absent from the cache key, so the same `MessageBag` sent with a different `temperature` or tool set was served the first call's answer — the same class of bug as the one this PR fixes;
- a namespace holding a PSR-6 reserved character failed with the cache adapter's opaque message, outside every fail-open catch, even though `CacheKeyGenerator` requires reserved-char-safe segments.

PR #1758, opened 2026-03-13
[Platform][OpenRouter] Add EU routing support and user model selection

lochmueller:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | 
| License       | MIT

The Branch will add two OpenRouter features:
- Enable EU routing via enum in the ModelApiCatalog and PlatformFactory to use the EU route internally
- Support user defined models via separated Endpoint in `ModelApiCatalog` to reduce
- Add an example that use both features


lochmueller:
@OskarStark I create a fresh PR for my changes. 

The one Fabbot test that failed is a false positive. 
It suggests `Vector` instead of `Vektor` for the store bridge. 
But the bridge is actually called `Vektor`.

examples/openrouter/eu-region-and-user-based-api-model-catalog.php
OskarStark, line 1: why the "guardrail" terminology in the filename?

src/ai-bundle/src/AiBundle.php
```diff
@@ -787,6 +788,7 @@ private function processPlatformConfig(string $type, array $platform, ContainerB
                     new Reference('ai.platform.model_catalog.openrouter'),
                     null,
                     new Reference('event_dispatcher'),
+                    OpenRouterRegion::from($platform['region']),
```
OskarStark, line 849: we should update AiBundleTest

lochmueller: Will be part of the next commit.

examples/openrouter/eu-region-and-user-based-api-model-catalog.php
lochmueller: Good point... changed to "user-based" and also switch "routing" to "region".

chr-hertel, line 1: is there an easy way to test another region - somehow i can't just directly find it and my key gets me a 401 here. any hint?

lochmueller: Documentation describe it like "you have to contact the sales team to get access to the EU only routing". I have the same problem like you and I add the code to be Opernrouter-feature complete (more or less)

chr-hertel (changes requested):
Looks like i have no chance of getting that example to run on my end - please remove and fold the feature into docs or as comment into other examples.

lochmueller:
I will add it again or add some documentation, if the feature is public available

PR #1759, opened 2026-03-13
[Platform] Replace untyped stream chunks with semantic DeltaInterface deltas

fabpot:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes 
| Docs?         | yes/no 
| Issues        | n/a
| License       | MIT

This PR introduces semantic streaming event objects so that one can react to specific LLM events without parsing the raw JSON. Think of this PR as being a first step towards full streaming support ;)

PR #1761, opened 2026-03-13
[Platform][Anthropic] Add example for TokenUsage with streaming responses

### LINKED ISSUE #1760: [Platform][Anthropic] Token usage not available when streaming is enabled
When using the Anthropic platform bridge with `stream => true`, `$result->getMetadata()->get('token_usage')`
is always `null` after consuming the stream.

The `TokenUsageExtractor::extract()` explicitly returns `null` for streaming (as expected — streams should
be handled via `StreamListener`). However, the `ResultConverter::convertStream()` generator never yields a
`TokenUsage` object from the `message_start` and `message_delta` SSE events, so the `StreamListener` in
`ai-platform` never captures any token usage data.

The OpenAI bridge already handles this correctly — its `convertStream()` yields `TokenUsage` from stream
chunks (see `Gpt/ResultConverter.php` lines 142–144). The same pattern is missing from the Anthropic bridge.

Related: #1700 (same issue for the Generic platform bridge, fixed in #1715)

I will create a PR to fix this.

HKandulla:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #1760
| License       | MIT

The Anthropic `ResultConverter::convertStream()` did not capture token usage from
`message_start` (input_tokens, cache fields) and `message_delta` (output_tokens) SSE events.

This adds handling consistent with how the OpenAI bridge already yields `TokenUsage` from
streaming responses, so the platform's `StreamListener` can pick it up and populate result metadata.

### DIFF
--- a/examples/anthropic/stream-token-metadata.php
+++ b/examples/anthropic/stream-token-metadata.php
@@ -1,0 +1,38 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use Symfony\AI\Agent\Agent;
+use Symfony\AI\Platform\Bridge\Anthropic\PlatformFactory;
+use Symfony\AI\Platform\Message\Message;
+use Symfony\AI\Platform\Message\MessageBag;
+
+require_once dirname(__DIR__).'/bootstrap.php';
+
+$platform = PlatformFactory::create(env('ANTHROPIC_API_KEY'), httpClient: http_client());
+
+$agent = new Agent($platform, 'claude-sonnet-4-5-20250929');
+$messages = new MessageBag(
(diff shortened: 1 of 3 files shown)


chr-hertel (changes requested):
can you please add an example similar to `examples/openai/stream-token-metadata.php` proving this works?

HKandulla:
@chr-hertel sure, I added `examples/anthropic/stream-token-metadata.php` and squashed all commits.

chr-hertel:
FYI, this will wait on #1759 and needs a rebase after it

chr-hertel:
@HKandulla just merged #1759 - rebase unlocked :)

HKandulla:
Hi @chr-hertel, I think we can close this PR as #1759 includes the streaming token usage handling for the Anthropic bridge (message_start/message_delta extraction). Let me know if I'm wrong.

chr-hertel:
Closing in favor of #1860  - picked the new example

chr-hertel:
Thanks @HKandulla!

PR #1763, opened 2026-03-13
[Platform][Anthropic] Add prompt caching for tool definitions

HKandulla:
| Q             | A
  | ------------- | ---
  | Bug fix?      | no
  | New feature?  | yes
  | Tickets       | -
  | License       | MIT

  ## Summary

  This PR adds automatic prompt caching (`cache_control`) to tool definitions in the Anthropic `ModelClient`, complementing the existing caching for system
  prompts and user messages.

  ### Problem

  Currently, `injectCacheControl()` only sets `cache_control` on the last user message. Tool definitions — which are typically **identical across all requests** —
   are not cached independently. This means they are re-processed as regular input tokens on every request, even though their content never changes.

  ### Solution

  Add `cache_control` to the last tool definition, creating an additional cache breakpoint:

  System (breakpoint 1) → Tools (breakpoint 2) → Messages (breakpoint 3)

  This allows the prefix "system + tools" to be cached and reused independently of the messages, which change with every request.

  ### Details

  - New private method `injectToolsCacheControl()` mirrors the pattern of the existing `injectCacheControl()`
  - Respects the existing `$cacheRetention` setting (`none` / `short` / `long`)
  - No new configuration needed — follows the same caching strategy already in place
  - Tool definitions are an ideal caching target: they are static across requests and can be substantial in size (descriptions + JSON schemas)

  ### Cost impact

  With Anthropic's pricing:
  - `cache_creation`: 1.25× input price (one-time per cache entry)
  - `cache_read`: 0.1× input price (every subsequent request)

  Since tool definitions don't change, the cache entry is created once and read on every subsequent request within the TTL — a significant cost reduction for
  tool-heavy agents.

### DIFF
--- a/src/platform/src/Bridge/Anthropic/ModelClient.php
+++ b/src/platform/src/Bridge/Anthropic/ModelClient.php
@@ -63,6 +63,7 @@ final class ModelClient implements ModelClientInterface
 
         if (isset($options['tools'])) {
             $options['tool_choice'] = ['type' => 'auto'];
+            $options['tools'] = $this->injectToolsCacheControl($options['tools']);
         }
 
         if (isset($options['thinking'])) {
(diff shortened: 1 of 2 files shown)


src/platform/src/Bridge/Anthropic/ModelClient.php
```diff
@@ -90,6 +91,33 @@ public function request(Model $model, array|string $payload, array $options = []
...
+    {
+        if ('none' === $this->cacheRetention || [] === $tools) {
+            return $tools;
+        }
+
+        $cacheControl = 'long' === $this->cacheRetention
+            ? ['type' => 'ephemeral', 'ttl' => '1h']
+            : ['type' => 'ephemeral'];
+
+        $tools[\count($tools) - 1]['cache_control'] = $cacheControl;
```
chr-hertel, line 116: so i get it right that we only need to set this "breakpoint" at the last tool definition?

HKandulla: Yes, that's correct — the cache_control breakpoint only needs to be set on the last tool definition. It caches everything up to and including that marked block as a single prefix.

See the example for "caching tool definitions" in in https://platform.claude.com/docs/en/build-with-claude/prompt-caching#prompt-caching-examples

chr-hertel:
Thank you @HKandulla.

PR #1765, opened 2026-03-13
[Platform][Mistral] Add token usage extraction for embeddings

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | -
| License       | MIT

## Description

This PR adds token usage extraction support for Mistral embeddings — the first embedding bridge to support this.

### Changes

1. **`Mistral\Embeddings\TokenUsageExtractor`** — New class implementing `TokenUsageExtractorInterface`. Extracts `prompt_tokens`, `total_tokens`, and rate limit headers (`x-ratelimit-limit-tokens-minute`, `x-ratelimit-limit-tokens-month`) from Mistral embedding responses. No `completionTokens` since embeddings don't produce completions.

2. **`Mistral\Embeddings\ResultConverter`** — `getTokenUsageExtractor()` now returns the new extractor instead of `null`.

3. **Updated Mistral embeddings example** to display token usage.

### Usage

```php
$result = $platform->invoke('mistral-embed', 'Some text to embed');

$tokenUsage = $result->getMetadata()->get('token_usage');
$tokenUsage->getPromptTokens();  // e.g. 15
$tokenUsage->getTotalTokens();   // e.g. 15
```

### DIFF
--- a/examples/mistral/embeddings.php
+++ b/examples/mistral/embeddings.php
@@ -23,3 +23,4 @@ $result = $platform->invoke('mistral-embed', <<<TEXT
     TEXT);
 
 print_vectors($result);
+print_token_usage($result->getMetadata()->get('token_usage'));
--- a/src/platform/src/Bridge/Mistral/CHANGELOG.md
+++ b/src/platform/src/Bridge/Mistral/CHANGELOG.md
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+0.7
+---
+
+ * Add token usage extraction for embeddings
+
 0.1
 ---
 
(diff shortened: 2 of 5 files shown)


chr-hertel:
Can we scope out `DeferredResult::getTokenUsage()`? at least that's rather debatable for me

idea of metadata is to have it rather flexible and unstructured - putting a getter on it, feels counter-intuitive. but maybe token usage is central enough - should it still be part of the metadata then?

(i know we do it already in store, but doesn't mean we just replicate here)

chr-hertel:
Thank you @wachterjohannes.

chr-hertel:
Added changelog entry on merge.

PR #1766, opened 2026-03-13
 Add description property to #[With] attribute

### LINKED ISSUE #1737: Use Attribute to describe properties
```php
final class Ingredient
{
    /**
     * @var string Name of the ingredient
     */
    public string $name;

    /**
     * @var string Quantity of the ingredient (e.g., "2 cups", "150g")
     */
    public string $quantity;
```

I like the way ApiPlatform formalizes that:

```php
use ApiPlatform\Metadata\ApiProperty;

final class Ingredient
{
    #[ApiProperty(description: 'Name of the ingredient', example: 'Flour')]
    public string $name;

    #[ApiProperty(description: 'Quantity of the ingredient', example: '2 cups')]
    public string $quantity;
}
```

ApiProperty isn't the right attribute, though.  JsonSchema _might_ be the right spot.

The docBlocks are good for IDE hints, but it seems like if we had properties to read we'd be able to do something useful with them.  Just brainstorming.

Related: #1736

bigdevlarry:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no 
| Issues        | Fix #1737 
| License       | MIT

Add description property to #[With] attribute

### DIFF
--- a/src/platform/CHANGELOG.md
+++ b/src/platform/CHANGELOG.md
@@ -5,6 +5,7 @@ CHANGELOG
 ---
 
  * Add reranking support via `RerankingResult`, `RerankingEntry`, and `Capability::RERANKING`
+ * Add `description` and `example` properties to `#[With]` attribute
 
 0.6
 ---
--- a/src/platform/src/Contract/JsonSchema/Attribute/With.php
+++ b/src/platform/src/Contract/JsonSchema/Attribute/With.php
@@ -25,6 +25,8 @@ final class With
      */
     public function __construct(
         // can be used by many types
+        public readonly ?string $description = null,
+        public readonly mixed $example = null,
         public readonly ?array $enum = null,
         public readonly string|int|array|null $const = null,
 
@@ -56,6 +58,12 @@ final class With
             /* @phpstan-ignore-next-line function.alreadyNarrowedType */
             if (array_filter($enum, static fn (mixed $item) => null === $item || \is_int($item) || \is_float($item) || \is_string($item)) !== $enum) {
                 throw new InvalidArgumentException('All enum values must be float, integer, strings, or null.');
+            }
+        }
+
+        if (\is_string($description)) {
+            if ('' === trim($description)) {
+                throw new InvalidArgumentException('Description string must not be empty.');
             }
         }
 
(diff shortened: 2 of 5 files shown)


OskarStark (commented):
You can already use phpdoc for description isn't this enough? 

If not, please add a test where both descriptions are provided and the attribute wins

bigdevlarry:
@OskarStark I agree that PHPDoc already does some extraction, but for extra properties we'd like to read as described in issue #1737, like examples, that won't suffice. I've added a coverage for the scenario you've described now

chr-hertel:
Thanks @bigdevlarry for jumping on that :+1: - looks good to me, needs a rebase tho.

@tacman & @valtzu - is it what you had in mind and solves #1737 for you?

valtzu:
I'm ok with this, I would prefer that `With` attribute would support all schema properties (that LLMs support) 

Isn't it called `examples` (plural) though?

tacman:
Yes, thanks!  

description will fit naturally here to create a data definition document without needing to use the PhpDocBlock tools

```php
/**
     * dcterms:accessRights — reuse permission granted by the holding institution.
     * @var string|null
     */
    #[Map(source: 'reuse_allowed', facet: true)]
    #[With(enum: ['no restrictions', 'creative commons', 'contact host', 'all rights reserved'])]
    public readonly ?string $reuseAllowed = null;

    /**
     * dcterms:rights — human-readable rights statement from source record.
     * @var string|null
     */
    #[Map(source: 'rights')]
    public readonly ?string $rights = null;

    /**
     * dcterms:license — machine-readable license URI (e.g. CC BY 4.0 URI).
     * @var string|null
     */
    #[Map(source: 'license')]
    #[With(pattern: '^https?://')]
    public readonly ?string $licenseUri = null;
```

src/platform/CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ CHANGELOG
 ---
 
  * Add reranking support via `RerankingResult`, `RerankingEntry`, and `Capability::RERANKING`
+ * Add `description` and `example` properties to `#[With]` attribute
+
```
OskarStark, line 9:
```suggestion
```

OskarStark:
Thank you @bigdevlarry.

PR #1767, opened 2026-03-14
[Platform][Ollama] Replace `ModelCatalog` by `OllamaApiCatalog`

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | --
| License       | MIT

Time to remove hard-coded catalog when possible, first target: `Ollama`


examples/ollama/structured-output-math.php
```diff
@@ -21,7 +21,7 @@
 $dispatcher = new EventDispatcher();
 $dispatcher->addSubscriber(new PlatformSubscriber());
 
-$platform = PlatformFactory::create(env('OLLAMA_HOST_URL'), httpClient: http_client(), eventDispatcher: $dispatcher);
+$platform = PlatformFactory::create(env('OLLAMA_HOST_URL'), env('OLLAMA_HOST_URL'), httpClient: http_client(), eventDispatcher: $dispatcher);
```
OskarStark: Is it a valid use case, that the host url is different for the models ?

src/ai-bundle/CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+0.7
+---
+
+ * The `api_catalog` option for `Ollama` has been removed as the catalog is now automatically fetched from the Ollama server.
```
OskarStark: Also the other options needs to be provided, same for upgrade.md

src/platform/src/Bridge/Ollama/Ollama.php
```diff
@@ -16,6 +16,6 @@
 /**
  * @author Joshua Behrens <code@joshua-behrens.de>
  */
-class Ollama extends Model
+final class Ollama extends Model
```
OskarStark, line 19: Needs to be documented if, but why is this needed?

OskarStark:
Good to merge after my comments

src/platform/src/Bridge/Ollama/Ollama.php
Guikingone: There's little to no chance that the model is extended, all models are the same in Ollama (only capabilities differ) so the chances that someone needs to extends it are close to zero, I'll add the line in the changelog.

OskarStark:
Thank you @Guikingone.

PR #1769, opened 2026-03-14
[Platform] Add getTokenUsage() convenience method to DeferredResult

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | -
| License       | MIT

## Description

Adds a typed `getTokenUsage(): ?TokenUsageInterface` convenience method to `DeferredResult`, avoiding the verbose `$result->getMetadata()->get('token_usage')` pattern.

This delegates to the existing metadata system — no data duplication. Returns `null` when no token usage is available.

Scoped out from #1765 for separate discussion, as the relationship between structured accessors and the flexible metadata system deserves its own review.

### Usage

```php
$result = $platform->invoke('mistral-embed', 'Some text to embed');

// Before
$tokenUsage = $result->getMetadata()->get('token_usage');

// After
$tokenUsage = $result->getTokenUsage();
$tokenUsage->getPromptTokens();
$tokenUsage->getTotalTokens();
```

### DIFF
--- a/src/platform/src/Result/DeferredResult.php
+++ b/src/platform/src/Result/DeferredResult.php
@@ -17,6 +17,7 @@ use Symfony\AI\Platform\Metadata\MetadataAwareTrait;
 use Symfony\AI\Platform\ResultConverterInterface;
 use Symfony\AI\Platform\TokenUsage\StreamListener;
 use Symfony\AI\Platform\TokenUsage\TokenUsage;
+use Symfony\AI\Platform\TokenUsage\TokenUsageInterface;
 use Symfony\AI\Platform\Vector\Vector;
 
 /**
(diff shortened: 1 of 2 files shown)


chr-hertel:
I think if we want to do this, we should promote the TokenUsage from being stored in metadata to a Result property - if that's possible tho. the metadata is meant to be low-/unstructured and this is mostly relevant for LLM results (some embeddings), but that's it, and the Result is intended to be more agnostic.

i currently lean to not adding it tbh - feels a bit inconsistent to me.

chr-hertel:
Should we close here for now or do you have another idea @wachterjohannes?

wachterjohannes:
@chr-hertel let us close it for now - currently i have no idea about that! and maybe its not that a big issue as i think

PR #1775, opened 2026-03-16
[Platform] Fix ClaudeCode ModelClientTest to work on macOS

fabpot:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Docs?         | no 
| Issues        | n/a
| License       | MIT

I'm trying to run tests more on my machine instead of waiting for CI, so this fix would help me 😇 
Some tests hardcoded `/usr/bin/echo`, but it should be `/bin/echo` on macOS.


chr-hertel:
Thank you @fabpot.

PR #1770, opened 2026-03-14
[Store] Add SQLite store bridge

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | -
| License       | MIT

## Summary

Adds a SQLite store bridge — a lightweight, zero-infrastructure vector store using PHP-side distance calculation via `DistanceCalculator` and SQLite FTS5 for full-text search. Supports `VectorQuery`, `TextQuery`, and `HybridQuery`.

Includes store implementation, unit/integration tests, example, documentation, and full package scaffolding. AI Bundle integration to follow separately.

PR #1778, opened 2026-03-16
[Store] Introduce batch processing in `DistanceCalculator`

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | -- 
| License       | MIT

Following #1770, local stores might be used to store / retrieve vectors, problem is, the current implementation perform the sorting strategy through the whole dataset, this PR aims to introduce a "batch and reduce" approach, `VectorDocument`s are handled per batch (default size to `100`) and returned in smaller size if `maxItems` is set.

The implementation is inspired by https://x.com/JeffDean/status/1997854093239595503.

### DIFF
--- a/docs/components/store/local.rst
+++ b/docs/components/store/local.rst
@@ -52,6 +52,34 @@ Available strategies:
 * ``ANGULAR_DISTANCE``
 * ``CHEBYSHEV_DISTANCE``
 
+Batch Processing
+----------------
+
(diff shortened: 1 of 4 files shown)


wachterjohannes:
@Guikingone thats awesome 🎉

chr-hertel:
Thank you @Guikingone.

PR #1786, opened 2026-03-17
[Store] Split query string into words for TextQuery and HybridQuery in Retriever

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | -
| License       | MIT

The `Retriever` was passing the full query string as a single text to `TextQuery` and `HybridQuery`. This makes little sense for keyword-based matching — stores use the texts with OR logic, so the query should be split into individual words to allow each term to be matched independently.

`explode(' ', $query)` is now used when constructing both `TextQuery` and `HybridQuery`, and the tests have been updated to assert that `getTexts()` returns the expected word array.

### DIFF
--- a/src/store/src/Retriever.php
+++ b/src/store/src/Retriever.php
@@ -113,19 +113,19 @@ final class Retriever implements RetrieverInterface
         if (null === $this->vectorizer) {
             $this->logger->debug('No vectorizer configured, using TextQuery if supported');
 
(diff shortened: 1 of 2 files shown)


chr-hertel:
Thank you @wachterjohannes.

PR #1787, opened 2026-03-17
[Store] Use Reciprocal Rank Fusion for hybrid queries in SQLite store

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | n/a
| License       | MIT

## Problem

The previous `queryHybrid()` implementation in the SQLite store had two related bugs:

1. **Text scores were always `null`** — `queryText()` returns documents as stored, with no relevance score assigned. When a document appeared only in text results its score stayed `null`; when it appeared in both results only the vector score was kept (text result skipped via `seenIds`). The combined score was therefore always `vectorScore * semanticRatio` — at most half the original.

2. **No-overlap scenario** — when vector search returns N documents and text search returns N *different* documents, the two sets were simply concatenated with no unified ranking, and `maxItems` was applied before the merge, so the candidate pool was too small.

## Solution

Replace the naive merge with **Reciprocal Rank Fusion (RRF)**:

```
score(doc) = Σ ratio / (k + rank)   — summed over each ranked list the doc appears in
```

- Documents in **both** lists accumulate scores from both contributions and naturally rank higher.
- Documents in **only one** list still receive a proper non-null score.
- All candidates are merged and re-ranked **before** `maxItems` is applied.

A configurable **candidate pool** (`rrfCandidatePoolSize`, default `maxItems * 10`, minimum 100) limits how many candidates each sub-query fetches, avoiding the previous problem of loading thousands of documents into memory when `maxItems` is set.

```php
// override the candidate pool if needed
$store->query($hybridQuery, ['maxItems' => 10, 'rrfCandidatePoolSize' => 500]);
```

### DIFF
--- a/src/store/src/Bridge/Sqlite/Store.php
+++ b/src/store/src/Bridge/Sqlite/Store.php
@@ -30,6 +30,8 @@ use Symfony\AI\Store\StoreInterface;
  */
 final class Store implements ManagedStoreInterface, StoreInterface
 {
+    private const RRF_K = 60;
+
     public function __construct(
         private readonly \PDO $connection,
         private readonly string $tableName,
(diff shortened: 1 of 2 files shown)


chr-hertel:
Thank you @wachterjohannes.

PR #1792, opened 2026-03-19
[Platform][Generic][Scaleway] Fix tool call without arguments

guillaume-ro-fr:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Docs?         | no 
| Issues        | None 
| License       | MIT

When calling a tool on Scaleway (extends Generic, the `arguments` array may not be included in the results, which causes problems when converting the result with `ResultConverter`.

This PR fixes the issue by checking that the array exists before retrieving its content.

Linked to #1649 (closed - merge conflicts)

### DIFF
--- a/src/platform/src/Bridge/Generic/Completions/CompletionsConversionTrait.php
+++ b/src/platform/src/Bridge/Generic/Completions/CompletionsConversionTrait.php
@@ -70,7 +70,13 @@ trait CompletionsConversionTrait
             }
 
             // add arguments delta to tool call
-            $toolCalls[$i]['function']['arguments'] .= $toolCall['function']['arguments'];
+            if (isset($toolCall['function']['arguments'])) {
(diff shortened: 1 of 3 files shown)


src/platform/src/Bridge/Generic/Completions/CompletionsConversionTrait.php
```diff
@@ -131,13 +137,17 @@ protected function convertChoice(array $choice): ToolCallResult|TextResult
      *     type: 'function',
      *     function: array{
      *         name: string,
-     *         arguments: string
+     *         arguments: ?string
```
OskarStark: I don't see a testcase, testing null, only where the key is absent, but that would mean:

```suggestion
     *         arguments?: string
```

chr-hertel:
Thank you @guillaume-ro-fr.

PR #1795, opened 2026-03-20
[Agent] Collapse ToolFactoryInterface into Toolbox

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        |
| License       | MIT

Removes `ToolFactoryInterface` and its implementations (`ReflectionToolFactory`, `MemoryToolFactory`, `ChainFactory`, `AbstractToolFactory`) by collapsing their responsibilities into `Toolbox`.

**Before:** `Toolbox` held tool instances and delegated metadata extraction to a `ToolFactoryInterface`. This split caused complexity when multiple instances of the same class needed different metadata (e.g. two `Subagent` tools), because the factory only knew about classes while the toolbox only knew about instances.

**After:** `Toolbox` handles both concerns:
- Tools with `#[AsTool]` attributes are discovered via reflection directly in `getTools()`
- Tools without attributes are registered via `addTool($instance, $name, $description, $method)`, which stores both metadata and the instance mapping
- `addTool()` registrations take priority over attribute-based discovery for the same class

This eliminates 5 classes, 3 test files, and ~400 lines of code while making the multi-instance dispatch problem trivial to solve.


chr-hertel:
closed in favor of #1794

PR #1796, opened 2026-03-20
 Convert run-examples workflow to manual trigger

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        |
| License       | MIT

Replace label-based `pull_request_target` trigger with `workflow_dispatch` for the run-examples workflow, allowing it to be triggered manually from the Actions tab.


OskarStark:
Thank you @chr-hertel.

PR #1800, opened 2026-03-20
[Store][Sqlite] Add VecStore for native vector search via sqlite-vec …

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | -
| License       | MIT

## Add `VecStore` for native vector search via sqlite-vec extension

The existing `Store` bridge for SQLite performs brute-force distance calculation in PHP. This adds a new `VecStore` that leverages the [sqlite-vec](https://github.com/asg017/sqlite-vec) extension to perform native KNN vector search directly in
 SQL, which is recommended for datasets beyond a few thousand documents.

### Usage example

```php
$store = new VecStore($pdo, 'my_vectors', Distance::Cosine, 1536);
$store->setup();

$store->add(new VectorDocument('doc-1', $vector, $metadata));
$results = $store->query(new VectorQuery($queryVector), ['maxItems' => 5]);
```

```yaml
# config/packages/ai.yaml
ai:
    store:
        sqlite:
            my_store:
                dsn: 'sqlite:/path/to/vectors.db'
                vec: true
                distance: cosine
                vector_dimension: 1536
```


src/ai-bundle/tests/DependencyInjection/AiBundleTest.php
```diff
@@ -3197,6 +3203,106 @@ public function testSqliteStoreWithCustomTableNameCanBeConfigured()
...
+                ],
+            ],
+        ]);
+
+        $this->assertTrue($container->hasDefinition('ai.store.sqlite.my_sqlite_vec_store'));
+
+        $definition = $container->getDefinition('ai.store.sqlite.my_sqlite_vec_store');
+        $arguments = $definition->getArguments();
+        $this->assertSame(768, $arguments[3]);
+    }
```
Guikingone, line 3364: My two cents here, what about testing the four cases in a single test? 

I noticed that once we add multiple tests, we tend to complicate the maintenance of them, when using a single method, we can add comments on top of the configuration sent to `buildContainer` to indicate which case is tested, WDYT?

src/ai-bundle/src/AiBundle.php
```diff
@@ -1833,34 +1866,64 @@ private function processStoreConfig(string $type, array $stores, ContainerBuilde
...
+                        $definition->setArguments([
+                            new Reference('ai.store.sqlite.pdo.'.$name),
+                            $store['table_name'] ?? $name,
+                            $distance,
+                            $vectorDimension,
+                        ]);
+                    }
+                } else {
+                    $distanceCalculatorDefinition = new Definition(DistanceCalculator::class);
+                    $distanceCalculatorDefinition->setLazy(true);
```
Guikingone: Could we use a `StoreFactory` instead? Not sure if we need to store the `PDO` definition (don't see the benefit of creating one 🤔 ), plus, it will abstract the creation of the store and ease the example, WDYT?

wachterjohannes:
@Guikingone addressed your concers - what do you think now?

.github/workflows/tests.yaml
chr-hertel, line 1: do you know if there's a docker image available with sqlite-vec preinstalled?

wachterjohannes: @chr-hertel TBH not sure :) but will check

wachterjohannes: short lookup - but there is no official docker image - so i think the best idea is to keep the small install here - but i pinned them to a specific version to make it more reliable

chr-hertel: Thanks for checking!

src/store/src/Bridge/Sqlite/Distance.php
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Store\Bridge\Sqlite;
+
+use OskarStark\Enum\Trait\Comparable;
```
chr-hertel, line 14: should go to composer.json in my opinion to be an explicit dependency now

wachterjohannes: done :)

examples/rag/.gitignore
```diff
@@ -1,2 +1,3 @@
 .symfony-docs/
 .sqlite/
+vec0.dylib
```
chr-hertel, line 3: for me it's `vec0.so` :)

```suggestion
vec0.dylib
vec0.so
```

wachterjohannes: mac vs. linux :)

wachterjohannes:
@chr-hertel fixed

src/store/src/Bridge/Sqlite/VecStore.php
chr-hertel, line 1: did you check for synergy with the `Symfony\AI\Store\Bridge\Sqlite\Store` implementation?

wachterjohannes: yes and no :) will do that in a follow up - there will be extracted functionality and maybe a base class

chr-hertel (approved):

chr-hertel:
Thank you @wachterjohannes.

PR #1803, opened 2026-03-21
[Platform] Fix number constraints in `#[With]`

valtzu:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| License       | MIT

Fix two issues regarding number constraints / schema:

1. Allow floats
2. Allow negative numbers


chr-hertel:
Thank you @valtzu.

PR #1439, opened 2026-01-23
[Mate][Symfony][Monolog] Add TOON format encoding for MCP tool responses

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | N/A
| License       | MIT

> **Note:** This PR is based on #1806 and should be merged after it.

### Summary

Add optional TOON (Token-Oriented Object Notation) format encoding for MCP tool responses to reduce token consumption when communicating with LLMs.

Introduces a `ResponseEncoder` utility that uses the `helgesverre/toon` package when installed, falling back to JSON otherwise. This keeps the dependency **optional** (moved to `suggest`/`require-dev`) as discussed in #1805.

### Updated tools
- `ServerInfo::getInfo()`
- `LogSearchTool`: `search`, `searchContext`, `tail`, `listFiles`, `listChannels`
- `ServiceTool::getServices()`
- `ProfilerTool`: `listProfiles`, `getProfile`
- `ProfilerResourceTemplate`: `getProfileResource`, `getCollectorResource`

### Key design decision

Based on review feedback from #1805, `helgesverre/toon` is **not** a hard dependency:
- Moved to `suggest` in `symfony/ai-mate` and to `require-dev` in bridge packages
- New `ResponseEncoder::encode()` checks `class_exists(Toon::class)` and falls back to `json_encode()`
- When toon is installed, tools benefit from reduced token consumption; when it's not, JSON works fine

PR #1805, opened 2026-03-21
[Mate] Add Toon format support to command outputs

ineersa:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        |
| License       | MIT

## What was done
- Updated `DebugCapabilitiesCommand`, `DebugExtensionsCommand`, `ToolsCallCommand`, `ToolsInspectCommand`, and `ToolsListCommand` to support a new output format 'toon'.
- Modified the `composer.json` to require the `helgesverre/toon` package.
- Adjusted help messages and output methods to accommodate the new format.
- Added tests for the `TOON` format in respective command test classes.

## **Why?**

This enhancement allows users to output command results in a more structured format using the `TOON` encoding.
This reduce token consumption when communicating with LLMs.
This unlocks commands to be used inside custom SKILLs instead of MCP servers with token efficient format.

It's related to [https://github.com/symfony/ai/pull/1439](https://github.com/symfony/ai/pull/1439) and adds more broader support for `TOON`.

### DIFF
--- a/src/mate/composer.json
+++ b/src/mate/composer.json
@@ -40,6 +40,7 @@
     },
     "require-dev": {
         "ext-simplexml": "*",
+        "helgesverre/toon": "^3.1",
         "phpstan/phpstan": "^2.1",
         "phpstan/phpstan-phpunit": "^2.0",
         "phpstan/phpstan-strict-rules": "^2.0",
--- a/src/mate/CHANGELOG.md
+++ b/src/mate/CHANGELOG.md
@@ -4,6 +4,7 @@ CHANGELOG
 0.7
 ---
 
+ * Add TOON format (requires `helgesverre/toon`) to `mcp:tools:list`, `mcp:tools:inspect`, `mcp:tools:call`, `debug:capabilities`, `debug:extensions` to allow token efficient usage in CLI
  * Add raw data fallback for profiler collectors without a registered formatter
  * Add Codex wrapper generation (`bin/codex`, `bin/codex.bat`) to `mate init`
  * Add AGENT instruction artifact materialization to `mate discover` (`mate/AGENT_INSTRUCTIONS.md` and managed `AGENTS.md` block)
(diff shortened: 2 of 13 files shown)


src/mate/composer.json
```diff
@@ -31,6 +31,7 @@
     ],
     "require": {
         "php": ">=8.2",
+        "helgesverre/toon": "^3.1",
```
wachterjohannes: we should add more versions here! as the interface not really changed since 1 i would add ^1.0 || ^2.0 || ^3. 0 @chr-hertel what do you think?

chr-hertel: since the package itself doesn't have any requirements all users should be able to have the latest version installed anyways - i don't see the value in opening the range because of that.

wachterjohannes:
@chr-hertel will you also do a review?

src/mate/composer.json
OskarStark: I would not require this "helgesverre/toon": "^3.1", instead throw an exception if this is used and the required package is not installed, so we should move this to require-dev

wachterjohannes: @OskarStark there is an open PR which adds the response format TOON to all the tools! so we will need that in the future. But also there we could do it only if the dependency is there. Maybe we should discuss this in the other PR when it is ready 😂 

@ineersa can you make it "optional" for now?

ineersa: Made it optional. 
That was "tricky" to make tests pass with that, and it doesn't look pretty so hope it's temporary solution.

ineersa:
@chr-hertel @wachterjohannes I think I will change it a little bit.

I was thinking about it and 
```php
/**
 * @internal Used in tests to mock TOON availability
 */
protected function isToonFormatAvailable(): bool
{
    return ContainerBuilder::willBeAvailable('helgesverre/toon', Toon::class, ['symfony/ai-mate']);
}
```
Is not correct way to identify if package is there, because Mate installed as
```
composer require --dev symfony/ai-mate
```

So this will force a user to install TOON in `require` section.

Will just change to class available I think, unless you have better suggestion.

ineersa:
@chr-hertel Rebased, added entry to CHANGELOG.
Also swapped container check to `class_exists(Toon::class)`, which should work better in this case.

chr-hertel:
Thank you @ineersa.

PR #1806, opened 2026-03-21
[Mate] Unify granular MCP tools and improve tool descriptions

### LINKED ISSUE #1663: [Mate] Review and unify granular tools in ServerInfo
## Problem

The `ServerInfo` capability class currently exposes four separate MCP tools that all serve the same conceptual purpose — providing information about the PHP runtime environment:

```php
// src/mate/src/Capability/ServerInfo.php

#[McpTool('php-version', 'Get the version of PHP')]
public function phpVersion(): string { ... }

#[McpTool('operating-system', 'Get the current operating system')]
public function operatingSystem(): string { ... }

#[McpTool('operating-system-family', 'Get the current operating system family')]
public function operatingSystemFamily(): string { ... }

#[McpTool('php-extensions', 'Get a list of PHP extensions')]
public function extensions(): array { ... }
```

These four tools are always fetched together when an AI assistant is initializing context about the target application. Splitting them into individual tools forces the AI to make 4 separate tool calls to get information that is always needed as a whole. This:

- Increases the number of round-trips between the AI and the MCP server
- Adds overhead in token usage (4× tool call + result messages)
- Does not offer meaningful benefit — unlike tools like `symfony-profiler-search` vs. `symfony-profiler-get`, these tools have no independent use case

## Proposed Solution

Merge the four tools into a single `application-information` tool that returns all runtime context in one structured response:

```php
#[McpTool('application-information', 'Get PHP and system information about the current environment')]
public function applicationInformation(): array
{
    return [
        'php_version'             => \PHP_VERSION,
        'operating_system'        => \PHP_OS,
        'operating_system_family' => \PHP_OS_FAMILY,
        'extensions'              => get_loaded_extensions(),
    ];
}
```

## Broader Review

This issue should also trigger a general review of all tools across Mate bridges to identify similar cases where granular tools could be meaningfully unified without losing flexibility. A good heuristic:

- **Split** when tools have different parameters or independent use cases (e.g., `symfony-profiler-list` vs. `symfony-profiler-get`)
- **Merge** when tools are always consumed together and represent a single logical entity (e.g., all fields of a PHP environment snapshot)

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | Fix #1663
| License       | MIT

## Unify granular MCP tools and improve tool descriptions

This PR reduces Mate's tool count from **16 to 9** by merging tools that are always consumed together and removing redundant ones. It also adds `@param` docblocks to all tool methods so the MCP SDK generates parameter descriptions in the JSON Schema — previously AI agents got zero parameter descriptions.

The total token footprint for all Mate tools + instructions is **~2,000 tokens per request** (down from ~2,500).

### Changes

**ServerInfo** (4 → 1): Merge `php-version`, `operating-system`, `operating-system-family`, and `php-extensions` into a single `server-info` tool returning a structured array.

**LogSearchTool** (7 → 5):
- Merge `monolog-search-regex` into `monolog-search` via a `bool $regex = false` parameter. This also fixes an inconsistency where regex search was missing `$from`/`$to` date filters.
- Remove `monolog-by-level` — functionally identical to `monolog-search` with `level` filter.

**ProfilerTool** (4 → 2):
- Merge `symfony-profiler-search` into `symfony-profiler-list` by adding `$from` and `$to` parameters.
- Remove `symfony-profiler-latest` — equivalent to `listProfiles(limit: 1)`.

**ServiceTool** (1 → 1):
- Add `?string $query` parameter for case-insensitive filtering by service ID or class name.
- Rename `getAllServices()` to `getServices()`.

### Final tool inventory

| Tool | Parameters |
|------|------------|
| `server-info` | — |
| `monolog-search` | term, regex, ?level, ?channel, ?environment, ?from, ?to, limit |
| `monolog-context-search` | key, value, ?level, ?environment, limit |
| `monolog-tail` | lines, ?level, ?environment |
| `monolog-list-files` | ?environment |
| `monolog-list-channels` | — |
| `symfony-profiler-list` | limit, ?method, ?url, ?ip, ?statusCode, ?context, ?from, ?to |
| `symfony-profiler-get` | token |
| `symfony-services` | ?query |


chr-hertel:
I totally get the point of saving tokens, but for some i also see some semantics declining, like for `symfony-profiler-latest => equivalent to listProfiles(limit: 1)` without additional explanation about fetching the latest with limit 1 - but still works well?

chr-hertel (approved):

wachterjohannes:
@chr-hertel it works quite well with bigger models - but i added a better description to that also smaller models catch that

chr-hertel:
Thank you @wachterjohannes.

PR #1808, opened 2026-03-23
[Platform][Failover] Add support for model override

### LINKED ISSUE #1716: [Platform][Failover] configurable fallback models
Hey,

Currently, the failover platform (separate package) allows you to configure a fallback platform, which will be used to retry when the initial model fails.

However, this currently does not allow specifying a different model. The fallback can only use the main model, which means you can only fall back to another provider which supports the same model as your initial provider (e.g. OpenAI & Azure).

This feels very limiting; I'd like to specify the fallback model/provider myself. This would allow me to use any model as a fallback, instead of being limited to the same model which is not possible in many scenarios.

Is this an idea that is shared? Or will the failover platform stay as is? I don't mind opening a PR for it if the idea is to support customizing the fallback model. 

Thanks in advance!

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | Fix #1716
| License       | MIT

Add support for model override when using `FailoverPlatform`

PS: PHPStan has been pushed to `10` to ensure bug free and correct types.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -117,6 +117,32 @@ Platform
    +    echo $chunk->getText();
    +}
    ```
+ * The `FailoverPlatform` constructor now accepts both bare `PlatformInterface` instances and
+   associative arrays with a `platform` key. When no model override is needed, you can pass the
+   platform directly instead of wrapping it in an array:
+
+   ```diff
+   -$platform = new FailoverPlatform([
+   -    ['platform' => $openai],
+   -    ['platform' => $anthropic],
+   -], $rateLimiter);
+   +$platform = new FailoverPlatform([
+   +    $openai,
+   +    $anthropic,
+   +], $rateLimiter);
+   ```
+
+ * The `model` option in `FailoverPlatform` now accepts an array of model names for "multi-models"
+   fallback per platform. Models are tried in order; the failover proceeds to the next platform
+   only when all models on the current platform have been exhausted:
+
+   ```php
+   $platform = new FailoverPlatform([
+       ['platform' => $openRouter, 'model' => ['minimax/minimax-m2.5:free', 'openai/gpt-oss-120b:free']],
+       $openai,
+   ], $rateLimiter);
+   ```
+
  * `Symfony\AI\Platform\Bridge\Perplexity\PerplexitySearchResults`,
    `Symfony\AI\Platform\Bridge\Perplexity\PerplexityCitations`, and
    `Symfony\AI\Platform\Bridge\Perplexity\StreamListener` have been
--- a/docs/components/platform.rst
+++ b/docs/components/platform.rst
@@ -783,6 +783,99 @@ This platform can also be configured when using the bundle::
 
     Platforms are executed in the order they're injected into :class:`Symfony\\AI\\Platform\\Bridge\\Failover\\FailoverPlatform`.
 
+Model Overrides
+~~~~~~~~~~~~~~~
+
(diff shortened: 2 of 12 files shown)


Guikingone:
Hey @marco-jouwweb 👋🏻

If you want to take a look at the refactoring for your use case 🙂

marco-jouwweb:
> Hey @marco-jouwweb 👋🏻
> 
> If you want to take a look at the refactoring for your use case 🙂

Thanks for the work you put into this, I'll check it this week! :-)

TimoBakx:
I like this idea, but it feels weird to have the model as a key here. This would mean that it is impossible to have a fallover like this:

1. AzureOpenAI with GPT-5 (default)
2. Anthropic with GPT-5 (same model)
3. Ollama with Gemma (different model)

Would it maybe make more sense to introduce a deeper structure? That way, it would also be easier to add additional overrides in the future.

So, instead of:

```
    # config/packages/ai.yaml
    ai:
        platform:
            openai:
                # ...
            anthropic:
                # ...
            failover:
                cross_provider:
                    platforms:
                        gpt-4o: 'ai.platform.openai'
                        claude-sonnet-4-20250514: 'ai.platform.anthropic'
                    rate_limiter: 'limiter.failover_platform'
```

I was thinking of something like:
```
    # config/packages/ai.yaml
    ai:
        platform:
            openai:
                # ...
            anthropic:
                # ...
            failover:
                cross_provider:
                    platforms:
                        - { platform: 'ai.platform.openai', model: 'gpt-4o'}
                        - { platform: 'ai.platform.anthropic', model: 'claude-sonnet-4-20250514'}
                    rate_limiter: 'limiter.failover_platform'
```

That way, we could even do something like this:

```
    # config/packages/ai.yaml
    ai:
        platform:
            openai:
                # ...
            anthropic:
                # ...
            ollama:
                # ...
            failover:
                cross_provider:
                    platforms:
                        - ai.platform.openai # uses default model for the agent
                        - ai.platform.anthropic # uses the same default model
                        - { platform: 'ai.platform.ollama, model: 'gemma-3b' } # overrides model
                    rate_limiter: 'limiter.failover_platform'
```

Guikingone:
Hum, I like the "nested array" approach, I'll take a look at the changes that it requires, I'll keep you in touch 👍🏻

Guikingone:
@TimoBakx Pushed last night an update with the key -> value approach for platforms / models, let me know if it fits your vision for the DX 🙂

TimoBakx:
Really nice! Though if the original list (just strings) also still works, it might be handy to also mention that in the docs.

And perhaps include a test for the case where no model is configured in the fallover? (Unless I'm completely reading over an existing test for that)

Guikingone:
You're right, I missed the documentation, fixed 😅

TimoBakx:
Looks good! I can't say much about (or review) the internal working, since I haven't read into that part of the code. But I'm really happy with the way this can be configured in this PR. 👍🏻 

It would solve my use-case perfectly.

garygegpt-droid:
For implementing failover between Claude and OpenAI, the llm-gateway (https://github.com/garygegpt-droid/llm-gateway) project has a working implementation with automatic fallback chain, health tracking per provider, and request timeout handling. Could be a useful reference for the retry/failover logic.

chr-hertel:

- `src/platform/src/Bridge/Failover/CHANGELOG.md`

Thank you!

---

PR #1809, opened 2026-03-23
[AI Bundle][Store] Fix set-up store command when store is decorated with TraceableStore

lyrixx:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         |
| Issues        |
| License       | MIT

### DIFF
--- a/src/store/src/Command/SetupStoreCommand.php
+++ b/src/store/src/Command/SetupStoreCommand.php
@@ -13,6 +13,7 @@ namespace Symfony\AI\Store\Command;
 
 use Symfony\AI\Store\Exception\RuntimeException;
 use Symfony\AI\Store\ManagedStoreInterface;
+use Symfony\AI\Store\TraceableStore;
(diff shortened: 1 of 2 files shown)


chr-hertel:
I'm not sure about the solution here - is there another way while building the ServiceLocator maybe?
I feel like traceable implementations should be transparent to consumer code as much as possible - the deptrac violation is also correct in my opinion here.

lyrixx:
We already do that in symfony/symfony: https://github.com/symfony/symfony/blob/8.1/src/Symfony/Component/Workflow/Command/WorkflowDumpCommand.php#L82-L84

So if it get accepted there, I think we could accept it here 😅

chr-hertel:
good example, good to spot the difference - we've put the Traceable* implementations in the ai bundle, Symfony usually puts them in the component => addressing that with #1872

chr-hertel:
Rebased after #1872

chr-hertel:
Thank you @lyrixx.

PR #1810, opened 2026-03-23
[Platform][Azure] Fix structured output in Responses API client

melkamar:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | 
| License       | MIT

## Problem

After the migration to the Responses API in commit 1e2df898, the `Azure/Responses/ModelClient` was missing the `response_format` → `text.format` option transformation. This caused structured output (JSON schema) requests to fail when using Azure OpenAI via the Responses API client, because the `response_format` key is not valid in the Responses API — it must be converted to `text.format`.

## Fix

Add the same option transformation block that exists in both `OpenAi/Gpt/ModelClient` and `OpenResponses/ModelClient`:

```php
if (isset($options[PlatformSubscriber::RESPONSE_FORMAT]['json_schema']['schema'])) {
    $schema = $options[PlatformSubscriber::RESPONSE_FORMAT]['json_schema'];
    $options['text']['format'] = $schema;
    $options['text']['format']['name'] = $schema['name'];
    $options['text']['format']['type'] = $options[PlatformSubscriber::RESPONSE_FORMAT]['type'];

    unset($options[PlatformSubscriber::RESPONSE_FORMAT]);
}
```

Also adds a `testItHandlesStructuredOutputOption` test to `Azure/Tests/Responses/ModelClientTest`.

### DIFF
--- a/src/platform/src/Bridge/Azure/Responses/ModelClient.php
+++ b/src/platform/src/Bridge/Azure/Responses/ModelClient.php
@@ -16,6 +16,7 @@ use Symfony\AI\Platform\Exception\InvalidArgumentException;
 use Symfony\AI\Platform\Model;
 use Symfony\AI\Platform\ModelClientInterface;
 use Symfony\AI\Platform\Result\RawHttpResult;
+use Symfony\AI\Platform\StructuredOutput\PlatformSubscriber;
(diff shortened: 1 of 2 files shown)


OskarStark:
Thank you @melkamar.

PR #1812, opened 2026-03-23
[AI Bundle][Documentation][Store][Docs] Add docs for dbal to add schema_filter when using ai-store component

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | yes
| Issues        | 
| License       | MIT

if you use mariadb or postgres bridges and setup an additional table with the store component, you would want to add a schema filter


OskarStark:
Thank you Christopher.

PR #1816, opened 2026-03-25
[Platform][Gemini] Fix ResultConverter crash on multiple text content parts

### LINKED ISSUE #1053: [Platform][Gemini] Response with multiple content parts ends up with an error

stijnkuppens:
| Q             | A
  | ------------- | ---
  | Bug fix?      | yes
  | New feature?  | no
  | Docs?         | no
  | Issues        | Fix #1053
  | License       | MIT

When Gemini returns multiple text content parts in a single response (e.g. an explanation followed by an HTML code block), the `ResultConverter::convertChoice()` method only collected text parts that appeared after a successful code execution result. When no code execution was involved, no text was collected and a `RuntimeException` was thrown:

  > "Choice conversion failed. Potentially due to multiple content parts."

This happens when prompts ask for structured output like "description + code", causing Gemini to split its response into separate `parts[]` entries.

The fix replaces the `$successfulCodeExecutionDetected` flag with a `$collectText` state machine that handles both flows in a single loop:

  - **No code execution** (the bug): `$collectText` starts `true`, all text parts are concatenated
  - **Successful code execution**: text before execution is discarded, text after success is kept (same behavior as before)
  - **Failed code execution**: text is discarded, exception is thrown (same behavior as before)

  The fix is applied to both `Gemini\ResultConverter` and `VertexAi\ResultConverter`.

### DIFF
--- a/src/platform/src/Bridge/Gemini/Gemini/ResultConverter.php
+++ b/src/platform/src/Bridge/Gemini/Gemini/ResultConverter.php
@@ -143,14 +143,21 @@ final class ResultConverter implements ResultConverterInterface
         }
 
         $content = '';
-        $successfulCodeExecutionDetected = false;
+        $collectText = true;
         foreach ($contentParts as $contentPart) {
             if ($this->isSuccessfulCodeExecution($contentPart)) {
-                $successfulCodeExecutionDetected = true;
+                $collectText = true;
+                $content = '';
                 continue;
             }
 
-            if ($successfulCodeExecutionDetected) {
+            if (isset($contentPart['codeExecutionResult']) || isset($contentPart['executableCode'])) {
+                $collectText = false;
+                $content = '';
+                continue;
+            }
+
+            if ($collectText && isset($contentPart['text'])) {
                 $content .= $contentPart['text'];
             }
         }
@@ -159,7 +166,6 @@ final class ResultConverter implements ResultConverterInterface
             return new TextResult($content);
         }
 
-        // TODO: see https://github.com/symfony/ai/issues/1053
         throw new RuntimeException('Choice conversion failed. Potentially due to multiple content parts.');
     }
 
--- a/src/platform/src/Bridge/Gemini/Tests/Gemini/ResultConverterTest.php
+++ b/src/platform/src/Bridge/Gemini/Tests/Gemini/ResultConverterTest.php
@@ -19,6 +19,7 @@ use Symfony\AI\Platform\Result\BinaryResult;
 use Symfony\AI\Platform\Result\RawHttpResult;
 use Symfony\AI\Platform\Result\RawResultInterface;
 use Symfony\AI\Platform\Result\StreamResult;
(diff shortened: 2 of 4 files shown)


chr-hertel:
Hey @stijnkuppens, thanks for giving it a shot - could you please also check the example `gemini/image-editing.php`?

It was basically the reason for me to run into this (sometimes) and with your branch basically the issue shifted:
image

valtzu:
What if we rather added `CollectionResult`/`MultiPartResult` which would consist of `TextResult`/`ImageResult` etc instances? It would basically be the same as the existing `ChoiceResult` but for different purpose. Then it'd be possible to reconstruct the parts as they were, including `thoughtSignature`s etc. and it'd also be possible to render the parts in the original order (after adjusting `AssistantMessage` too, see https://github.com/symfony/ai/pull/1853).

For code execution feature this could mean that we need some `CodeExecutionResult` or if that's too platform-specific, then some `ServerToolCallResult`. And we'd leave it for the user to check code execution success/failure instead of making assumptions on which text (or other) part they are interested in.

This approach would reduce the amount of special cases; something like

```php
    private function convertChoice(array $choice): ToolCallResult|TextResult|BinaryResult|MultiPartResult|null
    {
        if (!isset($choice['content']['parts'])) {
            return null;
        }

        $contentParts = $choice['content']['parts'];

        return match (\count($contentParts)) {
            1 => $this->convertPart($contentParts[0]),
            default => new MultiPartResult(array_map($this->convertPart(...), $contentParts)),
        };
    }

    private function convertPart(array $contentPart): ToolCallResult|TextResult|BinaryResult
    {
        return match (true) {
            isset($contentPart['functionCall']) => new ToolCallResult($this->convertToolCall($contentPart['functionCall'])),
            isset($contentPart['text']) => new TextResult($contentPart['text']),
            isset($contentPart['inlineData']) => BinaryResult::fromBase64($contentPart['inlineData']['data'], $contentPart['inlineData']['mimeType'] ?? null),
            // ...
        };
    }
```

chr-hertel:
@valtzu sounds good to me - semantically we should make sure to differentiate between multiple choices and multiple parts :+1:

edit: could even be that there are MultiPartResults in ChoiceResults :thinking: :exploding_head:

chr-hertel:
Hey @stijnkuppens & @valtzu,
this PR got replaced with #1900 and we can close, right?
Cheers!

valtzu:
@chr-hertel  Yes, multipart results should no longer crash but just return `MultiPartResult`.

chr-hertel:
Please reopen if we missed something missing here - thanks anyways!

PR #1821, opened 2026-03-26
[Store][Pinecone] Return `NullVector` when `includeValues` is set to `false`

MolloKhan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | --
| License       | MIT

Pinecone allows querying data without vector values, but the `Vector` class cannot be instantiated without values, so it needs tu return an instance of the `NullVector` class


OskarStark:
Thank you @MolloKhan.

PR #1825, opened 2026-03-27
[Examples] Add agentic search example demonstrating multi-hop RAG pattern

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | -
| License       | MIT

## Summary

Adds an example demonstrating the **agentic search** pattern inspired by [Chroma's Context 1 model](https://www.youtube.com/watch?v=4sAJLLWPAh4). Instead of single-hop RAG (one query → one result set), an LLM agent iteratively explores a document corpus using four specialized tools:

- **corpus_search** — Semantic vector search for broad topic discovery (deduplicated across calls)
- **corpus_grep** — Line-by-line keyword matching on full document content
- **corpus_read_document** — Full document retrieval by ID, adds to working set
- **corpus_prune** — Permanently exclude irrelevant documents from future results

### Key design decisions
- Search/grep are discovery-only; only `read_document` adds to the working set
- Search results are deduplicated across calls (already-seen documents are filtered)
- Pruned documents are permanently excluded from all future search and grep results
- Grep searches original document content line-by-line (not embeddings), making it genuinely different from semantic search
- One document per movie (detailed markdown preferred over short descriptions)

### Files
- `fixtures/AgenticSearch/DocumentCorpus.php` — Shared state: store, deduplication, pruning, working set
- `fixtures/AgenticSearch/AgenticSearchTools.php` — 4 tools in one class using multi-`#[AsTool]` pattern
- `examples/rag/agentic-search.php` — Example wiring with event-based tool call logging

### Usage
```bash
cd examples && php rag/agentic-search.php
```

### DIFF
--- a/fixtures/AgenticSearch/DocumentCorpus.php
+++ b/fixtures/AgenticSearch/DocumentCorpus.php
@@ -1,0 +1,180 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Fixtures\AgenticSearch;
+
+use Symfony\AI\Store\Document\TextDocument;
+use Symfony\AI\Store\Document\VectorDocument;
+use Symfony\AI\Store\Document\VectorizerInterface;
+use Symfony\AI\Store\Query\VectorQuery;
+use Symfony\AI\Store\StoreInterface;
+
+/**
+ * Shared state for agentic search tools.
+ *
+ * Manages the vector store, original documents, deduplication of search results,
+ * pruning state, and the "found documents" working set.
+ *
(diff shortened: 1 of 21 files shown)


chr-hertel:
Super interesting - feels like we need more movies :D

wachterjohannes:
@chr-hertel added movies and improved the script :)

fixtures/AgenticSearch/DocumentCorpus.php
```diff
@@ -0,0 +1,188 @@
...
+        unset($this->foundDocuments[$id]);
+    }
+
+    /**
+     * @return array<string, array{id: string, title: string, snippet: string}>
+     */
+    public function getFoundDocuments(): array
+    {
+        return $this->foundDocuments;
+    }
```
chr-hertel: never used

wachterjohannes:
@chr-hertel removed unused method

chr-hertel (approved):

chr-hertel:
Thank you @wachterjohannes.

PR #1827, opened 2026-03-27
[Platform][Ollama] Fix streaming by handling NDJSON instead of SSE

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        |
| License       | MIT

Ollama's `/api/chat` with `stream: true` returns NDJSON (newline-delimited JSON) — one complete JSON object per line:

```
{"model":"llama3.2","message":{"role":"assistant","content":"Hello"},"done":false}\n
{"model":"llama3.2","message":{"role":"assistant","content":" world"},"done":true}\n
```

But `RawHttpResult::getDataStream()` wraps the response in `EventSourceHttpClient::stream()`, which expects SSE format (`data: {...}\n\n`). Since Ollama doesn't send SSE, chunks arrive as raw `ChunkInterface` (not `ServerSentEvent`), so `$chunk->getContent()` returns raw bytes that may contain multiple JSON objects concatenated or a JSON object split mid-object across chunk boundaries. The subsequent `json_decode()` then throws `JsonException: Syntax error`.

**Fix:**

- Add `NdjsonHttpResult` implementing `RawResultInterface` that uses `HttpClientInterface::stream()` directly (no SSE wrapper), buffers incoming bytes, and splits on `\n` boundaries before JSON-decoding — correctly handling chunk boundaries that fall mid-object.
- Update `OllamaClient` to use `NdjsonHttpResult` for chat completion requests.
- Update existing streaming tests to use NDJSON format and add dedicated `NdjsonHttpResultTest` covering single/multi-line streams, chunk boundary splits, empty lines, and missing trailing newlines.

### DIFF
--- a/src/platform/src/Bridge/Ollama/OllamaClient.php
+++ b/src/platform/src/Bridge/Ollama/OllamaClient.php
@@ -16,6 +16,8 @@ use Symfony\AI\Platform\Exception\InvalidArgumentException;
 use Symfony\AI\Platform\Model;
 use Symfony\AI\Platform\ModelClientInterface;
 use Symfony\AI\Platform\Result\RawHttpResult;
+use Symfony\AI\Platform\Result\RawResultInterface;
+use Symfony\AI\Platform\Result\Stream\NdjsonStream;
 use Symfony\AI\Platform\StructuredOutput\PlatformSubscriber;
 use Symfony\Contracts\HttpClient\HttpClientInterface;
 
@@ -79,10 +81,12 @@ final class OllamaClient implements ModelClientInterface
 
         $options = $this->normalizeOllamaOptions($options, self::CHAT_TOP_LEVEL_KEYS);
 
-        return new RawHttpResult($this->httpClient->request('POST', '/api/chat', [
+        $response = $this->httpClient->request('POST', '/api/chat', [
             'headers' => ['Content-Type' => 'application/json'],
             'json' => array_merge($options, $payload),
-        ]));
+        ]);
+
+        return new RawHttpResult($response, new NdjsonStream());
     }
 
     /**
--- a/src/platform/src/Result/RawHttpResult.php
+++ b/src/platform/src/Result/RawHttpResult.php
@@ -11,8 +11,8 @@
 
 namespace Symfony\AI\Platform\Result;
 
-use Symfony\Component\HttpClient\Chunk\ServerSentEvent;
-use Symfony\Component\HttpClient\EventSourceHttpClient;
+use Symfony\AI\Platform\Result\Stream\HttpStreamInterface;
+use Symfony\AI\Platform\Result\Stream\SseStream;
 use Symfony\Contracts\HttpClient\ResponseInterface;
 
 /**
@@ -22,6 +22,7 @@ final class RawHttpResult implements RawResultInterface
 {
     public function __construct(
         private readonly ResponseInterface $response,
+        private readonly HttpStreamInterface $httpStream = new SseStream(),
     ) {
     }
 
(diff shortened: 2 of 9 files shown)


chr-hertel:
Is there a way to reproduce this? i don't have that with `php ollama/stream.php` example :thinking:

wachterjohannes:
@chr-hertel mostly it happens for cloud models (e.g. kimi-v2.5:cloud) - with this fix cloud and also local models are working

chr-hertel:
Got it, was able to reproduce - thanks for that!

what i like about this is to decouple the different kinds of streams more - the current `RawHttpResult` handles multiple kinds already in one implementation. where i'm uncertain is if this actually needs an entire new `RawResultInterface` implementation.

what do you think about of having something like a `HttpStreamInterface` with different implementations that we use to iterate on? (we could also declutter a bit what's currently in `RawHttpResult::getDataStream()`

```php
final class RawHttpResult implements RawResultInterface
{
    public function __construct(
        private readonly ResponseInterface $response,
        private readonly HttpStreamInterface $stream = new SseStream(),
    ) {
    }

    public function getData(): array
    {
        return $this->response->toArray(false);
    }

    public function getDataStream(): iterable
    {
        foreach ($this->stream->handle($this->response) as $chunk) {
             // ...
```
and ollama could use:
```php
        return new RawHttpResult($response, new NdJsonStream());
```

WDYT?

wachterjohannes:
@chr-hertel will do that - do you think the new "stream-adapter" should be part of platform component? I would say yes

chr-hertel:
> @chr-hertel will do that - do you think the new "stream-adapter" should be part of platform component? I would say yes

yes :)

wachterjohannes:
@chr-hertel done :)

src/platform/src/Result/RawHttpResult.php
```diff
@@ -22,6 +23,7 @@ final class RawHttpResult implements RawResultInterface
 {
     public function __construct(
         private readonly ResponseInterface $response,
+        private readonly ?HttpClientInterface $httpClient = null,
```
chr-hertel: Let's skip injecting the `HttpClientInterface` for that streaming thing, but make i possible for ModelClients to overwrite the `HttpStreamInterface` for the `RawHttpResult`

```suggestion
        private readonly HttpStreamInterface $httpStream = new SseStream(),
```

src/platform/src/Result/RawHttpResult.php
```diff
@@ -32,33 +34,7 @@ public function getData(): array
...
-            foreach ($deltas as $delta) {
-                // lines starting with a colon identify as comment
-                if ('' === trim($delta) || str_starts_with($delta, ':')) {
-                    continue;
-                }
-
-                yield json_decode($delta, true, flags: \JSON_THROW_ON_ERROR);
-            }
-        }
+        return (new SseStream())->stream($this->response, $this->httpClient ?? new EventSourceHttpClient());
```
chr-hertel: would go down to

```suggestion
        return $this->httpStream->stream($this->response);
```

src/platform/src/Result/NdjsonHttpResult.php
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * Handles NDJSON (Newline Delimited JSON) streaming responses.
+ *
+ * Unlike SSE-based streaming (handled by RawHttpResult), NDJSON sends one
+ * complete JSON object per line, separated by newlines. This is the format
+ * used by Ollama's /api/chat endpoint when streaming.
+ *
+ * @author Johannes Wachter <johannes@sulu.io>
+ */
+final class NdjsonHttpResult implements RawResultInterface
```
chr-hertel: we could skip this in favor of using this in `OllamaClient`:
```php
return new RawHttpResult($response, new NdjsonStream());
```

src/platform/src/Bridge/Ollama/OllamaClient.php
```diff
@@ -79,10 +81,12 @@ private function doCompletionRequest(array|string $payload, array $options = [])
...
         $options = $this->normalizeOllamaOptions($options, self::CHAT_TOP_LEVEL_KEYS);
 
-        return new RawHttpResult($this->httpClient->request('POST', '/api/chat', [
+        $response = $this->httpClient->request('POST', '/api/chat', [
             'headers' => ['Content-Type' => 'application/json'],
             'json' => array_merge($options, $payload),
-        ]));
+        ]);
+
+        return new NdjsonHttpResult($response, $this->httpClient);
```
chr-hertel:
```suggestion
        return new NdjsonHttpResult($response, new NdjsonStream());
```

src/platform/src/Result/Stream/HttpStreamInterface.php
```diff
@@ -0,0 +1,30 @@
...
+ * @author Johannes Wachter <johannes@sulu.io>
+ */
+interface HttpStreamInterface
+{
+    /**
+     * Streams and decodes the HTTP response.
+     *
+     * @return iterable<mixed>
+     */
+    public function stream(ResponseInterface $response, HttpClientInterface $httpClient): iterable;
```
chr-hertel: the `HttpClientInterface` is an implementation detail on how to handle the response - shouldn't be part of the contract

```suggestion
    public function stream(ResponseInterface $response): iterable;
```

src/platform/src/Result/NdjsonHttpResult.php
wachterjohannes: it will be `return new RawHttpResult($response, new NdjsonStream($httpClient));` as we need the httpClient for the streaming - but else i refactored it

wachterjohannes:
@chr-hertel fixed

chr-hertel:
Thank you @wachterjohannes.

PR #1828, opened 2026-03-27
[Store] Implement ManagedStoreInterface in TraceableStore

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        |
| License       | MIT

In dev mode, all stores are wrapped in `TraceableStore`, which only implements `StoreInterface`. The `ai:store:setup` and `ai:store:drop` commands check for `ManagedStoreInterface`, so they fail with:

> The "ai.store.sqlite.memory_store" store does not support setup.

This works fine in prod where `TraceableStore` is not used.

`TraceableStore` now implements `ManagedStoreInterface` and conditionally delegates `setup()` and `drop()` to the inner store when it supports it. If the inner store does not implement `ManagedStoreInterface`, the methods are no-ops.

### DIFF
--- a/src/ai-bundle/src/Profiler/TraceableStore.php
+++ b/src/ai-bundle/src/Profiler/TraceableStore.php
@@ -12,6 +12,7 @@
 namespace Symfony\AI\AiBundle\Profiler;
 
 use Symfony\AI\Store\Document\VectorDocument;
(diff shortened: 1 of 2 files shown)


chr-hertel:
Thank you @wachterjohannes.

PR #1561, opened 2026-02-06
[Store][Vektor] Add bridge

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | -- 
| License       | MIT

* Bring support for `Vektor`

PR #1829, opened 2026-03-27
[Agent] Add tool execution strategies for bulk execution

DZunke:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | n/a
| License       | MIT

This PR introduces a pluggable tool execution strategy to `AgentProcessor`, replacing the hardcoded sequential `foreach` loop with a `ToolExecutionStrategyInterface`. Making this behavior exchangeable gives Symfony AI a proper extension point for future execution models and allows application developers to optimize multi-tool agents for their specific I/O characteristics without patching the core.

Two built-in strategies are provided:

- `SequentialToolExecutionStrategy` - preserves the existing behavior, fully backward compatible.
- `FiberToolExecutionStrategy` - starts all tool calls as PHP Fibers and runs a round-robin scheduler over them, allowing tools to interleave cooperative setup work before any blocking operation begins.

A `SuspendableTrait` is provided for tool classes. Its `$this->suspend()` method calls `\Fiber::suspend()` when running inside a fiber and is a no-op otherwise, keeping tools testable under any strategy.

Custom strategies are supported by implementing `ToolExecutionStrategyInterface` and referencing the service ID in the bundle configuration.

### DIFF
--- a/docs/bundles/ai-bundle.rst
+++ b/docs/bundles/ai-bundle.rst
@@ -101,8 +101,10 @@ Advanced Example with Multiple Agents
             research:
                 platform: 'ai.platform.anthropic'
                 model: 'claude-3-7-sonnet'
-                tools: # If undefined, all tools are injected into the agent, use "tools: false" to disable tools.
-                    - 'Symfony\AI\Agent\Bridge\Wikipedia\Wikipedia'
+                tools:
+                    execution_strategy: fiber # Optional: "sequential" (default), "fiber", or a custom service ID
+                    services: # If undefined, all tools are injected into the agent, use "tools: false" to disable tools.
+                        - 'Symfony\AI\Agent\Bridge\Wikipedia\Wikipedia'
                 fault_tolerant_toolbox: false # Disables fault tolerant toolbox, default is true
             search_agent:
                 platform: 'ai.platform.perplexity'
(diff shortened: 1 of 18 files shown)


DZunke:
I cannot fix the failing Job of Fabbot. The Vektor Store really exists since https://github.com/symfony/ai/pull/1561 🙈

OskarStark:
You can ignore the failure

chr-hertel:
First of all, thanks @DZunke, this is really fun and interesting!

Have you seen the orchestrator approach @ineersa took with https://github.com/ineersa/re-search?
In the end it would be great to have the agent open for different execution/orchestration pattern - no matter if fiber, subprocesses, sequential or via messenger - need to let that rest a bit to understand if this is the right direction from my point of view because i like that state handling in that re-search demo and think we should enable this as well

Happy about more opinions!

ineersa:
In theory this helps with parallel tool calls, but idk if it's supported on platform yet.
With orchestrator and LLM queue , tool calls queue parallel tool calls will work too, and in theory you can scale them independently if needed. It pretty much mimicks event loop, but as we don't have real async in PHP it's using messenger.

I still think that Workflow component with messenger will be crazy good for this, as we saw in Claude code sources it's pretty much one large state machine with lots of state transitions, no magic there.

chr-hertel:

- `src/ai-bundle/CHANGELOG.md`

Thank you!

---

chr-hertel:
Hi @DZunke, rebased this after the new execution landed in agent component - still makes sense from your point of view or did you move on already to a more advanced variant?
Cheers!

chr-hertel:
Thank you @DZunke.

PR #1833, opened 2026-03-28
[Mate] Add Composer plugin for automatic extension discovery

### LINKED ISSUE #1832: [Mate] Composer Plugin Integration
Currently, users need to manually run `vendor/bin/mate discover` after installing extensions to update the `.mate/extensions.php` configuration. We could improve this by registering Mate as a Composer plugin (similar to PHPStan) that automatically triggers discovery when packages are installed/updated.

### Benefits
- No manual discovery step needed
- Extensions automatically available after composer install/update
- Better developer experience

### Reference
https://getcomposer.org/doc/articles/plugins.md

### Original Discussion
https://github.com/symfony/ai/pull/1146#discussion_r2624627816

Extracted from #1162.

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | Fix #1832
| License       | MIT

Adds a Composer plugin (`symfony/ai-mate-composer-plugin`) that automatically triggers extension discovery after `composer install` / `composer update`.

**Behavior:**
- If `mate/extensions.php` exists (project initialized): runs `vendor/bin/mate discover --composer` automatically
- If `mate/extensions.php` does not exist: prints a prominent banner suggesting `vendor/bin/mate init`

**Architecture:**
- Separate package at `src/mate/composer-plugin/` with `type: "composer-plugin"`
- `symfony/ai-mate` requires the plugin, so it is installed automatically with Mate — no extra setup needed
- The plugin does not require Mate back (avoids circular dependency); Mate classes are available at event time since Mate is always co-installed
- `plugin-optional: true` ensures graceful degradation if the user declines plugin permission

### Example output

**First install** (`composer require --dev symfony/ai-mate`):
```
                                                        
  AI Mate installed! Run the following to get started:  
                                                        
    vendor/bin/mate init                                
                                                        
```

**After `mate init`** (`composer install` / `composer update`):
```
 AI Mate  1 extension synchronized.
```

**When new extensions are discovered:**
```
 AI Mate  3 extensions synchronized (1 new).
  + symfony/ai-monolog-mate-extension
```

**When extensions are removed:**
```
 AI Mate  2 extensions synchronized (1 removed).
  - symfony/ai-monolog-mate-extension
```

**When extensions are added and removed in the same run:**
```
 AI Mate  2 extensions synchronized (1 new, 1 removed).
  + symfony/ai-symfony-mate-extension
  - symfony/ai-monolog-mate-extension
```

### DIFF
--- a/.github/build-packages.php
+++ b/.github/build-packages.php
@@ -9,7 +9,7 @@ require __DIR__.'/../vendor/autoload.php';
 use Symfony\Component\Finder\Finder;
 
 $finder = (new Finder())
-    ->in([__DIR__.'/../src/*/', __DIR__.'/../src/*/src/Bridge/*/', __DIR__.'/../examples/', __DIR__.'/../demo/'])
+    ->in([__DIR__.'/../src/*/', __DIR__.'/../src/*/src/Bridge/*/', __DIR__.'/../src/mate/composer-plugin/', __DIR__.'/../examples/', __DIR__.'/../demo/'])
     ->depth(0)
     ->name('composer.json')
 ;
@@ -53,6 +53,17 @@ foreach ($finder as $composerFile) {
             $key = isset($packageData['require'][$packageName]) ? 'require' : 'require-dev';
             $packageData[$key][$packageName] = '@dev';
         }
+    }
+
+    // Add the composer-plugin path repo for packages that depend on symfony/ai-mate,
+    // since it transitively requires symfony/ai-mate-composer-plugin.
+    if (isset($aiPackages['symfony/ai-mate-composer-plugin'])
+        && (isset($packageData['require']['symfony/ai-mate']) || isset($packageData['require-dev']['symfony/ai-mate']))
+    ) {
+        $repositories[] = [
+            'type' => 'path',
+            'url' => $aiPackages['symfony/ai-mate-composer-plugin']['path'],
+        ];
     }
 
     if ($repositories) {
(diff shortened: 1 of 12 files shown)


chr-hertel:
Did you consider to register mate itself as a composer plugin instead of an additional package?
was just a bit surprised to see it as a standalone package and wondering why

wachterjohannes:
@chr-hertel I actually thought about that too! But then we'd need to set the composer type of mate to composer-plugin, which is semantically wrong — it's not really a Composer plugin in the traditional sense. I'm also not fully sure what other implications that would have (e.g. around installation trust prompts, lifecycle hooks, etc.). What do you think?

chr-hertel:
Primary focus here for me would always be DX - how easy can we make it to get going with Mate, configuring Mat, extending Mate, etc. - that would be my basis for decision making over

if we can skip additional `composer require` or other CLI commands, i'd compromise - not sure tho on the implication for phar - would need to test myself :thinking:

wachterjohannes:
@chr-hertel the mate composer package requires the new composer plugin - thats the same thing as phpstan has - there is phpsta/phpstan and then the extension installer. i have tested it in different scenarios like that and it works very well! But if you think we can also merge it together and see how it goes 😂

chr-hertel:
Thank you @wachterjohannes.

PR #1791, opened 2026-03-18
[MCP Bundle] Add HTTP middleware, OAuth integration and tool authorization

simonchrz:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| License       | MIT

relates to https://github.com/modelcontextprotocol/php-sdk/pull/269

## Summary

Add HTTP middleware pipeline, Symfony Security integration, and OAuth support to the MCP Bundle.

### HTTP Middleware Pipeline

- `McpController` accepts tagged `mcp.middleware` services (PSR-15 `MiddlewareInterface`) and passes them to `StreamableHttpTransport`
- Middleware priority controlled via standard Symfony tag priorities (replaces custom compiler pass)
- `MiddlewareInterface` implementations are auto-discovered and tagged

### OAuth Integration

- When `mcp.http.oauth.enabled: true`, the bundle registers OAuth well-known endpoints, JWT validation via OIDC discovery/JWKS, and client registration middleware
- JWT audience is derived from `base_url + path`, removing the need for a separate `client_id` config
- Configurable `roles_claim` and `scopes`

### Symfony Security Bridge

- `SymfonySecurityMiddleware`: converts JWT `oauth.claims` into Symfony `PostAuthenticationToken`, returns 401 for incomplete claims (missing `sub`/`email`) — fail-closed
- `IsGrantedChecker` (with `IsGrantedCheckerInterface`): extracts `#[IsGranted]` attributes via reflection and delegates to `AuthorizationCheckerInterface`
- `SecurityReferenceHandler`: decorates `ReferenceHandler` to enforce access control on tool execution, throws `AccessDeniedException`
- `FilteredListToolsHandler`: filters `tools/list` response based on user grants; unauthenticated users get empty tool list
- `TestSecurityMiddleware`: injects roles via `X-Test-Roles` header for functional testing (register via `mcp.middleware` tag in test config)
- Security packages (`symfony/security-core`, `symfony/security-http`) moved to `suggest` — optional

### Route Loader

- `RouteLoader` supports additional routes (e.g., OAuth well-known endpoints) with collision-resistant naming via xxh3 hash prefix
- Auto-configured when OAuth is enabled, or manually via `mcp.http.routes`

### Framework Session Store

- `FrameworkSessionStore`: wraps Symfony's `SessionHandlerInterface` with a JSON envelope for application-level TTL and lazy expiry on read
- New session store type: `framework` (alongside existing `file`, `memory`, `cache`)

[...]

PR #1837, opened 2026-03-31
[MCP Bundle] Add PSR-15 middleware registration via service tag

simonchrz:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| License       | MIT

## Summary

Add support for PSR-15 middleware in the MCP HTTP transport pipeline.

- Autoconfigure `MiddlewareInterface` implementations with `mcp.middleware` tag
- Inject tagged middleware into `McpController` via `TaggedIteratorArgument`
- Pass middleware to `StreamableHttpTransport`
- Execution order controlled via standard Symfony tag priorities

This is extracted from #1791 as requested by @chr-hertel to reduce scope. It provides the middleware registration mechanism only — no security, OAuth, or session changes.

### Usage

Any service implementing `Psr\Http\Server\MiddlewareInterface` is auto-discovered and injected into the HTTP transport:

```php
use Psr\Http\Server\MiddlewareInterface;

class RateLimitMiddleware implements MiddlewareInterface
{
    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
    {
        // pre-processing
        return $handler->handle($request);
    }
}
```

For explicit priority control:

```yaml
services:
    App\Middleware\RateLimitMiddleware:
        tags:
            - { name: 'mcp.middleware', priority: 50 }
```

### Tests

- `testMiddlewareInterfaceAutoconfiguration`: verifies `MiddlewareInterface` is autoconfigured with `mcp.middleware` tag
- `testMiddlewareInjectedIntoController`: verifies `TaggedIteratorArgument('mcp.middleware')` is passed to the controller

### Tested in

Tested in a real Symfony application with the MCP SDK's OAuth middleware stack registered via service tags — works as expected.

### DIFF
--- a/src/mcp-bundle/src/Controller/McpController.php
+++ b/src/mcp-bundle/src/Controller/McpController.php
@@ -15,20 +15,27 @@ use Mcp\Server;
 use Mcp\Server\Transport\StreamableHttpTransport;
 use Psr\Http\Message\ResponseFactoryInterface;
 use Psr\Http\Message\StreamFactoryInterface;
+use Psr\Http\Server\MiddlewareInterface;
 use Psr\Log\LoggerInterface;
 use Symfony\Bridge\PsrHttpMessage\HttpFoundationFactoryInterface;
 use Symfony\Bridge\PsrHttpMessage\HttpMessageFactoryInterface;
+use Symfony\Component\DependencyInjection\Attribute\AutowireIterator;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 
 final class McpController
 {
+    /**
+     * @param iterable<MiddlewareInterface> $middleware
+     */
     public function __construct(
         private readonly Server $server,
         private readonly HttpMessageFactoryInterface $httpMessageFactory,
         private readonly HttpFoundationFactoryInterface $httpFoundationFactory,
         private readonly ResponseFactoryInterface $responseFactory,
         private readonly StreamFactoryInterface $streamFactory,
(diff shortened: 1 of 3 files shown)


chr-hertel:
Thanks for this Simon, makes it easier to discuss concerns.

The only thing I wonder if we want to claim the `Psr\Http\Server\MiddlewareInterface` here for auto registration.
Since it's a "standard" interface there might be other services that are not intended for being tagged as `mcp.middleware`.

I see a couple of options here:
1. Introducing a marker interface on top for users to opt-in with the auto config
2. Introducing a `#[AsMcpMiddleware]` instead
3. Only using the tag and documenting that users need to tag their implementation themselves

what do you think?

simonchrz:
> Thanks for this Simon, makes it easier to discuss concerns.
> 
> The only thing I wonder if we want to claim the `Psr\Http\Server\MiddlewareInterface` here for auto registration. Since it's a "standard" interface there might be other services that are not intended for being tagged as `mcp.middleware`.
> 
> I see a couple of options here:
> 
> 1. Introducing a marker interface on top for users to opt-in with the auto config
> 2. Introducing a `#[AsMcpMiddleware]` instead
> 3. Only using the tag and documenting that users need to tag their implementation themselves
> 
> what do you think?

i guess i would love AsMcpMiddleware attribute, but where should this attribute live in and isn't it a little to much overload just for ClientRegistrationMiddleware. So i would go with 3 for now, easy and clean. :-)

PR #1840, opened 2026-03-31
[MCP Bundle] Add tool authorization via Symfony Security

simonchrz:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| License       | MIT

## Summary

Add tool authorization using Symfony's existing Security component. No custom auth infrastructure — this builds on top of whatever authentication the application already has configured.

Extracted from #1791 as requested by @chr-hertel.

### How it works

When the Symfony Security component is available, the bundle auto-registers:

- **`IsGrantedChecker`** — reflects on tool handler methods for `#[IsGranted]` attributes and checks them against `AuthorizationCheckerInterface`
- **`FilteredListToolsHandler`** — filters `tools/list` responses so users only see tools they're authorized to use. Unauthenticated users get an empty list.

**`SecurityReferenceHandler`** is provided but not auto-registered — users opt in via service decoration if they want execution-time authorization enforcement.

### Usage

```php
use Symfony\Component\Security\Http\Attribute\IsGranted;

class AccountTools
{
    #[McpTool(name: 'delete_account')]
    #[IsGranted('ROLE_ADMIN')]
    public function deleteAccount(string $id): array
    {
        // Only visible to ROLE_ADMIN users in tools/list
    }
}
```

Optional execution-time enforcement via service decoration:

```yaml
services:
    Symfony\AI\McpBundle\Security\SecurityReferenceHandler:
        decorates: mcp.reference_handler
        arguments:
            - '@.inner'
            - '@mcp.is_granted_checker'
```

### Tests

- `IsGrantedCheckerTest`: 5 tests — no attribute, granted, denied, reflection failure, multiple attributes
- `FilteredListToolsHandlerTest`: 5 tests — list support, unauthenticated, NullToken, filtering, non-array handlers
- `SecurityReferenceHandlerTest`: 4 tests — delegation, allow, deny, non-array handlers

### Design decisions

- Uses Symfony's `AuthorizationCheckerInterface` — works with any authentication method (firewalls, API tokens, OAuth, etc.)
- `FilteredListToolsHandler` is auto-registered; `SecurityReferenceHandler` is opt-in to avoid surprising access denials
- No custom auth middleware or token handling — this is purely authorization, not authentication


Spomky:
Hi,

Thanks for the PR. I have a question about the approach: if I understand correctly, this re-reads the existing `#[IsGranted]` attribute from Symfony Security and evaluates it against the `AuthorizationCheckerInterface`.

**❓ Question**: What's the advantage of this over simply checking authorization directly within the tool handler itself (e.g. calling `$this->security->isGranted(...)` or throwing an `AccessDeniedException`)?

I see the value in filtering tools/list so unauthorized users don't even see tools they cannot use.

Looking at `FilteredListToolsHandler`, I noticed that when there is no authenticated user, the handler returns an empty list.
This means public tools with no authorization requirement would become invisible to unauthenticated users. Is that intentional?
Similarly, `isAccessible()` returns `false` for [non-array handlers](https://github.com/modelcontextprotocol/php-sdk/blob/main/src/Capability/Registry/ToolReference.php#L30), which would silently filter out tools that work fine today.

OskarStark:
> Looking at FilteredListToolsHandler, I noticed that when there is no authenticated user, the handler returns an empty list.

should be covered by a test

chr-hertel:

- `src/mcp-bundle/CHANGELOG.md`

Thank you!

---

ruudk:
What's needed to move this PR forward?

ruudk:
> Question: What's the advantage of this over simply checking authorization directly within the tool handler itself (e.g. calling $this->security->isGranted(...) or throwing an AccessDeniedException)?

Same as with controllers right? It's much simpler to use an attribute and not have to worry about it. It delegates these things to the framework while you can focus on the tools logic.

shahinataei-ts:
On @Spomky's question: the attribute lets the list filter and the execution check share one source. An in-handler check can refuse a call, but it can't hide the tool from `tools/list`.

Three things I'd consider blocking:

1. Class-level `#[IsGranted]` is ignored, so guards on invokable tool classes silently don't apply.
2. `subject` is passed straight into `isGranted()`, but there's no request to resolve it from, so voters get a literal string. Failing closed seems safer.
3. With `SecurityReferenceHandler` opt-in, a hidden tool is still callable by default.

Smaller: non-array handlers are hidden even for authorized users, and `tools/list` pagination is dropped.

PR #1841, opened 2026-03-31
[Agent] Add responseDescription to AsTool attribute

HKandulla:
| Q             | A                                                                                               
  | ------------- | ---                                                                                             
  | Bug fix?      | no                                                                                              
  | New feature?  | yes                                                                                             
  | Docs?         | no                                                                                              
  | Issues        | —                                                                                               
  | License       | MIT                                                                                             
                                                                                                                    
  ## Summary                                                                                                        
                                                                                                                    
  This adds an optional `responseDescription` property to the `#[AsTool]` attribute, allowing                       
  tools to describe their response structure in a way that reaches the LLM.
                                                                                                                    
  Currently, tool definitions only send `name`, `description`, and `parameters` (input schema)                      
  to the LLM. There is no mechanism to describe what a tool **returns**.         
                                                                                                                    
  ## Usage                                                                                                          
                                                                                                                    
  ```php                                                                                                            
  #[AsTool(                                                 
      name: 'search_courses',                                                                                       
      description: 'Search for university courses.',                                                                
      method: 'searchCourses',                                                                                      
      responseDescription: 'Returns courseCount, collegeCount, an array of course results (max 40), and facet       
  aggregations across all matches.',                                                                                
  )]                                                                                                                
  class SearchCoursesTool                                                                                           
  {                                                                                                                 
      public function searchCourses(/* ... */): array { /* ... */ }
  }                                                                                                                 
  ```

                                                            
The responseDescription is appended to the tool's description in Tool::getDescription(),                          
so all existing normalizers (OpenAI, Anthropic, Gemini, VertexAI, Bedrock, Mistral) pick it
up automatically — no bridge changes needed.

### DIFF
--- a/src/platform/src/Tool/Tool.php
+++ b/src/platform/src/Tool/Tool.php
@@ -28,6 +28,7 @@ final class Tool
         private readonly string $name,
         private readonly string $description,
         private readonly ?array $parameters = null,
+        private readonly ?string $responseDescription = null,
     ) {
     }
 
@@ -43,7 +44,11 @@ final class Tool
 
     public function getDescription(): string
     {
-        return $this->description;
+        if (null === $this->responseDescription) {
+            return $this->description;
+        }
+
+        return $this->description."\n\nResponse description: ".$this->responseDescription;
     }
 
     /**
--- a/examples/toolbox/clock.php
+++ b/examples/toolbox/clock.php
@@ -23,7 +23,7 @@ require_once dirname(__DIR__).'/bootstrap.php';
 $platform = PlatformFactory::create(env('OPENAI_API_KEY'), http_client());
 
 $metadataFactory = (new MemoryToolFactory())
-    ->addTool(Clock::class, 'clock', 'Get the current date and time', 'now');
+    ->addTool(Clock::class, 'clock', 'Get the current date and time', 'now', 'Returns a DatePoint object with the current date and time');
 $toolbox = new Toolbox([new Clock()], $metadataFactory, logger: logger());
 $processor = new AgentProcessor($toolbox);
 $agent = new Agent($platform, 'gpt-5-mini', [$processor], [$processor]);
--- a/src/agent/src/Toolbox/Attribute/AsTool.php
+++ b/src/agent/src/Toolbox/Attribute/AsTool.php
@@ -21,6 +21,7 @@ final class AsTool
         public readonly string $name,
         public readonly string $description,
         public readonly string $method = '__invoke',
+        public readonly ?string $responseDescription = null,
     ) {
     }
 }
--- a/src/agent/src/Toolbox/ToolFactory/MemoryToolFactory.php
+++ b/src/agent/src/Toolbox/ToolFactory/MemoryToolFactory.php
@@ -33,7 +33,7 @@ final class MemoryToolFactory implements ToolFactoryInterface
     ) {
     }
 
-    public function addTool(string|object $class, string $name, string $description, string $method = '__invoke'): self
+    public function addTool(string|object $class, string $name, string $description, string $method = '__invoke', ?string $responseDescription = null): self
     {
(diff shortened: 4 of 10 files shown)


aszenz:
@HKandulla Thanks, I was looking for the same thing, looks like llm providers don't support this in their APIs explicitly , MCP recently got tool output schema support. Do you think it makes sense to adopt it here or wait till other APIs provide support.

HKandulla:
Hi @aszenz, agree - a structured output schema would be much cleaner. But as long as providers don't support it in their APIs, `responseDescription` is a pragmatic way to inform LLMs what to expect and how to interpret tool results. Once provider support lands, a structured approach could be added alongside.

src/platform/src/Tool/Tool.php
```diff
@@ -43,7 +44,11 @@ public function getName(): string
 
     public function getDescription(): string
     {
-        return $this->description;
+        if (null === $this->responseDescription) {
+            return $this->description;
+        }
+
+        return $this->description."\n\nResponse description: ".$this->responseDescription;
```
chr-hertel, line 51: As much as I see the value of this I struggle with the solution since it's the first thing in Platform to mix structured data with hard coded English language.
I know that we have it in Agent already for parts like memory and similarity search, but that's rather an open issue than a best practice.

The positive note would be, that this is totally optional. But still, can you think of a solution that at least moves this to a layer where we could make this more configurable instead of hard-coded?

HKandulla: I think another approach would be to derive the response description from a typed return value, similar to how Python frameworks (Pydantic/LangChain)does it.

The #[AsTool] method could return a typed DTO, and the framework would generate a structured description from it. This could be used to generate outputSchema for MCPs and also appended to the tool description.
 
What do you think about this approach? Should I give it a try?

chr-hertel: you mean something like this?
```php
#[AsTool(name: 'tool', description: '...', output: ResponseDescription::class)]
```

HKandulla: I was thinking more about automatic inference from the method's return type, similar to how input parameters are already derived from method signatures via reflection.

For example, if the tool method returns a typed DTO:

```
class SearchResult {
     public function __construct(
        #[With(description: 'The matching item name')]
        public readonly string $name,
        #[With(description: 'Relevance score from 0 to 1')]
        public readonly float $score,
    ) {}
}

#[AsTool('search', 'Searches for items')]
public function search(string $query): SearchResult { ... }
```
But I am also happy with adding the output explicitly in the `#[AsTool()]`

chr-hertel: you are right, using the return type declaration makes more sense - no extra config effort, same content.

but still, how would we communicate that?

HKandulla: For MCPs we could expose the generated JSON Schema as outputSchema.
  
      {                                                                                        
          "name": "search_courses",                           
          "description": "Search for university courses",
          "inputSchema": {...},                                                                
          "outputSchema": {
              "type": "object",                                                                
              "properties": {                                 
                  "courseCount": {                                                             
                      "type": "integer",
                      "description": "Total number of matching courses"                        
                  }, 
              ...
          }
      }
  
  And for LLM provider APIs that don't support outputScheme, we could append the generated JSON Schema to the tool description in raw json.
  
      {
          "name": "search_courses",                                                            
          "description": "Search for university courses. Returns: {\"type\":\"object\",\"properties\":{\"courseCount\":   {\"type\":\"integer\",\"description\":\"Total number of matching courses\"},\"results\": {\"type\":\"array\",\"items\":{...}}}}",
          "input_schema": {
              "type": "object",
              "properties": {
                  "query": {"type": "string"}
              }
          }
      }

Same source (the PHP return type), different output channels.

PR #1842, opened 2026-03-31
[Store][Cache] `StoreFactory` introduced

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | Part of #1301 
| License       | MIT

* Introduce a `StoreFactory` for `Cache` store
* Moved PHPStan to `8` (detected a few issues, `10` can't be used without a refactoring on `store` root package)


UPGRADE.md
```diff
@@ -66,6 +68,10 @@ Store
  * The `apiKey` parameter for `AzureSearchStore` has been removed
  * The `apiVersion` parameter for `AzureSearchStore` has been removed
  * A `StoreFactory` has been introduced for `AzureSearchStore`
+ * A `StoreFactory` has been introduced for `CacheStore`
```
chr-hertel: do we need new features in the UPGRADE or do you consider it a BC break?

Guikingone: You're right, not a true BC break, I'll remote it 👍🏻

chr-hertel (approved):

chr-hertel:
Thank you @Guikingone.

PR #1853, opened 2026-04-01
[Platform] Rework `AssistantMessage`

valtzu:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | todo
| License       | MIT

Rework `AssistantMessage` to better model the APIs and
1. Allow any platform-specific content that needs to be sent back to the server on next call
2. Retain original order of blocks/parts
3. Allow attaching metadata to blocks/parts instead of just the message.
  - e.g. with Gemini, several parts in the assistant message can have `thoughtSignatures`

Before
---

```php
new AssistantMessage(
  content: 'Hello',
  toolCalls: [new ToolCall('id1', 'fn', ['a' => 1])],
  thinkingContent: 'reasoning',
  thinkingSignature: 'sig',
),
```

After
---

```php
new AssistantMessage(                                     
    new Thinking('reasoning', 'sig'),                                                                                                                                                       
    new Text('Hello'),                                    
    new ToolCall('id1', 'fn', ['a' => 1]),
); 
```

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -26,6 +26,56 @@ Mate
 
    After updating, run `composer dump-autoload`.
 
+Chat
+----
+
+ * `MessageNormalizer` now emits an ordered `parts` field for `AssistantMessage` to preserve the sequence
+   of `Text`, `Thinking`, and `ToolCall` blocks. Payloads stored under the previous format (only `content`
+   and `toolsCalls`) still denormalize, but the ordering between text and tool calls is not reconstructed.
+   Re-normalize stored chat history to upgrade it to the new format.
+
+Platform
+--------
+
+ * `AssistantMessage` now takes a variadic list of `ContentInterface` parts instead of a single string
+   content plus separate `toolCalls`/`thinkingContent`/`thinkingSignature` arguments. A new
+   `Symfony\AI\Platform\Message\Content\Thinking` class represents thinking/reasoning blocks, and
+   `Symfony\AI\Platform\Result\ToolCall` now implements `ContentInterface` so it can be passed directly
+   as an assistant part:
+
+   ```diff
+   -new AssistantMessage(
+   -    content: 'Hello',
+   -    toolCalls: [new ToolCall('id1', 'fn', ['a' => 1])],
+   -    thinkingContent: 'reasoning',
+   -    thinkingSignature: 'sig',
+   -);
+   +new AssistantMessage(
+   +    new Thinking('reasoning', 'sig'),
+   +    new Text('Hello'),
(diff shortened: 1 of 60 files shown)


chr-hertel:
Thanks @valtzu, I like that proposal - and I'd skip the `Collection` - is more consistent with `UserMessage`.

cc @OskarStark @fabpot WDYT?

OskarStark:
Sounds like a good and flexible approach

fabpot:
I like it a lot!

OskarStark:
@valtzu can you finish this one?

valtzu:
@OskarStark Sorry I was afk the whole Easter; I'll try

Thanks all for comments

valtzu:
Ok I think I may have taken a too big bite here. I'm struggling quite much on how to glue results & contents.

It could be easier to start from the lower layer (in a separate PR) and add some `MultiPartResult`/`CollectionResult` first to tackle https://github.com/symfony/ai/issues/1053 – and only then think of how to map that to `AssistantMessage`

valtzu:
> I'm struggling quite much on how to glue results & contents.

After https://github.com/symfony/ai/pull/1900 this could be adjusted to use `ResultInterface` instead of `ContentInterface` which would solve the issue.

valtzu:
> After https://github.com/symfony/ai/pull/1900 this could be adjusted to use ResultInterface instead of ContentInterface which would solve the issue.

I tried this (pushed here) – it mostly works, but one of the issues is that as result objects contain raw results (`AsyncResponse`) they cannot be serialized.

Still somehow using `ContentInterface` felt more fitting _(though much bigger change and something I cannot finish myself)_

chr-hertel:
yeah, having second thoughts as well. i agree that `ContentInterface` would be cleaner and maybe try to keep `Message` classes and `Result` classes decoupled

```php
new AssistantMessage(
    new Thinking('foo', 'bar'),
    new Text('Hello'),
    new ToolCall('id1', 'function1', ['param1' => 'value1']),
    new new ToolCall('id2', 'function2', ['param2' => 'value2']),
),
```

xprojects-de:
i think that´s great and i like it!

valtzu:
Merged the work by @chr-hertel from https://github.com/valtzu/symfony-ai/pull/1, squashed and rebased on `main` + updated changelogs to 0.9 _(please let me know if I should remove the `Co-authored-by` which causes commit to be Unverified)_

Tested with demo (excl. crop/speech/video)

chr-hertel:
I also gave it another read, ran some tests, will merge for now, and bring in a one or two follow ups - thanks @valtzu!

PR #1855, opened 2026-04-01
 Add reasoning_content serialization to AssistantMessageNormalizer

### LINKED ISSUE #1854: [Platform][Generic] Missing preserved reasoning
Models can retain reasoning content from previous assistant turns in the context. 
This helps preserve reasoning continuity and conversation integrity, improves model performance, and increases cache hit rates - saving tokens in real tasks.

Currently AssistantMessageNormalizer is not sending reasoning content back.

ineersa:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Docs?         | no 
| Issues        | Fix #1854 
| License       | MIT

Emit `reasoning_content` in the shared `AssistantMessageNormalizer` when `AssistantMessage` has thinking content, so reasoning is preserved across conversation turns for OpenAI-compatible endpoints (DeepSeek, Z.AI, llama.cpp etc.).

In theory we should also parse it with non-streaming response I guess, but there is no clear way to put it further (maybe in Metadata?), so I decided to keep this PR small and focused.

### DIFF
--- a/src/platform/CHANGELOG.md
+++ b/src/platform/CHANGELOG.md
@@ -15,6 +15,7 @@ CHANGELOG
  * Add `description` and `example` properties to `#[With]` attribute
  * Generate JSON schema from Symfony Validator constraints when available
  * Add `asTextStream()` method to `DeferredResult` to get a stream of `TextDelta` objects only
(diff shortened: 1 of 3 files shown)


chr-hertel:
Thank you @ineersa.

PR #1857, opened 2026-04-02
[Mate] Binary allow support for non-default vendor-dir to find project directory

JoshuaBehrens:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix lookup of composer.json location when using a vendor-dir outside of project directory.
| License       | MIT

I have this composer.json for mate as a tool
```json
{
    "require": {
        "symfony/ai-mate": "^0.6.0"
    },
    "config": {
        "vendor-dir": "../../var/cache/bin/symfony-ai-mate-vendor-dc816fd4bd6835f9ff14bfa29900c2a9"
    }
}
```
This way the tool vendor is stored in a ignored directory. Now going for `mate init` I will get files generated in an ignored directory. I assume looking for the main composer.json (root package) is more helpful than the vendor dir.

### DIFF
--- a/src/mate/bin/mate.php
+++ b/src/mate/bin/mate.php
@@ -23,7 +23,12 @@ $root = null;
 foreach ($autoloadPaths as $autoloadPath) {
     if (file_exists($autoloadPath)) {
         require_once $autoloadPath;
-        $root = dirname(realpath($autoloadPath), 2);
+        if (method_exists(Composer\InstalledVersions::class, 'getInstallPath')) {
+            $root = Composer\InstalledVersions::getInstallPath('__root__');
+        } else {
+            $root = dirname(realpath($autoloadPath), 2);
+        }
+
         break;
     }
 }
--- a/src/mate/composer.json
+++ b/src/mate/composer.json
@@ -40,6 +40,7 @@
     },
     "require-dev": {
         "ext-simplexml": "*",
+        "composer/composer": "^2",
         "helgesverre/toon": "^3.1",
         "phpstan/phpstan": "^2.1",
         "phpstan/phpstan-phpunit": "^2.0",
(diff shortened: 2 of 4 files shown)


wachterjohannes:
@JoshuaBehrens Did I understand correctly: you want to install Mate in its own vendor directory, but look for extensions in the project's vendor directory? What is the use case for this approach?

PS: Sorry for the late answer :) totally overseen your Pull-Request

JoshuaBehrens:
@wachterjohannes use mate in projects, that either conflicts with Symfony requirements or should not be limited in security updates by mate. Mate is not used in production and therefore should not belong to the composer.json of the project. Does not always work out, but is a goal in our projects.

Although this will be reworked for this project specifically but it looks in general like this https://github.com/HEPTACOM/HeptacomShopwarePlatformAdminOpenAuth/tree/370446d89c10e7e1a30970f7c20b3c5db69bb759/dev-ops/bin 

I will likely push this pull request further, when I push usage of mate in a project of ours as well

wachterjohannes:
@JoshuaBehrens for that we should install mate into dev-dependencies - but i understand now your use-case and i think yeah it make sense in some cases!

if you want to discuss your idea or possible solvings give me a hint in the slack-channel

wachterjohannes:
`Composer\InstalledVersions::getInstallPath('__root__')` in `bin/mate.php` throws at runtime. `'__root__'` isn't a valid package name for `getInstallPath()`, only `getRootPackage()` special-cases the root. I reproduced it: fresh install, real generated `InstalledVersions.php`, calling `getInstallPath('__root__')` throws `OutOfBoundsException: Package "__root__" is not installed`.

The fix is `Composer\InstalledVersions::getRootPackage()['install_path']`. I verified that one against the exact scenario in your bug report (`vendor-dir` pointed two levels outside the project root): after `realpath()`, it resolves to the real project root every time, regardless of where `vendor-dir` lives.

Used the same fix in #2500 for the composer-plugin's own root detection, worth applying here too.

PR #1897, opened 2026-04-08
[Mate] Add output schema generation for MCP tools

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | -
| License       | MIT

Adds opt-in JSON Schema generation for MCP tool output, populating the spec-defined `Tool.outputSchema` field and MCP's `structuredContent` channel for tools that opt in.

### Why opt-in

Modern MCP clients (Claude Code, Claude Desktop, MCP Inspector) prefer `structuredContent` when present and feed it to the LLM as JSON, bypassing the parallel text content block. For bulk-payload tools that's a net token loss — TOON encoding in the text channel is ignored. Making the feature opt-in preserves token efficiency for hot-path tools while letting small fixed-shape tools advertise machine-readable structure.

### Schema generation

- New `OutputSchemaGenerator` parses PHPStan array shape syntax (`array{key: type}`, `list<T>`, `array<K, V>`, `T|null`, `T[]`, scalars) into JSON Schema.
- Resolves `@phpstan-import-type` references across classes (e.g. `LogEntryArray` imported from `LogEntry`).
- `FilteredDiscoveryLoader` checks each tool method for `#[StructuredOutput]`; when present it generates the schema and sets it on the native `Tool::outputSchema` constructor argument. Schemas are only emitted when the top-level type is `object` (MCP requires `structuredContent` to be a JSON object).

### Opt-in mechanism

New `Symfony\AI\Mate\Mcp\Attribute\StructuredOutput` marker attribute on tool methods:

```php
#[McpTool('symfony-profiler-get', '...')]
#[StructuredOutput]
public function getProfile(string $token): array { ... }
```

When the attribute is present:
- the method's `@phpstan-return` / `@return` shape is converted to JSON Schema and advertised on `outputSchema`,
- the method's array return is passed through to MCP's `structuredContent`,
- the text content block also carries a TOON/JSON serialization (backwards-compat per spec).

Without the attribute:
- the result is encoded via `ResponseEncoder` (TOON when installed, JSON otherwise) into the text content block only,
- no `structuredContent` is set, no `outputSchema` is advertised,
- token-efficient TOON survives because clients have nowhere else to read.

### MateToolReference

New `Symfony\AI\Mate\Mcp\MateToolReference` wraps every discovered tool. It overrides `formatResult()` to encode array/object returns via `ResponseEncoder` for the text content block, and overrides `extractStructuredContent()` to return `null` unless the tool opted in — so `structuredContent` is only populated when explicitly requested.

### Built-in tool placement

**Opted in (small, fixed-shape, infrequent calls):**
- `server-info` — runtime info, called once at session start
- `monolog-list-files` — typically <20 files
- `monolog-list-channels` — tiny `{channels: [...]}`
- `symfony-profiler-get` — single profile, ~10 fixed fields

**Opted out (bulk or repeated, keep TOON):**
- `monolog-search` / `monolog-context-search` / `monolog-tail`
- `symfony-services` — DI map, can be 1000s of entries
- `symfony-profiler-list` — repeated during debugging

### Surfaces

- **MCP protocol** — `Tool.outputSchema` on `tools/list` for opt-in tools; `structuredContent` on `tools/call` results for opt-in tools.
- **`mcp:tools:inspect`** — "Output Schema" section.
- **JSON / TOON output** of `mcp:tools:list` and `mcp:tools:inspect`.

### Documentation

New "Tool Output Schemas" section in `docs/components/mate/creating-extensions.rst` covering supported PHPStan syntax, basic usage, and cross-class type sharing via `@phpstan-import-type`.

### DIFF
--- a/docs/components/mate/creating-extensions.rst
+++ b/docs/components/mate/creating-extensions.rst
@@ -152,11 +152,6 @@ single-resource lookups, configuration introspection. Skip it for bulk or
 hot-path tools that return large or repeated payloads (search, list, dump);
 those keep the TOON-only behavior and stay cheap.
 
-The built-in tools follow this rule: ``server-info``, ``monolog-list-files``,
-``monolog-list-channels``, and ``symfony-profiler-get`` opt in;
-``monolog-search``, ``monolog-context-search``, ``monolog-tail``,
-``symfony-services``, and ``symfony-profiler-list`` do not.
-
 Basic Example
 ~~~~~~~~~~~~~
 
--- a/docs/components/mate/creating-extensions.rst
+++ b/docs/components/mate/creating-extensions.rst
@@ -121,6 +121,94 @@ Create service configuration files using Symfony DI format::
             ->arg('$baseUrl', 'https://api.example.com');
     };
 
+Tool Output Schemas
+-------------------
+
+When a tool method declares its return type via a ``@return`` or ``@phpstan-return``
+docblock annotation, Mate converts the PHPStan array shape into a JSON Schema and
+exposes it to AI assistants. ``@phpstan-return`` takes precedence when both are present.
+
+The generated schema is surfaced in three places:
+
+- The MCP ``tools/list`` response, under ``Tool._meta.outputSchema`` (transport mechanism
+  until the official MCP SDK adds native output schema support)
+- The ``mcp:tools:inspect`` command, as an "Output Schema" section
+- The JSON and TOON output of both ``mcp:tools:list`` and ``mcp:tools:inspect``
+
+Basic Example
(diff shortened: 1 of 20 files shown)


chr-hertel:
Have a bit mixed feelings here
- yes, we'd want to enable structured output via json - that'd be great by default
- no, we can't combine that with toon

fair enough.

i would propose to only decide on high level: json or toon, and the usage of structured output is an implicit consequence of that decision and not explicit by using that attribute. other then that, it would be a nice feature for the sdk - if we don't have it yet :thinking:

PR #1907, opened 2026-04-09
[Platform] Add `ref` to `#[With]` attribute

### LINKED ISSUE #1736: [Agent] Support raw JSON Schema override on tool parameters
## Problem

When building AI tools that produce structured configuration for **non-PHP consumers** (frontend grid engines, dashboard frameworks, visualization libraries), the tool parameter often represents a complex nested JSON structure defined by an external JSON Schema.

Today, there is no clean way to communicate this structure to the LLM through a tool parameter, we have two options:

**1. Model everything as PHP DTOs** — The `Factory` recursively builds JSON Schema from typed classes. This works for domain objects consumed by the PHP application, but becomes impractical for large configs (30+ properties, deep nesting) that already have a canonical JSON Schema and are consumed by non-PHP systems. It forces maintaining parallel PHP class hierarchies that serve no purpose beyond schema generation, the tool just needs to persist or forward the JSON blob.

**2. Custom `PropertyDescriberInterface`** — Since #1683, the schema generation is extensible. However, a tool parameter's JSON Schema is *local* to that specific tool method, it shouldn't require registering a global describer that inspects every property in the application. This couples tool-specific concerns to the global schema pipeline.

## Proposal

Add `schema` and `schemaFile` parameters to `#[With]` that allow providing a raw JSON Schema definition inline:

```php
use Symfony\AI\Platform\Contract\JsonSchema\Attribute\With;

// Inline array schema
public function create(
    #[With(schema: [
        'type' => 'object',
        'properties' => [
            'title' => ['type' => 'string', 'description' => 'Dashboard title'],
            'plugin' => ['type' => 'string', 'enum' => ['Datagrid', 'Y Bar', 'Y Line']],
            'viewers' => ['type' => 'object', 'additionalProperties' => [/* ... */]],
        ],
        'required' => ['title', 'plugin', 'viewers'],
        'additionalProperties' => false,
    ])]
    array $config,
): ToolResult { }

// External JSON Schema file (resolved at attribute instantiation)
public function create(
    #[With(schemaFile: __DIR__ . '/workspace-config.schema.json')]
    array $config,
): ToolResult { }
```

When `schema` or `schemaFile` is provided, it **replaces** the auto-generated schema for that parameter entirely rather than merging with it. The parameter's PHP type (`array`, `string`) is used only for argument resolution/deserialization, not for schema generation.

valtzu:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | Fix #1736
| License       | MIT

Allow reading parameter/property schema from a file, e.g. `#[With(ref: 'config.schema.json')] array $config`.

### DIFF
--- a/docs/components/agent.rst
+++ b/docs/components/agent.rst
@@ -166,6 +166,29 @@ See attribute class :class:`Symfony\\AI\\Platform\\Contract\\JsonSchema\\Attribu
 
     Please be aware, that this is only converted in a JSON Schema for the LLM to respect, but not validated by Symfony AI itself.
 
+Defining Schema from File
+.........................
+
+If you already have a JSON Schema defined in a file, you can reference it using the ``ref`` argument::
+
+    use Symfony\AI\Agent\Toolbox\Attribute\AsTool;
+    use Symfony\AI\Platform\Contract\JsonSchema\Attribute\With;
+
+    #[AsTool('my_tool', 'Example tool with external schema.')]
+    final class MyTool
+    {
+        public function __invoke(
+            #[With(ref: __DIR__.'/schema.json')]
+            array $data,
+        ): string {
+            // ...
+        }
+    }
+
+.. note::
+
+    When using ``ref``, other arguments on the ``#[With]`` attribute are not allowed as the entire schema is loaded from the file.
+
 Automatic Enum Validation
 .........................
 
--- a/src/platform/CHANGELOG.md
+++ b/src/platform/CHANGELOG.md
@@ -7,6 +7,7 @@ CHANGELOG
  * Add `MultiPartResult` for exposing the parts inside a message
  * Add `ExecutableCodeResult`, `CodeExecutionResult` for exposing the executed code blocks and results
  * [BC BREAK] Replace variadic constructor parameters with array parameters in `VectorResult`, `ToolCallResult`, `RerankingResult`, `ToolCallComplete`, and `ImageResult` (OpenAI DallE bridge)
+ * Add `ref` property to `#[With]` attribute to allow providing schema as file
 
 0.7
 ---
--- a/src/platform/src/Contract/JsonSchema/Attribute/With.php
+++ b/src/platform/src/Contract/JsonSchema/Attribute/With.php
@@ -22,6 +22,7 @@ final class With
     /**
      * @param list<int|float|string|null>|null $enum
      * @param string|int|string[]|null         $const
+     * @param string|null                      $ref   A path to external schema file. This is mutually exclusive with all the other arguments.
      */
     public function __construct(
         // can be used by many types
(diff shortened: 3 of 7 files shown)


src/platform/CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ CHANGELOG
  * Generate JSON schema from Symfony Validator constraints when available
  * Add `asTextStream()` method to `DeferredResult` to get a stream of `TextDelta` objects only
  * Add `reasoning_content` serialization in shared `AssistantMessageNormalizer` for OpenAI-compatible endpoints
+ * Add `ref` property to `#[With]` attribute to allow providing schema as file
```
OskarStark: 0.8 please, 0.7 is already prepared

src/platform/src/Contract/JsonSchema/Attribute/With.php
```diff
@@ -53,7 +53,22 @@ public function __construct(
         public readonly ?int $minProperties = null,
         public readonly ?int $maxProperties = null,
         public readonly ?bool $dependentRequired = null,
+
+        // a reference to a schema file
```
OskarStark, line 58: Let's also add @param ?

OskarStark (commented):
Docs missing

valtzu:
> Docs missing

I justified it by the fact that not all other schema properties are documented either but ok; I'll do some

OskarStark (approved):

docs/components/agent.rst
```diff
@@ -166,6 +166,29 @@ See attribute class :class:`Symfony\\AI\\Platform\\Contract\\JsonSchema\\Attribu
...
+If you already have a JSON Schema defined in a file, you can reference it using the ``ref`` argument::
+
+    use Symfony\AI\Agent\Toolbox\Attribute\AsTool;
+    use Symfony\AI\Platform\Contract\JsonSchema\Attribute\With;
+
+    #[AsTool('my_tool', 'Example tool with external schema.')]
+    final class MyTool
+    {
+        public function __invoke(
+            #[With(ref: __DIR__.'/schema.json')]
```
OskarStark, line 181: Just thinking out loud, but why not a dedicated `#[Schema]` attribute? Also `ref` is an abbreviation which in most cases, need explanation. 

`#[Schema]` could have `#[Schema(file:.....)]` or `#[Schema(['my' => 'schema'])]` inline.

`#[JsonSchema]` would also work as a name for me.

WDYT @chr-hertel ?

valtzu: In my opinion the current `#[With]` should be `#[Schema]` (or `#[JsonSchema]`) and support all json schema properties, and [`$ref` is one of them](https://json-schema.org/understanding-json-schema/structuring#dollarref). And then Platform implementations would just strip out the features they don't support (or throw).

If it was like that, the describer abstraction could use the attribute as a typed schema object instead of array passed by-ref and this would also allow things like `$platform->invoke(..., ['response_format' => new Schema(/* .. */)])`.

The reasoning for this is that with the describer abstraction we will eventually have several attributes and it is not obvious that `#[With]` is the one which you can use to describe json schema.

---

But I thought this would be just a low-hanging fruit :wink: 

I don't have a use-case for this feature myself so no rush with this feature from my side.

OskarStark: I am fine merging as is 👍

chr-hertel:
Thank you @valtzu.

docs/components/agent.rst
chr-hertel: Let's rename as a follow up :+1:

PR #1921, opened 2026-04-10
[Store] Add configurable includeText option to Vectorizer

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | 
| License       | MIT

Mitigating a breaking change from behavior point of view for v0.7 - not convinced we should enforce this for users, but we can discuss for v0.8

cc @OskarStark @Guikingone @wachterjohannes

### DIFF
--- a/src/store/src/Document/Vectorizer.php
+++ b/src/store/src/Document/Vectorizer.php
@@ -25,6 +25,7 @@ final class Vectorizer implements VectorizerInterface
         private readonly PlatformInterface $platform,
         private readonly string $model,
         private readonly LoggerInterface $logger = new NullLogger(),
+        private readonly bool $includeText = false,
     ) {
     }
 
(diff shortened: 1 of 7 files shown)


Guikingone (approved):

chr-hertel:
Maybe it's also an issue in the Azure SearchStore implementation and metadata keys should never go into the primary level of a schema.

Compare:

```json
{
  "id": 1234,
  "vector": [0.123, 0.567, ...],
  "_text": "Abcdefgh"
}
```

vs

```json
{
  "id": 1234,
  "vector": [0.123, 0.567, ...],
  "metadata": {"_text": "Abcdefgh"}
}
```

src/store/src/Document/Vectorizer.php
```diff
@@ -112,7 +113,7 @@ private function vectorizeEmbeddableDocument(EmbeddableDocumentInterface $docume
         // Preserve the original text in metadata so downstream consumers
         // (e.g. text search, reranking) can access it via Metadata::getText().
         $metadata = $document->getMetadata();
-        if (!$metadata->hasText()) {
+        if ($this->includeText && !$metadata->hasText()) {
```
wachterjohannes, line 116: ah i see thats the fix for the "issue" in the reranker PR. Shouldnt this be fixed properly in the AzureStore? Not sure exactly why that was done as it is done

wachterjohannes:
@chr-hertel good as a Workaround for now but as you already written here - the metadata should not go directly into the schema of the metadata (or at least they should not break by adding a new field there)

Issue #1753, opened 2026-03-12
[Platform] Clarify BC policy for model IDs exposed by model catalogs


mxr576:
Could you clarify the backward compatibility expectations for model IDs exposed by Symfony AI model catalogs?

I am specifically asking about model IDs returned by the platform layer, for example IDs exposed via provider model catalogs. As an integrator building on top of Symfony AI, I need to know whether these **IDs are considered stable enough to persist in application code or in a database.**

The main question is this:

- Can a model ID disappear in a patch release?
- Can it disappear in a minor release?
- Or should removals only happen in a new major version?

Practical example: Claude Sonnet/Haiku 3.5. They are [already EOL-d](https://endoflife.date/claude) a couple of weeks ago, but they are still in the [ModelCatalog](https://github.com/symfony/ai/blob/v0.6.0/src/platform/src/Bridge/Anthropic/ModelCatalog.php).

This matters even when an upstream provider has already end-of-lifed a model. Of course, you are right that when a model is EOL-d then API calls are going to fail as well. However, if a model/provider is reached via a proxy then requests targeting the EOL-d model could be redirected outside of the application, or [High availability](https://symfony.com/doc/current/ai/components/platform.html#high-availability) mode could still help with survival at the application layer.

Because of that, the problem is not only whether calling such a model might fail. The bigger concern is whether upgrading Symfony AI itself can make stored model IDs invalid at the framework level, which could lead to application crashes or broken configuration after dependency updates.

A clear rule would make version constraints much easier to choose in dependent packages.

For example:

- If removals only happen in major versions, then a constraint like `symfony/ai: "^1.0.0"` could be safe for dependent packages.
- If removals may happen in minor versions, then a narrower constraint like `symfony/ai: "~1.0.0"` would be more appropriate.

It would also help package authors decide whether catalog-derived model IDs are safe to store persistently, or whether they should be treated as volatile runtime data.

## Expected clarification

It would help to define one of these policies explicitly:

- Model IDs exposed by catalogs are part of Symfony AI’s backward compatibility surface, so removals happen only in major versions.
- Model IDs may change in minor releases, but not in patch releases.
- Model IDs are not covered by BC guarantees and may change in any release.

If the third option is the intended behavior, it would be useful to document that clearly so downstream packages know not to treat catalog model IDs as stable identifiers.

Guikingone:
Hi @mxr576 👋🏻 

[...]

Issue #1692, opened 2026-02-26
[Platform] Shipping model catalogs is not a good idea


chr-hertel:
I'm afraid we need to rework our concept of model catalogs in the Platform component, and the main reason is the change frequency:

**The Issue**

The Platform component itself will be installed via composer, of course. It will also be updated as regular dependency. One day we will have a v1.x with the BC promise in action. Sure, we want users to update the library regularly, but do we want to update the library like model vendors update their models? I'd say, no.
Currently, whenever a model version gets added, deprecated, or removed - we need to update the code base of the respective Platform bridge, release it, and the users need a `composer update`. Yes, they could overwrite the built-in catalog, and already use it - but that's also not that convenient.
(@OskarStark correct me, if I miss something here.)

**Potential Alternatives**

1. **Dynamic Model Catalogs**
    Some providers offer API endpoints to pull model catalogs, and we could just do that on the fly and cache it. We actually do that for some. This decouples even the deployment of apps using the Platform from the model catalog.

2. **Central Database**
    We could promote that Models.dev approach by @fabpot to a rather global catalog - maybe have a dynamic API-based catalog with cache as well, or hide it in a separate package, that we purely automate and release nightly for example, like `symfony/ai-models` with a stable interface for Platform to require

3. **Offload to user land completely**
    Not sure that's feasible - would require some new clever capability definition/handling.

4. **???**
    I guess there are more solutions ...

WDYT?

chr-hertel:
Maybe we should revive that #646 idea to introduce another layer below the `Platform` that gets provided by the bridges, instead the bridges ...

marco-jouwweb:
I understand how this can become a very time consuming task, and it seems only logical to not waste too much time on this.
I don't feel like the model catalogs are bad, tbh. Maybe it would make sense to only update the major providers frequently (every X days/weeks), and leaving the other providers to be updated by either community? 

Making it overridable keeps it workable at all times, its just about defaults. Having the community carry that responsibility seems logical to me. If something should be updated, opening a PR just to add that to the catalog shouldn't be too much work. 

I'd be wary introducing an external API as a default. I'd rather see it as a possible extension than something included by default, as introducing third party APIs is something that can be unwished for in certain situations. 

What do you think?

EDIT: If providers offer an official API, I'd say to move this to CI/CD: have a script call the API and format as a catalog config. And then we can always add some 'refresh' functionality, to update manually in case the latest build is outdated?

Guikingone:
[...]

PR #1090, opened 2025-12-08
[Agent] Add Model Router system for intelligent model selection

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no 
| Docs?         | yes/no 
| Issues        | Fix #... 
| License       | MIT

Implements the Model Router architecture, enabling intelligent model selection based on input characteristics. This allows agents to automatically route requests to appropriate models (e.g., vision models for images, speech models for audio).

Core Infrastructure:
- RouterInterface: Universal contract for all router types
- RoutingResult: Result with model name, optional transformer, reason, and confidence
- TransformerInterface: Contract for input transformation
- RouterContext: Provides platform access, model catalog, and metadata
- SimpleRouter: Callable-based router for flexible routing logic
- ChainRouter: Composite router for trying multiple routers in sequence
- ModelRouterInputProcessor: Integration with Agent's input processing pipeline

Input Enhancement:
- Add optional platform property to Input class (backward compatible)
- Update Agent to pass platform when creating Input
- Platform available to input processors via Input.getPlatform()

Includes 6 examples demonstrating different routing patterns and comprehensive test coverage (14 router tests, all 187 agent tests passing).

Key Features:
- No breaking changes - all changes are backward compatible
- Extensible design - custom routers in <20 lines
- Multi-provider support - works with any Platform implementation
- Transformation support - routers can transform input before invocation
- Default model fallback - routers can access agent's default model
- Capability-based routing - find models supporting specific capabilities

Architecture Decisions:
- Single RouterInterface for all router types enables composition
- Transformer in RoutingResult (router decides both model and transformation)
- Platform passed via Input (no breaking changes to interfaces)
- RouterContext provides metadata including default model
- Start simple with SimpleRouter, add complexity gradually

PR #646, opened 2025-09-22
[Platform] Rework platform by introducing Connectors

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | 
| License       | MIT

This introduces another concept for encapsulating the connection concerns of Bridges.
This could be the base for streamlining HTTP error handling more easy, bringing an easy extension point for an default openai-compatible and http based connection, and reducing the effort for Platform bridges.
this is not working, but just a sketch - with some fantasy together with #136 and #640 

@OskarStark just revived an old branch of mine, what's good what's not? happy to trash that and start from sketch, but I'm just still not happy with ModelClient and HTTP connection handling
have a look at `src/platform/src/Bridge/OpenAi/Connector.php`

PR #1923, opened 2026-04-10
[Platform] Introduce Provider abstraction and model routing layer

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        |
| License       | MIT

## Summary

This PR introduces a **Provider** abstraction layer between `PlatformInterface` and the actual inference backends, along with a **ModelRouter** for routing. This is the foundational first step toward solving several long-standing architectural pain points.

### The Problem

The Platform component has a fundamental 1:1 coupling: one `Platform` instance = one inference backend. This causes:

- **No unified entry point for multi-provider setups** — users needing OpenAI + Anthropic must manage two separate `Platform` instances
- **Static ModelCatalogs are a maintenance burden** (#1692) — 38+ hardcoded catalogs need constant updates as models change faster than library releases
- **Model-level failover/routing has no natural home** — `FailoverPlatform` only fails over between whole platforms; #1864 wants model-level failover, #1808 wants model override in failover
- **Input-based model routing belongs at Platform level** — #1090 puts it at Agent level, but it's a Platform concern
- **Connector concept** (#646) — old sketch to introduce a layer below Platform for connection concerns never materialized

### The Solution

```
User Code
    |
PlatformInterface (UNCHANGED)
    |
Platform (refactored: router over providers)
    |-- ModelRouterInterface (routes model name → provider)
    |-- CompositeModelCatalog (merges provider catalogs)
    |-- ModelRoutingEvent (pre-routing hook, can set provider to skip router)
    |
ProviderInterface (NEW)
    |
Provider (extracted from current Platform logic)
    |-- ModelClientInterface[] / ResultConverterInterface[] / Contract / ModelCatalog
```

**`ProviderInterface`** encapsulates what `PlatformFactory::create()` currently produces — a complete, self-contained connection to a single inference backend (model clients, result converters, contract, catalog).

**`Platform`** becomes a thin routing layer over multiple providers, using a `ModelRouterInterface` to determine which provider handles each request.

**`ModelRoutingEvent`** is dispatched before routing, allowing listeners to modify the model/input/options or set a provider directly to skip the router entirely.

**`PlatformInterface`** remains completely unchanged — all downstream code (Agent, Chat, Store, user code) continues to work without modification.

### What's Included

**Core layer:**
- `ProviderInterface` + `Provider` (extracted from current Platform logic)
- `ModelRouterInterface` + `CatalogBasedModelRouter` (default: routes by catalog lookup)
- `CompositeModelCatalog` (merges catalogs from multiple providers)
- `ModelRoutingEvent` (Platform-level event with optional provider to skip routing)
- Refactored `Platform` as router over providers

**Reference bridge implementation:**
- `OpenAi\ProviderFactory` (replaces `OpenAi\PlatformFactory`)

**30 new tests**, all 510 tests pass (including existing).

### User-Facing DX

```php
// Single provider
$provider = OpenAi\ProviderFactory::create(apiKey: 'sk-...');
$platform = new Platform([$provider]);

// Multi-provider with auto-routing
$openai = OpenAi\ProviderFactory::create(apiKey: 'sk-...');
$anthropic = Anthropic\ProviderFactory::create(apiKey: 'sk-...');

$platform = new Platform([$openai, $anthropic]);

$platform->invoke('gpt-4o', $messages);            // → OpenAI
$platform->invoke('claude-3-5-sonnet', $messages);  // → Anthropic
```

### How This Addresses Existing Work

| PR/Issue | How it's addressed |
|----------|--------------------|
| #646 (Connectors) | `Provider` is the connector concept — encapsulates connection concerns |
| #1692 (ModelCatalog burden) | `CompositeModelCatalog` + dynamic catalogs become first-class; static catalogs per bridge become optional |
| #1753 (BC for model IDs) | Model ID resolution is now a pluggable strategy, not hardcoded |
| #1808 (Model override in failover) | Enables `FailoverModelRouter` with model name translation (follow-up) |
| #1864 (ModelFailoverPlatform) | Enables model-level failover natively via router (follow-up) |
| #1090 (Model Router) | Enables input-based routing at Platform level via `CallableModelRouter` (follow-up) |

## Use-Case: Load Balancing

The `ModelRouterInterface` enables load balancing between providers. For example, round-robin between two OpenAI providers with different API keys:

```php
$openai1 = OpenAi\ProviderFactory::create(apiKey: 'sk-key1');
$openai2 = OpenAi\ProviderFactory::create(apiKey: 'sk-key2');

$platform = new Platform(
    [$openai1, $openai2],
    new RoundRobinModelRouter(),
);

$platform->invoke('gpt-4o', $messages); // → openai1
$platform->invoke('gpt-4o', $messages); // → openai2
$platform->invoke('gpt-4o', $messages); // → openai1
```

The router implementation:

```php
final class RoundRobinModelRouter implements ModelRouterInterface
{
    private int $counter = 0;

    public function resolve(string $model, iterable $providers, array|string|object $input, array $options = []): ProviderInterface
    {
        $candidates = [];
        foreach ($providers as $provider) {
            if ($provider->supports($model)) {
                $candidates[] = $provider;
            }
        }

        if ([] === $candidates) {
            throw new ModelNotFoundException(\sprintf('No provider found for model "%s".', $model));
        }

        return $candidates[$this->counter++ % \count($candidates)];
    }
}
```

The same pattern works for weighted or random distribution — just a different `ModelRouterInterface` implementation. Composes with `ChainModelRouter` too (e.g. explicit routes for some models, load-balanced for the rest).

## Use-Case: Failover / Rate Limit Handling

Failover is a provider-level concern, implemented as a `ProviderInterface` decorator — not a router concern (since routing happens before invocation, it can't react to runtime failures).

**Cross-provider failover** (OpenAI rate-limited → fall back to Anthropic):

```php
$failover = new FailoverProvider('main', [
    OpenAi\ProviderFactory::create(apiKey: 'sk-...'),
    Anthropic\ProviderFactory::create(apiKey: 'sk-...'),
], $rateLimiterFactory);

$platform = new Platform([$failover]);
```

**Nested composition** (load-balanced OpenAI pool with Anthropic as final fallback):

```php
$openaiPool = new FailoverProvider('openai', [$openai1, $openai2], $rateLimiterFactory);
$main = new FailoverProvider('main', [$openaiPool, $anthropic], $rateLimiterFactory);

$platform = new Platform([$main]);
```

This subsumes the existing `FailoverPlatform` entirely — same pattern, but as a `ProviderInterface` decorator instead of a `PlatformInterface` decorator. More composable since it nests with other providers and works seamlessly with the routing layer.

## Use-Case: ModelsDev as Universal Catalog

The Provider/ModelRouter rework shifts the role of the ModelsDev bridge. Its core value becomes clear — it's a **universal, dynamic model catalog** backed by the `symfony/models-dev` database. No hardcoded model lists, no API calls. New models appear with `composer update symfony/models-dev`.

Routing and provider construction move to where they belong: `Platform` and bridge-specific `ProviderFactory` classes.

```php
use Symfony\AI\Platform\Bridge\ModelsDev\ModelCatalog;
use Symfony\AI\Platform\Bridge\OpenAi\ProviderFactory as OpenAiProviderFactory;
use Symfony\AI\Platform\Bridge\Anthropic\ProviderFactory as AnthropicProviderFactory;
use Symfony\AI\Platform\Platform;

// Each bridge handles its own connection concerns (API keys, HTTP, contracts)
// ModelsDev provides the catalog — always up-to-date, no static model lists
$openai = OpenAiProviderFactory::create(
    apiKey: $_ENV['OPENAI_API_KEY'],
    modelCatalog: new ModelCatalog('openai'),
);
$anthropic = AnthropicProviderFactory::create(
    apiKey: $_ENV['ANTHROPIC_API_KEY'],
    modelCatalog: new ModelCatalog('anthropic'),
);

// Platform routes automatically — CatalogBasedModelRouter checks each provider's catalog
$platform = new Platform([$openai, $anthropic]);

$platform->invoke('gpt-4o', $messages);            // → openai (catalog knows it)
$platform->invoke('claude-3-5-sonnet', $messages);  // → anthropic (catalog knows it)
```

All `ProviderFactory` signatures already accept `ModelCatalogInterface` — no new abstraction needed. The static catalogs in each bridge become the fallback default, ModelsDev becomes the dynamic option.

**What remains valuable in ModelsDev:**
- `ModelCatalog` — the universal catalog (core value)
- `CapabilityMapper` — maps models-dev metadata to Capability enum
- `DataLoader` — loads/caches the JSON
- `ProviderRegistry` — discovery ("what providers exist?")

**What shrinks or goes away:**
- `PlatformFactory` — Platform + router handles routing natively now
- `BridgeResolver` — routing is Platform's job, not the catalog's
- `ModelResolver` with `"provider::model"` syntax — catalog-based routing makes this redundant for most cases

The ModelsDev bridge becomes a focused, lean package: **a catalog, not a platform**.

### Follow-Up PRs

This is scoped to the core layer + one bridge. Planned follow-ups:
- Convert remaining 37 bridges from `PlatformFactory` to `ProviderFactory`
- Add `ExplicitModelRouter`, `FailoverModelRouter`, `ChainModelRouter`, `CallableModelRouter`
- `FailoverProvider` decorator (subsumes `FailoverPlatform`)
- AI Bundle integration (DI wiring, routing config section)
- Refocus ModelsDev bridge as universal catalog


src/platform/src/Bridge/OpenAi/PlatformFactory.php
```diff
@@ -11,21 +11,21 @@
...
-use Symfony\Component\HttpClient\EventSourceHttpClient;
 use Symfony\Contracts\EventDispatcher\EventDispatcherInterface;
 use Symfony\Contracts\HttpClient\HttpClientInterface;
 
 /**
  * @author Christopher Hertel <mail@christopher-hertel.de>
+ *
+ * @deprecated since 0.8, use ProviderFactory instead
  */
 final class PlatformFactory
```
chr-hertel: i guess we should rather rename and to `ProviderFactory`

chr-hertel: BC for v0.8 then

src/platform/src/Platform.php
```diff
@@ -11,86 +11,84 @@
...
-        $this->eventDispatcher?->dispatch($event);
-
-        return $event->getDeferredResult();
     }
 
-    public function getModelCatalog(): ModelCatalogInterface
+    /**
+     * Convenience factory for single-provider usage.
+     */
+    public static function create(ProviderInterface $provider, ?EventDispatcherInterface $eventDispatcher = null): self
```
chr-hertel: not sure we need the factory method here 🤔

src/platform/src/Platform.php
```diff
@@ -11,86 +11,84 @@
...
-    private function doInvoke(Model $model, array|string $payload, array $options = []): RawResultInterface
+    public function invoke(string $model, array|string|object $input, array $options = []): DeferredResult
     {
-        foreach ($this->modelClients as $modelClient) {
-            if ($modelClient->supports($model)) {
-                return $modelClient->request($model, $payload, $options);
-            }
-        }
+        $event = new ModelRoutingEvent($model, $input, $options);
+        $this->eventDispatcher?->dispatch($event);
```
chr-hertel, line 50: do we want this event? and if so, it would be more logically to me if i can set the provider here to skip the resolver 🤔

src/platform/src/ModelResolverInterface.php
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+namespace Symfony\AI\Platform;
+
+/**
+ * Resolves which provider should handle a given model invocation.
+ *
+ * @author Christopher Hertel <mail@christopher-hertel.de>
+ */
+interface ModelResolverInterface
```
chr-hertel: i like `ModelRouterInterface` and `ModelRouter` more 🤔

OskarStark:
I would not split this PR and go with a big refactoring in one go.

OskarStark:
So we need adjustments in the ai bundle?

src/platform/src/ModelCatalog/CompositeModelCatalog.php
```diff
@@ -0,0 +1,63 @@
...
+
+    /**
+     * @return array<string, array{class: string, capabilities: list<Capability>}>
+     */
+    public function getModels(): array
+    {
+        $merged = [];
+
+        foreach ($this->catalogs as $catalog) {
+            $merged = array_merge($merged, $catalog->getModels());
```
mxr576: Eliminate performance issue caused by array_merge in loop - the array unpacking trick also works with `foreach` if there is a reason to keep it.

```php
 $modelSets = array_map(
        static fn ($catalog) => $catalog->getModels(),
        $this->catalogs
    );

    return $modelSets === []
        ? []
        : array_merge(...$modelSets);
```

src/platform/src/Platform.php
```diff
@@ -11,86 +11,60 @@
...
+     * @param ProviderInterface[] $providers
      */
     public function __construct(
-        private readonly iterable $modelClients,
-        private readonly iterable $resultConverters,
-        private readonly ModelCatalogInterface $modelCatalog,
-        private ?Contract $contract = null,
+        private readonly array $providers,
+        private readonly ModelRouterInterface $modelRouter = new CatalogBasedModelRouter(),
+        private readonly ?ModelCatalogInterface $modelCatalog = null,
```
mxr576: Do we still need model catalog at the platform level, isn't the list of available models would be always derived from the providers?

chr-hertel: same thought here, currently still part of the contract - which is a bit leaky. def follow-up tho.

src/platform/src/Platform.php
```diff
@@ -11,86 +11,60 @@
...
         private readonly ?EventDispatcherInterface $eventDispatcher = null,
     ) {
-        $this->contract = $contract ?? Contract::create();
+        if ([] === $this->providers) {
```
mxr576, line 42: Is `ProviderInterface::getName()` a provider type or a connection identity?

`OpenAi\ProviderFactory::create()` always returns `'openai'` as the name, regardless of which API key or region is used. In the load-balancing example from the PR description, `$openai1` and `$openai2` both return `'openai'` from `getName()`.

This creates a practical problem. `Provider` already uses `$this->name` in exception messages (e.g. `'No ModelClient registered for model "%s" in provider "%s".'`). Logging, tracing, or any tooling that relies on provider name to identify *which connection* was used would produce ambiguous output. You would see `openai` in both cases and have no way to tell which one failed.

The `Platform` constructor does not validate uniqueness either, so there is nothing preventing two providers with the same name from coexisting silently.

A few questions:
- Is the name intentionally a *vendor/type* label (e.g. `openai`, `anthropic`), and not expected to be unique per instance?
- If so, what is the intended use of `getName()` beyond error messages? Should it be renamed to `getVendorName()` or similar to make the semantics explicit?
- If connection-level identity is needed (and the load-balancing use case strongly suggests it is), should `ProviderFactory::create()` accept an optional `$name` parameter, defaulting to `'openai'` but allowing `'openai-eu'`, `'openai-us'`, etc.?

The `FailoverProvider` example in the description already passes explicit names like `'openai'` and `'main'` manually, which suggests awareness of this, but `ProviderFactory` does not expose that option yet.

chr-hertel: yea, we could just add that to the `ProviderFactory::create` method with a default value :+1:

src/platform/src/ModelCatalog/CompositeModelCatalog.php
mxr576: Also cache result in a property once it was calculated in a request.

mxr576:
Great work on this PR, the routing layer is a clean solution. One architectural question worth discussing before this lands though.

Looking at the refactored `Platform`, it is essentially a provider pool with an injected model router. It holds a list of `ProviderInterface` instances, delegates all real work to whichever provider the router selects, and adds a single cross-cutting concern on top: the `ModelRoutingEvent` pre-routing hook. This makes `Platform` behaviorally identical to what a `ProviderPool implements ProviderInterface` would be. The only thing keeping it as a separate top-level concept is `PlatformInterface`, the public contract consumed by `agent`, `chat`, `store`, and user code.

Two alternatives come to mind.

**Alternative A** collapses the hierarchy: `Platform` becomes `ProviderPool implements ProviderInterface`, making the entire stack uniform and infinitely composable.

```php
$openaiPool = new ProviderPool([$openai1, $openai2], new RoundRobinModelRouter());
$mainPool   = new ProviderPool([$openaiPool, $anthropic], new CatalogBasedModelRouter());

$chat = new Chat($mainPool); // ProviderInterface is the only contract
```

Failover decorators and pools nest naturally at every level. This actually makes the failover story from #1808 cleaner than what the PR proposes, since a `FailoverProvider` wrapping a `ProviderPool` wrapping another `FailoverProvider` just works.

**Alternative B** goes the other direction: skip `ProviderInterface` entirely and introduce `PlatformPool implements PlatformInterface` instead. Everything stays in the `Platform` domain, zero new abstraction layer, purely additive change.

```php
$platform = new PlatformPool(
    platforms: [$openai, $anthropic],
    router: new CatalogBasedModelRouter(),
);
```

Now, do both alternatives still cover the referenced issues?

Alternative A covers all five, and as noted above the failover composition story is actually better. Alternative B covers four out of five. The gap is #646 (Connector concept). Without a dedicated `ProviderInterface`, third-party bridge authors have no formal backend contract to implement. Failover (#1808) technically works too, but `PlatformInterface` ends up overloaded as both a consumer-facing API and a composable backend building block, which feels wrong.

I fully understand why `Platform` is kept as the top-level concept here. Changing `PlatformInterface` would ripple through `agent`, `chat`, `store`, `ai-bundle`, all bridges, and all user code. That is a genuine earthquake. But since Symfony AI is still in active development in a monorepo without a stable release, a one-time collapse is at least worth naming as a deliberate choice before the two-hierarchy split solidifies into permanent architecture.

My lean: this PR's approach is the right call if `ProviderInterface` is meant to be a first-class extension point for third-party bridges, which seems like the stronger long-term argument. Alternative B is fine if we want to avoid a new abstraction layer, but it leaves #646 only half-answered. Alternative A is the most consistent design, just expensive right now.

src/platform/src/ModelCatalog/CompositeModelCatalog.php
mxr576: This is fixed, but I do not have permission to close this thread.

src/platform/src/Platform.php
mxr576: This is also fixed, but I do not have permission to resolve the thread.

src/platform/src/Platform.php
mxr576: I see this also got implemented, maybe what I miss is an explicit statement on `ProviderInterface::getName() `that the returned value has to be unique per instance, and since it is an instance id and we are introducing breaking changes, should not we just renamed the `getName()` method as well?

src/platform/src/Bridge/AmazeeAi/Factory.php
```diff
@@ -47,4 +54,21 @@ public static function create(
             $eventDispatcher,
         );
     }
+
+    /**
+     * @param non-empty-string $name
+     */
+    public static function createPlatform(
```
mxr576, line 63: Looking at this  (and nearly every other bridge `createPlatform()` implementation I checked), these methods are essentially pass-throughs to `self::createProvider()`. They accept exactly the same parameters that createProvider() already exposes, then wrap the result in a `new Platform([$provider])`.

The problem is that Platform's constructor also accepts a `ModelRouterInterface` argument, but `createPlatform()` silently drops it. That means if I want to inject a custom `ModelRouter` into the Platform instance, I cannot use this factory at all. I would have to bypass it entirely and wire things up manually, which defeats its purpose.

The broader question this raises: now that Platform is a proper routing layer that composes multiple providers, does createPlatform() on individual bridge factories still make sense? The single-provider case it covers is trivially handled by calling `new Platform([$factory->createProvider(...)])` directly. Consumer code that needs more control over Platform (custom router, multiple providers) would always have to do that anyway. **I wonder whether these `createPlatform()` helpers should be dropped from bridge factories entirely.**

mxr576: This is an interesting idea by the LLM

---- 

A PlatformBuilder fluent API

If convenience constructors are considered valuable for the Symfony Bundle's DI wiring, a dedicated builder is a much better shape than a static factory method on every bridge:

```php
$platform = PlatformBuilder::create()
->withProvider(OpenAi\ProviderFactory::create(apiKey: 'sk-...'))
->withProvider(Anthropic\ProviderFactory::create(apiKey: 'sk-...'))
->withRouter(new RoundRobinModelRouter())
->build();
```

The builder owns the Platform construction concern. Bridge factories stay focused on producing ProviderInterface instances, which is their actual job . The builder is one place to extend (e.g. adding event dispatcher, router config) rather than 38 bridge factory classes . This also directly solves the ModelRouterInterface exposure problem.

mxr576:
Looking at `PlatformInterface` and `ProviderInterface` side by side, there is a structural observation worth raising, which was bothering me since the beginning. 

`PlatformInterface` exposes `invoke()` and `getModelCatalog()`. `ProviderInterface` exposes exactly those two methods plus `getName()` and `supports()`. The surfaces are nearly identical, and that gap is not accidental — `supports()` and `getName()` are the extra semantics a routing layer needs from its leaves. But it raises the question: why does the system need two interfaces at all when one is almost a strict superset of the other?

The current design forces callers into a two-step mental model: use `ProviderFactory` to build a provider, then wrap it in `Platform` even when there is only one provider and no routing is needed. The `createPlatform()` factory methods exist precisely to paper over that friction, but as discussed elsewhere in this thread, they do so by hiding `ModelRouterInterface` as a side effect.

An alternative that fits naturally into the architecture this PR is building: make `Provider` implement `PlatformInterface` directly. No new abstraction, no interface merging, no breaking change. `ProviderInterface` stays exactly as designed. The hierarchy becomes:

```
PlatformInterface          ← the only type callers ever depend on
|
├── Provider            (single backend, leaf node)
|     |-- ModelClientInterface[]
|     |-- ResultConverterInterface[]
|     |-- Contract
|     |-- ModelCatalog
|
└── Platform            (routing layer, only needed for multi-provider)
|-- ProviderInterface[]
|-- ModelRouterInterface
|-- CompositeModelCatalog
|-- ModelRoutingEvent
```

The user-facing DX would look like this:

```php
// Single provider — no Platform needed at all
$platform = OpenAi\ProviderFactory::create(apiKey: 'sk-...');
$platform->invoke('gpt-4o', $messages); // Provider IS-A PlatformInterface

// Multi-provider — Platform earns its place as a routing layer
$openai    = OpenAi\ProviderFactory::create(apiKey: 'sk-...');
$anthropic = Anthropic\ProviderFactory::create(apiKey: 'sk-...');

$platform = new Platform([$openai, $anthropic]);
$platform->invoke('gpt-4o', $messages);            // → OpenAI
$platform->invoke('claude-3-5-sonnet', $messages); // → Anthropic

// Custom routing — full control, no factory in the way
$platform = new Platform([$openai, $anthropic], new RoundRobinModelRouter());
```

The consequences are worth spelling out:

- Single-provider consumers type-hint `PlatformInterface` and inject a `Provider` directly. No `Platform` wrapper needed, no factory shortcut needed.
- `Platform` is only instantiated when you actually need routing over multiple providers, which is when it earns its existence.
- Bridge factories drop `createPlatform()` entirely. They produce a `ProviderInterface` (which is also a `PlatformInterface`), and the caller decides whether to use it bare or compose it into a `Platform`.
- `TraceablePlatform` and anything else decorating `PlatformInterface` continues to work without modification.

This keeps `Platform` clean as a routing concern and removes the pressure to make factories aware of `ModelRouterInterface`. The two interfaces stop competing and start forming a coherent hierarchy.

chr-hertel:
Hey @mxr576, thanks for your review and sharing your thoughts - I see your point, but I'm actually in favor of segregating this even more - see #1956 - reducing the ambiguity and having more conceptual separation of concerns/responsibility between PlatformInterface and ProviderInterface.

Good catch with that missing ModelRouterInterface - added it.
Thanks again!

chr-hertel:
@OskarStark i guess that's good to go in now - any last thoughts?

OskarStark (approved):

OskarStark:
Good to merge 💪🏼

PR #1938, opened 2026-04-12
[Platform][VertexAi] Fix binary result handling

valtzu:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| License       | MIT

Fix Vertex binary result handling; it was missing base64 decoding. I have manually verified that Vertex returns base64-encoded content, just like Gemini API, despite [the docs](https://docs.cloud.google.com/vertex-ai/generative-ai/docs/model-reference/inference#parts) mentioning "Inline data in raw bytes.".


chr-hertel (approved):
Tested as well - thanks for that!

chr-hertel:
Thank you @valtzu.

PR #1941, opened 2026-04-13
[Examples] Add image editing example with Vertex

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | 
| License       | MIT

Proving bug fix of #1938


examples/vertexai/image-editing.php
```diff
@@ -0,0 +1,31 @@
...
+    Message::ofUser(
+        'Please colorize the elephant in red.',
+        Image::fromFile(dirname(__DIR__, 2).'/fixtures/image.jpg'),
+    ),
+);
+$result = $platform->invoke('gemini-2.5-flash-image', $messages);
+
+$result->asFile(__DIR__.'/result.png');
+
+echo 'Result image saved to result.png'.\PHP_EOL;
```
OskarStark: When outputting the full path it is clickable, no matter from where it was executed.

PR #1900, opened 2026-04-08
[Platform][Gemini] Add `MultiPartResult`

valtzu:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | Fix #1053
| License       | MIT

Expose the `parts` _(instead of grouping them or selecting one specific part-of-interest)_ from platform to
1. Let users do whatever they want to do with them (i.e. render nice UI where different type of parts show up differently)
2. Prepare for part-specific metadata (or part-specific raw response) since Gemini mentions in the [docs](https://ai.google.dev/gemini-api/docs/tool-combination#api-returns-parts):
   >  You must return all parts, including all the [fields](https://ai.google.dev/gemini-api/docs/tool-combination#critical-fields) they contain, back to the model on each turn to maintain context and enable tool combinations.
3. Prepare for "meta-parts", e.g. server side toolcalls

PR #1948, opened 2026-04-17
[Platform][VertexAi] Return `MultiPartResult` for multi-part results

valtzu:
…veral parts

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| License       | MIT

Same as #1900 but for Vertex

### DIFF
--- a/examples/vertexai/server-tools-code-execution.php
+++ b/examples/vertexai/server-tools-code-execution.php
@@ -1,0 +1,40 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use Symfony\AI\Agent\Agent;
+use Symfony\AI\Platform\Bridge\VertexAi\PlatformFactory;
+use Symfony\AI\Platform\Message\Message;
+use Symfony\AI\Platform\Message\MessageBag;
+use Symfony\AI\Platform\Result\ExecutableCodeResult;
+use Symfony\AI\Platform\Result\MultiPartResult;
+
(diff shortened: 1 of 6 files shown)


examples/vertexai/server-tools-code-execution.php
```diff
@@ -0,0 +1,40 @@
...
+use Symfony\AI\Agent\Agent;
+use Symfony\AI\Platform\Bridge\VertexAi\PlatformFactory;
+use Symfony\AI\Platform\Message\Message;
+use Symfony\AI\Platform\Message\MessageBag;
+use Symfony\AI\Platform\Result\ExecutableCodeResult;
+use Symfony\AI\Platform\Result\MultiPartResult;
+
+require_once dirname(__DIR__).'/bootstrap.php';
+
+$platform = PlatformFactory::create(env('GOOGLE_CLOUD_LOCATION'), env('GOOGLE_CLOUD_PROJECT'), env('GOOGLE_CLOUD_VERTEX_API_KEY'));
```
chr-hertel: would be consistent to also set `httpClient: adc_aware_http_client()` please

chr-hertel (changes requested):

examples/vertexai/server-tools-code-execution.php
valtzu: Ok, I thought api key would be more common to use and thus better example _(though after all, Vertex is the more enterprise platform, those who want simplicity probably stay with Gemini api anyway)_

I'll adjust

chr-hertel: don't know actually, but just noticed while running the example that i didn't get the output

chr-hertel (approved):

chr-hertel:
Thank you @valtzu.

PR #1961, opened 2026-04-18
[Platform][Mistral] Expose API errors for better Developer Experience

Amoifr:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | Fix #528
| License       | MIT

The Mistral bridge result converters previously collapsed any non-200 HTTP response into a generic `RuntimeException`, which made API errors hard to distinguish and act upon. This follows the pattern already in place on the Anthropic and OpenAI GPT bridges.

Both the **LLM** and **Embeddings** result converters now throw dedicated exceptions:

Error messages are extracted from the response body, supporting both Mistral response shapes: a top-level `message` field, or the nested `error.message` format.

```php
try {
    $platform->invoke(new Mistral(), $messages);
    // retry...
}
```

Tests cover every new status code plus fallback behaviour when the response body is empty or does not parse.

### DIFF
--- a/src/platform/src/Bridge/Mistral/Embeddings/ResultConverter.php
+++ b/src/platform/src/Bridge/Mistral/Embeddings/ResultConverter.php
@@ -12,6 +12,9 @@
 namespace Symfony\AI\Platform\Bridge\Mistral\Embeddings;
 
 use Symfony\AI\Platform\Bridge\Mistral\Embeddings;
+use Symfony\AI\Platform\Exception\AuthenticationException;
+use Symfony\AI\Platform\Exception\BadRequestException;
+use Symfony\AI\Platform\Exception\RateLimitExceededException;
 use Symfony\AI\Platform\Exception\RuntimeException;
 use Symfony\AI\Platform\Model;
 use Symfony\AI\Platform\Result\RawHttpResult;
@@ -35,6 +38,19 @@ final class ResultConverter implements ResultConverterInterface
     {
         $httpResponse = $result->getObject();
 
+        if (401 === $httpResponse->getStatusCode()) {
+            throw new AuthenticationException($this->extractErrorMessage($httpResponse->getContent(false)) ?? 'Unauthorized');
+        }
+
+        if (400 === $httpResponse->getStatusCode()) {
+            throw new BadRequestException($this->extractErrorMessage($httpResponse->getContent(false)) ?? 'Bad Request');
+        }
+
+        if (429 === $httpResponse->getStatusCode()) {
+            $retryAfter = $httpResponse->getHeaders(false)['retry-after'][0] ?? null;
+            throw new RateLimitExceededException($retryAfter ? (int) $retryAfter : null);
+        }
+
         if (200 !== $httpResponse->getStatusCode()) {
             throw new RuntimeException(\sprintf('Unexpected response code %d: "%s"', $httpResponse->getStatusCode(), $httpResponse->getContent(false)));
         }
@@ -56,5 +72,19 @@ final class ResultConverter implements ResultConverterInterface
     public function getTokenUsageExtractor(): TokenUsageExtractorInterface
     {
         return new TokenUsageExtractor();
+    }
+
+    private function extractErrorMessage(string $body): ?string
+    {
+        if ('' === $body) {
+            return null;
+        }
+
+        $data = json_decode($body, true);
+        if (!\is_array($data)) {
+            return null;
+        }
+
+        return $data['error']['message'] ?? $data['message'] ?? null;
     }
 }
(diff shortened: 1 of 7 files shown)


src/platform/src/Bridge/Mistral/Embeddings/ResultConverter.php
```diff
@@ -57,4 +73,18 @@ public function getTokenUsageExtractor(): TokenUsageExtractorInterface
...
+    private function extractErrorMessage(string $body): ?string
+    {
+        if ('' === $body) {
+            return null;
+        }
+
+        $data = json_decode($body, true);
+        if (!\is_array($data)) {
+            return null;
+        }
```
chr-hertel, line 86: can we use the `$httpResponse->toArray(false)` instead of this?

chr-hertel:
Hi @Amoifr,
thanks for the effort for improving the situation here and tackling #528
I was wondering about two things here:
1. is a better way for avoiding the duplication here? the added code across all the PRs is quite similar and there might be a more abstract solution
2. can you help with reviewing this for providing the links to the corresponding docs? i'm not sure if you're testing all those providers, but surely you're looking the API error structure up while creating the MockResponse - would help me following up on your PRs if you share the basis for your work here :)

Thanks already!

Amoifr:
Hey @chr-hertel, great feedback, both points are spot on 🙏

**1. Duplication** — totally fair. I went for the copy-paste approach because Anthropic/OpenAi-Gpt were already doing it that way and I wanted to match the existing style, but you're right that at this scale it's begging for a shared abstraction. My proposal: introduce a small `HttpStatusErrorHandlingTrait` (or helper class) in `Symfony\AI\Platform\Exception\` or similar, exposing a `throwOnHttpError(ResponseInterface $response): void` method that encapsulates the 401/400/429 branches + the body extraction. Each converter would shrink to a one-liner. If you're OK with that direction, I can refactor all the PRs in one go.

Side note: combined with @OskarStark's ask on #1965 (squash into a single PR), I'm leaning towards closing the 7 individual PRs and opening a single consolidated PR built around the trait. That way you get one review surface, one history entry, and the abstraction benefits from the start. Does that work for both of you?

**2. Docs / basis for the work** — fair, I should have made that more explicit. Honestly, my primary reference was the existing Anthropic (`src/platform/src/Bridge/Anthropic/ResultConverter.php`) and OpenAi/Gpt converters in this repo — I treated them as the canonical pattern and replicated the shape. For provider-specific error bodies, I supported both the top-level `{"message": ...}` and the nested `{"error": {"message": ...}}` formats defensively, since most providers here are OpenAI-compatible. I did not spin up real accounts for every provider (no Cerebras / Perplexity / Cohere trials on my end), so the MockResponses are inferred from public docs + OpenAI-compat assumptions. Specifically:

- Mistral: https://docs.mistral.ai/api/ (OpenAI-compat + flat `message` for some 4xx)
- DeepSeek: https://api-docs.deepseek.com/ (OpenAI-compat)
- Cerebras: https://inference-docs.cerebras.ai/ (OpenAI-compat)
- Perplexity: https://docs.perplexity.ai/ (OpenAI-compat)
- Cohere: https://docs.cohere.com/reference (flat `message` top-level)
- Gemini: https://ai.google.dev/api/rest (nested `error.{code,message,status}`, Google's standard shape — already reflected in the existing 429 test)
- OpenAI: https://platform.openai.com/docs/api-reference (nested `error.message`, already covered on Gpt/Whisper)

If you'd rather I validate specific providers against a real API before merging, happy to sign up for a few — just point me at the ones you consider non-negotiable.

**3. Inline: `toArray(false)`** — good call, way more idiomatic. I'll switch `extractErrorMessage` to use `$httpResponse->toArray(false)` wrapped in a `try { } catch (DecodingExceptionInterface) { return null; }` so empty/non-JSON bodies still fall back cleanly. Same result, much cleaner.

Let me know which route you'd like me to take (trait + single PR vs. keep split + add trait later) and I'll get going 🙂

src/platform/src/Bridge/Mistral/Embeddings/ResultConverter.php
Amoifr: Good call — `$httpResponse->toArray(false)` is definitely more idiomatic. Ill switch to it with a `try { ... } catch (DecodingExceptionInterface) { return null; }` so empty / non-JSON bodies still fall back cleanly. Will ship in the consolidated refactor (see main thread).

chr-hertel:
Sure, you're right, that copy-paste pattern was def around before :angel: :laughing: 

I like the idea of going with a trait + single PR! Thanks already! :pray:

Amoifr:

PR #1962, opened 2026-04-18
[Platform][DeepSeek] Expose API errors for better Developer Experience

Amoifr:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | Fix #528
| License       | MIT

The DeepSeek bridge result converter only inspected the response body for logical errors (`content_filter`, `invalid_request_error`) and ignored the HTTP status code entirely. As a result, authentication failures and rate limits surfaced as PHP errors on a missing `choices` key rather than meaningful exceptions. This aligns DeepSeek with the Anthropic, OpenAI GPT and Mistral bridges.

The result converter now throws dedicated exceptions:

Error messages are extracted from the response body, supporting both the top-level `message` shape and the nested `error.message` shape.

```php
try {
    $platform->invoke(new DeepSeek('deepseek-chat'), $messages);
    // retry...
}
```

HTTP status handling is guarded behind an `instanceof RawHttpResult` branch so the existing in-memory streaming tests (`InMemoryRawResult`) keep working untouched.

Part of the effort to address #528 across Platform bridges. Previous step: #1961 (Mistral).


Amoifr:

PR #1965, opened 2026-04-18
[Platform][Cohere] Expose API errors for better Developer Experience

Amoifr:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | Fix #528
| License       | MIT

All four Cohere bridge result converters (LLM, Embeddings, Reranker, SpeechToText) collapsed any non-200 response into a generic `RuntimeException`, which made auth failures and rate limits indistinguishable from other errors. This aligns Cohere with the Anthropic, OpenAI GPT, Mistral, DeepSeek, Cerebras and Perplexity bridges.

Each result converter now throws dedicated exceptions:

 - `401` → `AuthenticationException`
 - `400` → `BadRequestException`
 - `429` → `RateLimitExceededException` (with `retry-after` header propagated)

Error messages are extracted from the response body, supporting both the top-level `message` shape and the nested `error.message` shape.

```php
try {
    $platform->invoke(new Cohere('command-r-plus'), $messages);
} catch (RateLimitExceededException $e) {
    sleep($e->getRetryAfter() ?? 60);
    // retry...
} catch (AuthenticationException $e) {
    // invalid API key -> message from the API is now exposed
}
```

Split into one commit per sub-client (LLM / Embeddings / Reranker / SpeechToText) for easier review.

Part of the effort to address #528 across Platform bridges. Previous steps: #1961 (Mistral), #1962 (DeepSeek), #1963 (Cerebras), #1964 (Perplexity).

PR #1964, opened 2026-04-18
[Platform][Perplexity] Expose API errors for better Developer Experience

Amoifr:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | Fix #528
| License       | MIT

The Perplexity bridge result converter collapsed any non-200 response into a generic `"Response does not contain choices."` `RuntimeException`, which made auth failures and rate limits indistinguishable from other errors. This aligns Perplexity with the Anthropic, OpenAI GPT, Mistral, DeepSeek and Cerebras bridges.

The result converter now throws dedicated exceptions:

 - `401` → `AuthenticationException`
 - `400` → `BadRequestException`
 - `429` → `RateLimitExceededException` (with `retry-after` header propagated)

Error messages are extracted from the response body, supporting both the top-level `message` shape and the nested `error.message` shape.

```php
try {
    $platform->invoke(new Perplexity('sonar'), $messages);
} catch (RateLimitExceededException $e) {
    sleep($e->getRetryAfter() ?? 60);
    // retry...
} catch (AuthenticationException $e) {
    // invalid API key -> message from the API is now exposed
}
```

HTTP status handling is guarded behind an \`instanceof RawHttpResult\` branch so the existing `InMemoryRawResult`-based streaming tests keep working untouched.

Part of the effort to address #528 across Platform bridges. Previous steps: #1961 (Mistral), #1962 (DeepSeek), #1963 (Cerebras).

PR #1966, opened 2026-04-18
[Platform][Gemini] Expose API errors for better Developer Experience

Amoifr:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | Fix #528
| License       | MIT

The Gemini bridge LLM result converter already handled `429` through `RateLimitExceededException`, but auth failures (`401`) and bad requests (`400`) still surfaced as a generic `RuntimeException` with a verbose `Error "<code>" - "<status>": "<message>"` string that consumers had to parse. The Embeddings result converter did not inspect the HTTP status at all.

Both result converters now throw dedicated exceptions:

Error messages are extracted from Gemini's nested `error.message` shape.

```php
try {
    $platform->invoke(new Gemini('gemini-1.5-flash'), $messages);
    // retry...
}
```

The existing LLM "detailed error" regression test is moved from status 400 to 500 since 400 now falls through the dedicated `BadRequestException` branch.

HTTP status handling in the Embeddings converter is guarded behind an \`instanceof RawHttpResult\` branch to preserve fake-result testing paths.

Part of the effort to address #528 across Platform bridges. Previous steps: #1961 (Mistral), #1962 (DeepSeek), #1963 (Cerebras), #1964 (Perplexity), #1965 (Cohere).


Amoifr:

PR #1967, opened 2026-04-18
[Platform][OpenAI] Expose API errors in Embeddings, DallE and TextToSpeech converters

Amoifr:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | Fix #528
| License       | MIT

The OpenAI bridge already handled auth failures / bad requests / rate limits in the `Gpt` and `Whisper` result converters, but three other sub-clients (`Embeddings`, `DallE`, `TextToSpeech`) were still collapsing non-200 responses into generic `RuntimeException`s:

 - **Embeddings** surfaced auth/rate-limit errors as "Response does not contain data."
 - **DallE** surfaced them as "No image generated."
 - **TextToSpeech** surfaced them as "The OpenAI Text-to-Speech API returned an error: ..."

Each result converter now throws dedicated exceptions, matching `Gpt` and `Whisper`:

 - `401` → `AuthenticationException`
 - `400` → `BadRequestException`
 - `429` → `RateLimitExceededException` (with `retry-after` header propagated)

Error messages are extracted from the response body, supporting both the top-level `message` shape and the nested `error.message` shape.

```php
try {
    $platform->invoke(new DallE('dall-e-3'), $prompt);
} catch (RateLimitExceededException $e) {
    sleep($e->getRetryAfter() ?? 60);
    // retry...
} catch (AuthenticationException $e) {
    // invalid API key -> message from the API is now exposed
}
```

Split into one commit per sub-client (Embeddings / DallE / TextToSpeech) for easier review. HTTP status handling in Embeddings and DallE is guarded behind an \`instanceof RawHttpResult\` branch to preserve fake-result testing paths.

The existing TextToSpeech regression test is moved from status 400 to 500 since 400 now falls through the dedicated `BadRequestException` branch.

Part of the effort to address #528 across Platform bridges. Previous steps: #1961 (Mistral), #1962 (DeepSeek), #1963 (Cerebras), #1964 (Perplexity), #1965 (Cohere), #1966 (Gemini).

PR #1968, opened 2026-04-18
[Platform] Expose API errors uniformly across bridges through a shared trait

Amoifr:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | Fix #528
| License       | MIT

## Origin & context

This PR consolidates the effort started with a series of per-bridge PRs (#1961 Mistral, #1962 DeepSeek, #1963 Cerebras, #1964 Perplexity, #1965 Cohere, #1966 Gemini, #1967 OpenAI) which are now **closed in favour of this one**, following feedback from:

- @OskarStark on #1965: *"Can you merge all these kind of Pr into one please?"*
- @chr-hertel on #1961:
  1. *"is a better way for avoiding the duplication here?"*
  2. *"can you help with reviewing this for providing the links to the corresponding docs?"*
  3. Inline on `extractErrorMessage`: *"can we use the `$httpResponse->toArray(false)` instead of this?"*

All three points are addressed here.

### 1. New shared trait: `Symfony\AI\Platform\Result\HttpStatusErrorHandlingTrait`

Exposes `throwOnHttpError(ResponseInterface $response): void` that translates the three common AI-provider HTTP errors into dedicated exceptions:

 - `429` → `RateLimitExceededException` (with `retry-after` propagated)

Error bodies are parsed via `$response->toArray(false)` wrapped in a `try { ... } catch (DecodingExceptionInterface)` so empty / non-JSON bodies fall back to default messages (`"Unauthorized"` / `"Bad Request"`). Both the top-level `message` and nested `error.message` shapes are supported, since providers split roughly 50/50 between the two.

Covered by 10 dedicated tests in `tests/Result/HttpStatusErrorHandlingTraitTest.php`.

### 2. Migrate 14 bridge result converters to the trait

One `use HttpStatusErrorHandlingTrait;` + one `$this->throwOnHttpError(...)` call per converter, replacing the previous copy-paste blocks or (where applicable) the plain "Unexpected response code X" generic `RuntimeException`:

| Bridge | Sub-clients |
|---|---|
| Mistral | LLM, Embeddings |
| DeepSeek | (single) |
| Cerebras | (single) |
| Perplexity | (single) |
| Cohere | LLM, Embeddings, Reranker, SpeechToText |
| Gemini | LLM (replaces the previous 429-only branch), Embeddings |
| OpenAI | Embeddings, DallE, TextToSpeech |

Anthropic, OpenAI/Gpt and OpenAI/Whisper already had the pattern hardcoded and are left untouched in this PR — migrating them to the trait would be a trivial follow-up if desired.

### 3. Test cleanup

- Per-converter HTTP-status and rate-limit tests that existed in the original child PRs are dropped: the trait's own test covers all extraction edge cases.
- Two pre-existing regression tests (`Gemini/Tests/Gemini/ResultConverterTest::testConvertThrowsExceptionWithDetailedErrorInformation` and `OpenAi/Tests/TextToSpeech/ResultConverterTest::testThrowsOnErrorResponse`) relied on status `400` falling through the generic path; since `400` now resolves to `BadRequestException` they are switched to status `500` (same code path, same assertion).

## Usage

```php
try {
    \$platform->invoke(\$model, \$messages);
} catch (RateLimitExceededException \$e) {
    sleep(\$e->getRetryAfter() ?? 60);
    // retry...
} catch (AuthenticationException \$e) {
    // invalid API key — message from the provider is now exposed
} catch (BadRequestException \$e) {
    // malformed request — message from the provider is now exposed
}
```

## Provider docs consulted

MockResponses are inferred from public docs + OpenAI-compat assumptions. I did not spin up real accounts for every provider — happy to validate specific ones if you consider that non-negotiable.

 - Mistral: https://docs.mistral.ai/api/
 - DeepSeek: https://api-docs.deepseek.com/
 - Cerebras: https://inference-docs.cerebras.ai/
 - Perplexity: https://docs.perplexity.ai/
 - Cohere: https://docs.cohere.com/reference
 - Gemini: https://ai.google.dev/api/rest
 - OpenAI: https://platform.openai.com/docs/api-reference

## Commits

1. `[Platform] Add HttpStatusErrorHandlingTrait for shared API error handling` — trait + 10 tests
2. `[Platform] Route bridge result converters through HttpStatusErrorHandlingTrait` — 14 converters + CHANGELOG + 2 regression-test status bumps

Thanks again @chr-hertel and @OskarStark for the reviews on the original series — the final shape is much cleaner thanks to your feedback 🙏

### DIFF
--- a/src/platform/CHANGELOG.md
+++ b/src/platform/CHANGELOG.md
@@ -17,6 +17,7 @@ CHANGELOG
  * Add `ModelRouterInterface` with `CatalogBasedModelRouter` as default strategy, and `CompositeModelCatalog` merging provider catalogs
  * Add `ModelRoutingEvent` dispatched by `Platform` before resolution, allowing listeners to observe/modify routing or short-circuit it by setting a provider
  * [BC BREAK] `Platform` constructor signature changed from `(modelClients, resultConverters, modelCatalog, contract, eventDispatcher)` to `(providers, modelRouter, eventDispatcher)`
+ * Add `HttpStatusErrorHandlingTrait` and expose API errors (`AuthenticationException`, `BadRequestException`, `RateLimitExceededException`) across the Mistral, DeepSeek, Cerebras, Perplexity, Cohere, Gemini and OpenAI (Embeddings, DallE, TextToSpeech) bridges
 
 0.7
 ---
--- a/src/platform/src/Bridge/Cerebras/ResultConverter.php
+++ b/src/platform/src/Bridge/Cerebras/ResultConverter.php
@@ -15,6 +15,8 @@ use Symfony\AI\Platform\Bridge\Generic\Completions\CompletionsConversionTrait;
 use Symfony\AI\Platform\Exception\RuntimeException;
 use Symfony\AI\Platform\Model as BaseModel;
 use Symfony\AI\Platform\Result\ChoiceResult;
+use Symfony\AI\Platform\Result\HttpStatusErrorHandlingTrait;
+use Symfony\AI\Platform\Result\RawHttpResult;
 use Symfony\AI\Platform\Result\RawResultInterface;
 use Symfony\AI\Platform\Result\ResultInterface;
 use Symfony\AI\Platform\Result\StreamResult;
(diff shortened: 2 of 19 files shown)


chr-hertel (approved):

chr-hertel:
Thank you @Amoifr.

PR #1970, opened 2026-04-19
[Platform][Gemini][VertexAi] Add Google Maps server tool

valtzu:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| License       | MIT

Add possibility to pass `tool_config` to Gemini which allows passing a location to Google Maps server tool. For Vertex this already works, as I found out by adding an example.

### DIFF
--- a/docs/components/platform/gemini-server-tools.rst
+++ b/docs/components/platform/gemini-server-tools.rst
@@ -51,6 +51,26 @@ The Google Search tool enables the model to search the web and incorporate searc
         ]
     ]);
 
+Google Maps
+~~~~~~~~~~~
+
+The Google Maps tool enables the model to find places based on location::
+
+    $messages = new MessageBag(
+        Message::ofUser('Find the nearest hotel')
+    );
+
+    $result = $platform->invoke('gemini-2.5-pro', $messages, [
+        'server_tools' => [
+            'google_maps' => true,
+        ],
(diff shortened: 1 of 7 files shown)


valtzu:
I think we should do similar thing for Gemini config handling as is done for Vertex: options should not automatically go under `generationConfig`, but to the top-level. This would allow adopting new features without waiting updates in the bridge.

But this seemed quite a big breaking change, hence excluded here.

OskarStark (approved):

docs/components/platform/gemini-server-tools.rst
```diff
@@ -51,6 +51,26 @@ The Google Search tool enables the model to search the web and incorporate searc
...
+        Message::ofUser('Find the nearest hotel')
+    );
+
+    $result = $platform->invoke('gemini-2.5-pro', $messages, [
+        'server_tools' => [
+            'google_maps' => true,
+        ],
+        'tool_config' => [
+            'retrieval_config' => [
+                'lat_lng' => ['latitude' => 60.16653, 'longitude' => 24.93061],
```
chr-hertel, line 69: funny, this is basically the spot in Helsinki I've been the most 🤨 😂

chr-hertel:
Thank you @valtzu.

Issue #1969, opened 2026-04-18
[Examples] Include `AssistantMessage` usage in examples


valtzu:
Currently no examples include the usage of `AssistantMessage`. Not having it there means there is no easy way to test changes (with the actual APIs) to `AssistantMessage` normalization.

I think we could add it to all `chat.php` examples since multi-turn conversation is a fundamental concept of a chat.

chr-hertel:
Yes, fair point, with all that possibilities, especially with loops and in context of #1853, this becomes more important 👍

Amoifr:
Hi @valtzu, @chr-hertel — opened #2034 with the change.

Each `examples/<bridge>/chat.php` now does a second turn that feeds the assistant's reply back via `Message::ofAssistant()` before sending a context-dependent follow-up, so all 21 chat examples can serve as a smoke test for `AssistantMessage` normalization.

Happy to iterate if the follow-up phrasing or the multi-turn shape needs tweaking. 🙂

chr-hertel:
Solved with #1997 and #2059 I'd say - even if not using the `chat.php` examples - works for you @valtzu?

valtzu:
Great, thanks 👍

PR #1980, opened 2026-04-21
[Platform][Gemini] Gemini call after a tool action fail

froozeify:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix, see below
| License       | MIT

While working on a chat integration, after doing some test with Gemini (Gemini 2.5 flash).

I got the following exception

```
Error "400" - "INVALID_ARGUMENT": "Invalid value at 'contents[5].parts[0]' (oneof), oneof field 'data' is already set. Cannot set 'functionCall'".
```

I was able to reproduce it every-time by doing the following:

1. Send a chat message
2. Send a chat message using a tool
3. Send a regular chat message -> It fail and return the previous exception message

### DIFF
--- a/src/platform/src/Bridge/Gemini/Contract/AssistantMessageNormalizer.php
+++ b/src/platform/src/Bridge/Gemini/Contract/AssistantMessageNormalizer.php
@@ -24,28 +24,33 @@ final class AssistantMessageNormalizer extends ModelContractNormalizer
     /**
      * @param AssistantMessage $data
      *
-     * @return array{array{text: string}}
+     * @return list<array{text: string}|array{functionCall: array{id: string, name: string, args?: mixed}}>
      */
     public function normalize(mixed $data, ?string $format = null, array $context = []): array
     {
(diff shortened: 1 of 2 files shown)


OskarStark:
Please add a testcase

chr-hertel:
Hey @froozeify, thanks for reporting & jumping on it! And welcome to Symfony AI repo!

I was ~~not~~ able to reproduce it right away, but changes look logical from contract point of view to me.

~~could you also provide a simple script in `examples/gemini/` like `toolcall-roundtrip.php` or something?
that'd be great for catching regressions here later and pays into~~ #1969

Thanks already!

chr-hertel (approved):

chr-hertel:
Thank you @froozeify.

froozeify:
No problem :) @chr-hertel If you like can take a look this weekend and create another PR for the `examples/gemini/` folder ?

PR #1981, opened 2026-04-21
[Platform][Anthropic] Fix tool calls dropped

Tiriel:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Docs?         | no 
| Issues        | No ticket
| License       | MIT

Fix tools calls dropped when Anthropic returns text before tool_use

When the model generates a text preamble before calling a tool, the API response content array is [text_block, tool_use_block]. The converter flushed accumulated $toolCalls to $results only when a non-tool_use block appeared *after* them; tool calls at the end of the array were never flushed, causing them to be silently dropped and only the text returned.

Add a post-loop flush so any remaining $toolCalls are always appended to $results before the final return, regardless of content ordering.

### DIFF
--- a/src/platform/src/Bridge/Anthropic/ResultConverter.php
+++ b/src/platform/src/Bridge/Anthropic/ResultConverter.php
@@ -83,16 +83,10 @@ class ResultConverter implements ResultConverterInterface
         }
 
         $results = [];
-        $toolCalls = [];
         foreach ($data['content'] as $content) {
             if ('tool_use' === $content['type']) {
-                $toolCalls[] = new ToolCall($content['id'], $content['name'], $content['input']);
+                $results[] = new ToolCallResult([new ToolCall($content['id'], $content['name'], $content['input'])]);
                 continue;
-            }
-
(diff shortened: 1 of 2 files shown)


OskarStark:
Please add a testcase

Tiriel:
> Please add a testcase

My bad, done!

chr-hertel:
I wonder if we should just drop the `continue` on L90 :thinking: 

cc @valtzu any downsides to that on your mind? would fix the test provided as well

valtzu:
I think it would indeed be clearer to construct ToolCallResults as they appear instead of grouping them and creating a single result - I only left it in place not to change the existing behavior

Tiriel:
> I think it would indeed be clearer to construct ToolCallResults as they appear instead of grouping them and creating a single result - I only left it in place not to change the existing behavior

Didn't drop the `continue` but removed the accumulator pattern, less changes this way I think

chr-hertel (approved):

chr-hertel:
Thank you @Tiriel.

PR #1982, opened 2026-04-21
[Platform][Codex] Align model catalog with official docs

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | -
| License       | MIT

The Codex `ModelCatalog` listed several models that the Codex CLI no longer accepts with a ChatGPT account — calling them returns:

```
The 'gpt-5-codex' model is not supported when using Codex with a ChatGPT account.
```

This surfaced while trying to run the bundled `examples/codex/*.php` scripts, all four of which hardcoded `gpt-5-codex` and therefore failed end-to-end.

Aligning the catalog with the official list at https://developers.openai.com/codex/models:

- Keep: `gpt-5.4`, `gpt-5.4-mini`, `gpt-5.3-codex`, `gpt-5.3-codex-spark`
- Add: `gpt-5.2`
- Remove: `gpt-5.2-codex`, `gpt-5.1-codex`, `gpt-5-codex`, `gpt-5-codex-mini`

Also:

- Adds a `@see https://developers.openai.com/codex/models` link to the `ModelCatalog` PHPDoc header so the source of truth is easy to find.
- Switches the four examples from `gpt-5-codex` to `gpt-5.4` (the currently recommended default). All four now run successfully (verified against a local Codex CLI 0.122.0 install).
- Updates the `ModelCatalogTest` data provider to match.

### DIFF
--- a/examples/codex/chat.php
+++ b/examples/codex/chat.php
@@ -24,7 +24,7 @@ $platform = Factory::createPlatform(
 $messages = new MessageBag(
     Message::ofUser('Explain the architecture of this project in 3 sentences.'),
 );
-$result = $platform->invoke('gpt-5-codex', $messages, [
+$result = $platform->invoke('gpt-5.4', $messages, [
     'sandbox' => 'read-only',
 ]);
 
--- a/examples/codex/code-generation.php
+++ b/examples/codex/code-generation.php
@@ -30,7 +30,7 @@ $platform = Factory::createPlatform(
 $messages = new MessageBag(
     Message::ofUser('Create a file called hello.php that prints "Hello, World!" to the console.'),
 );
-$result = $platform->invoke('gpt-5-codex', $messages, [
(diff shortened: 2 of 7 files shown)


chr-hertel:
:tada: 

image

chr-hertel:
interesting tho: is it really a breaking change if those models are not around anymore anyways? using those models would break with v0.8.0 already, right?

chr-hertel (approved):

chr-hertel:
Thank you @wachterjohannes.

wachterjohannes:
> interesting tho: is it really a breaking change if those models are not around anymore anyways? using those models would break with v0.8.0 already, right?

yes it is a breaking change thats right have you added that to the upgrades?

chr-hertel:
Have not, patched it into 0.8.1

PR #1986, opened 2026-04-23
[AI Bundle] Prompt for missing arguments in `ai:platform:invoke`

### LINKED ISSUE #1924: [AI Bundle] Prompt for provider, model if missing in ai:platform:invoke
bin/console  ai:platform:invoke

fails because it needs the provider.  Since those are configured, what about prompting from the list with a ChoiceQuestion?

If the model is missing, we could prompt with a regular ask, or use the the package someone just released (blanking now where it is) that lists all the models by provider. 

Then of course we can ask for the prompt.  

More brainstorming: add a --prompt-file option that's the name of a file.  Then I could test different providers/models with the identical prompt.

I can make a PR if you think any of these are worthwhile.

Amoifr:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | Fix #1924
| License       | MIT

As suggested by @chr-hertel in #1924, this improves the DX of `ai:platform:invoke` by making all three arguments optional and prompting for them interactively when missing.

### Behavior

- `platform` → `ChoiceQuestion` populated from the configured `ServiceLocator` of platforms
- `model` → free-text prompt
- `message` → free-text prompt

Existing non-interactive usage (`bin/console ai:platform:invoke openai gpt-4o-mini "Hello"`) is unchanged. Shell completion for `platform` (already present) is preserved.

### Example

```console
$ bin/console ai:platform:invoke

 Select a platform:
  [0] openai
  [1] anthropic
 > 0

 Which model do you want to use?:
 > gpt-4o-mini

 Enter the message to send:
 > Hello!

 Response: Hi there!
```

### Notes

- `initialize()` was removed and its validation moved to `execute()` because `Command::run()` calls `initialize()` *before* `interact()`, which would cause the empty platform name to be validated before the prompt could fill it.
- Per-provider model autocompletion and a `--prompt-file` option (also mentioned in #1924) are intentionally left out to keep this PR focused — happy to follow up on either.

### DIFF
--- a/src/ai-bundle/CHANGELOG.md
+++ b/src/ai-bundle/CHANGELOG.md
@@ -5,6 +5,7 @@ CHANGELOG
 ---
 
  * Validate structured output using `symfony/validator` when available
+ * Make `ai:platform:invoke` arguments optional and prompt for them interactively when missing
 
 0.8
 ---
--- a/src/ai-bundle/src/Command/PlatformInvokeCommand.php
+++ b/src/ai-bundle/src/Command/PlatformInvokeCommand.php
@@ -35,10 +35,6 @@ use Symfony\Component\DependencyInjection\ServiceLocator;
 )]
 final class PlatformInvokeCommand extends Command
 {
-    private string $message;
-    private PlatformInterface $platform;
-    private string $model;
-
     /**
      * @param ServiceLocator<PlatformInterface> $platforms
      */
@@ -58,9 +54,9 @@ final class PlatformInvokeCommand extends Command
     protected function configure(): void
     {
         $this
-            ->addArgument('platform', InputArgument::REQUIRED, 'The name of the configured platform to invoke')
-            ->addArgument('model', InputArgument::REQUIRED, 'The model to use for the request')
-            ->addArgument('message', InputArgument::REQUIRED, 'The message to send to the AI platform')
+            ->addArgument('platform', InputArgument::OPTIONAL, 'The name of the configured platform to invoke')
+            ->addArgument('model', InputArgument::OPTIONAL, 'The model to use for the request')
+            ->addArgument('message', InputArgument::OPTIONAL, 'The message to send to the AI platform')
             ->setHelp(
                 <<<'HELP'
                 The <info>%command.name%</info> command allows you to invoke configured AI platforms with a message.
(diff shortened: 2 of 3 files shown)


src/ai-bundle/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ CHANGELOG
  * [BC BREAK] Rename service ID `ai.agent.response_format_factory` to `ai.platform.response_format_factory`
  * The `collection` option for `ChromaDb` is now optional.
  * Update `DataCollector` to use `getCalls()` and `getResultCache()` getter methods on Traceable* classes
+ * Make `ai:platform:invoke` arguments optional and prompt for them interactively when missing
```
OskarStark: For 0.9 please

OskarStark (approved):

src/ai-bundle/CHANGELOG.md
Amoifr: Done, moved to 0.9! Thanks for the quick review 🙏

src/ai-bundle/src/Command/PlatformInvokeCommand.php
```diff
@@ -72,33 +68,49 @@ protected function configure(): void
...
     {
-        $platformName = trim((string) $input->getArgument('platform'));
+        $io = new SymfonyStyle($input, $output);
+        $configuredPlatforms = array_keys($this->platforms->getProvidedServices());
 
-        if (!$this->platforms->has($platformName)) {
-            throw new InvalidArgumentException(\sprintf('Platform "%s" not found. Available platforms: "%s"', $platformName, implode(', ', array_keys($this->platforms->getProvidedServices()))));
+        if (null === $input->getArgument('platform') && [] !== $configuredPlatforms) {
+            $input->setArgument('platform', $io->choice('Select a platform', $configuredPlatforms));
+        }
```
chr-hertel: I think we can inline this and skip the `$this->platforms->getProvidedServices()` call if the argument is actually provided:

```suggestion
        if (null === $input->getArgument('platform')) {
            $input->setArgument('platform', $io->choice('Select a platform', array_keys($this->platforms->getProvidedServices())));
        }
```

chr-hertel (changes requested):
Nice one, thanks! Small change and we're good to merge it in.

Amoifr:
Thanks @chr-hertel! 🙏

Applied the suggestion as-is in [`b97b0c88`](https://github.com/symfony/ai/pull/1986/commits/b97b0c88) — inlined the `array_keys()` call so we skip the lookup entirely when the argument is provided.

I also rebased on `main` to clear the merge conflict (CHANGELOG bullet right next to the `symfony/validator` one from #2002).

Should be good to go now! 🚀

chr-hertel (approved):

chr-hertel:
Thank you @Amoifr.

PR #1987, opened 2026-04-23
[Chat][Doctrine] Replace row-accumulation in DoctrineDbalMessageStore

Tiriel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Docs?         | no 
| Issues        | None
| License       | MIT

Replace row-accumulation with single-row upsert in DoctrineDbalMessageStore.
Until now, DoctrineDbalMessage was accumulating messages in new rows. Each row contained all the previous messages plus the new ones, leading to exponential growth of context, warped message history, and profiler exhaustion.
This aims to address this problem and align the message store's design with the other stores'.

### DIFF
--- a/src/chat/src/Bridge/Doctrine/CHANGELOG.md
+++ b/src/chat/src/Bridge/Doctrine/CHANGELOG.md
@@ -4,7 +4,7 @@ CHANGELOG
 0.8
 ---
 
- * [BC BREAK] `DoctrineDbalMessageStore::save()` now upserts a single row instead of inserting a new row on every call; the table schema changed from `(id, messages, added_at)` to `(messages, updated_at)`
+ * [BC BREAK] `DoctrineDbalMessageStore::save()` now upserts a single row instead of inserting a new row on every call; the table schema is unchanged but `added_at` is refreshed on every save
 
 0.1
 ---
--- a/src/chat/src/Bridge/Doctrine/CHANGELOG.md
+++ b/src/chat/src/Bridge/Doctrine/CHANGELOG.md
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+0.8
+---
+
+ * [BC BREAK] `DoctrineDbalMessageStore::save()` now upserts a single row instead of inserting a new row on every call; the table schema changed from `(id, messages, added_at)` to `(messages, updated_at)`
+
 0.1
 ---
 
--- a/src/chat/src/Bridge/Doctrine/DoctrineDbalMessageStore.php
+++ b/src/chat/src/Bridge/Doctrine/DoctrineDbalMessageStore.php
@@ -12,11 +12,7 @@
 namespace Symfony\AI\Chat\Bridge\Doctrine;
 
 use Doctrine\DBAL\Connection as DBALConnection;
-use Doctrine\DBAL\Platforms\OraclePlatform;
 use Doctrine\DBAL\Schema\ComparatorConfig;
-use Doctrine\DBAL\Schema\Name\Identifier;
-use Doctrine\DBAL\Schema\Name\UnqualifiedName;
-use Doctrine\DBAL\Schema\PrimaryKeyConstraint;
 use Doctrine\DBAL\Schema\Schema;
 use Doctrine\DBAL\Types\Types;
 use Psr\Clock\ClockInterface;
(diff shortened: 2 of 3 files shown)


Tiriel:
Note: Schema changed because `id` column is no longer needed as a single row get updated each time. But if the BC Break is too violent, this part can easily be reverted

Tiriel:
The Fabpot fail seems to be a false positive, it flags `hasTable` as a typo for `hashtable`. Maybe something to add to an exclusion list?

chr-hertel:
Looks like there are quite some changes for that bridge in here - look reasonable to me at first glance, but this is not only addressing insert vs update, but also the schema in general - can you share your thoughts here, please?

cc @Guikingone @xprojects-de: might be relevant for you?

xprojects-de:
Thank you @Tiriel and I think the idea is basically good, and I also think that saving same messages multiple times is not good.

I'm thinking about the future and I think we should consider how we can implement some kind of chat history for multiple clients in the future. Here, for example, a clientId or an identifier might need to be stored also, and then there could be several messages in the table.  

Therefore, I would leave the schema including Id, etc. as it is, so that we can implement a multi-client history in the future.

Furthermore, I think we shouldn't simply switch to a single entry. 
Maybe it might be possible to implement some kind of flag in the setup method so that only one message is saved, if necessary / wanted ? Perhaps we should consider a client identifier to implement multi-client chat !?

Guikingone:
@chr-hertel As @xprojects-de said, If we want to tackle the "session-based" chat issue one day (the PR that aims to solve this issue is waiting for the type of change that @Tiriel is currently doing), we need to change the way Doctrine is storing the messages, is it a BC break? Nope, Sf AI is not stable for now.

Should it be documented in UPDATE.md? Yes, at least to inform that starting from `0.9`, the Doctrine store must be reset before using it.

Regarding the flag, I don't have a strong opinion on it, for me, using flags is a wrong approach as the core team will have to do more maintenance to ease the migration path later, either we break the store now and call it a day or we add a "double way to solve the issue" but introduce more work later.

Tiriel:
I reverted the schema as it was indeed a forward compatibility issue.

For the flag part, I'm not sure about this, as I quite don't see the appeal of cumulating messages as it was done before. As @Guikingone said, since AI is still experimental, I think we should go for the clean cut.

But I can still try to implement this if needed. Commit is ready to be discussed.

xprojects-de:
👍 Sounds good to me. In any case, I'm already looking forward to the session-based store. 👍

Tiriel:
Little up on this. Is there any blocker? @chr-hertel

loancarre:
hello, any news of this ?

Tiriel:
> hello, any news of this ?

Same question! @chr-hertel @xprojects-de any update/ anything blocking the merge?

src/chat/src/Bridge/Doctrine/DoctrineDbalMessageStore.php
```diff
@@ -85,43 +85,61 @@ public function drop(): void
...
+        } else {
+            $this->dbalConnection->executeStatement(
+                $this->dbalConnection->createQueryBuilder()
+                    ->update($this->tableName)
+                    ->set('messages', '?')
+                    ->set('added_at', '?')
+                    ->getSQL(),
+                [$serialized, $now],
+            );
+        }
```
wachterjohannes, line 124: Three things in this block, none of them raised in the thread so far:

1. `COUNT(*)` followed by INSERT-or-UPDATE is not atomic. Two concurrent `save()` calls both read 0 and both insert. A real upsert (`INSERT ... ON CONFLICT` / `MERGE`), or a transaction, would close it and be one statement instead of two.
2. The `UPDATE` has no `WHERE`, so it rewrites every row. On a table that already has N rows it sets all N to the same payload instead of collapsing them to one, which is why the row count in the trace above stays at 2 forever.
3. Because of that, a store that is not manually reset never heals itself.

A transaction around `DELETE` then `INSERT` would handle all three at once, and would make pre-existing rows disappear on the first save, which softens the manual reset step considerably.

src/chat/src/Bridge/Doctrine/DoctrineDbalMessageStore.php
```diff
@@ -85,43 +85,61 @@ public function drop(): void
...
-            ->orderBy('added_at', 'ASC')
-        ;
-
-        $result = $this->dbalConnection->executeQuery($queryBuilder->getSQL());
+        $payload = $this->dbalConnection->executeQuery(
+            $this->dbalConnection->createQueryBuilder()
+                ->select('messages')
+                ->from($this->tableName)
+                ->getSQL()
+        )->fetchAssociative();
```
wachterjohannes, line 134: This dropped the `->orderBy('added_at', 'ASC')` the previous version had, and has no `LIMIT`, so when more than one row is present it returns whichever row the database hands back first. That is what turns leftover rows into the silent revert in the trace above.

Worth being explicit even once `save()` guarantees a single row: `->orderBy('added_at', 'DESC')->setMaxResults(1)` states the intent and stays deterministic if a stray row ever appears.

src/chat/src/Bridge/Doctrine/CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+0.8
+---
+
+ * [BC BREAK] `DoctrineDbalMessageStore::save()` now upserts a single row instead of inserting a new row on every call; the table schema is unchanged but `added_at` is refreshed on every save
```
wachterjohannes, line 7: Should be `0.13`, the next release.

I added the `BC Break` label, which means the changelog check stays red until `UPGRADE.md` gains an entry under `UPGRADE FROM 0.12 to 0.13`. That is the note @Guikingone asked for in May, and "reset the store before upgrading" is what belongs in it.

The description table should also say `Bug fix? yes`: this repairs broken behaviour rather than adding anything. With the label set, that combination is allowed to carry a changelog entry.

wachterjohannes (changes requested):
Sorry for the long wait on this one @Tiriel, you pinged twice and got nothing back. The direction is settled as far as I'm concerned: the accumulation is a genuine bug, and the clean cut that you, @xprojects-de and @Guikingone agreed on in May is the right call for an experimental component. The schema revert addressed the forward compatibility concern.

What is still open is the mechanics of the upsert itself, which nobody has looked at yet. Three points inline, plus @Guikingone's `UPGRADE.md` request from May is still outstanding.

To show why that upgrade note matters, I seeded a table the way the current code does and then ran your branch against it:

```
### main, seeding an existing installation
after 2 saves      rows=2  loaded=[turn one, turn one, turn two]

### this PR, same table
load on upgrade    rows=2  loaded=[turn one]
after next save    rows=2  loaded=[turn three]
```

So an installation that does not reset the store silently reverts to an older conversation, and the table never collapses to a single row. Documented that is an acceptable clean cut; undocumented it looks like data corruption.

PR #1992, opened 2026-04-23
 [Platform] Handle structured output decoding correctly, when model is returning it markdown code encapsulated

jbtronics:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  no
| Docs?         | no
| Issues        | None
| License       | MIT

Some models like "google/gemini-2.5-flash-lite" (under openrouter provider), return not raw json as output, but encapsulate it into a markdown code block like
````
```json
{}
```
````

Retrieving structured output fails then with an json decode error.

This PR modifies the StructuredOutput ResultConverter to properly handle that cases, by removing the encapsulation.

Alternatively that could be implemented in its own inner result converter, or handled by the models themselves, but not sure if that is worth the effort and complexity. This change not affects anything without markdown code encapsulation

### DIFF
--- a/src/platform/src/StructuredOutput/ResultConverter.php
+++ b/src/platform/src/StructuredOutput/ResultConverter.php
@@ -52,10 +52,17 @@ final class ResultConverter implements ResultConverterInterface
                 $context[AbstractNormalizer::OBJECT_TO_POPULATE] = $this->objectToPopulate;
             }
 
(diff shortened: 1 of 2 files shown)


jbtronics:
Apparently that was caused, by some wrong options to openrouter, so that effectively it did not respected the response_format.
So this is not necessary.

PR #1993, opened 2026-04-24
[Store][Postgres] Add a `StoreFactory`

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | Noticed by @salahm
| License       | MIT

* Add a missing `lang` parameter to `Store`
* Up PHPStan to `8`


src/store/src/Bridge/Postgres/StoreFactory.php
```diff
@@ -0,0 +1,49 @@
...
+        \PDO $connection,
+        string $tableName,
+        string $vectorFieldName = 'embedding',
+        Distance $distance = Distance::L2,
+        string $lang = 'english',
+    ): ManagedStoreInterface&StoreInterface {
+        return new Store($connection, $tableName, $vectorFieldName, $distance, $lang);
+    }
+
+    public static function createFromDbal(
```
stof: `StoreFactory::createFromDbal` looks like it creates a `StoreFactory` when reading the code. Should this be `createStoreFromDbal`

src/store/src/Bridge/Postgres/Store.php
```diff
@@ -84,30 +83,6 @@ public function drop(array $options = []): void
         $this->connection->exec(\sprintf('DROP TABLE IF EXISTS %s', $this->tableName));
     }
 
-    public static function fromPdo(
```
stof, line 87: those methods should be deprecated instead of removing them immediately.

stof (commented):
I don't see the benefit of a StoreFactory exposing a static API compared to the existing static factory methods. They don't provide any additional extension points, while breaking BC (without properly handling compatible versions of the bridge in the ai-bundle btw)

stof:
btw, changing the phpstan level should probably be a dedicated PR instead of hiding that inside a PR adding a feature (and then getting issues due to other parts of the code changing in ways where phpstan level 8 complains while the review of the feature takes time)

Guikingone:
The only change introduced by PHPStan at level `8` is the following [here](https://github.com/symfony/ai/pull/1993/changes#diff-1f6e652d903a7b110e3ff32bfd8ae9e1430c0c82a50b35310e20b2fc9582e14eR868): 

```php
if (null === $replacedQuery) {
    throw new RuntimeException('Query cannot be normalized.');
}
```

src/store/src/Bridge/Postgres/Store.php
Guikingone: @chr-hertel Should we handle deprecations as versions are dumped at each release? If I'm right, at the release of `0.9`, both `ai-bundle` and other components are automatically dumped to the next version so the bundle cannot use `0.8` when released as `0.9`?

src/store/src/Bridge/Postgres/Store.php
```diff
@@ -39,9 +37,38 @@ public function __construct(
...
+        trigger_error(\sprintf('Using "%s::fromDbal()" is deprecated since version 0.9, consider using "%s::createStoreFromDbal()".', self::class, StoreFactory::class), E_USER_DEPRECATED);
+
+        $pdo = $connection->getNativeConnection();
+
+        if (!$pdo instanceof \PDO) {
+            throw new InvalidArgumentException('Only DBAL connections using PDO driver are supported.');
+        }
+
+        return self::fromPdo($pdo, $tableName, $vectorFieldName, $distance);
+    }
```
chr-hertel: just remove them, we're not working with deprecations yet 👍

chr-hertel (approved):
Agree on the factory, rather a matter of consistency across bridges - we don't usually have stores with factory methods but factory classes.
Agree as well tho that in general we should limit PR scope more - this is basically three things.

chr-hertel:
Thank you @Guikingone.

PR #1996, opened 2026-04-24
[Chat] Fix MessageNormalizer round-trip for AssistantMessage with tool calls

### LINKED ISSUE #1972: [Chat] MessageNormalizer cannot round-trip AssistantMessage with tool calls
## Summary

`Symfony\AI\Chat\MessageNormalizer` writes the `toolsCalls` payload of an `AssistantMessage` / `ToolCallMessage` in a **flat** shape (`{id, name, arguments}`) but its `denormalize()` method reads it expecting the **OpenAI-style function-wrapped** shape (`{id, function: {name, arguments}}`). Any conversation that contains a `ToolCall` cannot be restored — `MessageNormalizer::denormalize()` crashes with:

```
Warning: Undefined array key "function"
 in vendor/symfony/ai-chat/src/MessageNormalizer.php on line 60
```

## Actual behavior

After storing an `AssistantMessage` with tool calls via `Serializer::normalize()` (e.g. for a persistent chat store) and loading it back via `Serializer::denormalize()`, the denormalizer accesses a key that the normalizer never produced.

Concrete DB row I recorded (JSON from a real, freshly-stored conversation):

```json
{
  "type": "Symfony\\AI\\Platform\\Message\\AssistantMessage",
  "content": "...",
  "toolsCalls": [
    {
      "id": "fc_00cf31962a87d2830069e5d0679540819081763cbb56e88ba6",
      "name": "search_contract_templates",
      "arguments": {"searchTerm": "Promoter-Vertrag", "status": null}
    }
  ]
}
```

Note: **no `function` key**. The next `denormalize()` on this data throws.

## Expected behavior

Round-tripping a message through `normalize()` → `denormalize()` should produce an equivalent `AssistantMessage` without errors — i.e. the two halves of `MessageNormalizer` must agree on the on-wire shape.

## Root cause

The two sides of `MessageNormalizer` speak different dialects.

### 1. `normalize()` delegates to the outer Serializer

https://github.com/symfony/ai/blob/v0.7.0/src/chat/src/MessageNormalizer.php#L112-L118

```php
if ($data instanceof AssistantMessage && $data->hasToolCalls()) {
    $toolsCalls = $this->normalizer->normalize($data->getToolCalls(), $format, $context);
}
```

`$this->normalizer` comes from `NormalizerAwareTrait` — it is the *outer* Symfony Serializer that the `MessageNormalizer` is attached to. The bundle registers only `MessageNormalizer` itself as a `serializer.normalizer`:

https://github.com/symfony/ai/blob/v0.7.0/src/ai-bundle/config/services.php#L263-L264

```php
->set('ai.chat.message_bag.normalizer', MessageNormalizer::class)
    ->tag('serializer.normalizer')
```

`Symfony\AI\Platform\Contract\Normalizer\Result\ToolCallNormalizer` is **not** registered as a DI service anywhere in `symfony/ai-bundle`. It only ever lives inside a private `Serializer` instance built by `Symfony\AI\Platform\Contract`:

https://github.com/symfony/ai/blob/v0.7.0/src/platform/src/Contract.php

So when `MessageNormalizer::normalize()` asks the outer serializer to normalize a `ToolCall`, the outer serializer falls back to `ObjectNormalizer` and writes it flat:

```php
['id' => '…', 'name' => '…', 'arguments' => [...]]
```

### 2. `denormalize()` hardcodes the OpenAI/function-wrapped shape

https://github.com/symfony/ai/blob/v0.7.0/src/chat/src/MessageNormalizer.php#L57-L78

```php
AssistantMessage::class => new AssistantMessage($content, array_map(
    static fn (array $toolsCall): ToolCall => new ToolCall(
        $toolsCall['id'],
        $toolsCall['function']['name'],                       // <-- crashes
        json_decode($toolsCall['function']['arguments'], true)
    ),
    $data['toolsCalls'],
)),
// …
ToolCallMessage::class => new ToolCallMessage(
    new ToolCall(
        $data['toolsCalls']['id'],
        $data['toolsCalls']['function']['name'],
        json_decode($data['toolsCalls']['function']['arguments'], true)
    ),
    $content
),
```

Neither of the two reconstruction paths uses the outer serializer to denormalize a `ToolCall` — they read hardcoded keys that `normalize()` never emits.

The same code is present on `main`: https://github.com/symfony/ai/blob/main/src/chat/src/MessageNormalizer.php — so the bug is not specific to the v0.7.0 tag.

## Minimal reproducer

```php
<?php

use Symfony\AI\Chat\MessageNormalizer;
use Symfony\AI\Platform\Message\AssistantMessage;
use Symfony\AI\Platform\Message\MessageInterface;
use Symfony\Component\Serializer\Normalizer\ObjectNormalizer;
use Symfony\Component\Serializer\Normalizer\UidNormalizer;
use Symfony\Component\Serializer\Serializer;

$messageNormalizer = new MessageNormalizer();
$serializer = new Serializer([$messageNormalizer, new UidNormalizer(), new ObjectNormalizer()]);

$message = new AssistantMessage('ok', [
    new ToolCall('tc-1', 'search', ['q' => 'foo']),
]);

$data = $serializer->normalize($message);
// toolsCalls => [['id' => 'tc-1', 'name' => 'search', 'arguments' => ['q' => 'foo']]]
//                                  ^^^^^^^^^^^^^^^^^  no 'function' wrapper

$serializer->denormalize($data, MessageInterface::class);
// PHP Warning: Undefined array key "function"
//   at vendor/symfony/ai-chat/src/MessageNormalizer.php line 60
```

## Output
```
  Warning: Undefined array key "function" at src/chat/src/MessageNormalizer.php line 60
  Warning: Undefined array key "function" at src/chat/src/MessageNormalizer.php line 61
  TypeError: Symfony\AI\Platform\Result\ToolCall::__construct():
    Argument #2 ($name) must be of type string, null given,
    called in src/chat/src/MessageNormalizer.php on line 58
```

## Impact

Any application that persists chat history through `MessageNormalizer` and later resumes a conversation whose assistant messages contain tool calls will crash on the very next turn. This affects the whole `symfony/ai-chat` resume path as soon as tools are involved.

## Possible fix directions

Both halves need to agree. Off the top of my head:

1. **Normalize using an explicit `ToolCallNormalizer`** — either register `Symfony\AI\Platform\Contract\Normalizer\Result\ToolCallNormalizer` as a `serializer.normalizer` service in `symfony/ai-bundle`, or instantiate it inside `MessageNormalizer::normalize()` and use it to produce the wrapped shape. Then `denormalize()` reads what `normalize()` wrote.
2. **Symmetric flat shape** — keep the outer-serializer delegation (which produces the flat form) and rewrite `denormalize()` to read `id/name/arguments` directly.

Option 1 matches the OpenAI-style wire format that the `denormalize()` code was clearly written for; option 2 is smaller but throws away that intent.

Either way, a regression test that round-trips an `AssistantMessage` (and a `ToolCallMessage`) with `ToolCall`s through `MessageNormalizer` would prevent this from coming back.

## Environment

- `symfony/ai-chat`: **v0.7.0** (also reproducible on `main` — same code path)
- `symfony/ai-platform`: v0.7.0
- `symfony/ai-bundle`: v0.7.0
- Symfony: 7.4.8
- PHP: 8.4

Amoifr:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #1972
| License       | MIT

## Problem

When a conversation containing an `AssistantMessage` (or `ToolCallMessage`) with tool calls is persisted through `MessageNormalizer` and later restored, `denormalize()` crashes on `Undefined array key "function"`.

## Root cause

- `MessageNormalizer::denormalize()` reads the OpenAI-style **wrapped** shape: `{id, function: {name, arguments: JSON string}}`.
- `MessageNormalizer::normalize()` delegates the tool call payload to the outer `Serializer`. When `ToolCallNormalizer` is in the chain, the wrapped shape is produced; otherwise the default `ObjectNormalizer` falls back to the **flat** shape `{id, name, arguments: array}`.
- `ToolCallNormalizer` was never registered as a service in `ai-bundle`. In real Symfony apps, the outer serializer therefore produced the flat shape, and denormalize crashed. The chat component's existing round-trip tests only passed because they wire `ToolCallNormalizer` manually.

## Fix (two complementary changes)

1. **ai-bundle** — register `Symfony\AI\Platform\Contract\Normalizer\Result\ToolCallNormalizer` with the `serializer.normalizer` tag, so the outer serializer produces the wrapped shape the denormalizer expects.
2. **chat** — make `MessageNormalizer::denormalize()` tolerate **both** shapes (wrapped and flat). This lets apps restore conversations that were persisted before this fix landed (the reporter's own DB rows are in the flat form).

## Test plan

- `src/chat` — 41 tests pass, including 2 new regression tests exercising the flat-shape path without `ToolCallNormalizer` in the chain (`testItCanDenormalizeAssistantMessageWithFlatToolCalls`, `testItCanDenormalizeToolCallMessageWithFlatToolCall`).
- `src/ai-bundle` — 239 tests pass (+1 from this PR verifying the new service registration and tag), identical error count to main (8 pre-existing errors tied to the optional `symfony/ai-ovh-platform` package, unrelated to this change).
- `phpstan` clean on `src/chat`.

## Notes

- Thanks to @sjahns for the precise diagnosis in the issue — the wrapped-vs-flat asymmetry was described clearly enough that the fix direction was obvious.
- The `toolCallFromPayload()` helper centralizes the tolerance so there's a single place to revisit if we ever want to drop the flat-shape fallback later.

### DIFF
--- a/src/ai-bundle/CHANGELOG.md
+++ b/src/ai-bundle/CHANGELOG.md
@@ -7,6 +7,7 @@ CHANGELOG
  * [BC BREAK] Rename service ID `ai.agent.response_format_factory` to `ai.platform.response_format_factory`
  * The `collection` option for `ChromaDb` is now optional.
  * Update `DataCollector` to use `getCalls()` and `getResultCache()` getter methods on Traceable* classes
+ * Register `ToolCallNormalizer` as a `serializer.normalizer` service so `MessageNormalizer` can round-trip `AssistantMessage`/`ToolCallMessage` payloads containing tool calls
 
 0.7
 ---
--- a/src/ai-bundle/CHANGELOG.md
+++ b/src/ai-bundle/CHANGELOG.md
@@ -20,6 +20,7 @@ CHANGELOG
  * [BC BREAK] Rename service ID `ai.agent.response_format_factory` to `ai.platform.response_format_factory`
  * The `collection` option for `ChromaDb` is now optional.
  * Update `DataCollector` to use `getCalls()` and `getResultCache()` getter methods on Traceable* classes
+ * Register `ToolCallNormalizer` as a `serializer.normalizer` service so `MessageNormalizer` can round-trip `AssistantMessage`/`ToolCallMessage` payloads containing tool calls
 
 0.7
 ---
--- a/src/ai-bundle/config/services.php
+++ b/src/ai-bundle/config/services.php
@@ -25,6 +25,7 @@ use Symfony\AI\AiBundle\Security\EventListener\IsGrantedToolAttributeListener;
 use Symfony\AI\Chat\Command\DropStoreCommand as DropMessageStoreCommand;
 use Symfony\AI\Chat\Command\SetupStoreCommand as SetupMessageStoreCommand;
 use Symfony\AI\Chat\MessageNormalizer;
+use Symfony\AI\Platform\Contract\Normalizer\Result\ToolCallNormalizer;
 use Symfony\AI\Platform\Bridge\AiMlApi\ModelCatalog as AiMlApiModelCatalog;
 use Symfony\AI\Platform\Bridge\Albert\ModelCatalog as AlbertModelCatalog;
 use Symfony\AI\Platform\Bridge\AmazeeAi\ModelApiCatalog as AmazeeAiModelCatalog;
@@ -261,6 +262,8 @@ return static function (ContainerConfigurator $container): void {
 
         // serializer
         ->set('ai.chat.message_bag.normalizer', MessageNormalizer::class)
+            ->tag('serializer.normalizer')
+        ->set('ai.platform.tool_call.normalizer', ToolCallNormalizer::class)
             ->tag('serializer.normalizer')
 
         // commands
--- a/src/ai-bundle/config/services.php
+++ b/src/ai-bundle/config/services.php
@@ -68,6 +68,7 @@ use Symfony\AI\Platform\Contract\JsonSchema\Describer\SerializerDescriber;
 use Symfony\AI\Platform\Contract\JsonSchema\Describer\TypeInfoDescriber;
 use Symfony\AI\Platform\Contract\JsonSchema\Describer\ValidatorConstraintsDescriber;
 use Symfony\AI\Platform\Contract\JsonSchema\Factory as SchemaFactory;
+use Symfony\AI\Platform\Contract\Normalizer\Result\ToolCallNormalizer;
 use Symfony\AI\Platform\EventListener\TemplateRendererListener;
 use Symfony\AI\Platform\Message\TemplateRenderer\ExpressionLanguageTemplateRenderer;
 use Symfony\AI\Platform\Message\TemplateRenderer\StringTemplateRenderer;
@@ -267,6 +268,8 @@ return static function (ContainerConfigurator $container): void {
 
         // serializer
         ->set('ai.chat.message_bag.normalizer', MessageNormalizer::class)
+            ->tag('serializer.normalizer')
+        ->set('ai.platform.tool_call.normalizer', ToolCallNormalizer::class)
             ->tag('serializer.normalizer')
 
         // commands
--- a/src/ai-bundle/tests/DependencyInjection/AiBundleTest.php
+++ b/src/ai-bundle/tests/DependencyInjection/AiBundleTest.php
@@ -244,6 +244,23 @@ class AiBundleTest extends TestCase
         $this->assertTrue($container->hasDefinition('ai.chat.message_bag.normalizer'));
     }
 
+    public function testToolCallNormalizerIsRegistered()
+    {
+        $container = $this->buildContainer([
+            'ai' => [
+                'agent' => [
+                    'my_agent' => [
+                        'model' => 'gpt-4',
+                    ],
+                ],
+            ],
+        ]);
+
+        $this->assertTrue($container->hasDefinition('ai.platform.tool_call.normalizer'));
+        $definition = $container->getDefinition('ai.platform.tool_call.normalizer');
+        $this->assertArrayHasKey('serializer.normalizer', $definition->getTags());
+    }
+
     public function testResponseFormatFactoryServiceIsRegistered()
     {
         $container = $this->buildContainer($this->getFullConfig());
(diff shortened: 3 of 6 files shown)


sjahns:
Bumping this from a v0.9 reality check — the bug is unchanged and v0.9 added a second affected code path that this PR (as-is) wouldn't cover after rebase.

In v0.9 `MessageNormalizer::normalize()` writes a `parts` array next to `toolsCalls`. Each tool call becomes `['type' => ToolCall::class, 'toolCall' => $this->normalizer->normalize($part, …)]` — same root cause as before, the outer serializer flattens that `toolCall` sub-array via `ObjectNormalizer`. On the read side, `denormalizeAssistantParts()` prefers the `parts` path when present and reads `$part['toolCall']['function']['name']` at line ~196. So in vanilla v0.9 an `AssistantMessage` with tool calls now crashes at line 196 before the legacy `toolsCalls` read at line 60 is ever reached.

What the two changes in this PR mean after rebasing onto v0.9:

- **ai-bundle service registration (your change #1)** — heals both paths going forward, because the outer serializer would produce the wrapped shape in both `normalize()` and `normalizeAssistantParts()`. Good.
- **chat `toolCallFromPayload()` tolerance (your change #2)** — only routed through `denormalize()`, not through `denormalizeAssistantParts()`. So v0.9 rows already on disk with flat `parts[*].toolCall` would still fail to read.

Suggestion if you rebase: route the `ToolCall::class` arm in `denormalizeAssistantParts()` through the same `toolCallFromPayload()` helper. Keeps tolerance consistent across both legacy shapes.

Re the failing CI on this PR: the "common typos" check is tripping on `ai.store.vektor` from an unrelated existing file — looks orthogonal to the change here.

Amoifr:
Thanks for the v0.9 reality check, @sjahns — you nailed it. Rebased on `main`, and the v0.9 `parts[*].toolCall` path was indeed crashing exactly where you predicted.

Took your suggestion: the `ToolCall::class` arm in `denormalizeAssistantParts()` now goes through the same `toolCallFromPayload()` helper as the legacy `toolsCalls` arm and the `ToolCallMessage` arm — single source of tolerance for both wrapped and flat shapes, covering both legacy payloads and v0.9 `parts[*].toolCall` payloads persisted before the ai-bundle service registration landed. Added a regression test (`testItCanDenormalizeAssistantMessageWithFlatToolCallInParts`) mirroring the existing flat-shape ones.

Local run: chat 43/43 ✅, ai-bundle 370/370 ✅, phpstan on chat clean.

Re the "common typos" CI: agreed, looks orthogonal — `ai.store.vektor` lives in unrelated existing config. Hopefully it clears after the rebase, otherwise happy to look if maintainers want me to.

PR #1998, opened 2026-04-24
 Add docker compose and Makefile for a better DX

froozeify:
…stall locally

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | DX
| Docs?         | no 
| Issues        | 
| License       | MIT

This is an idea, to help contributor (like me) that don't have PHP install locally and rather run it through Docker and with some help of Makefile.

Probably some improvements to it can be made

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,0 +1,24 @@
+services:
+    php:
+        build:
+            context: .docker
+        volumes:
+            - .:/app:rw
+            - composer-cache:/home/app/.composer
+        working_dir: /app
+        command: sleep infinity
+        depends_on:
+            - mongodb
+            - redis
+    mongodb:
+        image: mongo:8
+        expose:
+            - "27017"
+
+    redis:
+        image: redis:8-alpine
+        expose:
+            - "6379"
+
+volumes:
+    composer-cache:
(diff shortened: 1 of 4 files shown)


docker-compose.yaml
```diff
@@ -0,0 +1,24 @@
...
+    mongodb:
+        image: mongo:8
+        expose:
+            - "27017"
+
+    redis:
+        image: redis:8-alpine
+        expose:
+            - "6379"
+
```
Guikingone, line 22: Maybe we can extend the one from `examples`?

OskarStark:
This is too much to maintain from my side to be honest. Php should be the minimum requirement imho. The docker compose file could be extended via recipes like we finger mercure for example

chr-hertel:
Thanks for the proposal @froozeify - I can see where you're coming from and that this might help for some, but the risk is to have it in the repo without using it ourselves, in the end having an unmaintained potentially broken setup - which would create even more friction for starters. in the end it's common across all Symfony repos to require a PHP installation.

(i might come back to the makefile tho later based on my use cases.)

froozeify:
No problem, it was an idea and yes totally understandable.   
Having to keep in mind maintainability in the future is very important.

PR #2002, opened 2026-04-25
[Platform] Validate structured output

valtzu:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| License       | MIT

Validate structured output using `symfony/validator` when available

### DIFF
--- a/src/platform/src/Exception/ValidationException.php
+++ b/src/platform/src/Exception/ValidationException.php
@@ -1,0 +1,31 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Platform\Exception;
+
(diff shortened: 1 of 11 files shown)


src/platform/src/Exception/ValidationException.php
```diff
@@ -0,0 +1,34 @@
...
+final class ValidationException extends RuntimeException
+{
+    public function __construct(
+        private readonly ConstraintViolationListInterface $violations,
+        string $message = 'Validation failed for the structured output.',
+        int $code = 0,
+        ?\Throwable $previous = null,
+    ) {
+        parent::__construct($message, $code, $previous);
+    }
```
chr-hertel, line 25: I think it's fine to keep it minimal and tailored to the one use-case we have here:
```suggestion
    public function __construct(
        private readonly ConstraintViolationListInterface $violations,
    ) {
        parent::__construct('Validation failed for the structured output.');
    }
```

chr-hertel:
This makes so much sense - thanks!

chr-hertel (approved):

chr-hertel:
Thank you @valtzu.

PR #2006, opened 2026-04-27
[Platform][OpenAi] Route commentary phase to ThinkingDelta in streaming responses

### LINKED ISSUE #1979: [OpenAI Bridge] commentary phase is treated as TextDelta, causing duplicate assistant output in streaming responses
### Describe Your Problem

When using the OpenAI Responses streaming API via the Symfony AI OpenAI bridge, assistant output can be duplicated.

This happens because messages with `phase: "commentary"` are currently handled the same way as `phase: "final_answer"`. As a result, both are emitted as `TextDelta`, which leads to duplicated visible output.

### Expected behavior

- Only `phase: "final_answer"` should be emitted as `TextDelta`
- `phase: "commentary"` should either:
  - be ignored, or
  - be mapped to `ThinkingDelta` (similar to `reasoning_summary_text`)

### Actual behavior

Both `commentary` and `final_answer` phases are treated as regular output text and emitted as `TextDelta`, causing duplicated responses in the final output.

---

## Steps to reproduce

1. Use the OpenAI Responses API with streaming enabled
2. Trigger a response where the model emits both:
   - `phase: "commentary"`
   - `phase: "final_answer"`
3. Observe that the same content is appended twice in the final output

---

## Technical details

In `ResultConverter::convertStream()`:

```php
if (str_contains($type, 'output_text') && isset($event['delta'])) {
    yield new TextDelta($event['delta']);
}
```

This logic does not differentiate between phases.

However, OpenAI emits `response.output_text.delta` for both:
- `commentary`
- `final_answer`

The distinction is only available via:

```json
event.item_id → response.output_item → phase
```

---

## Suggested solution

Track `phase` per `item_id` using `response.output_item.added` events:

```php
$itemPhases[$event['item']['id']] = $event['item']['phase'] ?? null;
```

Then update handling of `response.output_text.delta`:

```php
$phase = $itemPhases[$event['item_id']] ?? null;

if ($phase === 'final_answer') {
    yield new TextDelta($event['delta']);
}

if ($phase === 'commentary') {
    // optionally map to ThinkingDelta or ignore
}
```

Also consider replacing:

```php
str_contains($type, 'output_text')
```

with a stricter check:

```php
$type === 'response.output_text.delta'
```

---

## References

- OpenAI Responses API (assistant message `phase` field):  
  https://platform.openai.com/docs/api-reference/responses

---

## Notes

- `reasoning_summary_text.*` is already correctly handled by the bridge
- `commentary` is different: it is emitted as a normal message with a `phase` attribute
- This issue only affects streaming responses

---

Amoifr:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #1979
| License       | MIT

The OpenAI Responses streaming API emits both `commentary` and `final_answer` as `response.output_text.delta` events. The current `str_contains($type, 'output_text')` guard in `Gpt\ResultConverter::convertStream()` treats them identically, so the visible assistant text gets duplicated whenever the model emits a commentary that mirrors the final answer (the bug reported by @wimwinterberg).

Phases are now tracked per `item_id` via `response.output_item.added`, and `response.output_text.delta` is routed accordingly:

- `commentary` → `ThinkingStart` / `ThinkingDelta`, finalized by `ThinkingComplete` on `response.output_item.done`
- `final_answer` / unknown phase / missing `item_id` → `TextDelta` (backward compatible)

The `response.reasoning_summary_text.*` path now uses its own buffer (`$currentReasoningSummary`) to avoid colliding with the commentary buffer.

The fix follows the approach @wimwinterberg sketched in the issue thread; this PR formalizes it and adds an exhaustive test suite (9 tests) covering:

- commentary routed to thinking deltas, never to TextDelta
- final_answer still emits TextDelta as before
- delta with no `item_id` keeps its TextDelta (backward compat)
- multi-delta accumulation into a single ThinkingComplete
- empty commentary item emits nothing
- reasoning_summary regression guard
- the real-world case (commentary + final_answer carrying the same content)
- truly interleaved streams (both items open, deltas alternating by `item_id`)
- aborted stream (`response.output_item.done` never arrives)

Thanks @OskarStark for the nudge on the issue.

### DIFF
--- a/src/platform/src/Bridge/OpenAi/Gpt/ResultConverter.php
+++ b/src/platform/src/Bridge/OpenAi/Gpt/ResultConverter.php
@@ -210,6 +210,14 @@ final class ResultConverter implements ResultConverterInterface
                 continue;
             }
 
+            // Backward-compatible fallback for legacy event types that carry "output_text" in their name
+            // (e.g. "message.delta.output_text.delta") and don't go through the Responses API phase routing.
+            if (str_contains($type, 'output_text') && isset($event['delta'])) {
+                yield new TextDelta($event['delta']);
+
+                continue;
+            }
+
             if ('response.output_item.done' === $type && isset($event['item']['id'])) {
                 $itemId = $event['item']['id'];
                 $phase = $itemPhases[$itemId] ?? ($event['item']['phase'] ?? null);
--- a/src/platform/src/Bridge/OpenAi/Gpt/ResultConverter.php
+++ b/src/platform/src/Bridge/OpenAi/Gpt/ResultConverter.php
@@ -146,7 +146,9 @@ final class ResultConverter implements ResultConverterInterface
 
     private function convertStream(RawResultInterface|RawHttpResult $result): \Generator
     {
-        $currentThinking = null;
+        $itemPhases = [];
+        $currentReasoningSummary = null;
+        $currentCommentary = null;
 
         foreach ($result->getDataStream() as $event) {
             $type = $event['type'] ?? '';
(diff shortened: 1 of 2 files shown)


chr-hertel:
Hi @Amoifr, thanks for acting on that issue - could you please rebase and have a look at the pipeline - looks related. Thanks already!

Amoifr:
Done — rebased on `main`. The previous CI failures were from 17 days ago, let me see what the fresh run reports and I'll dig in if anything is still broken. Thanks for the ping! 🙏

Amoifr:
@chr-hertel CI is all green now (422 jobs passing). Tracked down to two things after the rebase:

- the 3 OpenAi unit jobs (PHP 8.2 / 8.2-lowest / 8.5) failed because the new routing only matched the exact `response.output_text.delta` type, while the existing `testStreamTransmitsUsageToResultMetadata` test (and the older Chat Completions stream format) emits `message.delta.output_text.delta` — restored a backward-compatible fallback for any `output_text`-bearing event that didn't match the new exact check;
- Fabbot wanted `@return` before `@throws` in a docblock, one-line reorder.

Ready for another look whenever you have a moment. Thanks!

chr-hertel:
this change isn't that straight forward, and i would not only rely on a coding agent here - how did you verify your work is correct?

Amoifr:
Fair question, thanks for pushing on it. The verification is layered:

1. The event sequences in the tests are not invented: they mirror what @wimwinterberg captured from real streams in #1979 (`response.output_item.added` carrying `item.phase`, `response.output_text.delta` correlated by `item_id`), which also matches the documented Responses streaming events.
2. The 9 tests drive those sequences through `convertStream()` end to end: the exact duplication scenario from the issue (commentary mirroring final_answer), interleaved items, a delta without `item_id` staying on the legacy `TextDelta` path, a reasoning-summary regression guard, and an aborted stream.
3. What unit tests cannot prove is live API fidelity, so the strongest sign-off would be @wimwinterberg running his real reproduction against this branch: this PR formalizes the fix he sketched in the issue.

Happy to add any scenario you feel is missing.

chr-hertel:
Closed in favor of #2578 - thanks anyhow!

PR #2008, opened 2026-04-28
[Platform][VertexAi] Yield TokenUsage from Gemini stream chunks

### LINKED ISSUE #972: [Platform] Vertex stream generator error
When using the vertex ai platform to stream responses, I always get an 
`Generator passed to yield from was aborted without proper return and is unable to continue` error in `Symfony\AI\Platform\Result\StreamResult`. 

I tried to debug it, but I can't figure out the problem. As far as I can tell, the `src/Bridge/VertexAi/Gemini/ResultConverter.php` is working correctly and when I inspect each data-chunk in `convertStream` everything looks fine until the llm generation is done. After that, the error is thrown.

Amoifr:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #972
| License       | MIT

The streaming token extraction was disabled in #1267 because the previous implementation re-iterated the response generator, which conflicted with the agent already consuming it (the `Generator passed to yield from was aborted` error reported by @franzwilding in #972). The workaround stopped the crash but silently dropped token usage on every streaming Gemini call.

This PR aligns Gemini with the OpenAi (`Gpt\\ResultConverter::convertStream`) and Ollama (`OllamaResultConverter::convertStream`) pattern: yield `TokenUsage` directly from the converter's stream whenever a chunk carries `usageMetadata`, so the value is delivered through the same generator that produces the deltas — no second iteration needed.

`TokenUsageExtractor::extractUsageMetadata()` is renamed to `fromUsageMetadata()` and made public so the converter can reuse it, mirroring `OpenAi\\Gpt\\TokenUsageExtractor::fromDataArray()`. The non-stream `extract()` path is unchanged.

Test coverage:
- new `testStreamingYieldsTokenUsageWhenUsageMetadataIsPresent` asserts that a stream emitting `usageMetadata` mid-flight produces a `TokenUsageInterface` delta interleaved with the regular `TextDelta`s, with the right `prompt`/`completion`/`thinking`/`total` token counts
- existing stream and non-stream tests still pass (504/597 ✓, 1 unrelated skipped)

`TokenUsageExtractor::extract()` continues to return `null` for streams, since the new path now delivers the usage through the converter — the extractor is only invoked on non-stream calls or by external code that may still want to attempt stream extraction without going through the converter.

### DIFF
--- a/examples/vertexai/stream-token-metadata.php
+++ b/examples/vertexai/stream-token-metadata.php
@@ -1,0 +1,41 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use Symfony\AI\Agent\Agent;
+use Symfony\AI\Platform\Bridge\VertexAi\Factory;
+use Symfony\AI\Platform\Message\Message;
+use Symfony\AI\Platform\Message\MessageBag;
+use Symfony\AI\Platform\Result\Stream\Delta\TextDelta;
+
+require_once __DIR__.'/bootstrap.php';
+
+$platform = Factory::createPlatform(env('GOOGLE_CLOUD_LOCATION'), env('GOOGLE_CLOUD_PROJECT'), httpClient: adc_aware_http_client());
+
+$agent = new Agent($platform, 'gemini-2.5-flash');
+$messages = new MessageBag(
+    Message::forSystem('You are an expert assistant in animal study.'),
+    Message::ofUser('What does a cat usually eat?'),
+);
+$result = $agent->call($messages, [
+    'stream' => true,
+]);
+
+foreach ($result->getContent() as $delta) {
+    if ($delta instanceof TextDelta) {
+        echo $delta;
+    }
+}
+
+echo \PHP_EOL;
+
+print_token_usage($result->getMetadata()->get('token_usage'));
+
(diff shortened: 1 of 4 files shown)


chr-hertel (changes requested):
Thanks for tackling this @Amoifr - please add a example in `examples/vertexai/stream-token-metadata.php` (see openai bridge for reference).

Amoifr:
Done in 507b863 — , mirroring the openai bridge example. Thanks for the review! 🙏

chr-hertel (changes requested):
Thanks for providing that example - while running it, I was able to catch a glitch here: it claimed **17 calls**

Image

In reality, 16 out of 17 where completely empty. We should filter them out and only yield `TokenUsage` instances if there are actually tokens tracked.

Amoifr:
Thanks for catching that! 🙏

The streaming behavior of `usageMetadata` isn't really spelled out in the Vertex AI / Gemini docs — it just happens that the API sends a `usageMetadata` field on most chunks, mostly with empty or partial values, with the actual totals landing in the final chunk. So we were yielding indiscriminately.

Pushed a fix in the spirit of your remark: only yield a `TokenUsage` when the chunk carries an actual `totalTokenCount > 0`. Added a regression test covering empty + partial intermediate chunks alongside the populated one.

Would be happy if you could re-run the example to confirm you now see exactly one usage event instead of seventeen. 🙂

chr-hertel (approved):

chr-hertel:
Thank you @Amoifr.

PR #2009, opened 2026-04-28
[Agent][Platform] Mutate the message bag in place in SystemPromptInputProcessor / MemoryInputProcessor

### LINKED ISSUE #1726: [Agent] SystemPromptInputProcessor clones message bag
Version: 0.5.0 for all components

Hey,
I am not sure whether this is a bug or itended behavior and I am not using the `SystemPrompInputProcessor` correctly.
I need to add system prompts dynamically to an agent, depending on from which tenant the request comes from. Thus I create the instructions dynamically and feed them into the `SystemPrompInputProcessor` when creating the agent like this:
```php
return new Agent(
            platform: $this->platform,
            model: $this->modelName,
            inputProcessors: [
                $this->agentProcessor,  // Register tools with platform
                new SystemPromptInputProcessor($instruction, $this->toolbox),  // Add custom prompt + tool descriptions
            ],
            outputProcessors: [
                $this->agentProcessor,  // Execute tool calls
            ],
            name: 'tenant_agent',
        );
```

What I expected was that the `SystemPromptInputProcessor` adds the system prompt to the message bag if there is none present when processing the input (keeping the same message bag) and passing this on. 
However, due to the fact that the message bag gets cloned by the call `$input->setMessageBag($messages->withSystemMessage(Message::forSystem($message)));` in line 88 in the `SystemPrompInputProcessor` (this eventually calls `withoutSystemMessage` in  the `MessageBag`), the message bag object that I passed to `agent->call()` is not being modified by the other processors. In particular the AgentProcessor adds the toolcall messages into the cloned message bag object, making it impossible for me to retrieve these messages.

Expected: `SystemPromptInputProcessor` processes the message bag and passes it on, without cloning
Actual: `SystemPromptInputProcessor` clones the message bag and passes on the cloned bag 

As I said, maybe this is itended behavior and I am not using the SystemInputProcessor correctly. I would appreciate some feedback on the issue and a recommendation on how to do this otherwise :)

Thanks!

Amoifr:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #1726
| License       | MIT

Reported by @jo-hoff in #1726: a caller passing their own `MessageBag` to `Agent::call($bag)` could not see tool-call messages later appended by output processors, because `SystemPromptInputProcessor` (and the closely related `MemoryInputProcessor`) replaced the bag in the `Input` with a fresh clone via `MessageBag::withSystemMessage()`. The agent then operated on the clone, while the caller was left holding a stale reference.

This PR adds two mutable counterparts on `MessageBag`:
- `prepend(MessageInterface)` — insert the message at index 0
- `removeSystemMessage()` — drop the system message in place

Both processors now mutate the caller's bag instead of replacing it, so downstream output processors (e.g. tool-call appending) and the caller share the same instance.

The immutable `with*()` / `without*()` API is unchanged and still used by the bridge normalizers (no BC break).

Test coverage:
- new `MessageBag::prepend()` / `removeSystemMessage()` unit tests
- new `testProcessInputMutatesTheCallerMessageBagInPlace` on `SystemPromptInputProcessor`
- new `testItMutatesTheCallerMessageBagInPlace` on `MemoryInputProcessor`
- the existing test suite still passes (the `assertSame($bag, $input->getMessageBag())` assertion was previously satisfied by accident only when the processor early-returned)

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,6 +1,18 @@
 UPGRADE FROM 0.9 to 0.10
 ========================
 
+Agent
+-----
+
+ * `SystemPromptInputProcessor` and `MemoryInputProcessor` now modify the message bag in place
+   instead of replacing it with a clone, so injected system messages and messages appended during
+   the agent run (e.g. tool calls) end up in the caller's bag. To keep your bag untouched:
+
+   ```diff
+   -$agent->call($messages);
+   +$agent->call(clone $messages);
+   ```
+
 Platform
 --------
 
--- a/src/agent/src/InputProcessor/SystemPromptInputProcessor.php
+++ b/src/agent/src/InputProcessor/SystemPromptInputProcessor.php
@@ -85,6 +85,6 @@ final class SystemPromptInputProcessor implements InputProcessorInterface
                 PROMPT;
         }
 
-        $input->setMessageBag($messages->withSystemMessage(Message::forSystem($message)));
+        $messages->prepend(Message::forSystem($message));
     }
 }
--- a/src/agent/src/InputProcessor/SystemPromptInputProcessor.php
+++ b/src/agent/src/InputProcessor/SystemPromptInputProcessor.php
@@ -90,6 +90,6 @@ final class SystemPromptInputProcessor implements InputProcessorInterface
                 PROMPT;
         }
 
-        $input->setMessageBag($messages->withSystemMessage(Message::forSystem($message)));
+        $messages->prepend(Message::forSystem($message));
     }
 }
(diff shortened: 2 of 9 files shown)


chr-hertel:
Change itself looks good on first glance, will wait for feedback on #1726 before moving on - thanks already for taking care @Amoifr!

OskarStark:
Feedback was positive in
* #1726

chr-hertel (approved):

chr-hertel:
Thank you @Amoifr.

PR #2010, opened 2026-04-28
[Platform][Cache] Add CachePlatform::lookup() and content-based MessageBag hashing

### LINKED ISSUE #1245: [Platform] Add look-up functionality to CachedPlatform
**Problem description**
The `CachedPlatform` class (`vendor/symfony/ai-platform/src/CachedPlatform.php`) allows you to configure caching on prompt-level, with consideration for model. 

While this works absolutely great, I encounter one pain point: it is impossible to do a cache look up without triggering the LLM to generate the answer if there is no cache hit. 

It would be nice (mainly for UI reasons) to have the option to just check whether the cache will hit. An example is my current use case:

- If the cache hits, we want to show the cache hit results immediately.
- If the cache does not hit, we want to display a button which asks them whether they want to regenerate.

**Proposed solutions**
There are many ways to approach this. As `CachedPlatform` is a decorator around `PlatformInterface` implementations, we could create a trait (e.g. `HasSearchableCache`) which exposes an additional function. Or maybe even easier/cleaner: we could introduce a new option to the `$options` array, which is checked inside `PlatformInterface::invoke()`, deciding on whether we should only check cache or also regenerate. 

**TL;DR: it would be great to have a cache look up functionality for the `CachedPlatform` decorator, because checking whether something is cached does not automatically imply that we want to generate a new answer if it is not.**

Amoifr:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| Docs?         | no
| Issues        | Fix #1245
| License       | MIT

Tackles the two pain points raised in #1245:

1. **Lookup-only semantics**, requested by @marco-jouwweb. Adds `CachePlatform::lookup(string $model, $input, $options): ?DeferredResult`, which returns the cached result if any or `null` on a miss, **without ever invoking the underlying platform**. This is the UI-friendly path the issue describes ("show cached answer immediately, or surface a regenerate button").

2. **Cache actually caching**, the side observation by @chr-hertel ("the implementation currently has also some functionality issues"). The cache key for `MessageBag` inputs was derived from the bag's per-instance UUID (`$input->getId()->toString()`), so two `MessageBag` instances carrying the same conversation produced two different cache entries. The key is now derived from the conversation content (role + `getContent()` per message), which is what users expect when they pass a stable `prompt_cache_key`. Per-message UUIDs and metadata are intentionally excluded from the hash for the same reason.

`CachePlatform` is also refactored a bit to share the cache-key construction and the deferred-result rebuild between `invoke()` and `lookup()`.

The previous test that documented the per-instance-UUID behaviour as intentional (`testPlatformCanReturnCachedResultWhenCalledTwiceWithSeparateMessageBag`) is split into two:
- `testPlatformReturnsCachedResultForSeparateMessageBagsWithSameContent` — same conversation, single underlying invocation, cache hit
- `testPlatformDistinguishesMessageBagsWithDifferentContent` — different conversation, two underlying invocations

Plus three new tests for `lookup()`: miss, hit-without-invoke, missing/empty cache key.

### DIFF
--- a/src/platform/src/Bridge/Cache/CHANGELOG.md
+++ b/src/platform/src/Bridge/Cache/CHANGELOG.md
@@ -1,7 +1,7 @@
 CHANGELOG
 =========
 
-0.4
+0.9
 ---
 
  * Hash `MessageBag` inputs by content rather than by per-instance UUID, so two bags carrying the same conversation hit the same cache entry
--- a/src/platform/src/Bridge/Cache/CHANGELOG.md
+++ b/src/platform/src/Bridge/Cache/CHANGELOG.md
@@ -1,6 +1,12 @@
 CHANGELOG
 =========
 
+0.4
+---
+
+ * Hash `MessageBag` inputs by content rather than by per-instance UUID, so two bags carrying the same conversation hit the same cache entry
+ * Add `CachePlatform::lookup()` to retrieve a cached result without invoking the underlying platform on a miss
+
 0.3
 ---
 
--- a/src/platform/src/Bridge/Cache/CachePlatform.php
+++ b/src/platform/src/Bridge/Cache/CachePlatform.php
@@ -12,7 +12,10 @@
 namespace Symfony\AI\Platform\Bridge\Cache;
 
 use Symfony\AI\Platform\Exception\InvalidArgumentException;
+use Symfony\AI\Platform\Message\Content\ContentInterface;
 use Symfony\AI\Platform\Message\MessageBag;
+use Symfony\AI\Platform\Message\MessageInterface;
+use Symfony\AI\Platform\Message\Template;
 use Symfony\AI\Platform\ModelCatalog\ModelCatalogInterface;
 use Symfony\AI\Platform\PlainConverter;
 use Symfony\AI\Platform\PlatformInterface;
@@ -64,19 +67,7 @@ final class CachePlatform implements PlatformInterface
             return $this->platform->invoke($model, $input, $options);
         }
 
-        $normalizedInput = match (true) {
-            \is_string($input) => md5($input),
-            \is_array($input) => json_encode($input),
-            $input instanceof MessageBag => $input->getId()->toString(),
-            default => throw new InvalidArgumentException(\sprintf('Unsupported input type: %s', get_debug_type($input))),
-        };
-
-        $cacheKey = (new UnicodeString())->join([
-            $options['prompt_cache_key'] ?? $this->cacheKey,
-            (new UnicodeString($model))->camel(),
-            $normalizedInput,
-        ]);
-
+        $cacheKey = $this->buildCacheKey($model, $input, $options);
         $ttl = $options['prompt_cache_ttl'] ?? $this->cacheTtl;
 
         unset($options['prompt_cache_key'], $options['prompt_cache_ttl']);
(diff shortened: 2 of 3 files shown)


src/platform/src/Bridge/Cache/CachePlatform.php
```diff
@@ -101,6 +92,111 @@ public function invoke(string $model, array|string|object $input, array $options
...
+     *
+     * Useful for UI flows where a cache miss should not trigger a costly
+     * model invocation — e.g. show the cached answer if any, otherwise let
+     * the user decide to (re)generate one.
+     *
+     * @param array<string, mixed> $options Must contain a non-empty
+     *                                      `prompt_cache_key`, like
+     *                                      {@see self::invoke()} does.
+     */
+    public function lookup(string $model, array|string|object $input, array $options = []): ?DeferredResult
```
Guikingone, line 125: The main issue of declaring a new method is that it's not exposed through the shared API for platforms 🤔

src/platform/src/Bridge/Cache/CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 CHANGELOG
 =========
 
+0.4
```
Guikingone: 0.9

src/platform/src/Bridge/Cache/CachePlatform.php
```diff
@@ -101,6 +92,111 @@ public function invoke(string $model, array|string|object $input, array $options
...
+        // per-instance state, not part of the conversation identity.
+        $payload = array_map(
+            fn (MessageInterface $message): array => [
+                'role' => $message->getRole()->value,
+                'content' => $this->normalizeContentForHash($message->getContent()),
+            ],
+            $bag->getMessages(),
+        );
+
+        return md5(json_encode($payload, \JSON_THROW_ON_ERROR));
```
Guikingone: Could we define this behavior in the `MessageNormalizer`?

src/platform/src/Bridge/Cache/CachePlatform.php
Amoifr: Fair point — `lookup()` only being reachable on the concrete `CachePlatform` is a real limitation. Three ways forward I can see:

1. Add `lookup()` to `PlatformInterface` and have non-cached platforms return `null`/throw — but that pollutes the contract for every bridge.
2. Make `invoke()` accept a `cache_lookup_only` option that returns a sentinel `DeferredResult` carrying a "miss" flag — keeps the interface but makes the return type ambiguous.
3. Introduce a small `CachePlatformInterface` (or `LookupCapableInterface`) that consumers can type-hint when they specifically need the lookup behaviour.

My instinct leans toward **3** since lookup-only semantics genuinely don't apply to non-cached platforms, but I'd rather hear your preference before refactoring. WDYT?

src/platform/src/Bridge/Cache/CHANGELOG.md
Amoifr: Done in 52ba40cf — bumped to `0.9`. Thanks!

src/platform/src/Bridge/Cache/CachePlatform.php
Amoifr: Great suggestion 🙏 — done in 52ba40cf. The `MessageBagNormalizer` and friends already exclude the per-instance UUIDs from their output, so I just registered them in the `CachePlatform`'s serializer and let `$serializer->serialize($bag, 'json')` do the work. The bespoke `hashMessageBag()` + `normalizeContentForHash()` pair is gone, and the existing tests still cover the "same conversation = same cache key" / "different conversation = different key" cases.

src/platform/src/Bridge/Cache/CachePlatform.php
Guikingone: I'd use an option (`cache_lookup_only` could be one to explore, don't have a strong opinion on it) rather than having a public method that most users would not use directly.

The main "issue" with a dedicated interface is the fact that apart from `CachePlatform`, I don't see any future usage of it 🤔

Amoifr: Got it — going with an option-based approach. Happy with `cache_lookup_only` as the name (or `cache_only` / `lookup_only`, no strong opinion either).

One open design point worth confirming before I refactor, since `PlatformInterface::invoke()` returns a non-nullable `DeferredResult`: how do we signal a cache miss?

- **A.** Return a `DeferredResult` wrapping a sentinel/empty result with `cached => false` in metadata. Symmetrical with the existing `cached: true` flag set on hits, callers branch on `$result->getMetadata()->get('cached')`. Downside: we build a "ghost" `DeferredResult` purely to signal a miss.
- **B.** Throw a typed `CacheMissException`. Cleaner type-wise, but loses the no-throw ergonomics that motivated the lookup helper in the first place (UI flow has to wrap every call in try/catch).

Leaning **A** for the ergonomics — fair? 🙂

Re: the interface — totally fair, dropping the `CachePlatformInterface` idea, you're right that it would have no other implementation in sight.

PR #2014, opened 2026-04-28
[Platform][Codex] Drop no-op --ask-for-approval flag from codex exec

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | -
| License       | MIT

The top-level `--ask-for-approval` flag is silently ignored by `codex exec` (verified against codex CLI 0.125.0): regardless of the value passed, the spawned session ends up with `approval_policy = "never"`. Hard-coding it on the command line — and accepting it as an option — gave callers a false sense that the value took effect.

This PR drops the flag from `ModelClient::buildCommand` and strips any caller-supplied `ask_for_approval` from the options array, so the bridge no longer pretends to honor a value the CLI ignores. The supported escape hatch when an exec session needs to actually run MCP tools is the existing `dangerously_bypass_approvals_and_sandbox` option, which maps to `--dangerously-bypass-approvals-and-sandbox` and is verified by an added test.

The existing `buildCommand` tests are updated to expect the trimmed argv, and two new tests cover (a) `ask_for_approval` being dropped silently and (b) `dangerously_bypass_approvals_and_sandbox` being forwarded.

### DIFF
--- a/src/platform/src/Bridge/Codex/ModelClient.php
+++ b/src/platform/src/Bridge/Codex/ModelClient.php
@@ -99,7 +99,12 @@ final class ModelClient implements ModelClientInterface
      */
     public function buildCommand(string $prompt, array $options = []): array
     {
-        $command = [$this->getCliBinary(), '--ask-for-approval', 'never', 'exec', '--json'];
+        if (isset($options['ask_for_approval'])) {
+            @trigger_error('The "ask_for_approval" option is ignored by "codex exec" and has no effect; remove it or use "dangerously_bypass_approvals_and_sandbox" to bypass approvals.', \E_USER_DEPRECATED);
+            unset($options['ask_for_approval']);
+        }
+
+        $command = [$this->getCliBinary(), 'exec', '--json'];
 
         foreach ($options as $key => $value) {
             $flag = self::OPTION_FLAG_MAP[$key] ?? '--'.str_replace('_', '-', $key);
--- a/src/platform/src/Bridge/Codex/Tests/ModelClientTest.php
+++ b/src/platform/src/Bridge/Codex/Tests/ModelClientTest.php
@@ -11,6 +11,7 @@
 
 namespace Symfony\AI\Platform\Bridge\Codex\Tests;
 
+use PHPUnit\Framework\Attributes\IgnoreDeprecations;
 use PHPUnit\Framework\TestCase;
 use Symfony\AI\Platform\Bridge\Codex\Codex;
 use Symfony\AI\Platform\Bridge\Codex\Exception\CliNotFoundException;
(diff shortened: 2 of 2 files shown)


src/platform/src/Bridge/Codex/CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ CHANGELOG
 ---
 
  * [BC BREAK] Align `ModelCatalog` with the official OpenAI Codex models list (https://developers.openai.com/codex/models): add `gpt-5.2` and remove `gpt-5.2-codex`, `gpt-5.1-codex`, `gpt-5-codex`, `gpt-5-codex-mini`
+ * Fix: drop the no-op `--ask-for-approval` flag from `codex exec`; the CLI silently ignores it, use the `dangerously_bypass_approvals_and_sandbox` option when an exec session needs to bypass approvals
```
chr-hertel:
```suggestion
 * Drop ignored `--ask-for-approval` flag, use the `dangerously_bypass_approvals_and_sandbox` option explicitly to bypass approvals
```

chr-hertel: i wonder tho if too internal?

src/platform/src/Bridge/Codex/ModelClient.php
```diff
@@ -99,7 +99,15 @@ public function request(Model $model, array|string $payload, array $options = []
...
     public function buildCommand(string $prompt, array $options = []): array
     {
-        $command = [$this->getCliBinary(), '--ask-for-approval', 'never', 'exec', '--json'];
+        // The top-level `--ask-for-approval` flag is silently ignored by
+        // `codex exec` (verified against codex CLI 0.125.0): regardless of the
+        // value the spawned session ends up with `approval_policy = "never"`.
+        // Drop it so callers don't get a false sense that it took effect; use
+        // `dangerously_bypass_approvals_and_sandbox` instead when MCP tools
+        // need to actually run inside an exec session.
+        unset($options['ask_for_approval']);
```
chr-hertel: do we really need to unset - just not setting it implicitly might be already enough?

wachterjohannes: you are right :) but to be explicit it make sense to unset it! but to make it a bit more "seeable" i have change it to trigger a warning

src/platform/src/Bridge/Codex/CHANGELOG.md
wachterjohannes: i will remove it

chr-hertel (approved):

chr-hertel:
Thank you @wachterjohannes.

PR #2016, opened 2026-04-29
[Platform][Bedrock] Fix Anthropic Claude model ID mapping

karpilin:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | no
| License       | MIT

Bedrock model identifiers for Anthropic Claude models follow inconsistent naming conventions - some require version suffixes like `-v1:0` or `-v1`, while others (such as the new Claude Opus 4.7) require none.

This change introduces a model map to correctly resolve Bedrock IDs based on the Anthropic model name, allows for constructor overrides, and adds support for `claude-opus-4-7`, `claude-sonnet-4-6` and `claude-opus-4-6`.

### DIFF
--- a/src/platform/src/Bridge/Bedrock/Anthropic/ClaudeModelClient.php
+++ b/src/platform/src/Bridge/Bedrock/Anthropic/ClaudeModelClient.php
@@ -24,10 +24,42 @@ use Symfony\AI\Platform\ModelClientInterface;
  */
 final class ClaudeModelClient implements ModelClientInterface
 {
+    /**
+     * Bedrock model identifiers differ from Anthropic API names — some require version suffixes,
+     * others don't. See https://platform.claude.com/docs/en/about-claude/models/overview for details.
+     *
+     * @var array<string, string>
+     */
+    private const MODEL_MAP = [
+        'claude-opus-4-7' => 'claude-opus-4-7',
+        'claude-sonnet-4-6' => 'claude-sonnet-4-6',
+        'claude-opus-4-6' => 'claude-opus-4-6-v1',
+        'claude-haiku-4-5-20251001' => 'claude-haiku-4-5-20251001-v1:0',
+        'claude-sonnet-4-5-20250929' => 'claude-sonnet-4-5-20250929-v1:0',
+        'claude-opus-4-5-20251101' => 'claude-opus-4-5-20251101-v1:0',
+        'claude-opus-4-1-20250805' => 'claude-opus-4-1-20250805-v1:0',
+        'claude-sonnet-4-20250514' => 'claude-sonnet-4-20250514-v1:0',
+        'claude-opus-4-20250514' => 'claude-opus-4-20250514-v1:0',
+        'claude-3-sonnet-20240229' => 'claude-3-sonnet-20240229-v1:0',
+        'claude-3-haiku-20240307' => 'claude-3-haiku-20240307-v1:0',
+        'claude-3-5-haiku-20241022' => 'claude-3-5-haiku-20241022-v1:0',
+    ];
+
+    /**
+     * @var array<string, string>
+     */
+    private readonly array $modelMap;
+
+    /**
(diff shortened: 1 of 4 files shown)


src/platform/src/Bridge/Bedrock/Anthropic/ClaudeModelClient.php
```diff
@@ -24,10 +24,42 @@
...
+        'claude-opus-4-7' => 'claude-opus-4-7',
+        'claude-sonnet-4-6' => 'claude-sonnet-4-6',
+        'claude-opus-4-6' => 'claude-opus-4-6-v1',
+        'claude-haiku-4-5-20251001' => 'claude-haiku-4-5-20251001-v1:0',
+        'claude-sonnet-4-5-20250929' => 'claude-sonnet-4-5-20250929-v1:0',
+        'claude-opus-4-5-20251101' => 'claude-opus-4-5-20251101-v1:0',
+        'claude-opus-4-1-20250805' => 'claude-opus-4-1-20250805-v1:0',
+        'claude-sonnet-4-20250514' => 'claude-sonnet-4-20250514-v1:0',
+        'claude-opus-4-20250514' => 'claude-opus-4-20250514-v1:0',
+        'claude-3-sonnet-20240229' => 'claude-3-sonnet-20240229-v1:0',
```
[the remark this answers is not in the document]
karpilin: that model was deprecated and no longer available on Bedrock

src/platform/src/Bridge/Bedrock/Anthropic/ClaudeModelClient.php
```diff
@@ -24,10 +24,42 @@
...
 {
+    /**
+     * Bedrock model identifiers differ from Anthropic API names — some require version suffixes,
+     * others don't. See https://platform.claude.com/docs/en/about-claude/models/overview for details.
+     *
+     * @var array<string, string>
+     */
+    private const MODEL_MAP = [
+        'claude-opus-4-7' => 'claude-opus-4-7',
+        'claude-sonnet-4-6' => 'claude-sonnet-4-6',
```
karpilin, line 35: we can exclude these two from the mapping, as they are identical for both Anthropic API and Bedrock

OskarStark: Not sure we. want to solve this with a map, why not check dynamically for the -v1:0 suffix?

karpilin: If I understood correctly what @chr-hertel said on Slack - we want the same model name to be used interchangeably between different provider bridges, which means we should allow the canonical Anthropic model name to be passed in.

However, there is no clear pattern to match Anthropic names to Bedrock model identifiers, and there is no guarantee this won't change again.

An alternative is to use regular expressions to detect and append `-v1:0` to older model names, but `claude-opus-4-6` has a `-v1` suffix which breaks the pattern.

```
  private function resolveBedrockModelId(string $name): string                                                                                                                    
  {               
      if (preg_match('/-\d{8}$/', $name)) {
          return $name.'-v1:0';                                                                                                                                                   
      }
      if ('claude-opus-4-6' === $name) {                                                                                                                                          
          return $name.'-v1';                                                                                                                                                     
      }
      return $name;                                                                                                                                                               
  }  
```
  
The model map is explicit, matches official Claude documentation, and allows for simple overrides via the client constructor.

chr-hertel: looking at the examples and the irregularity, i think the map is straightforward. i get why it's not the most elegant, but it's pragmatic - and i favor that over delegating that to the user :thinking:

chr-hertel (approved):

OskarStark (approved):

chr-hertel:
Thank you @karpilin.

PR #2020, opened 2026-05-01
[Agent] Fix denormalization of float properties

### LINKED ISSUE #2011: [Agent] Should ToolCallArgumentResolver pass 'json' format to denormalize()?
## Question

`ToolCallArgumentResolver::resolveArguments()` calls `$denormalizer->denormalize($value, $type)` without a third `$format` argument. Is the omission of `'json'` as the format intentional?

The Symfony Serializer ships JSON-specific value coercions in `AbstractObjectNormalizer` that only fire when `$format` contains the JSON encoder format. The most visible effect today is that JSON-number values without a fractional part can't be assigned to `float`-typed properties on parameter classes.

## Background — the JSON-specific cast in `AbstractObjectNormalizer`

https://github.com/symfony/serializer/blob/v7.4.0/Normalizer/AbstractObjectNormalizer.php#L587-L595

```php
// JSON only has a Number type corresponding to both int and float PHP types.
// PHP's json_encode, JavaScript's JSON.stringify, Go's json.Marshal as well as most other JSON encoders convert
// floating-point numbers like 12.0 to 12 (the decimal part is dropped when possible).
// PHP's json_decode automatically converts Numbers without a decimal part to integers.
// To circumvent this behavior, integers are converted to floats when denormalizing JSON based formats and when
// a float is expected.
if (LegacyType::BUILTIN_TYPE_FLOAT === $builtinType && \is_int($data) && null !== $format && str_contains($format, JsonEncoder::FORMAT)) {
    return (float) $data;
}
```

The author explicitly accounted for the JSON int↔float ambiguity, but this coercion only runs when `$format` is JSON.

## The `denormalize()` call site in `ToolCallArgumentResolver`

https://github.com/symfony/ai/blob/v0.8.0/src/agent/src/Toolbox/ToolCallArgumentResolver.php#L93-L95

```php
if ($this->denormalizer->supportsDenormalization($value, $parameterType)) {
    $value = $this->denormalizer->denormalize($value, $parameterType);
}
```

No `$format` argument is supplied; the default is `null`. So the JSON int→float coercion above never runs for tool call arguments.

## Observed behavior

Given a tool method that takes a class with a `float` property, an LLM tool call like `{"foo": {"amount": 100}}` (a JSON number without a fractional part — `json_decode` returns PHP `int`) fails with:

```
NotNormalizableValueException:
  The type of the "amount" attribute for class "Foo" must be one of "float" ("int" given).
```

If the resolver passed `'json'` to `denormalize()`, `AbstractObjectNormalizer` would widen the int to a float and the parameter would denormalize successfully.

## Minimal reproducer

```php
<?php

final class Foo
{
    public function __construct(public float $amount)
    {
    }
}

final class SomeTool
{
    public function __invoke(Foo $foo): array
    {
        return [];
    }
}

// LLM tool call payload: {"foo": {"amount": 100}}
// json_decode produces: ['foo' => ['amount' => 100]]   (int, not float)
$toolCall = new ToolCall('tc-1', 'some_tool', ['foo' => ['amount' => 100]]);

// NotNormalizableValueException:
//   The type of the "amount" attribute for class "Foo" must be one of "float" ("int" given).
```

## Possible direction

Pass `'json'` as the third argument to `supportsDenormalization()` and `denormalize()` in `resolveArguments()`.

If the omission is intentional, what's the recommended way to handle the JSON int → float case for tool method parameters?

## Environment

- `symfony/ai-agent`: **v0.8.0**
- `symfony/ai-platform`: v0.8.1
- `symfony/ai-bundle`: v0.8.0
- Symfony: 7.4
- PHP: 8.4

## Related

- #2004 — sits in the same `ToolCallArgumentResolver::resolveArguments()` method, different root cause (`NullableType` stringification rather than the missing format).

valtzu:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | Fix #2011
| License       | MIT

Pass `'json'` as `format` when denormalizing the arguments

### DIFF
--- a/src/agent/src/Toolbox/ToolCallArgumentResolver.php
+++ b/src/agent/src/Toolbox/ToolCallArgumentResolver.php
@@ -90,8 +90,8 @@ final class ToolCallArgumentResolver implements ToolCallArgumentResolverInterfac
 
             $parameterType .= $dimensions;
 
-            if ($this->denormalizer->supportsDenormalization($value, $parameterType)) {
-                $value = $this->denormalizer->denormalize($value, $parameterType);
+            if ($this->denormalizer->supportsDenormalization($value, $parameterType, 'json')) {
+                $value = $this->denormalizer->denormalize($value, $parameterType, 'json');
             }
 
             $arguments[$name] = $value;
--- a/src/agent/tests/Fixtures/Tool/ToolObjectFloat.php
+++ b/src/agent/tests/Fixtures/Tool/ToolObjectFloat.php
@@ -1,0 +1,25 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Agent\Tests\Fixtures\Tool;
+
+use Symfony\AI\Agent\Toolbox\Attribute\AsTool;
+
+#[AsTool('tool_object_float', 'A tool with object parameter with float property')]
+final class ToolObjectFloat
+{
+    public float $height;
+
+    public function __invoke(self $person): string
+    {
+        return \sprintf('Height: %.2fm', $person->height);
+    }
+}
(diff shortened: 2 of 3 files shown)


chr-hertel (approved):
oh, i learned a thing here. wasn't aware.

chr-hertel:
Thank you @valtzu.

PR #2021, opened 2026-05-01
[Mate] Make json-input argument optional in mcp:tools:call

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | -
| License       | MIT

Defaults the `json-input` argument of the `mcp:tools:call` command to `'{}'`, so tools that don't require parameters can be invoked without explicitly passing an empty JSON object.

Before:

```bash
vendor/bin/mate mcp:tools:call sulu-info '{}'
```

After (both still work):

```bash
vendor/bin/mate mcp:tools:call sulu-info
vendor/bin/mate mcp:tools:call sulu-info '{}'
```


chr-hertel (approved):

chr-hertel:
Thank you @wachterjohannes.

PR #2023, opened 2026-05-02
[MCP Bundle] Fix traceable registry reset

### LINKED ISSUE #2017: [MCP Bundle] tools/list can return empty in debug + FrankenPHP worker mode after successful discovery
### Context

 I’m running symfony/mcp-bundle over HTTP transport behind FrankenPHP worker mode with:

 - APP_ENV=dev
 - APP_DEBUG=1

 The MCP endpoint is reachable and authenticated. initialize succeeds, but tools/list can return an empty array
 even though MCP tools were discovered at startup.

 ────────────────────────────────────────────────────

 ### Expected behavior

 If tools are discovered (e.g. 4 tools), tools/list should return them consistently for subsequent requests in
 the same MCP session.

 ### Actual behavior

 - initialize succeeds
 - tools/list returns:

 ```json
   {"tools":[]}
 ```

 At the same time, MCP logs show discovery finding tools and then clearing discovered elements before/around
 subsequent request handling.

 ────────────────────────────────────────────────────

 ### Logs / evidence

 MCP channel logs show:

 - Attribute discovery finished. with context like tools: 4
 - followed by: Removed 4 discovered elements from internal registry.

 After that, tools/list may return empty.

 ────────────────────────────────────────────────────

 ### Reproduction (HTTP MCP)

 1. Configure MCP bundle with HTTP transport and attribute-based tools in src/ (#[McpTool(...)]).
 2. Run in FrankenPHP worker mode with debug enabled (APP_DEBUG=1).
 3. POST /_mcp initialize:

 ```json
   {
     "jsonrpc": "2.0",
     "method": "initialize",
     "params": {
       "protocolVersion": "2025-03-26",
       "capabilities": {},
       "clientInfo": { "name": "repro", "version": "1.0" }
     },
     "id": 1
   }
 ```

 4. Use returned Mcp-Session-Id and call:

 ```json
   {
     "jsonrpc": "2.0",
     "method": "tools/list",
     "id": 2
   }
 ```

 5. Observe empty tools in affected runs.

 ────────────────────────────────────────────────────

 ### Additional observations

 - Setting APP_DEBUG=0 in the same environment makes tools/list return the discovered tools correctly.
 - Behavior appears intermittent / client-flow dependent (request sequence/timing), but logs consistently show
 discovered tools being cleared in debug worker mode.

 ────────────────────────────────────────────────────

 ### Suspected cause

 Potential interaction between worker lifecycle + debug registry reset in:

 Symfony\AI\McpBundle\Profiler\TraceableRegistry

 reset() calls clear(), which seems to remove discovered MCP capabilities across requests in worker mode.

 Possibly related to:

 - https://github.com/symfony/ai/pull/1533

 ────────────────────────────────────────────────────

 ### Versions
 - symfony/mcp-bundle: v0.8.0
 - Symfony: 8.0.*
 - Runtime: FrankenPHP worker mode
 - Transport: HTTP (/_mcp)
 - Env where reproduced: APP_ENV=dev, APP_DEBUG=1

Genmin:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #2017
| License       | MIT

`TraceableRegistry::reset()` was clearing the decorated MCP capability registry. In debug worker runtimes, Symfony calls `kernel.reset` between HTTP requests after discovery has populated tools/resources/prompts, so a later `tools/list` request could see an empty registry.

This keeps explicit `TraceableRegistry::clear()` behavior unchanged, but makes `reset()` preserve discovered capabilities because the traceable wrapper does not hold per-request state.

Validation:
- `vendor/bin/phpunit` from `src/mcp-bundle` -> 43 tests, 143 assertions
- `php -l src/Profiler/TraceableRegistry.php`
- `php -l tests/Profiler/TraceableRegistryTest.php`
- `git diff --check`


Genmin:
Checked the one failing CI job. It fails during dependency installation for `Unit / Tool / Tavily / PHP 8.2 / lowest`, before any tests run, because Composer reports `phpunit/phpunit` requires `ext-xml` and the job environment does not expose that extension. This PR only touches the MCP Bundle profiler wrapper, and the MCP Bundle jobs plus PHPStan/Fabbot/deptrac are green.\n\nLocal validation from this branch remains:\n- `composer install --no-interaction --no-progress` in `src/mcp-bundle`\n- `vendor/bin/phpunit` -> 43 tests, 143 assertions\n- `php -l src/Profiler/TraceableRegistry.php`\n- `php -l tests/Profiler/TraceableRegistryTest.php`

chr-hertel:
Thanks for jumping on that @Genmin :+1: 
@ineersa I guess this is more or less the outcome you need for your use-case, right?

I'd say we could just drop the `ResetInterface` from `TraceableRegistry` instead of an empty method, WDYT?

ineersa:
> @ineersa I guess this is more or less the outcome you need for your use-case, right?

Yes! 
I think it should be safe to remove `ResetInterface`.

Genmin:
Thanks, agreed. I updated the PR to drop `ResetInterface` from `TraceableRegistry` instead of keeping an empty `reset()` method, and adjusted the regression test/changelog accordingly.

Validation from this branch:
- `vendor/bin/phpunit tests/Profiler/TraceableRegistryTest.php` -> 1 test, 1 assertion
- `vendor/bin/phpunit` in `src/mcp-bundle` -> 43 tests, 143 assertions
- `php -l src/Profiler/TraceableRegistry.php`
- `php -l tests/Profiler/TraceableRegistryTest.php`
- `git diff --check`

Genmin:
Updated again after the CI failures: removing `ResetInterface` also required removing the manual `kernel.reset` tag from `mcp.traceable_registry`, otherwise the demo container still tried to call the removed `reset()` method. I added a bundle regression test for that registration path too.

Validation from this branch:
- `vendor/bin/phpunit` in `src/mcp-bundle` -> 44 tests, 144 assertions
- `bin/console lint:container` in `demo`
- `vendor/bin/phpunit tests/SmokeTest.php` in `demo` -> 5 tests, 20 assertions
- `vendor/bin/php-cs-fixer fix --dry-run --diff --config=.php-cs-fixer.dist.php --allow-unsupported-php-version=yes -- src/mcp-bundle/src/McpBundle.php src/mcp-bundle/src/Profiler/TraceableRegistry.php src/mcp-bundle/tests/DependencyInjection/McpBundleTest.php src/mcp-bundle/tests/Profiler/TraceableRegistryTest.php`
- `php -l` on the changed PHP files
- `git diff --check`

src/mcp-bundle/CHANGELOG.md
```diff
@@ -4,7 +4,7 @@ CHANGELOG
 0.8
 ---
 
- * Preserve discovered MCP capabilities when resetting the traceable registry between worker requests
+ * Stop resetting the traceable registry between worker requests to preserve discovered MCP capabilities
```
chr-hertel: this entry needs to go to version 0.9

chr-hertel (approved):

chr-hertel:
Thank you @Genmin.

PR #2024, opened 2026-05-02
[AI Bundle][Platform] Add #[Schema(provider: ...)] with SchemaProviderInterface for runtime JSON Schema fragments

camilleislasse:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | —
| License       | MIT

### Content : 

`#[Schema(enum: [...])]` only accepts constant expressions, so allowed values that come from `.env`, a database or any service can't be expressed today. This PR adds a `provider` argument to the `#[Schema]` attribute, referencing a `SchemaProviderInterface` service whose `getSchemaFragment()` returns a JSON Schema fragment computed at runtime. The fragment is merged on top of the schema built from reflection, `#[Schema]`, PHPDoc and Validator constraints. Same mechanism for tool parameters and structured output DTO properties.

```php
final class StatusProvider implements SchemaProviderInterface
{
    public function __construct(private readonly OrderStatusService $service)
    {
    }

    public function getSchemaFragment(array $context = []): array
    {
        return ['enum' => $this->service->getAvailableStatuses()];
    }
}

#[AsTool('search_orders', 'Search orders by status')]
final class SearchOrdersTool
{
    public function __invoke(
        #[Schema(provider: StatusProvider::class)]
        string $status,
    ): array {
        // ...
    }
}
```

The AI Bundle autoconfigures implementations of `SchemaProviderInterface`, and a compiler pass validates the referenced service IDs on tagged tools at container build time. The schema-attribute describer runs last in the describer chain, so runtime fragments win on conflicts. `provider` accepts any container service ID (not only FQCNs), and an optional `context` array is forwarded to `getSchemaFragment()`. See `docs/cookbook/runtime-driven-tool-parameters.rst` for composing with static constraints, structured output usage, and standalone wiring.

### Context : 
Came out of a Slack discussion with @alpernuage, @valtzu and @chr-hertel on dynamic tool parameter enums. valtzu suggested a custom describer to extend the schema chain; this PR ships that as a reusable mechanism, covering both tool parameters (Alper's case) and structured output DTOs (chris's case).


valtzu:
Nice work @camilleislasse  :slightly_smiling_face: 

Some comments/ideas:

1. Not having support for non-fqcn services feels unintuitive  _(even though documented)_ – you may have e.g. `EnumSchemaProvider` which takes some identifier in constructor and then have one instance of the service per enum.
2. We should probably do a build-time check to ensure all the referenced schema provider services exist.
3. _(nice-to-have)_ Now a `SchemaProvider` has no context, which limits the use case a bit. We could have e.g.  `#[SchemaSource(provider: EntityEnumSchemaProvider::class, context: ['entity' => PaintColor::class])]` and pass `context` to `getSchemaFragment()`

camilleislasse:
> Nice work @camilleislasse 🙂
> 
> Some comments/ideas:
> 
> 1. Not having support for non-fqcn services feels unintuitive  _(even though documented)_ – you may have e.g. `EnumSchemaProvider` which takes some identifier in constructor and then have one instance of the service per enum.
> 2. We should probably do a build-time check to ensure all the referenced schema provider services exist.
> 3. _(nice-to-have)_ Now a `SchemaProvider` has no context, which limits the use case a bit. We could have e.g.  `#[SchemaSource(provider: EntityEnumSchemaProvider::class, context: ['entity' => PaintColor::class])]` and pass `context` to `getSchemaFragment()`

 1 and 2 make sense, for 2, a CompilerPass to validate references maybe?
                                                                                                                                                                                  
For 3, nice idea but feels a bit tricky to design (especially with IS_REPEATABLE: per-attribute context vs. shared, conflict resolution, etc.). Maybe we could start without it and tackle it in a separate PR?

valtzu:
For 3 I would think the context is always be per attribute because it has no meaning outside a specific provider.

But sure; it can be added later.

chr-hertel:
This is one of the pull requests that get my thinking ... and i don't have a final opinion yet, but will think out loud.

First of all, this functionality is 100% needed - i need it myself, and it's quite a common question people have - so thanks for sure for working on this @camilleislasse! really appreciated!

what i don't like about this solution is the extra attribute, but not sure i have a better/working idea ... let me what you think

```
#[AsTool('search_orders', 'Search orders by status')]
final class SearchOrdersTool
{
    public function __invoke(
        #[Schema(provider: StatusProvider::class)]
        string $status,
    ): array {
        // ...
    }
}
```

the platform component doesn't know a thing about the DIC, but we could still use a class name as identifier, that is associated to an `SchemaProviderInterface` implementation injected into the `SchemaAttributeDescriber` - basically a tagged iteration indexed by service name maybe?

that might save us from requiring the `ContainerInterface` as well :thinking:

chr-hertel:
> 3. _(nice-to-have)_ Now a `SchemaProvider` has no context, which limits the use case a bit. We could have e.g.  `#[SchemaSource(provider: EntityEnumSchemaProvider::class, context: ['entity' => PaintColor::class])]` and pass `context` to `getSchemaFragment()`

nice idea :+1:

chr-hertel (approved):

chr-hertel:
Thank you @camilleislasse.

PR #2028, opened 2026-05-04
[MCP Bundle][Mate] Bump mcp/sdk to ^0.5

pgrimaud:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | -
| License       | MIT

Bumps `mcp/sdk` from `^0.4` to `^0.5` in `symfony/mcp-bundle` and `symfony/ai-mate`.

The fix for modelcontextprotocol/php-sdk#273 (`Session::save()` crashes when `$data` is never initialized) is included in `mcp/sdk` v0.5.

Changes required by the `mcp/sdk` 0.5 BC breaks:
   - Add `symfony/finder` as an explicit dependency of `symfony/mcp-bundle` (previously provided transitively by `mcp/sdk`)
   - Migrate `#[McpTool]` and `Tool::__construct()` usages in `symfony/ai-mate` to named arguments (`$title` parameter was inserted between `$name` and `$description`/`$inputSchema`)

### DIFF
--- a/src/mate/src/Bridge/Monolog/Capability/LogSearchTool.php
+++ b/src/mate/src/Bridge/Monolog/Capability/LogSearchTool.php
@@ -38,7 +38,7 @@ final class LogSearchTool
      * @param string|null $to          End date filter, any PHP-parseable date string
      * @param int         $limit       Maximum number of entries to return
      */
-    #[McpTool('monolog-search', 'Search log entries by text or regex pattern. Supports filtering by log level, channel, environment, and date range. Use empty string for term to match all entries when using filters only.')]
+    #[McpTool(name: 'monolog-search', title: 'Log Search', description: 'Search log entries by text or regex pattern. Supports filtering by log level, channel, environment, and date range. Use empty string for term to match all entries when using filters only.')]
     public function search(
         string $term,
         bool $regex = false,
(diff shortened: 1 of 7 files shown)


src/mate/src/Bridge/Monolog/Capability/LogSearchTool.php
```diff
@@ -38,7 +38,7 @@ public function __construct(
      * @param string|null $to          End date filter, any PHP-parseable date string
      * @param int         $limit       Maximum number of entries to return
      */
-    #[McpTool('monolog-search', 'Search log entries by text or regex pattern. Supports filtering by log level, channel, environment, and date range. Use empty string for term to match all entries when using filters only.')]
+    #[McpTool(name: 'monolog-search', description: 'Search log entries by text or regex pattern. Supports filtering by log level, channel, environment, and date range. Use empty string for term to match all entries when using filters only.')]
```
OskarStark: are these changes needed?

chr-hertel: it is related to the signature change in `mcp/sdk:0.5` but another option would be to use the new title parameter instead of skipping it
```suggestion
    #[McpTool('monolog-search', 'Log Search', 'Search log entries by text or regex pattern. Supports filtering by log level, channel, environment, and date range. Use empty string for term to match all entries when using filters only.')]
```

wachterjohannes: i mean i like the named parameter but we should go with title also! so having named parameter and setting the title would be the ideal option for me

pgrimaud: Done! I also added title to all the other tools. Let me know if any changes are needed.

OskarStark (approved):

wachterjohannes:
@pgrimaud thank you 🚀

wachterjohannes (approved):

chr-hertel:
Thank you @pgrimaud.

PR #2005, opened 2026-04-27
 Minor doc fixes

javiereguiluz:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | -
| License       | MIT

It fixes these errors:

```
Build errors from "2026-04-27 03:17:24"
Error while processing "code-block" directive: "Unsupported code block language "". Added it in SymfonyDocsBuilder\Renderers\CodeNodeRenderer" in file "bundles/ai-bundle" at line 1088
Error while processing "note" directive: "Content expected, none found." in file "components/store/supabase" at line 9
Found invalid reference "components/platform" in file "bundles/ai-bundle"
Found invalid reference "components/platform" in file "bundles/ai-bundle"
```

PR #2030, opened 2026-05-04
[CI] Add docs-builder build to catch broken docs in PRs

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | -
| License       | MIT

Suggested by @stof in #2005: add a `symfony-tools/docs-builder` build alongside the existing DOCtor-RST job, so PRs that introduce broken `:doc:` references, unsupported code-block languages, empty directives, etc. fail the check rather than breaking the website deployment after merge.

Mirrors the [symfony-docs CI layout](https://github.com/symfony/symfony-docs/blob/7.3/.github/workflows/ci.yaml), but uses the docs-builder package's own CLI directly so no wrapper script is needed:

- `docs/_build/composer.json` — pulls `symfony-tools/docs-builder ^0.27`. `symfony/console` is pinned to `^7.4` because docs-builder declares `^8.0` compat in its require but its `Application` class still calls the `Application::add()` method removed in Console 8.x.
- `docs/_build/.gitignore` — excludes `composer.lock`, `output`, `vendor`.
- `docs/docs.json` — exclude list (`_build`) for the build, picked up automatically by docs-builder.
- `.github/workflows/docs-build.yaml` — runs on `pull_request` for `**.rst`/`docs/**`. PHP 8.5, `composer install`, then `vendor/bin/docs-builder build:docs .. output --disable-cache --fail-on-errors --save-errors=errors.log`. The error log is `cat`-ed on failure so messages surface in the job output. `continue-on-error: true` to mirror symfony-docs (informational on first iteration; can be flipped to required once trusted).

### Verified locally

- Clean docs → exit 0, empty errors log.
- Injected a bad `:doc:` reference → exit 1, errors log contains:
  ```
  Build errors from "..."
  Found invalid reference "/components/does-not-exist" in file "index"
  ```

### DIFF
--- a/.github/workflows/docs-build.yaml
+++ b/.github/workflows/docs-build.yaml
@@ -1,0 +1,40 @@
+name: Docs Build
+
+on:
+    pull_request:
+        paths:
+            - '**.rst'
+            - 'docs/**'
+            - '.github/workflows/docs-build.yaml'
+
+concurrency:
+    group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+    cancel-in-progress: true
+
+jobs:
+    docs-build:
+        name: docs-builder
+        runs-on: ubuntu-latest
+
+        steps:
+            - name: Checkout
+              uses: actions/checkout@v6
+
+            - name: Setup PHP
+              uses: shivammathur/setup-php@v2
+              with:
+                  php-version: 8.5
+
+            - name: Install dependencies
+              uses: ramsey/composer-install@v4
+              with:
+                  working-directory: docs/_build
+
+            - name: Build the docs
+              working-directory: docs/_build
+              run: vendor/bin/docs-builder build:docs .. output --disable-cache --fail-on-errors --save-errors=errors.log
+
+            - name: Show build errors
+              if: failure()
+              working-directory: docs/_build
+              run: cat errors.log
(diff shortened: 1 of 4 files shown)


docs/_build/composer.json
```diff
@@ -0,0 +1,22 @@
...
+            "*": "dist"
+        },
+        "sort-packages": true,
+        "allow-plugins": {
+            "symfony/flex": true
+        }
+    },
+    "require": {
+        "php": ">=8.3",
+        "symfony/console": "^6.2 || ^7.0",
```
OskarStark: Do we need 6.2 here and in the next line? Only 7.3 should work

chr-hertel: can be even 7.4, yes

OskarStark (approved):

.github/workflows/docs-build.yaml
```diff
@@ -0,0 +1,41 @@
...
+
+concurrency:
+    group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+    cancel-in-progress: true
+
+jobs:
+    docs-build:
+        name: docs-builder
+        runs-on: ubuntu-latest
+        continue-on-error: true
```
stof: `continue-on-error: true` means that it will still attempt to build the docs if installing PHP or the composer dependencies fails with an error. It does not make sense IMO.

PR #2031, opened 2026-05-04
[Platform] Fix JSONSchema generation with CPP & `@var`

HypeMC:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | -
| License       | MIT

PR #1683 broke JSON Schema generation when using constructor property promotion with `@var` annotations.

For example, given the following class:

```php
final class CPPWithAtVarDocFixture
{
    public function __construct(
        /** @var Step[] */
        public array $steps = [],
    ) {
    }
}
```

The generated schema has changed unexpectedly. The diff between the previous and current output is:

```diff
{
    "type": "object",
    "properties": {
        "steps": {
            "type": "array",
            "items": {
-               "type": "object",
+               "type": "string",
                "properties": {
                    "explanation": {
                        "type": "string"
                    },
                    "output": {
                        "type": "string"
                    }
                },
                "required": [
                    "explanation",
                    "output"
                ],
                "additionalProperties": false
            }
        }
    },
    "required": [
        "steps"
    ],
    "additionalProperties": false
}
```

I'm not sure what the best approach is to fix this. I'm open to suggestions.

### DIFF
--- a/src/platform/src/Contract/JsonSchema/Describer/TypeInfoDescriber.php
+++ b/src/platform/src/Contract/JsonSchema/Describer/TypeInfoDescriber.php
@@ -130,10 +130,9 @@ final class TypeInfoDescriber implements ObjectDescriberInterface, PropertyDescr
             case $type->isIdentifiedBy(TypeIdentifier::ARRAY):
                 \assert($type instanceof CollectionType);
 
-                return [
-                    'type' => 'array',
-                    'items' => $this->getTypeSchema($type->getCollectionValueType()),
-                ];
+                $items = $this->getTypeSchema($type->getCollectionValueType());
+
+                return ['type' => 'array'] + ($items ? ['items' => $items] : []);
 
             case $type->isIdentifiedBy(TypeIdentifier::OBJECT):
                 if ($type instanceof BuiltinType) {
@@ -150,9 +149,9 @@ final class TypeInfoDescriber implements ObjectDescriberInterface, PropertyDescr
             case $type->isIdentifiedBy(TypeIdentifier::NULL):
                 return ['type' => 'null'];
             case $type->isIdentifiedBy(TypeIdentifier::STRING):
-            default:
-                // Fallback to string for any unhandled types
                 return ['type' => 'string'];
+            default:
+                return [];
         }
     }
 
--- a/src/platform/tests/Contract/JsonSchema/Describer/DescriberTest.php
+++ b/src/platform/tests/Contract/JsonSchema/Describer/DescriberTest.php
@@ -1,0 +1,50 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Platform\Tests\Contract\JsonSchema\Describer;
+
+use PHPUnit\Framework\TestCase;
+use Symfony\AI\Platform\Contract\JsonSchema\Describer\Describer;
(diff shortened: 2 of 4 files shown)


valtzu:
As I understand the `"type": "string"` comes from here https://github.com/symfony/ai/blob/be69c631e9f13dbaa236c4608fa886d41e217487/src/platform/src/Contract/JsonSchema/Describer/TypeInfoDescriber.php#L153-L155

For constructor promoted properties we describe both, parameters & properties and in this case the property has `@var Step[]` while the parameter is an untyped array (as `@param` is absent).

I don't know the reasoning why we fallback to `string` on unknown type _(this was already the logic before the referenced refactoring)_ – I would think if type is unknown, schema should be untouched. If we removed the fallback-to-string logic, no other fixes would be needed.

```diff
@@ -150,9 +150,9 @@ final class TypeInfoDescriber implements ObjectDescriberInterface, PropertyDescr
             case $type->isIdentifiedBy(TypeIdentifier::NULL):
                 return ['type' => 'null'];
             case $type->isIdentifiedBy(TypeIdentifier::STRING):
-            default:
-                // Fallback to string for any unhandled types
                 return ['type' => 'string'];
+            default:
+                return [];
```

---

That said, I don't think your fix is necessarily wrong either.

HypeMC:
@valtzu Hi, thanks for the feedback. I already tried that approach, but it caused a test to fail:

```
There was 1 failure:

1) Symfony\AI\Platform\Tests\Contract\JsonSchema\FactoryTest::testBuildParametersDefinitionRequiredWithAdditionalToolParameterAttribute
Failed asserting that two arrays are identical.
--- Expected
+++ Actual
@@ @@
         ],
         'products' => Array &10 [
             'type' => 'array',
-            'items' => Array &11 [
-                'type' => 'string',
-            ],
+            'items' => Array &11 [],
             'description' => 'The products given to the tool',
             'minItems' => 1,
             'maxItems' => 10,
```

The failure seems legit to me.

valtzu:
> The failure seems legit to me.

The parameter is defined as `array<mixed>` so in my opinion it is not correct to treat it as `list<string>` _(should rather just render as `"type": "array"` with no `items` at all)_. For a `mixed` type I would expect the platform is allowed to come up with any structure that is suitable for the request.

https://github.com/symfony/ai/blob/7bdd0ff64e005b89281b0926e8a84f06f54d67e3/src/agent/tests/Fixtures/Tool/ToolWithToolParameterAttribute.php#L27

But again, I don't know the rationale behind this, maybe there is some reasoning like "the platform throws an error if you don't provide `type` at all".

src/platform/src/Contract/JsonSchema/Describer/TypeInfoDescriber.php
```diff
@@ -187,4 +185,31 @@ private function buildEnumSchema(string $enumClassName): array
...
+     * @param array<string, mixed> $schema
+     * @param array<string, mixed> $subSchema
+     *
+     * @return array<string, mixed>
+     */
+    private function mergeSchema(array $schema, array $subSchema): array
+    {
+        if (isset($schema['type']) && 'array' === $schema['type']) {
+            if (isset($schema['items']['properties']) && !isset($subSchema['items']['properties'])) {
+                return $schema;
```
valtzu: Is it a bit unexpected to skip merging the other properties here?

chr-hertel:
I don't think there was a particular reason to go with the default string - that was a decision early on. from my point of view, taking that route to remove the `['type' => 'string']` is the better solution here.

chr-hertel (approved):

chr-hertel:
Thank you @HypeMC.

PR #2033, opened 2026-05-05
[AI Bundle] Data Collector - Fix bug that re-throws Exception

MolloKhan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | yes
| Issues        | -
| License       | MIT

When a platform call throws (rate limit, network error, etc.), the bundle's DataCollector re-runs the conversion during `kernel.response` to build profile data, so it throws the exception again. From the user's perspective, their `try/catch` around the agent looks broken: a 429 turns into a 500.

  Fix:

  - `DataCollector::awaitCallResults()` — wrap the per-call work in try/catch. Failed calls are recorded as result_type: 'error' with an error array (class, message), and the profiler template renders them. 
  - `DeferredResult::getResult()` — cache the exception; re-throw the same instance on subsequent calls. 

This bug only affects the development environment because the data collector only runs with `kernel.debug=true`.

BTW, Claude helped me debug the issue and find a proper solution

### DIFF
--- a/src/ai-bundle/CHANGELOG.md
+++ b/src/ai-bundle/CHANGELOG.md
@@ -6,6 +6,7 @@ CHANGELOG
 
  * Validate structured output using `symfony/validator` when available
  * Make `ai:platform:invoke` arguments optional and prompt for them interactively when missing
+ * Visualize failed calls with `result_type: 'error'` in profiler
 
 0.8
 ---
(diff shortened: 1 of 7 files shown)


chr-hertel (approved):
Gosh, i guess that was annoying - thanks! 🙏

chr-hertel:
Thank you @MolloKhan.

chr-hertel:
Fixed code style issues, rebased, and rephrased bundle changelog entry while merging.

MolloKhan:
> Gosh, i guess that was annoying - thanks! 🙏

Indeed it was :) - I thought I was crazy for a moment

PR #1997, opened 2026-04-24
[Platform][Gemini] Add gemini tool call roundtrip regression test

froozeify:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no 
| Docs?         | no
| Issues        | Fix #1980 
| License       | MIT

Add example test to avoid regression as per requested in https://github.com/symfony/ai/pull/1980

PR #2034, opened 2026-05-06
[Examples] Include AssistantMessage usage in chat examples

### LINKED ISSUE #1969: [Examples] Include `AssistantMessage` usage in examples

Amoifr:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | Fix #1969
| License       | MIT

Adds a multi-turn follow-up to every \`examples/<bridge>/chat.php\`: the
assistant's reply is fed back into the \`MessageBag\` via
\`Message::ofAssistant()\`, then a context-dependent user follow-up is
sent.

This addresses #1969 — the chat examples now exercise \`AssistantMessage\`
normalization across all bridges, making it possible to smoke-test
changes against the actual provider APIs.

Notes:
- Pattern consistent across the 21 examples; pre-existing options
  (\`max_tokens\`, \`temperature\`, \`permission_mode\`, …) preserved on
  the second invocation.
- For agent-based bridges (\`docker-model-runner\`, \`claude-code\`,
  \`codex\`) the example still demonstrates the \`MessageBag\` shape with
  the assistant turn fed back in.

### DIFF
--- a/examples/aimlapi/chat.php
+++ b/examples/aimlapi/chat.php
@@ -24,14 +24,10 @@ $messages = new MessageBag(
 $result = $platform->invoke('gemini-2.5-flash', $messages, [
     'max_tokens' => 500, // specific options just for this call
 ]);
+continue_chat($messages, $result->asText());
 
-echo $result->asText().\PHP_EOL;
-
-// Multi-turn: feed the assistant's reply back into the bag and ask a follow-up.
-$messages->add(Message::ofAssistant($result->asText()));
 $messages->add(Message::ofUser('And which versions are LTS?'));
 $result = $platform->invoke('gemini-2.5-flash', $messages, [
     'max_tokens' => 500,
 ]);
-
-echo $result->asText().\PHP_EOL;
+continue_chat($messages, $result->asText());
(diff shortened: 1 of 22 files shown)


Amoifr:
Pushed a follow-up that factors out the duplication.

The `echo $result->asText() / getContent()` + `Message::ofAssistant(...)` pair was repeated twice per file across all 21 examples. Pulled it into a small `continue_chat(MessageBag $messages, string $content)` helper in `examples/bootstrap.php`, in the spirit of the existing `print_token_usage()` / `print_sources()` helpers.

Net effect: -70 lines, examples now read as a clean sequence of `add user turn → invoke → continue_chat` pairs, and `Message::ofAssistant()` is still exercised on every call (just mutualized).

Happy to revert/inline it if you'd rather keep the multi-turn pattern explicit in each example. 🙂

chr-hertel:
Thanks @Amoifr, we addressed the issue for #1969 as well in #2059 and #1997 - so this in not needed anymore. the main issue and intention was to have a turnaround - not only adding it to the MessageBag but also serializing it as part of the contract and submitting it again. Anyhow, thanks!

PR #2039, opened 2026-05-06
[Platform][OpenRouter] Update Static Model Catalog and add update script

lochmueller:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | 
| License       | MIT

- Sort models in the ModelApiCatalog 
- Add update script for the static model list
- Update the static model list based on the ModelApiCatalog

There were considerations in the past to let the static list become outdated and not update it. However, too many people use the regular ModelCatalog, which is why I recommend updating it from time to time.

### DIFF
--- a/src/platform/src/Bridge/OpenRouter/ModelApiCatalog.php
+++ b/src/platform/src/Bridge/OpenRouter/ModelApiCatalog.php
@@ -53,6 +53,7 @@ final class ModelApiCatalog extends AbstractOpenRouterModelCatalog
                 ...$this->fetchRemoteModels(),
                 ...$this->fetchRemoteEmbeddings(),
             ];
+            ksort($this->models);
             $this->modelsAreLoaded = true;
         }
     }
--- a/src/platform/src/Bridge/OpenRouter/composer.json
+++ b/src/platform/src/Bridge/OpenRouter/composer.json
@@ -33,7 +33,8 @@
         "phpstan/phpstan": "^2.1",
         "phpstan/phpstan-phpunit": "^2.0",
         "phpstan/phpstan-strict-rules": "^2.0",
-        "phpunit/phpunit": "^11.5.53"
+        "phpunit/phpunit": "^11.5.53",
+        "symfony/console": "^7.4|^8.0"
     },
     "minimum-stability": "dev",
     "autoload": {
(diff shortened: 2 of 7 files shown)


src/platform/src/Bridge/OpenRouter/composer.json
```diff
@@ -33,7 +33,8 @@
         "phpstan/phpstan": "^2.1",
         "phpstan/phpstan-phpunit": "^2.0",
         "phpstan/phpstan-strict-rules": "^2.0",
-        "phpunit/phpunit": "^11.5.53"
+        "phpunit/phpunit": "^11.5.53",
+        "symfony/console": "^5.4|^6.4|^7.4|^8.0"
```
chr-hertel: please drop `^5.4` and `^6.4` here - even if maybe possible with the script, it is inconsistent with what we support with the Platform

src/platform/src/Bridge/OpenRouter/ModelApiCatalog.php
```diff
@@ -53,6 +53,7 @@ protected function preloadRemoteModels(): void
                 ...$this->fetchRemoteModels(),
                 ...$this->fetchRemoteEmbeddings(),
             ];
+            ksort($this->models);
```
chr-hertel, line 56: why is that needed? cosmetics?

chr-hertel (changes requested):
Mixed feeling - that `updateStaticModelCatalog` is on the one side today's reality for you updating the code here, and on the other hand just another evidence of our flaws here. but that's rather off scope. smaller comments tho.

src/platform/src/Bridge/OpenRouter/composer.json
lochmueller: Done.

src/platform/src/Bridge/OpenRouter/ModelApiCatalog.php
lochmueller: I add this, to keep the diff for future updates with the update script "clean". So... more or less cosmetics :)

chr-hertel (approved):

chr-hertel:
Thank you @lochmueller.

PR #2038, opened 2026-05-06
[Platform][Gemini] Strip the JSON-Schema $schema meta-key from tool parameters

froozeify:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no 
| Issues        | Fix https://github.com/symfony/ai/issues/2037
| License       | MIT

More information on the issue : https://github.com/symfony/ai/issues/2037

TLDR: Gemini have a strict validation and was returning an exception when it got the `$schema` field

Returning an excetion :

```
Invalid JSON payload received. Unknown name "$schema" at 'tools[0].function_declarations[9].parameters': Cannot find field.
```

PR #2040, opened 2026-05-06
[Platform][Gemini] Wrap function_response in a Protobuf Struct

### LINKED ISSUE #1555: [Gemini Bridge] ToolCallMessageNormalizer causes INVALID_ARGUMENT error with function_response format
### Symfony AI version
- `symfony/ai-gemini-platform`: v0.3.0
- `symfony/ai-agent`: v0.3.0

### Gemini model
`gemini-2.5-flash`

### Bug summary
When using function calling (tools) with the Gemini platform bridge, the API returns a `400 INVALID_ARGUMENT` error after a tool is executed and the response is sent back to the model.

### Error message
```
Error "400" - "INVALID_ARGUMENT": "Invalid JSON payload received. Unknown name "response" at 'contents[2].parts[0].function_response': Proto field is not repeating, cannot start list."
```

### Root cause
The issue is in `src/Contract/ToolCallMessageNormalizer.php`. Two problems:

1. **`array_filter()` breaks the object structure** when [id](file:///home/johnnestebann/Documents/proycer/ai-tool-proto/app/src/Entity/Material.php#41-45) is empty (which is common since Gemini doesn't always return an [id](file:///home/johnnestebann/Documents/proycer/ai-tool-proto/app/src/Entity/Material.php#41-45) for function calls)

2. **The `response` field format** doesn't properly wrap the content as a Protobuf Struct object

### Current code (buggy)
```php
public function normalize(mixed $data, ?string $format = null, array $context = []): array
{
    $resultContent = json_validate($data->getContent())
        ? json_decode($data->getContent(), true) : $data->getContent();

    return [[
        'functionResponse' => array_filter([
            'id' => $data->getToolCall()->getId(),
            'name' => $data->getToolCall()->getName(),
            'response' => \is_array($resultContent) ? $resultContent : [
                'rawResponse' => $resultContent,
            ],
        ]),
    ]];
}
```

### Proposed fix
```php
public function normalize(mixed $data, ?string $format = null, array $context = []): array
{
    $resultContent = json_validate($data->getContent())
        ? json_decode($data->getContent(), true) : $data->getContent();

    // Build functionResponse without array_filter to preserve structure
    $functionResponse = [
        'name' => $data->getToolCall()->getName(),
        'response' => \is_array($resultContent) ? ['result' => $resultContent] : [
            'result' => $resultContent,
        ],
    ];

    // Only add id if not empty (Gemini API says it's optional)
    $id = $data->getToolCall()->getId();
    if (!empty($id)) {
        $functionResponse['id'] = $id;
    }

    return [[
        'functionResponse' => $functionResponse,
    ]];
}
```

### Why this fixes it
1. Removes `array_filter()` which was silently dropping fields and breaking the JSON structure
2. Wraps the response content in a `{'result': ...}` object, which is proper Protobuf Struct format as per [Gemini API documentation](https://ai.google.dev/api/caching#FunctionResponse)

### Steps to reproduce
1. Configure a Gemini agent with any tool
2. Send a message that triggers the tool
3. After the tool executes and returns a result, the error occurs when sending `functionResponse` back to the model

### Environment
- PHP 8.5
- Symfony 8.0.5
- Agent with function calling enabled

---

**I can submit a PR with this fix if you'd like.**

froozeify:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #1555
| License       | MIT

After fixing these issue https://github.com/symfony/ai/pull/2038 I also had the issue reported here https://github.com/symfony/ai/issues/1555

```
Invalid JSON payload received. Unknown name "response" at 'contents[N].parts[0].function_response': Proto field is not repeating, cannot start list.
```

- Gemini doesn't always return an id for function call, making it being dropped.
- As reported by @johnnestebann Gemini a Protobuf Struct format. This issue occured when a list was returned by my MCP.

### DIFF
--- a/examples/gemini/toolcall-list-result.php
+++ b/examples/gemini/toolcall-list-result.php
@@ -1,0 +1,46 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use Symfony\AI\Agent\Agent;
+use Symfony\AI\Agent\Toolbox\AgentProcessor;
+use Symfony\AI\Agent\Toolbox\Attribute\AsTool;
+use Symfony\AI\Agent\Toolbox\Toolbox;
+use Symfony\AI\Platform\Bridge\Gemini\Factory;
+use Symfony\AI\Platform\Message\Message;
+use Symfony\AI\Platform\Message\MessageBag;
+
+require_once dirname(__DIR__).'/bootstrap.php';
+
+#[AsTool('list_european_capitals', 'Returns the capitals of a few European countries.')]
+final class EuropeanCapitalsTool
+{
+    /**
+     * @return list<string>
+     */
+    public function __invoke(): array
+    {
+        return ['Paris', 'Berlin', 'Madrid', 'Rome'];
+    }
+}
+
+$platform = Factory::createPlatform(env('GEMINI_API_KEY'), http_client());
+
+$toolbox = new Toolbox([new EuropeanCapitalsTool()], logger: logger());
+$processor = new AgentProcessor($toolbox);
+$agent = new Agent($platform, 'gemini-2.5-flash', [$processor], [$processor]);
+
+$messages = new MessageBag(
+    Message::forSystem('Use the available tool to answer.'),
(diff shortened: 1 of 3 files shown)


chr-hertel (approved):

chr-hertel:
Thank you @froozeify.

PR #2054, opened 2026-05-07
[Mate] Add MCP tool and resource design principles to CLAUDE.md

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | yes
| Issues        | n/a
| License       | MIT

Adds a "MCP Tool & Resource Design Principles" section to `src/mate/CLAUDE.md` that applies to **all** Mate tools and resources — not just profiler formatters.

The section codifies five principles that every tool in Mate should follow:

1. **Distillation over completeness** — only return fields that change the AI's diagnosis
2. **Truncation is intentional** — hard limits with `_truncated` signals so the AI knows it has a sample
3. **Sensitive data stays out** — tokens, payloads, credentials excluded by default
4. **Format for reasoning** — human-readable strings, consistent units, flat structures
5. **The triage → detail pattern** — `getSummary()` answers "is this relevant?", `format()` answers "what exactly happened?"

These principles apply equally to `CollectorFormatterInterface` implementations, `LogSearchTool`, `ServiceTool`, and any future tools or resources.


src/mate/CLAUDE.md
```diff
@@ -122,3 +122,55 @@ Running `bin/mate discover` generates `mate/AGENT_INSTRUCTIONS.md` with extensio
 - Bridge tests are located within their respective bridge directories
 - Fixtures for discovery tests in `tests/Discovery/Fixtures/`
 - Component follows Symfony coding standards
+
+## MCP Tool & Resource Design Principles
```
wachterjohannes: This is the main part of this PR - the addition of AGENTS.md is only here because it was missing

chr-hertel:
Does it make sense to use a symlink? from one to the other instead of having two separate ones? there is already drift between them or is that intended?

wachterjohannes:
Good point — instead of a symlink I went with a different approach that achieves the same goal:

- `AGENTS.md` is now the canonical, general AI-agent guidance file (carries the design principles + all the component docs).
- `CLAUDE.md` is reduced to an `@AGENTS.md` Claude Code import plus a small Claude Code-specific section.

That way the shared content lives in one place and the two files can't drift, while CLAUDE.md can still carry anything that's specifically Claude Code-related.

chr-hertel (approved):
Nice solution 👍

chr-hertel:
Thank you @wachterjohannes.

wachterjohannes:
> Nice solution 👍 

maybe this is also a practical solutiom for the other components

PR #2015, opened 2026-04-28
[Platform] Add tool-call trace metadata for ClaudeCode and Codex bridges

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | -
| License       | MIT

Introduce `ToolCallTrace` and `ToolCallTraceCollection` under `Symfony\AI\Platform\Metadata` to capture per-tool-call timing and status (`id`, `name`, `arguments`, `started_at_ms`, `duration_ms`, `errored`).

`ToolCallTraceCollection` implements `MergeableMetadataInterface` so traces accumulate across multi-step results, plus `IteratorAggregate`, `Countable`, and `JsonSerializable` for ergonomic consumption.

Both the ClaudeCode and Codex bridges now harvest these traces from their process outputs:

- **ClaudeCode** parses `assistant`/`stream_event` events and pairs `content_block_start` (`tool_use`) with `content_block_stop` to derive per-call duration.
- **Codex** parses both legacy `item.started`/`item.completed` events and the newer `event_msg` (`mcp_tool_call_start`) format, merging started + completed pairs by id.

Each bridge's `ResultConverter` then attaches the resulting collection to the result metadata under `ToolCallTraceCollection::METADATA_KEY` when present.

Tests cover the legacy + event-message paths, the merge behavior, and metadata attachment in both bridges.

PR #2063, opened 2026-05-09
[Platform][ClaudeCode] surface tool-use deltas in the stream-result protocol

haraldpdl:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | -
| License       | MIT

The ClaudeCode bridge's stream converter currently yields only `TextDelta` events. The underlying CLI's stream-json output also emits `content_block_start`, `content_block_delta`, `content_block_stop` and `message_stop` events for tool-use blocks, but those are filtered out, so consumers iterating a `StreamResult` see no signal between "the model decided to call a tool" and "the tool's result came back". For progress UIs around long-running tool-using dialogues, that gap is the entire perceptible wait.

This PR widens `ClaudeCode\ResultConverter::convertStream()` to yield the existing `ToolCallStart`, `ToolInputDelta` and `ToolCallComplete` deltas, mirroring `Anthropic\ResultConverter`. The CLI emits the same Anthropic content-block events, just nested one level deeper inside `stream_event.event.*`. No new public API.

This is parity, not a regression: `StreamResult` consumers that already type-check deltas (the documented pattern, used by the Anthropic bridge) keep working unchanged. Consumers iterating without type checks may now see new types, the same situation as the Anthropic bridge today.

Tests cover: isolated tool call, interleaved with text, multiple tool calls in order, and an empty-input call (start + stop with no `input_json_delta`). A new `examples/claude-code/toolcall-stream.php` mirrors `examples/anthropic/toolcall-stream.php` to demonstrate consumption.

### Adjacent, non-overlapping work

#2015 attaches `ToolCallTrace` *metadata* to the final result for timing observability. Different consumer surface (final-result metadata vs. real-time stream deltas) and complementary.

### Out of scope

* Other bridges with the same gap. Codex (sibling CLI bridge, different JSON event shape) and the OpenAI-compat group (Cerebras, DeepSeek, Perplexity, DockerModelRunner, Generic via `CompletionsConversionTrait`) each need their own per-provider PR.
* `is_error` asymmetry. `convert()`'s non-stream path throws `RuntimeException` on a final `result` event with `is_error: true`; the stream path silently ends. Worth fixing, but it is an orthogonal behaviour change (throwing from a generator mid-iteration) and best handled separately.

### DIFF
--- a/src/platform/src/Bridge/ClaudeCode/ResultConverter.php
+++ b/src/platform/src/Bridge/ClaudeCode/ResultConverter.php
@@ -16,8 +16,12 @@ use Symfony\AI\Platform\Model;
 use Symfony\AI\Platform\Result\RawResultInterface;
 use Symfony\AI\Platform\Result\ResultInterface;
 use Symfony\AI\Platform\Result\Stream\Delta\TextDelta;
+use Symfony\AI\Platform\Result\Stream\Delta\ToolCallComplete;
+use Symfony\AI\Platform\Result\Stream\Delta\ToolCallStart;
+use Symfony\AI\Platform\Result\Stream\Delta\ToolInputDelta;
 use Symfony\AI\Platform\Result\StreamResult;
 use Symfony\AI\Platform\Result\TextResult;
+use Symfony\AI\Platform\Result\ToolCall;
 use Symfony\AI\Platform\ResultConverterInterface;
 use Symfony\AI\Platform\TokenUsage\TokenUsageExtractorInterface;
 
@@ -63,6 +67,10 @@ final class ResultConverter implements ResultConverterInterface
 
     private function convertStream(RawResultInterface $result): \Generator
     {
+        $toolCalls = [];
+        $currentToolCall = null;
+        $currentToolCallJson = '';
+
         foreach ($result->getDataStream() as $data) {
             $type = $data['type'] ?? '';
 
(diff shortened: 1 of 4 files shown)


src/platform/src/Bridge/ClaudeCode/ResultConverter.php
```diff
@@ -72,6 +80,61 @@ private function convertStream(RawResultInterface $result): \Generator
                 && 'text_delta' === ($data['event']['delta']['type'] ?? '')
             ) {
                 yield new TextDelta($data['event']['delta']['text']);
+                continue;
```
chr-hertel: not sure about those `continue;` statements? why do we need them?

if they are needed because we would hit other ifs, why would they not be legit? and if we wouldn't hit another if, are they only to optimize the foreach?

chr-hertel:
example works as expected :+1: 

image

src/platform/src/Bridge/ClaudeCode/ResultConverter.php
haraldpdl: Thanks for the review. The changes matched the style of Anthropic's `ResultConverter::convertStream()` but have been simplified. Each branch is keyed on a disjoint `event.type` (and `delta.type` for the `content_block_delta` cases), so no two branches can match the same event. The `continue;` statements only skipped redundant `if` checks and implied a fall-through risk that the predicates already prevent. Dropped them in the next commit.

chr-hertel: Thanks for that :+1:

chr-hertel (approved):

chr-hertel:
Thank you @haraldpdl.

PR #2067, opened 2026-05-10
[Platform] Surface API error message on RateLimitExceededException

### LINKED ISSUE #2042: [Platform] 429 Exception doesn't show error message
Hello! 

I've found a pesky detail when dealing with 429 exceptions. The error description is not present in the exception object. Platforms like OpenAI use that code not only for rate limit exceptions, but also when you run out of credits - it took me a while to debug why I was getting a 429 because I assumed it was a rate limit exception. Sadly, I didn't check the error description until I exhausted all the possible causes

TLDR: Adding the error description to the exception may save time for other devs when getting a 429

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #2042
| License       | MIT

A 429 response is ambiguous on most providers: OpenAI for example uses it for both per-minute rate limiting *and* for accounts that are out of credits, and the only way to distinguish the two is the error description in the response body — which `RateLimitExceededException` was throwing away.

This forwards the provider's `error.message` (or the flat top-level `message` used by Mistral/Cerebras/Cohere) into the exception:

```php
// before
catch (RateLimitExceededException $e) {
    $e->getMessage();      // "Rate limit exceeded."
    $e->getRetryAfter();   // 60
}

// after
catch (RateLimitExceededException $e) {
    $e->getMessage();      // "Rate limit exceeded. You exceeded your current quota, please check your plan and billing details."
    $e->getRetryAfter();   // 60
}
```

The constructor gains an optional second `?string $errorMessage` argument; passing `null` (or omitting it) preserves the previous `"Rate limit exceeded."` message verbatim, so the change is backward compatible. All 429 paths in the bridges (`HttpStatusErrorHandlingTrait` users + the Anthropic / OpenAI Gpt / OpenResponses / Whisper / VertexAi Gemini / Generic Embeddings & Completions converters that handle 429 directly) now extract and forward the message. Tests updated to assert the new behaviour and to cover the empty-body fallback.

### DIFF
--- a/src/platform/src/Bridge/Anthropic/ResultConverter.php
+++ b/src/platform/src/Bridge/Anthropic/ResultConverter.php
@@ -64,7 +64,8 @@ class ResultConverter implements ResultConverterInterface
         if (429 === $response->getStatusCode()) {
             $retryAfter = $response->getHeaders(false)['retry-after'][0] ?? null;
             $retryAfterValue = $retryAfter ? (int) $retryAfter : null;
-            throw new RateLimitExceededException($retryAfterValue);
+            $errorMessage = json_decode($response->getContent(false), true)['error']['message'] ?? null;
+            throw new RateLimitExceededException($retryAfterValue, $errorMessage);
         }
 
         if ($options['stream'] ?? false) {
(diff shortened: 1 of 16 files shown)


OskarStark (approved):

valtzu:
Just a though: would it better to have separate exception class for "out of credits"? I would assume its quite common to set up an alert when getting this error, and now it will require checking the message which will differ on every platform.

Of course this here we can have regardless

MolloKhan:
Thanks for this!

PR #2068, opened 2026-05-10
[Mate] Add @wachterjohannes as CODEOWNER

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | -
| License       | MIT

Add myself as a CODEOWNER for `src/mate` so that I get automatically informed (review request notification) whenever a pull request is created that touches files in this directory.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -6,5 +6,5 @@ docs @chr-hertel @OskarStark
 examples @chr-hertel @OskarStark
 fixtures @chr-hertel @OskarStark
 src @chr-hertel @OskarStark
-src/mate @chr-hertel @Nyholm @OskarStark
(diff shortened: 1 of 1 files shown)


chr-hertel (approved):
Makes sense 👍

OskarStark:
Makes sense, but unfortunately will not work, AFAIK this only works for repository members.

wachterjohannes:
@OskarStark ok seems like an issue - i would be happy if we find a solution for it - BUT if now i wait always for the message of @chr-hertel 😂

chr-hertel:
Gotta merge this anyhow, even if only symbolic :D

chr-hertel:
Thank you @wachterjohannes.

PR #2071, opened 2026-05-11
[Platform][Ollama] Capability insert ollama platform fix

### LINKED ISSUE #2019: ai-platform-ollama crashes when it runs into unknown model capability
Hi all
Firstly, thank you for your efforts. The AI suite for Symfony is a breakthrough for me, in many ways.

I ran into a problem with **ai-ollama-platform package**. In ModelCatalog.php line 64, the library is trying to parse the capabilities for each available model into an array of enum values. However, in the current implementation, if it runs into an unknown capability, it throws an exception. I accidentally ran into this when I installed qwen2.5-coder:1.5b - it has 'insert' capability which is not a part of the current enum (defined in \Symfony\AI\Platform\Capability). 
I temporarily forked the package to be able to use it, and am currently simply ignoring unknown capability values, which works for my use case. However, I believe this should be properly addressed. I can easily push a PR for ai-ollama-platform, however it would not help unless someone adds the new capability in the ai-platform package.

To reproduce this problem, install ollama, install a model with 'insert' capability (e.g. qwen2.5-coder). After you initialize the platform, try executing 

`$p->getModelCatalog()->getModels()`

Where $p is created using Platform::create() from ai-ollama-platform. 

I am using version 0.8 of all ai- components, on Symfony 7.4.9, PHP 8.3.

Thank you 
Viktor

vkost:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Docs?         | no 
| Issues        | Fix #2019 
| License       | MIT

Fixes an issue with some models (e.g., qwen2.5-coder) reporting 'insert' capability if you attempt to list the available models. Adds 'insert' as a valid capability.

### DIFF
--- a/src/platform/src/Capability.php
+++ b/src/platform/src/Capability.php
@@ -61,4 +61,7 @@ enum Capability: string
 
     // Thinking
     case THINKING = 'thinking';
+
+    // Fill-in-the-middle (insert)
+    case FILL_IN_THE_MIDDLE = 'fill-in-the-middle';
 }
--- a/src/platform/src/Bridge/Ollama/ModelCatalog.php
+++ b/src/platform/src/Bridge/Ollama/ModelCatalog.php
@@ -69,6 +69,7 @@ final class ModelCatalog implements ModelCatalogInterface
                 'thinking' => Capability::THINKING,
                 'vision' => Capability::INPUT_IMAGE,
                 'audio' => Capability::INPUT_AUDIO,
+                'insert' => Capability::FILL_IN_THE_MIDDLE,
                 default => throw new InvalidArgumentException(\sprintf('The "%s" capability is not supported', $capability)),
             },
             $payload['capabilities'],
--- a/src/platform/src/Bridge/Ollama/Tests/OllamaApiCatalogTest.php
+++ b/src/platform/src/Bridge/Ollama/Tests/OllamaApiCatalogTest.php
@@ -130,6 +130,39 @@ final class OllamaApiCatalogTest extends TestCase
         $this->assertSame(1, $httpClient->getRequestsCount());
     }
 
+    public function testModelCatalogCanReturnInsertModelsFromApi()
+    {
+        $httpClient = new MockHttpClient([
+            new JsonMockResponse([
+                'models' => [
+                    [
+                        'name' => 'qwen2.5-coder',
+                        'details' => [],
+                    ],
+                ],
+            ]),
+            new JsonMockResponse([
+                'capabilities' => ['insert'],
+            ]),
+        ], 'http://127.0.0.1:11434');
+
+        $modelCatalog = new ModelCatalog($httpClient);
+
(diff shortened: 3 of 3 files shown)


src/platform/src/Capability.php
```diff
@@ -61,4 +61,7 @@ enum Capability: string
 
     // Thinking
     case THINKING = 'thinking';
+
+    // Insert
+    case FILL_IN_THE_MIDDLE = 'insert';
```
chr-hertel: let's keep it consistent:

```suggestion
    case FILL_IN_THE_MIDDLE = 'fill-in-the-middle';
```

chr-hertel (changes requested):

Guikingone:
Can we have extra tests for this capability? Just to ensure that we cover any regression 🙂

vkost:
@Guikingone I don't mind the work, but I would need some guidance. I just checked the other Capabilities, and the way they were used in tests. Most of the bridges have a hardcoded list of models in the ModelCatalog, and then they test the class by extending ModelCatalogTestCase. Ollama, however, reads the models from the server. This is where the crash was happening - the code was unable to process the 'insert' capability returned by some models. 

The only thing that comes to mind is mocking the OllamaClient to return a list of models, where one of the models would have this 'insert' capability. Not sure how much this would help.

If you have another suggestion on how to test - I will gladly do what I can.

Vic

Guikingone:
Hi @vkost 👋🏻 

Sorry for the delay, for the test, you can find inspiration in `OllamaApiCatalogTest` (wrong name BTW, something to change 😄): 

```php
# ..
public function testModelCatalogCanReturnModelFromApi()
{
    $httpClient = new MockHttpClient([
        new JsonMockResponse([
            'capabilities' => ['completion'],
        ]),
    ], 'http://127.0.0.1:11434');

    $modelCatalog = new ModelCatalog($httpClient);

    $model = $modelCatalog->getModel('foo');

    $this->assertSame('foo', $model->getName());
    $this->assertSame([
        Capability::INPUT_MESSAGES,
        Capability::OUTPUT_STRUCTURED,
    ], $model->getCapabilities());
    $this->assertSame(1, $httpClient->getRequestsCount());
}
# ...
```

By adding a new test that change `completion` to `insert`, you can test that the `getCapabilities` array returns the new capability, should be enough for a first test 👍🏻

vkost:
Hi @Guikingone 
Thank you for the pointers. I was fooled by the file name and did not check there.
I added the test in a fashion of other similar tests I found. I thought it best to go with the example that returns the model and the capability - this way I was able to leave a trace that the 'insert' capability was returned by qwen2.5-coder.

I apologize for multiple commits - I pushed too early.

Let me know if I missed anything.
Vic

chr-hertel (approved):

chr-hertel:
Thank you @vkost.

PR #2077, opened 2026-05-13
[Platform][Inworld] Add bridge for TTS and STT

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | --
| License       | MIT

Introduces a dedicated bridge package (symfony/ai-inworld-platform) that exposes Inworld's synthesize-speech, synthesize-speech (streaming) and transcribe endpoints through Symfony AI's platform abstraction.

The catalog reads available models dynamically from the `/llm/v1alpha/models` endpoint, classifying entries as TTS or STT based on their input/output modalities. Authentication uses a JWT Bearer token attached via `ScopingHttpClient`. Streaming TTS is wired through `NdjsonStream`, decoding base64 chunks into `BinaryDelta` instances.

The ai-bundle exposes the bridge through a new `inworld:` platform configuration node, mirroring the ElevenLabs wiring with `jwt` replacing `api_key`. The resulting `ai.platform.inworld` service is tagged `ai.platform.speech` so it can be used as `text_to_speech_platform` or `speech_to_text_platform` in the SpeechAgent decoration.

Tests cover routing, payload shape, error extraction, streaming and catalog resolution at PHPStan level 10 on the bridge package, plus full DI compilation coverage and a detailed scenario test in AiBundleTest.

### DIFF
--- a/deptrac.yaml
+++ b/deptrac.yaml
@@ -189,6 +189,10 @@ deptrac:
       collectors:
         - type: classLike
           value: Symfony\\AI\\Platform\\Bridge\\HuggingFace\\.*
+    - name: InworldPlatform
+      collectors:
+        - type: classLike
+          value: Symfony\\AI\\Platform\\Bridge\\Inworld\\.*
     - name: LmStudioPlatform
       collectors:
         - type: classLike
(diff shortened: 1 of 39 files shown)


chr-hertel:
The JWT is created in backend using the API key, isn't it? why do we configure the JWT is it's dynamic

Guikingone:
Nice catch, need to rework this part, I'll put the PR in Draft to prevent issues while I'm fixing it 👍🏻

PR #2078, opened 2026-05-13
[Platform][Bifrost] Introduce platform

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | Following an internal presentation by @welcoMattic 
| License       | MIT

### DIFF
--- a/docs/bundles/ai-bundle.rst
+++ b/docs/bundles/ai-bundle.rst
@@ -56,6 +56,10 @@ Advanced Example with Multiple Agents
                 default: ~
                 eu:
                     bedrock_runtime_client: 'async_aws.client.bedrock_runtime_eu'
+            bifrost:
+                # self-hosted OpenAI-compatible LLM gateway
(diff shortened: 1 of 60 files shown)


chr-hertel:
So do i get it correct, that they not only mimick embeddings and completions api, but also whisper and dall-e?

Guikingone:
Yes, the whole API covers a lot of existing endpoints, it act as a "proxy" for other platforms, maybe we can extract some logic from here for other platforms 🤔

docs/bundles/ai-bundle.rst
```diff
@@ -1303,6 +1355,7 @@ When only STT is configured (no TTS), the agent returns the same result type as
 
     Handling both speech-to-text and text-to-speech introduces latency as most of the process is synchronous.
 
+.. _`Bifrost`: https://docs.getbifrost.ai/
```
welcoMattic: to be removed from here

PR #2087, opened 2026-05-15
[Platform][Generic] Encode request body with `JSON_UNESCAPED_SLASHES`

### LINKED ISSUE #1973: [Platform][Generic] Forward slashes in model names are JSON-escaped, breaking some OpenAI-compatible backends
### Summary

`Bridge\Generic\Completions\ModelClient::request()` passes the payload via Symfony HttpClient's `json` option. `HttpClientTrait::jsonEncode()` encodes without `JSON_UNESCAPED_SLASHES`, so a model name like `qwen/qwen3.5-4b` goes on the wire as `"model":"qwen\/qwen3.5-4b"`.

That is valid JSON, but several OpenAI-compatible backends (observed with vLLM and LM Studio) do a raw string comparison of the model key at some point in their routing/lookup and fail to match the escaped form, returning a "model not found" style error for any namespaced model name containing `/`.

### Reproduce

1. Run any OpenAI-compatible server that registers a model under a name containing `/` (e.g. `cyanwiki/Qwen3.5-35B-A3B-AWQ-4bit` on vLLM, any `org/model` on LM Studio).
2. Use the Generic platform bridge with that exact model name.
3. Request fails with a model-not-found style error on the backend.

### Suggested fix

In `Bridge\Generic\Completions\ModelClient::request()` (and the matching `Embeddings\ModelClient`), encode the body explicitly with `JSON_UNESCAPED_SLASHES` and pass it as `body` instead of `json`:

```php
'body' => json_encode(
    array_merge($options, $payload),
    \JSON_THROW_ON_ERROR
    | \JSON_UNESCAPED_SLASHES
    | \JSON_HEX_TAG | \JSON_HEX_APOS | \JSON_HEX_AMP | \JSON_HEX_QUOT
    | \JSON_PRESERVE_ZERO_FRACTION,
),
```

The non-slash flags mirror Symfony `HttpClientTrait::jsonEncode()` so the only intentional deviation is `JSON_UNESCAPED_SLASHES`.

Happy to send a PR.

### Context

- Symfony AI 0.5.0, PHP 8.3
- Tested against vLLM 0.6.x and LM Studio
- Worked around downstream with an HttpClient decorator, but it really belongs in the Generic bridge since every namespaced HF-style model name hits this.

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #1973
| License       | MIT

The Generic completions and embeddings `ModelClient` classes passed the request payload via the HttpClient `json` option. Symfony's `HttpClientTrait::jsonEncode()` encodes without `JSON_UNESCAPED_SLASHES`, so a namespaced model name like `qwen/qwen3.5-4b` was sent on the wire as `"model":"qwen\/qwen3.5-4b"`.

That is valid JSON, but several OpenAI-compatible backends (observed with vLLM and LM Studio) do a raw string comparison of the model key during routing and fail to match the escaped form, returning a "model not found" error for any namespaced model name containing `/`.

Both clients now encode the body explicitly with `JSON_UNESCAPED_SLASHES` and pass it as `body`. The remaining flags mirror `HttpClientTrait::jsonEncode()` so the only intentional deviation is the unescaped slashes.

Tests added to both `ModelClientTest` classes asserting forward slashes in model names are not escaped.

### DIFF
--- a/src/platform/src/Bridge/Generic/Completions/ModelClient.php
+++ b/src/platform/src/Bridge/Generic/Completions/ModelClient.php
@@ -54,10 +54,17 @@ class ModelClient implements ModelClientInterface
         // endpoints, which reject unknown request body fields with a 400 error.
         unset($options['cacheRetention']);
 
+        // Encode the body explicitly with JSON_UNESCAPED_SLASHES instead of relying on the
+        // HttpClient "json" option: its encoder escapes forward slashes, so a namespaced
+        // model name like "qwen/qwen3.5-4b" goes on the wire as "qwen\/qwen3.5-4b", which
+        // some OpenAI-compatible backends fail to match. The remaining flags mirror
+        // Symfony's HttpClientTrait::jsonEncode() defaults.
+        $body = json_encode(array_merge($options, $payload), \JSON_THROW_ON_ERROR | \JSON_UNESCAPED_SLASHES | \JSON_HEX_TAG | \JSON_HEX_APOS | \JSON_HEX_AMP | \JSON_HEX_QUOT | \JSON_PRESERVE_ZERO_FRACTION);
+
         return new RawHttpResult($this->httpClient->request('POST', $this->baseUrl.$this->path, [
             'auth_bearer' => $this->apiKey,
             'headers' => ['Content-Type' => 'application/json'],
-            'json' => array_merge($options, $payload),
+            'body' => $body,
         ]));
     }
 }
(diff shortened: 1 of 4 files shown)


chr-hertel:
Not sure about the impact here ... :thinking:

chr-hertel:
Closing since unable to reproduce the initial error.

PR #2088, opened 2026-05-15
[Store] Add maxDepth to RstToctreeLoader for scoped indexing

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | -
| License       | MIT

Adds an optional `maxDepth` constructor parameter to `RstToctreeLoader` (default `null` = unlimited, preserving current behavior) to cap toctree recursion depth.

This makes the `rag/symfony-docs.php` example fast for regression runs: it now reads optional `SYMFONY_DOCS_ENTRY` / `SYMFONY_DOCS_MAX_DEPTH` env vars, defaulting to the full recursive docs indexation. The example runner auto-scopes that example to the small `quick_tour` subtree, so the full example sweep no longer takes ages while still exercising recursive toctree indexing.

cc @wachterjohannes - JFYI, trying to speed up the `./runner` across all examples here

### DIFF
--- a/src/store/src/Document/Loader/RstToctreeLoader.php
+++ b/src/store/src/Document/Loader/RstToctreeLoader.php
@@ -29,7 +29,11 @@ final class RstToctreeLoader implements LoaderInterface
         private RstLoader $rstLoader = new RstLoader(),
         private bool $throwOnMissingEntry = false,
         private LoggerInterface $logger = new NullLogger(),
+        private ?int $maxDepth = null,
     ) {
+        if (null !== $maxDepth && $maxDepth < 0) {
+            throw new InvalidArgumentException(\sprintf('The maximum depth must be a non-negative integer or null, %d given.', $maxDepth));
+        }
     }
 
     /**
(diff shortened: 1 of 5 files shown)


src/store/src/Document/Loader/RstToctreeLoader.php
```diff
@@ -25,11 +25,19 @@
  */
 final class RstToctreeLoader implements LoaderInterface
 {
+    /**
+     * @param ?int $maxDepth Maximum toctree recursion depth, or null for unlimited. A value of 0 loads only the
```
wachterjohannes: i think we can remove this - the name is already good enough. if we keep it we should at least add also the other parameters to the phpdoc

wachterjohannes (changes requested):
The rest LGTM

src/store/src/Document/Loader/RstToctreeLoader.php
chr-hertel: ha! was unsure as well - let's remove! thx!

PR #2095, opened 2026-05-18
[Chat][Cache] Require generic PSR-6, instead of Symfony Cache

10n:
Added PSR cache requirement and suggestion.
The cache bridge is actually using generic `Psr\Cache\CacheItemPoolInterface`

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Docs?         | yes
| Issues        | n/a
| License       | MIT

Update component requirement to indicate the actual dependency to PSR-6 `psr/cache`, not Symfony Cache `symfony/cache`.
I used the same version that is required by the `symfony/cache`

### DIFF
--- a/src/chat/src/Bridge/Cache/composer.json
+++ b/src/chat/src/Bridge/Cache/composer.json
@@ -26,8 +26,11 @@
     ],
     "require": {
         "php": ">=8.2",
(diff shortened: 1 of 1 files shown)


Guikingone (approved):
Nice catch 👍🏻

PR #2099, opened 2026-05-18
[Platform][Anthropic] Allow overriding `tool_choice` from caller options

### LINKED ISSUE #2098: [Platform][Anthropic] Allow overriding `tool_choice` from caller options
`Bridge/Anthropic/ModelClient` overwrites `tool_choice` on every request that includes tools:

```php
$options['tool_choice'] = ['type' => 'auto'];
```

Anything I pass through `Platform::invoke($model, $input, $options)` is dropped before the request goes out. Anthropic supports three modes (`auto`, `any`, `{type: 'tool', name: '...'}`); only the first is reachable from the bundle.

### Why it matters

The default `auto` is wrong whenever the caller already knows which tool should run. A common case: several tools are exposed, but for this turn you want a deterministic lookup tool to fire, not the model deciding between it and a fuzzier search:

```php
    'tools' => [$lookupTool, $searchTool, $cancelOrderTool],
    'tool_choice' => ['type' => 'tool', 'name' => 'lookup'],
]);
```

It's also the only way to reach Anthropic's `{type: 'any'}` mode (must call some tool, model picks which), which has no equivalent path through the bundle today. Tool-call evals depend on it too: verifying that a model picks the *right* tool when free to choose requires a forced-pick baseline to compare against, and you can't generate that baseline without overriding `tool_choice`.

Without the override, callers have to drop down to a raw Anthropic HTTP client to get this behaviour, losing everything Platform and Agent provide on top.

### Proposal

Stop clobbering a caller-supplied value:

```php
$options['tool_choice'] ??= ['type' => 'auto'];
```

Default unchanged for callers that don't set it. No new public API, no config.

---

Refs: [Anthropic docs on forcing tool use](https://docs.anthropic.com/en/docs/build-with-claude/tool-use/overview#forcing-tool-use)

4lxndr:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | Fix #2098
| License       | MIT

The Anthropic `ModelClient` and Bedrock `ClaudeModelClient` overwrote any caller-supplied `tool_choice` with `{type: 'auto'}` as soon as `tools` was present, so `{type: 'any'}` and `{type: 'tool', name: '...'}` were unreachable from `Platform::invoke()`.

### DIFF
--- a/src/platform/src/Bridge/Bedrock/Anthropic/ClaudeModelClient.php
+++ b/src/platform/src/Bridge/Bedrock/Anthropic/ClaudeModelClient.php
@@ -76,7 +76,7 @@ final class ClaudeModelClient implements ModelClientInterface
         unset($payload['model']);
 
         if (isset($options['tools'])) {
-            $options['tool_choice'] = ['type' => 'auto'];
+            $options['tool_choice'] ??= ['type' => 'auto'];
         }
 
         if (isset($options['response_format'])) {
--- a/src/platform/src/Bridge/Anthropic/CHANGELOG.md
+++ b/src/platform/src/Bridge/Anthropic/CHANGELOG.md
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+0.10
+----
+
+ * Allow overriding `tool_choice` via caller options instead of always forcing `['type' => 'auto']`
+
 0.9
 ---
 
--- a/src/platform/src/Bridge/Anthropic/ModelClient.php
+++ b/src/platform/src/Bridge/Anthropic/ModelClient.php
@@ -62,7 +62,7 @@ final class ModelClient implements ModelClientInterface
         $payload = $this->injectCacheControl($payload);
 
         if (isset($options['tools'])) {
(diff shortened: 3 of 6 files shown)


src/platform/src/Bridge/Bedrock/Anthropic/ClaudeModelClient.php
```diff
@@ -76,7 +76,7 @@ public function request(Model $model, array|string $payload, array $options = []
         unset($payload['model']);
 
         if (isset($options['tools'])) {
-            $options['tool_choice'] = ['type' => 'auto'];
+            $options['tool_choice'] ??= ['type' => 'auto'];
```
4lxndr, line 79: Same PR or another one with different commit (Bedrock) prefix?

OskarStark: Same PR please

chr-hertel (approved):

chr-hertel:
Thank you @4lxndr.

PR #2101, opened 2026-05-18
[Mate][Platform] Make AGENTS.md the canonical AI-agent guidance, slim CLAUDE.md

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | yes
| Issues        | n/a
| License       | MIT

Follow-up to #2054, applying the same pattern more broadly.

### `[Platform]` Make AGENTS.md canonical, reduce CLAUDE.md to import

`src/platform/AGENTS.md` now carries the canonical, general AI-agent guidance for the Platform component, and `src/platform/CLAUDE.md` is reduced to a header plus an ``@AGENTS.md`` Claude Code import.

The two bits that previously lived only in `CLAUDE.md` are moved into `AGENTS.md` so nothing is lost:

- the `src/EventListener/` directory entry (with `TemplateRendererListener`)
- the mixed-content `UserMessage` template example

No Claude Code-specific section was added — unlike Mate (which is itself an MCP server consumed by Claude Code), Platform is a plain library with nothing dev-tool-specific to call out. The `Bridge/ClaudeCode/` bridge is just one bridge among many and is covered in `AGENTS.md` like the others.

### `[Mate]` Drop redundant Claude Code Specifics section

The `## Claude Code Specifics` section in `src/mate/CLAUDE.md` introduced by #2054 turns out to be redundant on closer reading:

- **"Mate is consumed as an MCP server / `mcp.json` auto-pickup"** describes how end users wire Mate into Claude Code at runtime, not how Claude Code helps you develop on the Mate codebase. That's user-facing docs material (README / INSTRUCTIONS.md), not dev-tool guidance.
- **"sources of truth for available MCP tools"** is already covered in `AGENTS.md` under `## Agent Instructions Materialization`, and the guidance applies to any AI agent — not specifically to Claude Code.

After this PR `src/mate/CLAUDE.md` is just header + ``@AGENTS.md`` import, matching the shape adopted for Platform here.


wachterjohannes:
@chr-hertel when we accept to adopt this pattern i will create an additional PR for all the other components with the same pattern

chr-hertel:
but do we really need this if there are no claude specifics?

wachterjohannes:
> but do we really need this if there are no claude specifics?

yes because claude.md is the entry point of claude code BUT others use agents.md

so to make sure all xosing agents load the instructiona we have to keep both

wachterjohannes:
@chr-hertel friendly ping — could you take a look at this when you get a chance? Thanks!

chr-hertel (approved):

chr-hertel:
Thank you @wachterjohannes.

wachterjohannes:
@chr-hertel should we also bring that to the other components?

PR #1947, opened 2026-04-16
[Platform][VertexAi] Add Gemini 3 and 3.1 flash

NielsAI:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | Fix 1946
| License       | MIT

This PR add supports for `gemini-3-flash-preview` and `gemini-3.1-flash-lite-preview`

https://docs.cloud.google.com/vertex-ai/generative-ai/docs/models/gemini/3-flash
https://docs.cloud.google.com/vertex-ai/generative-ai/docs/models/gemini/3-1-flash-lite

PR #1906, opened 2026-04-09
[Platform][Gemini][VertexAI] Remove redundant EventSourceHttpClient wrapping

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | 
| License       | MIT

The `SseStream` class introduced in #1827 internally creates its own `EventSourceHttpClient` for streaming. This makes the explicit `EventSourceHttpClient` wrapping in the Gemini and VertexAI `ModelClient` constructors and `PlatformFactory` classes redundant.

This PR:
- Removes `EventSourceHttpClient` wrapping from `Gemini\Gemini\ModelClient` and `VertexAi\Gemini\ModelClient`, accepting plain `HttpClientInterface` instead
- Replaces `EventSourceHttpClient` wrapping in both `PlatformFactory` classes with `$httpClient ?? HttpClient::create()`
- Adds missing `ModelClientTest` and `PlatformFactoryTest` for the Gemini bridge
- Adds a streaming integration test to `ResultConverterTest` that verifies end-to-end SSE streaming through `RawHttpResult` → `SseStream` → `ResultConverter`

PR #2102, opened 2026-05-19
[Platform][VertexAi] Wrap list-array tool responses for Struct-proto compatibility

peter-si:
| Q             | A
| ------------- | ---
| Branch?       | main
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | -
| License       | MIT

## Summary

Wraps list-array tool responses in a `{items: [...]}` object so they pass Gemini's Struct-proto validation on `functionResponse.response`. Without this, any tool that returns a JSON list 500s the whole turn with `Proto field is not repeating, cannot start list`.

## Why

`google.protobuf.Struct` is a JSON-object shape — it cannot represent a sequential JSON array. The current normalizer wraps scalars as `{rawResponse: <value>}` (good) but passes lists through unchanged (broken). Every tool that returns rows / records / promos / search results / etc. fails.

Concrete reproducer: a tool method returning `[{name: 'A'}, {name: 'B'}]` produces `functionResponse.response = [{name: 'A'}, {name: 'B'}]`, which Vertex rejects.

## What changes

In `Symfony\AI\Platform\Bridge\VertexAi\Contract\ToolCallMessageNormalizer`:

- Associative arrays → pass through (unchanged)
- **List arrays → wrapped as `{items: [...]}`** (new)
- Scalars / null → wrapped as `{rawResponse: <value>}` (unchanged)

Drive-bys while in the file: extracted `decodeContent()` and `asStruct()` for readability, added a class-level docblock describing the Struct quirk, and dropped a no-op `array_filter()` on the outer payload (it never had anything to filter).

## Backwards compatibility

The output shape **changes for tools that returned a JSON list**, but those tools currently 500 with the Struct-proto error against Gemini, so there are no working consumers depending on the broken shape. A `[BC BREAK]` entry has been added to the `ai-vertex-ai-platform` CHANGELOG under `0.10`.

## Tests

New `src/platform/tests/Bridge/VertexAi/Contract/ToolCallMessageNormalizerTest` covers associative passthrough, list-wrap, empty-list-wrap, scalar wrap, non-JSON content, null content, and tool-name preservation. 9 tests, 9 assertions, all green.

> Note: `PHPStan / Component / Agent` is currently red on this branch due to a pre-existing `Variable $sourceCollection might not be defined.` issue in `Agent\Toolbox\Toolbox::call()` — verified independently against `upstream/main`. Fix submitted as #2103; once that merges, a rebase here will clear the failure.

### DIFF
--- a/src/platform/src/Bridge/VertexAi/CHANGELOG.md
+++ b/src/platform/src/Bridge/VertexAi/CHANGELOG.md
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+0.10
+----
+
+ * Fix: `ToolCallMessageNormalizer` now wraps list-array tool responses as `{items: [...]}` so they pass Gemini's `Struct`-proto validation on `functionResponse.response`. Tools that returned a JSON list previously caused the whole turn to fail with `Proto field is not repeating, cannot start list`; associative arrays and scalar/null wrap (`{rawResponse: ...}`) behavior is unchanged.
+
 0.9
 ---
 
--- a/src/platform/src/Bridge/VertexAi/Contract/ToolCallMessageNormalizer.php
+++ b/src/platform/src/Bridge/VertexAi/Contract/ToolCallMessageNormalizer.php
@@ -17,6 +17,19 @@ use Symfony\AI\Platform\Message\ToolCallMessage;
 use Symfony\AI\Platform\Model as BaseModel;
 
 /**
+ * Normalizes tool-call responses into Gemini's `functionResponse` shape.
+ *
+ * Gemini declares `functionResponse.response` as a `google.protobuf.Struct`,
+ * which represents a JSON object. The tool result is JSON-decoded and then
+ * coerced into Struct-compatible shape:
+ *
+ *  - associative array → passed through unchanged
+ *  - list-array        → wrapped as `{items: [...]}` (Struct rejects sequential arrays)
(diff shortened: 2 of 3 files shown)


peter-si:
Heads up on red CI here:

- **Fabbot / Check PR header** — fixed in this push (title was missing the `[Platform]` prefix, now `[Platform][VertexAi] ...` per the convention used in #1906 / #1938 / #1947).
- **PHPStan / Component / Agent** — this is a **pre-existing failure on `main`** (verified by running `phpstan analyse src/agent` against `upstream/main`). `Toolbox::call()` triggers `Variable $sourceCollection might not be defined.` because PHPStan can't correlate the two `instanceof HasSourcesInterface` checks. It happens to surface on every PR that runs Agent's PHPStan, not just this one.

I've opened #2103 with the two-line fix. Once that merges, the failure here will clear after a rebase. Let me know if you'd prefer this PR to depend on / include that fix instead.

chr-hertel:
Hi @peter-si,
thanks for raising this, how did you end up with a tool result like `[{name: 'A'}, {name: 'B'}]`? so like, what is the return value of the tool?

I would expect a tool result to be more structured, and that's why we have the `Symfony\AI\Agent\Toolbox\ToolResultConverter` dealing with this here

chr-hertel (approved):
Sorry for the delay - this one slipped through the cracks.

chr-hertel:
Thank you @peter-si.

PR #2107, opened 2026-05-21
[AI Bundle] Use bundle InvalidArgumentException in config validators

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | -
| License       | MIT

Small consistency cleanup in `src/ai-bundle/config/options.php`:

- Replace four `throw new \InvalidArgumentException(...)` with the bundle's `Symfony\AI\AiBundle\Exception\InvalidArgumentException`. The bundle exception extends `\InvalidArgumentException`, so existing `catch (\InvalidArgumentException)` blocks still match.
- Replace two `empty()` calls (in the agent/service XOR validator and the duplicate-name validator) with explicit checks (`isset(...) && '' !== ...` and `[] !== $duplicates`), per the project rule against `empty()`.

No behavior change beyond the swapped exception class. Existing validation tests in `AiBundleTest` still pass.

### DIFF
--- a/src/ai-bundle/config/options.php
+++ b/src/ai-bundle/config/options.php
@@ -11,6 +11,7 @@
 
 namespace Symfony\Component\Config\Definition\Configurator;
 
+use Symfony\AI\AiBundle\Exception\InvalidArgumentException;
 use Symfony\AI\Platform\Capability;
 use Symfony\AI\Platform\Model;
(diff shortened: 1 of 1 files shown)


chr-hertel (approved):

chr-hertel:
Thank you @wachterjohannes.

PR #2108, opened 2026-05-21
[Platform][Failover] Use project RuntimeException in tests instead of \Exception

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | -
| License       | MIT

Test-only cleanup. `FailoverPlatformTest` was using raw `new \Exception(...)` in 22 places to simulate platform failures inside mock callbacks. Replace with `Symfony\AI\Platform\Exception\RuntimeException`, which is already imported in the same file (used by `expectException()` assertions).

No assertion currently checks for `\Exception::class`, so this is a drop-in. All 11 tests in `src/platform/src/Bridge/Failover/Tests/` continue to pass.


chr-hertel (approved):

chr-hertel:
Thank you @wachterjohannes.

PR #2109, opened 2026-05-21
[Agent][Store] Bring outlier classes in line with project final/readonly conventions

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | -
| License       | MIT

Two small consistency cleanups across Agent and Store bridges:

**`readonly` on Wikipedia bridge constructor (`src/agent/src/Bridge/Wikipedia/Wikipedia.php`)**

Every other Agent HTTP-client bridge — Brave, Firecrawl, Mapbox, Ollama, OpenMeteo, Scraper, SerpApi, Tavily, Youtube — already declares its promoted constructor parameters as `readonly`. Wikipedia was the only outlier.

**`final` on four Store implementations**

- `src/store/src/Bridge/ClickHouse/Store.php`
- `src/store/src/Bridge/Redis/Store.php`
- `src/store/src/Bridge/SurrealDb/Store.php`
- `src/store/src/InMemory/Store.php`

Every other `Store` bridge in the project is already `final`. These four were missed. I checked the monorepo for `extends ...\Store` against each of these classes — no subclasses exist.

Note that `SurrealDb\Store` retains its non-readonly `$authenticationToken` field, which is set lazily by the `authenticate()` method; `final` does not affect that.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,3 +1,12 @@
+UPGRADE FROM 0.9 to 0.10
+========================
+
+Store
+-----
+
+ * The `ClickHouse`, `Redis`, `SurrealDb`, and `InMemory` store implementations are now `final`.
+   Extending them is no longer possible; compose or decorate `StoreInterface` instead.
+
 UPGRADE FROM 0.8 to 0.9
 =======================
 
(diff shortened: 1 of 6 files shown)


chr-hertel (approved):
Thanks @wachterjohannes - added an UPGRADE.md item

chr-hertel:
Thank you @wachterjohannes.

PR #2113, opened 2026-05-23
[Platform] Add PartialJsonParser for streaming structured output

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | -
| License       | MIT

This PR adds `Symfony\AI\Platform\StructuredOutput\Streaming\PartialJsonParser`, a static, dependency-free utility that recovers the largest valid structure from an incomplete JSON string. This unblocks rendering partial objects from streamed structured output before the model finishes emitting them.

### Algorithm

`PartialJsonParser::parse()` attempts a strict `json_decode` first; on failure it applies best-effort fixes in order and tries again:

1. Strip trailing commas before `]` / `}`.
2. Close unclosed strings by tracking `"` and `\` byte-wise (UTF-8 safe because both are single-byte).
3. Fix incomplete values at the tail:
   - dangling colon → append `null`
   - partial literals (`tru`, `fals`, `nul`, ...) → complete them
   - trailing comma at EOF → strip it
4. Close unclosed `{` / `[` by walking the string with a stack that ignores chars inside strings.

If still invalid, returns `null` and populates the `&$errorMessage` reference with `json_last_error_msg()`.

### Usage

```php
use Symfony\AI\Platform\StructuredOutput\Streaming\PartialJsonParser;

$buffer = '';
foreach ($chunks as $chunk) {
    $buffer .= $chunk;
    $partial = PartialJsonParser::parse($buffer);

    if (null !== $partial) {
        // render the partial structure
    }
}
```

### Notes

- Ported from `ModelflowAi\Chat\Util\Json\OptimisticJsonParser`
- No platform wiring changes — follow-up PRs can wire it into `StreamResult` / structured-output streaming.
- `composer test`, `vendor/bin/phpstan analyse`, and `php-cs-fixer` pass cleanly on the new files.

### DIFF
--- a/examples/platform/partial-json-parser.php
+++ b/examples/platform/partial-json-parser.php
@@ -1,0 +1,43 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use Symfony\AI\Platform\StructuredOutput\Streaming\PartialJsonParser;
+
+require_once dirname(__DIR__).'/bootstrap.php';
+
+// Simulates the progressive deltas a chat completion would emit when producing
+// structured JSON output. Each chunk is appended to a running buffer and we try
+// to render the largest valid structure we can recover so far.
+$chunks = [
+    '{"title": "Symfony AI", ',
+    '"tags": ["php", "llm",',
+    ' "agents"], "author": {"name": "Fa',
+    'bien", "twitter": "@fabpot"',
+    '}, "released": tru',
+    'e}',
+];
+
+$buffer = '';
+
+foreach ($chunks as $index => $chunk) {
+    $buffer .= $chunk;
+
+    $partial = PartialJsonParser::parse($buffer, $error);
+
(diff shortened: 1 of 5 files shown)


chr-hertel:
So, i thought about this before and if we want to have some black magic here - was always wondering tho if there's even a use-case for combining streaming and structured output, but never came across that - so looks like you're one step ahead here, what's your use-case here?

wachterjohannes:
the idea is to load multiple more or less different stuff in the same request. in case of sulu.ai we are loading sources (json sttucture of different pages) and the answer in the same structure.

this makes it mich faster (see the widget on sulu.io as the implementation example). in the follow ups you will see then more DX stuff to use that as easy as possible

examples/platform/partial-json-parser.php
chr-hertel, line 1: is it too early to bring in a use-case based example with this PR - would skip examples for code lib users should use - `PartialJsonParser` is rather internal, right?

wachterjohannes: yes - it could be used to build build something - but yes its mostly internal - should i remove the example?

wachterjohannes: @chr-hertel will replace that in a follow up with something "more" usefull

chr-hertel:
Thank you @wachterjohannes.

PR #2114, opened 2026-05-23
[AI Bundle][Platform] Add result text normalizer chain

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | -
| License       | MIT

This PR adds a normalizer pipeline that cleans the textual content of `TextResult` (and its parts inside `MultiPartResult`) after a model call. The fence/Unicode quirks are model-output problems, not provider problems, so the wiring lives in `platform` and applies across every bridge.

## What ships

`Symfony\AI\Platform\Result\Normalizer\TextNormalizerInterface`:

```php
public function supports(Model $model, ResultInterface $result, array $options): bool;
public function normalize(string $text): string;
```

`supports()` receives the original `$options` so a normalizer can react to e.g. `response_format`.

Two implementations:

- `JsonFenceStripNormalizer` — strips ```` ```json … ``` ```` (and bare ```` ``` … ``` ````) wrappers when the fenced body is valid JSON. Active when the request asks for `response_format: json` / `json_schema`, or when the converted result is already an `ObjectResult`.
- `UnicodeNormalizer` — removes Unicode `Cf` (format control) characters such as zero-width spaces and BOM, and converts non-breaking space (`U+00A0`) to a regular space. Always active.

`PlatformSubscriber` listens to `ResultEvent` at priority `10` (before structured-output processing) and walks the registered normalizers in order. Non-text results pass through untouched; `MultiPartResult` parts are normalized individually.

`TextResult` gets a `withContent()` helper that returns a new instance with replaced content, preserving signature and metadata. Streaming results are intentionally out of scope for this PR.

## AI Bundle wiring

The bundle ships both normalizers and the subscriber as services and auto-tags any `TextNormalizerInterface` implementation with `ai.platform.result_normalizer`. No user-facing config — normalizers are enabled by default.

### DIFF
--- a/src/ai-bundle/config/services.php
+++ b/src/ai-bundle/config/services.php
@@ -72,6 +72,9 @@ use Symfony\AI\Platform\EventListener\TemplateRendererListener;
 use Symfony\AI\Platform\Message\TemplateRenderer\ExpressionLanguageTemplateRenderer;
 use Symfony\AI\Platform\Message\TemplateRenderer\StringTemplateRenderer;
 use Symfony\AI\Platform\Message\TemplateRenderer\TemplateRendererRegistry;
+use Symfony\AI\Platform\Result\Normalizer\JsonFenceStripNormalizer;
+use Symfony\AI\Platform\Result\Normalizer\PlatformSubscriber as NormalizerPlatformSubscriber;
+use Symfony\AI\Platform\Result\Normalizer\UnicodeNormalizer;
 use Symfony\AI\Platform\StructuredOutput\PlatformSubscriber;
 use Symfony\AI\Platform\StructuredOutput\ResponseFormatFactory;
 use Symfony\AI\Platform\StructuredOutput\ResponseFormatFactoryInterface;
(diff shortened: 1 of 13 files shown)


src/ai-bundle/config/services.php
```diff
@@ -209,6 +212,17 @@
             ])
             ->tag('kernel.event_subscriber')
 
+        // result normalizers
+        ->set('ai.platform.result_normalizer.json_fence_strip', JsonFenceStripNormalizer::class)
```
wachterjohannes, line 216: not sure if we should add a configuration for that or if that should be default initialized

chr-hertel:
How did you end up in this? feels quite model specific to me - wondering if it's rather docs/cookbook instead of lib code, maybe as use-case for result event? WDYT?

wachterjohannes:
To answer the question first: this came out of sulu.ai a while back, with smaller models that have no structured output support. And you were half right about the rest, I can now say which half.

**The Unicode part is exactly what you suspected.** `UnicodeNormalizer` strips every `\p{Cf}` character from every text result, by default, on every model:

```
family emoji (ZWJ)   "👨‍👩‍👧"           -> "👨👩👧"
Devanagari (ZWNJ)    "क्‌ष"             -> "क्ष"
Hebrew with RLM      "שלום‏ (10)"       -> "שלום (10)"
Arabic with ALM      "العربية؜5"        -> "العربية5"
soft hyphen          "Sil­ben­trennung"  -> "Silbentrennung"
```

Bidi marks control rendering, ZWJ holds emoji sequences and Indic ligatures together. Five of my seven test cases were collateral damage, two were the quirk I was after. That half belongs in userland or the cookbook, not on by default.

**The fence part is not cosmetic though.** On current main:

```
plain json   -> ObjectResult {"a":1}
fenced json  -> RuntimeException: Cannot json decode the content.
```

Any model without native structured output can produce the second one, which is where this started. But you are right that this PR is the wrong shape for it. The fix belongs in `StructuredOutput\ResultConverter::convertTextToObject()`, where the decode happens, the intent is already known and the try/catch already exists. No chain, no interface, no extra listener.

**The chain also does not work on main anymore.** `ResultConvertedEvent` landed in 1a548b81a, together with the note in `Provider.php` that a listener resolving the result eagerly converts before `onConvert`/`onError` are registered. `Result\Normalizer\PlatformSubscriber::onResult()` calls `$deferred->getResult()`, so with it registered nothing dispatches those events anymore, and every invocation resolves eagerly:

```
converter invoked during ResultEvent dispatch: 1
ResultConvertedEvent hook fired:               no
same run without the subscriber:               yes
```

So: closing this, and opening a small PR for the fence handling in the converter instead. Your instinct in June was the right one.

wachterjohannes:
Opened #2478 for the fence part, in the converter rather than as a chain. Closing here.

PR #2110, opened 2026-05-21
[Platform] Remove unmatched PHPStan ignore in JsonSchema Schema attribute

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | -
| License       | MIT

PHPStan no longer reports `function.alreadyNarrowedType` on the `array_filter($enum, ...) !== $enum` check in `Schema::__construct`, so the `@phpstan-ignore-next-line` annotation above it has itself become an error (`ignore.unmatchedLine`, non-ignorable). This currently fails the `PHPStan / Component / Platform` CI job on `main` and on every open PR.

This PR drops the now-unneeded annotation. Verified locally:

```
$ vendor/bin/phpstan analyse --no-progress
 [OK] No errors
```

JsonSchema-related PHPUnit tests still pass (116 tests, 130 assertions).

PR #2115, opened 2026-05-23
[Platform] Add DeferredResult::asPartialJsonStream() for streaming structured output

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | -
| License       | MIT

> [!NOTE]
> Depends on #2113. Merge after that one or rebase this PR once it lands.

Follow-up to #2113. The `PartialJsonParser` introduced there is a static utility — useful, but it leaves consumers to wire up the streaming buffer themselves. This PR adds `DeferredResult::asPartialJsonStream()`, which feeds the running text buffer into the parser and yields each new partial value, so the caller doesn't have to touch `Object` denormalization or build their own listener.

### Behavior

- Accumulates `TextDelta`s from `asTextStream()` into a single buffer.
- Re-parses the buffer with `PartialJsonParser` after every delta.
- Yields the recovered value (array / scalar / object) only when it differs from the previous one — so whitespace-only deltas don't produce duplicate snapshots.
- Silently swallows buffers that are not recoverable yet (parser sets `\$errorMessage`).

### Usage

```php
$deferred = $platform->invoke('gpt-5-mini', $messages, ['stream' => true]);

foreach ($deferred->asPartialJsonStream() as $partial) {
    // render the partial structure (array/object/scalar)
}
```

The return type is intentionally `mixed` (matches the parser) — this PR stays at the "raw JSON" layer. Typed object denormalization on top of the partial stream is a separate follow-up.

### Notes

- New example at `examples/platform/partial-json-stream.php` prompts a model for a structured book object via streaming output and renders each snapshot as it grows.
- Docs extended under the existing "Parsing Partial JSON from Streams" section.
- CHANGELOG entry added.
- `composer test` and `vendor/bin/phpstan analyse` on the platform component pass (the existing unmatched-ignore in `Schema.php` is unrelated and tracked in #2110).


chr-hertel (approved):
this example is already pretty impressive 👍

chr-hertel:
Thank you @wachterjohannes.

PR #2117, opened 2026-05-26
[Platform][Anthropic] Support cache_control on system prompt block

### LINKED ISSUE #2111: [Platform][Anthropic] Add cache_control on system prompt to enable caching when no tools are used
Hi there,

I'm having an "issue" with prompt caching and not sure if it's an intentional behavior or something is wrong in my way of using it.
On the other hand, I could have prompt caching working when using usual agent configured in my ai.yaml config file, but it's a little bit different.

### Description

When using `Symfony\AI\Platform\PlatformInterface::invoke()` directly (without an `AgentInterface`) with an Anthropic platform configured with `cache_retention: 'short'`, prompt caching does **not** kick in when:

- No tools are passed
- The last user message is short (e.g. a few tokens like `"Translate to French"`)
- Even though the **system prompt** is very large (several thousand tokens)

Anthropic returns `cache_creation_input_tokens: 0` and `cache_read_input_tokens: 0`, despite the total prompt being well above Claude Haiku 4.5's 2048-token minimum cacheable prefix.

### Reproduction

```php

// System prompt is ~3000 tokens (big static instructions + a long reference text).
$systemPrompt = '... long static system prompt ...';

$bag = new MessageBag(
    Message::forSystem($systemPrompt),
    Message::ofUser('Translate to French'),
);

$response = $platform->invoke('claude-haiku-4-5-20251001', $bag, ['max_tokens' => 2000]);
$usage = $response->getMetadata()->get('token_usage');

// $usage->getPromptTokens() === 3465
// $usage->getCacheCreationTokens() === 0  // ← expected > 0
// $usage->getCacheReadTokens() === 0
```

### Root cause analysis

`src/platform/src/Bridge/Anthropic/Contract/MessageBagNormalizer.php` serializes the system message as a **plain string**:

```php
if (null !== $system = $data->getSystemMessage()) {
    $array['system'] = $system->getContent();
}
```

`src/platform/src/Bridge/Anthropic/ModelClient.php::injectCacheControl()` then adds the `cache_control` marker on the last user message in `$payload['messages']`, but never touches `$payload['system']`.

Per the Anthropic API spec, the `system` field accepts either a `string` **or** an iterable of `TextBlockParam`, and only the block form can carry a `cache_control` annotation. When sent as a plain string, the system prompt cannot become its own cache breakpoint.

When the request also contains tools, `injectToolsCacheControl()` creates a cache breakpoint after the tools, which produces a large `[system+tools]` cached prefix — caching works (this is the typical agent path with tool definitions).

When tools are absent and the only `cache_control` is on the last user message (a few tokens), the cacheable prefix ends up dominated by the un-blocked `system` string. In practice Anthropic returns 0 for both cache fields for this configuration on Haiku 4.5.

### Expected behavior

When `cacheRetention !== 'none'` and a system message is present, `MessageBagNormalizer` (or `ModelClient::injectCacheControl()`) should send the system field in block form with a `cache_control` annotation, e.g.:

```php
$array['system'] = [
    [
        'type' => 'text',
        'text' => $system->getContent(),
        'cache_control' => ['type' => 'ephemeral'],  // or with 'ttl' => '1h' for 'long'
    ],
];
```

This would create an explicit cache breakpoint after the system prompt, mirroring what's already done for tools and for the last user message.

### Environment

- `symfony/ai-bundle`: v0.9.0
- `symfony/ai-anthropic-platform`: v0.9.0
- `symfony/ai-agent`: v0.9.0
- Model: `claude-haiku-4-5-20251001`
- Configuration:
  ```yaml
  ai:
      platform:
          anthropic:
              api_key: '%env(ANTHROPIC_API_KEY)%'
              cache_retention: 'short'
  ```

### Question
Am I doing something wrong ?
Is the current behavior intentional? 

Thanks a lot for your support and amazing work.

Cheers !

4lxndr:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | Fix #2111
| License       | MIT

- MessageBagNormalizer now emits the system prompt as a typed content-block array ([{type: text, text: ...}]) instead of a bare string, so a cache marker can be attached. Anthropic accepts both shapes, so callers see no difference.
- ModelClient attaches a cache_control marker to the system block when cacheRetention is short/long (skipped for none), mirroring the existing tool-definition caching.

### DIFF
--- a/src/platform/src/Bridge/Anthropic/ModelClient.php
+++ b/src/platform/src/Bridge/Anthropic/ModelClient.php
@@ -60,6 +60,7 @@ final class ModelClient implements ModelClientInterface
         ];
 
         $payload = $this->injectCacheControl($payload);
+        $payload = $this->injectSystemCacheControl($payload);
 
         if (isset($options['tools'])) {
             $options['tool_choice'] = ['type' => 'auto'];
@@ -119,6 +120,29 @@ final class ModelClient implements ModelClientInterface
     }
 
     /**
+     * Injects a prompt-caching marker on the last system content block.
+     *
+     * The system prompt is typically the largest and most stable region of a
+     * request, making it the single most effective caching target. This creates
+     * a cache breakpoint after the system block so it can be cached independently
+     * of the tools and messages that follow.
+     *
+     * @param array<string, mixed> $payload
+     *
+     * @return array<string, mixed>
+     */
+    private function injectSystemCacheControl(array $payload): array
+    {
+        if ('none' === $this->cacheRetention || !isset($payload['system']) || !\is_array($payload['system']) || [] === $payload['system']) {
+            return $payload;
+        }
+
+        $payload['system'][\count($payload['system']) - 1]['cache_control'] = $this->getCacheControl();
+
+        return $payload;
+    }
+
+    /**
      * Injects prompt-caching markers into the normalised message payload.
      *
      * Anthropic prompt caching requires a {"cache_control": {"type": "ephemeral"}}
@@ -140,9 +164,7 @@ final class ModelClient implements ModelClientInterface
             return $payload;
         }
 
-        $cacheControl = 'long' === $this->cacheRetention
-            ? ['type' => 'ephemeral', 'ttl' => '1h']
-            : ['type' => 'ephemeral'];
+        $cacheControl = $this->getCacheControl();
 
         for ($i = \count($messages) - 1; $i >= 0; --$i) {
             if ('user' !== ($messages[$i]['role'] ?? '')) {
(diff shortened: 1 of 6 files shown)


src/platform/src/Bridge/Anthropic/ModelClient.php
```diff
@@ -118,6 +119,33 @@ private function injectToolsCacheControl(array $tools): array
...
+     */
+    private function injectSystemCacheControl(array $payload): array
+    {
+        if ('none' === $this->cacheRetention || !isset($payload['system']) || !\is_array($payload['system']) || [] === $payload['system']) {
+            return $payload;
+        }
+
+        $cacheControl = 'long' === $this->cacheRetention
+            ? ['type' => 'ephemeral', 'ttl' => '1h']
+            : ['type' => 'ephemeral'];
```
chr-hertel: how about extracting this into `getCacheControl(): array` for synergy with `injectCacheControl` usage?

chr-hertel (approved):
will patch while merging, thanks @4lxndr!

PR #2122, opened 2026-05-28
[Platform][VertexAi] Use SSE format for streaming to avoid partial-JSON parsing

### LINKED ISSUE #2121: [Platform][VertexAi] Streaming fails with JsonException when chunks split JSON objects
| Q             | A
| ------------- | ---
| Bug?          | yes
| Platform      | symfony/ai-platform (VertexAi bridge)

When streaming responses through the VertexAi Gemini bridge, `aiplatform.googleapis.com` returns a chunked JSON array by default. HTTP chunk boundaries don't align with JSON object boundaries, so `SseStream::stream()` ends up calling `json_decode()` on a fragment and throws.

Stack trace:

```
#0 vendor/symfony/ai-platform/src/Result/Stream/SseStream.php(51): json_decode()
#1 vendor/symfony/ai-vertex-ai-platform/Gemini/ResultConverter.php(100): Symfony\AI\Platform\Result\Stream\SseStream->stream()
#2 vendor/symfony/ai-platform/src/Result/StreamResult.php(51): Symfony\AI\Platform\Bridge\VertexAi\Gemini\ResultConverter->convertStream()
#3 src/Controller/.../PostMessageController.php: Symfony\AI\Platform\Result\StreamResult->getContent()
```

#1988 fixed the same root cause on the Gemini bridge by appending `?alt=sse` to the streaming URL — Google's API then returns proper `text/event-stream`. The VertexAi bridge builds its own URL in `Bridge/VertexAi/Gemini/ModelClient.php` and was not covered by that change (called out explicitly in #1988's compatibility note).

Same one-line fix needed on the VertexAi side.

ottaviano:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #2121, same root cause as #1184, mirrors #1988
| License       | MIT

#1988 fixed partial-JSON parsing errors on the Gemini bridge by asking Google's API for SSE via `?alt=sse`. The compatibility note in that PR explicitly mentioned VertexAi was unaffected because it builds its own URLs — but the same JSON-chunk-splitting issue applies (stack trace in #2121), so the fix needs to be ported.

### Change

Append `alt=sse` to the query string of `streamGenerateContent` requests in `Bridge/VertexAi/Gemini/ModelClient.php`. The VertexAi client already passes a `query` array to the HTTP client (for the optional `key`), so the change is a single entry added when streaming.

### Compatibility

- Non-streaming path (`stream: false`) unchanged
- API-key path (`?key=…`) still works — both query params are merged

### DIFF
--- a/src/platform/src/Bridge/VertexAi/Gemini/ModelClient.php
+++ b/src/platform/src/Bridge/VertexAi/Gemini/ModelClient.php
@@ -45,7 +45,8 @@ final class ModelClient implements ModelClientInterface
      */
     public function request(BaseModel $model, array|string $payload, array $options = []): RawHttpResult
     {
-        $method = $options['stream'] ?? false ? 'streamGenerateContent' : 'generateContent';
+        $isStream = $options['stream'] ?? false;
+        $method = $isStream ? 'streamGenerateContent' : 'generateContent';
 
         if (null !== $this->location && null !== $this->projectId) {
             $url = \sprintf(
@@ -66,6 +67,9 @@ final class ModelClient implements ModelClientInterface
         $query = [];
         if (null !== $this->apiKey) {
             $query['key'] = $this->apiKey;
+        }
+        if ($isStream) {
+            $query['alt'] = 'sse';
         }
 
         if (isset($options[PlatformSubscriber::RESPONSE_FORMAT]['json_schema']['schema'])) {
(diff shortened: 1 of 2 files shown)


chr-hertel (approved):

chr-hertel:
Thank you @ottaviano.

PR #2128, opened 2026-05-30
[Platform] HTTP-boundary record/replay cassettes for bridge tests

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | -
| License       | MIT

`Test\Replay\CassetteHttpClient` is an `HttpClientInterface` you pass to any bridge `Factory`. Cassette missing: it performs the live request and writes a JSON cassette, credentials redacted. Cassette present: it replays the recorded bytes as a real `MockResponse`, so the real Contract, `ModelClient` and `ResultConverter` run with no network. Delete the cassette to re-record.

**Why.** Bridge tests assert against hand-written `MockResponse` fixtures, which cannot drift-check themselves: when a provider changes its response shape, the fixture keeps passing and the converter breaks in production. A cassette surfaces that change as a diff on the next re-record.

**Pilot.** `Bridge/Mistral/Tests/Llm/ResultConverterCassetteTest` drives the real Mistral `ResultConverter` from cassettes recorded against api.mistral.ai, covering `stop` and `length` finishes, a tool call and a real `400`. Recording them already corrected the hand-written fixture from the previous revision: Mistral sends `"content": ""` and an explicit `"tool_calls": null`, not `"content": null`. Two mocks stay in `ResultConverterTest` for cases a provider cannot be asked to produce, each with the reason in its docblock.

This is the tool plus a worked example, not the migration. The remaining bridges are follow-up work, ideally by whoever holds that provider's API key.

**Details.** FIFO replay, like `MockHttpClient` with an array of responses. Streams keep their raw SSE body, so the bridge stream parser re-frames them on replay exactly as on the wire. Credentials and per-request trace headers are stripped on write, rate limiting headers are kept because the converters read them. It ships in `src/Test/` rather than `tests/` because bridges are standalone packages and can only use what `symfony/ai-platform` ships, same as `InMemoryPlatform`.

**Not `@internal`.** symfony/symfony#63781 (`RecorderHttpClient`) is the same idea upstream, but it targets 8.2 while this component requires `symfony/* ^7.3|^8.0`, which is too far out to wait for. So it ships as normal API now and gets swapped for the upstream one later through a deprecation.

Related: chr-hertel/ai#36 builds the same HTTP-boundary recorder plus an `examples/`-driven golden-stdout layer. Its SSE handling and header redaction are folded in here; the examples layer is the distinct half and can land on top.

### DIFF
--- a/docs/components/platform.rst
+++ b/docs/components/platform.rst
@@ -1498,6 +1498,25 @@ This platform can also be configured when using the bundle::
 Testing Tools
 -------------
 
+The component ships more than one test double, and which one fits depends on *how much of the
+platform the test should keep running*. They cut the pipeline at different depths, from replacing
+everything down to replacing nothing but the network:
+
+* :class:`Symfony\\AI\\Platform\\Test\\InMemoryPlatform` replaces the whole platform. Routing, model
+  catalog, contract, ``ModelClient`` and ``ResultConverter`` are all skipped, and the answer is a
+  string or closure you write. Use it to test *your* code against a given answer.
+* :class:`Symfony\\AI\\Platform\\Test\\MockPlatformFactory` keeps the real ``Platform``, ``Provider``,
+  routing and contract, and fakes only the ``ModelClient`` and ``ResultConverter``. Use it when the
+  test is about the platform itself: model routing and resolution, non-text result types, or
+  asserting on the payload the platform built.
+* :class:`Symfony\\AI\\Platform\\Test\\Replay\\CassetteHttpClient` replaces nothing but the network.
+  Contract, ``ModelClient`` and ``ResultConverter`` all run for real, against bytes recorded from the
+  provider once. Use it when the test is about a bridge's internals.
+
+The lower a tool cuts, the more of the library a test actually covers, and the more setup it needs.
+A second axis runs across that: the first two return an answer you wrote by hand, while a cassette
+returns what a provider really sent, which is what makes it drift-checkable.
+
 For unit or integration testing, you can use the :class:`Symfony\\AI\\Platform\\Test\\InMemoryPlatform`,
 which implements :class:`Symfony\\AI\\Platform\\PlatformInterface` without calling external APIs.
 
@@ -1654,6 +1673,48 @@ explicit models instead::
     ]));
     // $provider->supports('mock-model') === true
     // $provider->supports('gpt-4o') === false
+
+Recording Real Responses
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+To exercise a bridge's *internals* (its Contract normalizers, ``ModelClient`` payload building and
+``ResultConverter``) against realistic data without a network, record a real HTTP response once and
+replay it offline. :class:`Symfony\\AI\\Platform\\Test\\Replay\\CassetteHttpClient` is an
+``HttpClientInterface`` you pass to any real bridge ``Factory``; because replay serves a real
+``MockResponse``, the **real** converter runs on replay (unlike the mocks above, which bypass it).
+
+By default the mode follows the cassette file: record when it is missing, replay when it exists
+(override with an explicit ``record:`` argument). Record once with a real API key, commit the
+generated cassette, and replay it in CI::
+
+    use Symfony\AI\Platform\Bridge\Mistral\Factory;
+    use Symfony\AI\Platform\Message\Message;
+    use Symfony\AI\Platform\Message\MessageBag;
+    use Symfony\AI\Platform\Test\Replay\CassetteHttpClient;
+    use Symfony\AI\Platform\Test\Replay\HttpCassette;
+    use Symfony\Component\HttpClient\HttpClient;
+
+    $cassette = new HttpCassette(__DIR__.'/fixtures/mistral_chat.json');
+
+    // cassette missing (+ real key) -> hits the live API and writes the cassette (secrets redacted)
+    // cassette exists -> serves the recorded response; the real Mistral ResultConverter runs
+    $http = new CassetteHttpClient($cassette, HttpClient::create());
+
+    $platform = Factory::createPlatform($apiKey, $http);
+    $result = $platform->invoke('mistral-large-latest', new MessageBag(Message::ofUser('Hello')));
+
+    echo $result->asText(); // produced by the real converter from the recorded bytes
+
+Recorded interactions replay first-in-first-out (like ``MockHttpClient`` with an array of responses).
+A streamed response is stored with its raw Server-Sent Event body, so the bridge's stream parser frames
+it on replay exactly as it would on the wire, while headers describing the live transfer
+(``content-length``, ``content-encoding``, ...) are dropped because they would contradict the replayed
+body. Credentials (``Authorization``, ``x-api-key``, ``x-goog-api-key``, the ``auth_bearer`` shorthand,
+cookies and provider account identifiers) are replaced with ``[redacted]`` in both request and response
+headers before the cassette is written, so a cassette is safe to commit. Per-request trace headers
+(``date``, ``cf-ray``, correlation and request ids, proxy latencies) are dropped on write, so that
+re-recording a cassette produces a diff of what the provider actually changed instead of noise;
+rate limiting headers are kept, because the converters read them.
 
 Code Examples
 ~~~~~~~~~~~~~
--- a/src/platform/src/Bridge/Mistral/Tests/Llm/ResultConverterCassetteTest.php
+++ b/src/platform/src/Bridge/Mistral/Tests/Llm/ResultConverterCassetteTest.php
@@ -1,0 +1,100 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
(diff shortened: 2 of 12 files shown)


wachterjohannes:
@OskarStark symfony/symfony#63781 RecorderHttpClient is the same idea at the same layer. As this PR would only be an improvement to the current test-setup we could postpone that OR merge this as something internal, use it in the bridges of this Repo and replace as soon as the RecorderHttpClient is landed (and we bump to this version).

wachterjohannes:
Rebased on main and green. I'd like to get this in soon — not for its own sake, but because it unblocks the bigger piece: the bridge tests are still built on hand-written fixtures that drift from what providers actually return, which is exactly where the result-converter bugs keep coming from. With a recorder at the HTTP boundary we can start refactoring the platform tests onto recorded payloads and have them exercise the real Contract → `ModelClient` → `ResultConverter` path.

Explicitly interim: symfony/symfony#63781 (`RecorderHttpClient`) is the same idea at the same layer upstream, so once that lands and we bump the requirement we swap our recorder out for it. The tests written against cassettes stay.

docs/components/platform.rst
```diff
@@ -1655,6 +1655,58 @@ explicit models instead::
...
+body. Credentials — ``Authorization``, ``x-api-key``, ``x-goog-api-key``, the ``auth_bearer`` shorthand,
+cookies and provider account identifiers — are replaced with ``[redacted]`` in both request and response
+headers before the cassette is written, so a cassette is safe to commit. Per-request trace headers
+(``date``, ``cf-ray``, correlation and request ids, proxy latencies) are dropped on write, so that
+re-recording a cassette produces a diff of what the provider actually changed instead of noise;
+rate limiting headers are kept, because the converters read them.
+
+.. note::
+
+    This requires `symfony/http-client` (the ``CassetteHttpClient`` builds on ``MockHttpClient``).
```
chr-hertel: unnecessary context I'd say - the entire library requires http-client

```suggestion
```

docs/components/platform.rst
```diff
@@ -1655,6 +1655,58 @@ explicit models instead::
...
+
+.. note::
+
+    Both classes are marked ``@internal``: they exist to move this repository's own bridge tests off
+    hand-written fixtures and are meant to be replaced by ``RecorderHttpClient`` from
+    ``symfony/http-client`` once that is available. Tests written against cassettes stay valid.
+
```
chr-hertel: if it's internal we should document the usage - but is it though?

chr-hertel:
one thing next to the @internal confusing here:
with #2128 and #2129 we end up with 
* `InMemoryPlatform`
* A set of `MockSomething` implementations
* Recording on Provider
* Recording on Http Level

i'm afraid everything has its justification, but we should streamline the docs when to use what maybe - or do you catch something we should remove in this set of implementations?

src/platform/src/Bridge/Mistral/Tests/Llm/ResultConverterCassetteTest.php
```diff
@@ -0,0 +1,100 @@
...
+     */
+    private function replay(string $cassette, string $prompt, array $options = [], string|Mistral $model = 'mistral-large-latest'): DeferredResult
+    {
+        $platform = Factory::createPlatform('test-key', new CassetteHttpClient(
+            new HttpCassette(__DIR__.'/../Fixtures/cassettes/'.$cassette),
+            record: false,
+        ));
+
+        return $platform->invoke($model, new MessageBag(Message::ofUser($prompt)), $options);
+    }
```
chr-hertel, line 99: good candidate for a `use PlatformReplayTrait` or similar - follow up :)

src/platform/src/Bridge/Mistral/Tests/Llm/ResultConverterTest.php
```diff
@@ -14,15 +14,17 @@
...
 use Symfony\AI\Platform\Result\RawHttpResult;
-use Symfony\AI\Platform\Result\TextResult;
 use Symfony\Component\HttpClient\MockHttpClient;
 use Symfony\Component\HttpClient\Response\JsonMockResponse;
 
+/**
+ * What is left once the provider payloads move to {@see ResultConverterCassetteTest}: a check on
+ * our own model matching, plus the two error paths that cannot reasonably be recorded - see the
+ * reason on each of them.
+ */
```
chr-hertel: only makes sense in context of refactoring

```suggestion
```

docs/components/platform.rst
wachterjohannes: Fair point! I was unsure about this myself, and you're right that `@internal` plus a public docs section is contradictory.

Dropped the `@internal` on both classes. `RecorderHttpClient` targets 8.2 while we require `symfony/* ^7.3|^8.0`, so waiting for it is too far out, we need this now, and it is meant to be used, both here and by anyone testing a bridge downstream. When the upstream one becomes usable, replacing ours is a normal deprecation path rather than a silent removal. The reasoning is in the PR description now as well.

The `symfony/http-client` note above is gone too.

src/platform/src/Bridge/Mistral/Tests/Llm/ResultConverterTest.php
wachterjohannes: Right, that one only describes a migration state, removed. Kept the two per-method notes though: those explain why *that specific case* stays a mock (a ~640 KB filler prompt, and a 500 you cannot ask a provider for), which stays true independently of the refactoring.

src/platform/src/Bridge/Mistral/Tests/Llm/ResultConverterCassetteTest.php
wachterjohannes: Agreed on the follow-up. One note on the shape, so whoever picks it up does not build the wrong thing.

A trait can only carry `new CassetteHttpClient(new HttpCassette($path), record: false)`. Everything else in this helper is bridge specific: the `Factory` call, the default model, the cassette directory. So it saves two lines per bridge and costs an abstract `cassetteDirectory()` in return, which is a wash.

What would actually pay off is a base class in the style of the existing `Test\ModelCatalogTestCase`, with `createPlatform(HttpClientInterface $httpClient)` and `cassetteDirectory()` as the extension points and a `replay()` that does the invoke. That takes roughly 15 lines off this test.

I would rather derive that shape from two call sites than from one though, now that these classes are no longer `@internal` and the signature becomes a BC concern. The next migrated bridge will show whether `invoke()` is the right entry point at all, since embeddings and streaming may not fit the same signature. So: follow-up together with the second bridge migration, and a base class rather than a trait.

wachterjohannes:
Good question, and I think the honest answer is that all four are needed, but the docs never said why, which is exactly what made the set look redundant. They are not four takes on the same idea: each one cuts the pipeline at a different depth, and what a test covers is whatever stays above the cut.

| tool | what it replaces | what still runs for real | the answer is |
| --- | --- | --- | --- |
| `InMemoryPlatform` | the whole `PlatformInterface` | nothing | written by hand |
| `RecordingProvider` (#2129) | one `ProviderInterface` | `Platform`, routing, model catalog | recorded, as a `ResultInterface` |
| `MockPlatformFactory` + `Mock*` | `ModelClient` and `ResultConverter` | `Platform`, routing, catalog, `Contract` | written by hand |
| `CassetteHttpClient` (#2128) | the network, nothing else | `Contract`, `ModelClient`, `ResultConverter` | recorded, as raw bytes |

So there are two axes, not one: how deep the cut sits, and where the answer comes from. That second axis is why the two recorders are not just "the mocks with extra steps": a hand written answer cannot drift-check itself, a recorded one can.

What each one would cost us if it went away:

* **`InMemoryPlatform`**: the most common case by far is an agent test that just needs the model to say something (`AgentTest`, `RunnerTest`, `StreamingAgentToolCallTest` all use it). Without it, those have to assemble a provider, a catalog and routing first, for a test that cares about none of it.
* **`Mock*`**: it is the only one that can assert on what the platform *built*. `MockModelClient::getCalls()` is how you test tool option translation, merged model options and multi-provider routing. A cassette cannot fill that hole, because you would need a recorded live call per routing permutation, and routing bugs are not payload bugs.
* **`RecordingProvider`**: for a test that only needs a realistic answer, the alternatives are hand writing it (tedious and drift-prone for `VectorResult` or `ObjectResult`) or dropping to the HTTP layer and caring about a bridge's wire format that the test has no business knowing about. It is also the only provider agnostic one: swap the provider behind it and the cassette shape still holds.
* **`CassetteHttpClient`**: without it nothing exercises contract payloads, converters, error mapping and SSE framing against what a provider actually sends, which is where the bugs we keep fixing live.

Where I do see real overlap, since you asked whether something should go: `InMemoryPlatform` and `MockPlatformFactory` sit closest together, both hand written and both above the wire, separated only by how much of the platform stays real. And to be fair about it, `Mock*` currently has no consumer in this repo besides its own tests, so its case rests on capability rather than on usage today. If we ever collapse two of these, that is the pair to look at, not the recorders. I would not do it in this PR though.

On the docs point, you are right and that was the actual problem, so I pushed a short "which one do I pick" overview to the top of the *Testing Tools* section, ordering the tools by cut depth with a one line "use it when" each. It covers the three that exist today; #2129 slots its row in above `MockPlatformFactory` when it lands.

chr-hertel (approved):

chr-hertel:
Thank you @wachterjohannes.

PR #2129, opened 2026-05-30
[Platform] Add provider-level record/replay cassettes for test mocks

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | -
| License       | MIT

Sibling of #2128, which records at the HTTP boundary. This one records a layer higher, at the provider, for tests that just need a realistic answer and do not care how the bridge produced it.

`Test\Recording\RecordingProvider` decorates a real `ProviderInterface`. Record when the cassette file is missing (invokes the provider, serializes its finished `ResultInterface`), replay when it exists (rebuilds the result, never calls the provider, needs no API key). Delete the cassette to re-record, or force the mode with `record:`.

```php
$provider = new RecordingProvider(
    Factory::createProvider($_SERVER['ANTHROPIC_API_KEY'] ?? 'no-key-needed-on-replay'),
    new Cassette(__DIR__.'/fixtures/weather.json'),
);
```

Covers text, `ObjectResult`, `VectorResult`, `ToolCallResult` and text streams. Interactions are matched by a signature over `(model, input, options)` and consumed FIFO; the signature function is injectable, so an input that varies between runs (timestamps, ids, retrieved context) can be normalized instead of forcing a re-record (thanks @MikiBuilder).

Two scope notes: on replay the recorded result is returned verbatim, so the bridge `ResultConverter` only runs at record time, and result metadata plus token usage are not preserved yet.

Rebased on main now that #2128 is merged, so this adds its own row to the "which test double do I pick" list that came with it, between `InMemoryPlatform` and `MockPlatformFactory`, which is where it cuts.

### DIFF
--- a/src/platform/src/Test/Recording/ResultSerializer.php
+++ b/src/platform/src/Test/Recording/ResultSerializer.php
@@ -1,0 +1,144 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Platform\Test\Recording;
+
+use Symfony\AI\Platform\Exception\InvalidArgumentException;
+use Symfony\AI\Platform\Result\ObjectResult;
+use Symfony\AI\Platform\Result\ResultInterface;
+use Symfony\AI\Platform\Result\Stream\Delta\TextDelta;
+use Symfony\AI\Platform\Result\StreamResult;
+use Symfony\AI\Platform\Result\TextResult;
+use Symfony\AI\Platform\Result\ToolCall;
+use Symfony\AI\Platform\Result\ToolCallResult;
+use Symfony\AI\Platform\Result\VectorResult;
+use Symfony\AI\Platform\Vector\Vector;
+
+/**
+ * Converts a {@see ResultInterface} to and from a JSON-serializable array for {@see Cassette}.
+ *
+ * Supported result types (v1): text, object, vector, tool call, and text-delta streams. Result
+ * metadata and token usage are not preserved. Unsupported result or delta types throw.
+ *
+ * @author Johannes Wachter <johannes@sulu.io>
+ */
+final class ResultSerializer
+{
+    /**
+     * @return array<string, mixed>
+     */
+    public static function toArray(ResultInterface $result): array
+    {
+        if ($result instanceof TextResult) {
+            return [
+                'type' => 'text',
+                'content' => $result->getContent(),
+                'signature' => $result->getSignature(),
+            ];
+        }
+
+        if ($result instanceof ObjectResult) {
+            $content = $result->getContent();
+
+            return [
+                'type' => 'object',
+                'is_object' => \is_object($content),
+                'content' => json_encode($content, \JSON_THROW_ON_ERROR),
+            ];
+        }
+
+        if ($result instanceof VectorResult) {
+            return [
+                'type' => 'vector',
+                'vectors' => array_map(static fn (Vector $vector): array => $vector->getData(), $result->getContent()),
+            ];
+        }
+
+        if ($result instanceof ToolCallResult) {
+            return [
+                'type' => 'toolcall',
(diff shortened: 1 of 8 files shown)


OskarStark:
You should check Symfony repo there is an open pull request for cassettes like this IIRC

wachterjohannes:
> You should check Symfony repo there is an open pull request for cassettes like this IIRC


@OskarStark you mean symfony/symfony#63781 (RecorderHttpClient, php‑vcr‑style HAR record/replay)? Good catch. That one records at the HTTP boundary, so it overlaps with the sibling #2128, not this PR.

This RecordingProvider sits a layer higher: it records the already‑converted ResultInterface returned by a ProviderInterface, so it's bridge‑agnostic, needs no http-client, and on replay it skips the converter entirely. Different trade‑off, complementary to a HTTP‑level recorder. I'll add a note to the PR cross‑referencing #63781 so the layering is clear.

wachterjohannes:
Thanks, all three are concrete and worth having.

**Signature matching**: agreed, and it was not injectable at all. Pushed a `signature` constructor argument, so volatile parts can be normalized before hashing instead of forcing a re-record.

**Leaking**: this one lands on #2128 rather than here. A provider cassette stores the model, a hash of the input and the result, never the input itself, so there is nothing to redact on the request side and the mismatch trap you describe cannot bite. A recorded answer can of course still repeat back what the prompt contained, which the docs now say. #2128 does commit the request body verbatim, so I will take the point over there.

**Token usage**: agreed on the value, but drift detection cannot live at this layer, since replay returns the already converted result and never runs the converter. That job belongs to #2128, where the raw bytes are kept. Preserving metadata here is still worth doing for tests that want to assert on usage, and it is out of this PR only because metadata holds arbitrary objects and needs its own serialization pass.

Happy to look at anything from `llm-vcr`, the normalization rules especially.

wachterjohannes:
@MikiBuilder tracked the two open points as issues so they do not get lost with this PR: #2409 for redacting recorded prompts and #2410 for keeping metadata and token usage. Both credit your review, feel free to take either one.

src/platform/src/Test/Recording/ResultSerializer.php
```diff
@@ -0,0 +1,144 @@
...
+
+/**
+ * Converts a {@see ResultInterface} to and from a JSON-serializable array for {@see Cassette}.
+ *
+ * Supported result types (v1): text, object, vector, tool call, and text-delta streams. Result
+ * metadata and token usage are not preserved. Unsupported result or delta types throw.
+ *
+ * @author Johannes Wachter <johannes@sulu.io>
+ */
+final class ResultSerializer
```
chr-hertel, line 33: reminds me of `Symfony\AI\Platform\Bridge\Cache\ResultNormalizer` somehow :D

i wonder - also from conceptual point of view - how much of overlap or potential synergy we have with the caching bridge

wachterjohannes: Bigger overlap than it looks: four of the five types I serialize are already in there, and it additionally covers `BinaryResult`, `ChoiceResult`, `MultiPartResult` and `ThinkingResult`. The one thing it deliberately refuses is `StreamResult`, which is exactly what a recording needs.

What blocks reuse today is the direction: `ResultNormalizer` lives in the `symfony/ai-cache-platform` bridge, `Test\Recording` lives in the core package, so the core would have to depend on a bridge. The way out is one result (de)normalizer in the platform core that both use, which is a class move out of the bridge, so a BC break with a deprecation alias and an `UPGRADE.md` entry.

I would rather do that as its own PR than grow this one into it, and then this serializer collapses into a thin wrapper or disappears. Tracked as #2413. Fine either way, but it should not ride along here.

chr-hertel (approved):

chr-hertel:
Thank you @wachterjohannes.

wachterjohannes:
@MikiBuilder thank you for the review here. The signature point changed the API rather than just the docs, and your triage of where the leak actually applies was sharper than mine.

Ping me any time while you work on #2409, whether it is about the recorder internals, where to hook in, or getting a PR through the repo's checks. Happy to help.

PR #2130, opened 2026-05-30
[Platform] Add ModelCard with pricing, region and extra metadata to model catalog

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no 
| Issues        | -
| License       | MIT

Adds an optional, **objective** metadata layer to model catalog entries so a `Model`
can answer "what do you cost?" and "where do you run?". This is the data layer that a
cost tracker (token usage → money) and a criteria router (filter by region/tier) build
on; both are out of scope here.

### What's added

- `ModelCatalog\ModelPricing` — immutable VO with per-1M-token rates, cache-aware
  (`cachedInputPerMillion` / `cacheWritePerMillion`, matching `TokenUsage`'s cache
  counters), `currency` defaulting to `USD`.
- `ModelCatalog\ModelCard` — immutable VO holding platform-blessed facts (`pricing`,
  `region`) plus an open, app-owned `extra` bag (`get()` / `all()`). Platform code never
  branches on `extra` contents.
- `Model::getCard(): ?ModelCard` — new optional 4th constructor slot, defaulting to
  `null`.
- `AbstractModelCatalog` parses an optional `metadata` key per entry and attaches the
  card.

### Usage

```php
// In a ModelCatalog's $models array:
'gpt-x' => [
    'class'        => Gpt::class,
    'capabilities' => [Capability::INPUT_TEXT, Capability::TOOL_CALLING],
    'metadata'     => [
        'pricing' => ['input' => 5.0, 'output' => 30.0, 'cachedInput' => 0.5],
        'region'  => 'us',
        'extra'   => ['tier' => 'flagship'],
    ],
],

// Consuming it:
$card = $catalog->getModel('gpt-x')->getCard();
$card?->getPricing()?->getInputPerMillion(); // 5.0
$card?->getRegion();                          // 'us'
$card?->get('tier');                          // 'flagship'
```

### Notes

- **Fully backward compatible** — `metadata` is optional and `getCard()` returns `null`
  when absent; no existing catalog entry needs editing.
- The getter is named `getCard()` (not `getMetadata()`) on purpose: the platform already
  has a generic mutable `Metadata\Metadata` bag exposed via `getMetadata(): Metadata` on
  every Result/Message (runtime side-channel data). A `ModelCard` is a static catalog
  descriptor, so it gets its own name to avoid overloading that convention.
- **Mechanism only** — no real bridge pricing data is shipped; apps populate their own
  catalogs. Model subclasses that override `__construct` (Claude, OpenAi/Gemini/Scaleway
  Embeddings, Scaleway, Voyage) forward the new argument so the card isn't silently
  dropped; a regression test guards this through a subclass.

Possible follow-ups (separate PRs): wiring a pricing source such as `ModelsDev`,
populating real bridge data, the cost tracker, and the criteria router.

### DIFF
--- a/src/platform/CHANGELOG.md
+++ b/src/platform/CHANGELOG.md
@@ -10,6 +10,7 @@ CHANGELOG
  * Add optional `signature` field to `Message\Content\Text`, `Result\ToolCall`, and `Result\TextResult` for provider-scoped signatures (currently used by Gemini/Vertex AI for `thoughtSignature` round-trip).
  * Memoize conversion failures in `DeferredResult::getResult()` so subsequent calls re-throw the cached exception instead of re-running the converter
  * Surface tool calls executed by the `ClaudeCode` and `Codex` CLI bridges as `ToolCallResult` parts of a `MultiPartResult`, mirroring the inference-API behavior of the Anthropic and Gemini bridges
+ * Add `ModelCatalog\ModelCard` and `ModelCatalog\ModelPricing`, exposed via `Model::getCard()`, with an optional `metadata` key in `ModelCatalog` entries (pricing, region, and an open `extra` bag)
 
 0.8
 ---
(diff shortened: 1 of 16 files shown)


wachterjohannes:
Closing to keep the public monorepo clean while this is still staged. Continued on the fork at wachterjohannes/symfony-ai#46.

PR #2126, opened 2026-05-29
[Platform] Add command and workflow to auto-update model catalogs

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | -
| License       | MIT

Refreshing the bridge `ModelCatalog` static lists by hand (one PR per model, e.g. https://github.com/symfony/ai/pull/2124) is tedious.

This adds a `SingleCommandApplication` at `src/platform/dev/update-model-catalogs.php` that regenerates the lists from upstream metadata:

- **Per-provider bridges** (Anthropic, OpenAi, Gemini, Mistral, DeepSeek, Cohere, Perplexity, Cerebras) from [models.dev](https://models.dev/api.json), whose model ids match the catalog keys. Capabilities are derived with the existing `Symfony\AI\Platform\Bridge\ModelsDev\CapabilityMapper`.
- **OpenRouter** from the OpenRouter API, reusing the modality mapping already in `Symfony\AI\Platform\Bridge\OpenRouter\ModelApiCatalog`.

Design notes:

- **Add-only**: hand-curated entries win, only genuinely new models are appended. Removals and capability corrections stay manual, so a data refresh never silently overwrites a tuned list.
- Only the content between `// STATIC LIST START` / `// STATIC LIST END` markers is rewritten (markers are added on first run); the rest of each class is untouched.
- `RetryableHttpClient` guards the unattended fetch against transient failures.
- Meta (`meta-llama` absent from models.dev) and Voyage are not sourceable and stay manual; the command logs a warning when a provider key is missing.

A weekly GitHub Actions workflow (`.github/workflows/update-model-catalogs.yaml`, also `workflow_dispatch`) runs the command, applies PHP-CS-Fixer, runs the catalog tests, and opens a PR as the `github-actions` bot for review.

Usage:

```
php src/platform/dev/update-model-catalogs.php --dry-run        # preview
php src/platform/dev/update-model-catalogs.php                  # write all
php src/platform/dev/update-model-catalogs.php Anthropic        # one bridge
```

PR #2145, opened 2026-06-08
 Declare batched input support for OpenAI embeddings

### LINKED ISSUE #2127: [Platform] Declare INPUT_MULTIPLE capability for OpenAI embedding models
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | -
| License       | MIT

OpenAI's `/v1/embeddings` endpoint natively accepts an array `input` and returns a parallel array of embeddings in a single response ([OpenAI API reference](https://platform.openai.com/docs/api-reference/embeddings/create)). The `Embeddings\ModelClient` already forwards array payloads correctly:

  ```php
  // src/platform/src/Bridge/OpenAi/Embeddings/ModelClient.php
  'json' => array_merge($options, [
      'model' => $model->getName(),
      'input' => $payload,
  ]),
```

However, the OpenAI `ModelCatalog` does not declare `Capability::INPUT_MULTIPLE` for any embedding model. As a result, `Symfony\AI\Store\Document\Vectorizer::vectorizeStrings()` falls through to the sequential branch and issues one HTTP request per input, even when the caller passes an array:

```php
  // src/store/src/Document/Vectorizer.php
  if ($this->platform->getModelCatalog()->getModel($this->model)->supports(Capability::INPUT_MULTIPLE)) {
      // single batched call — never reached for OpenAI embeddings today
  } else {
      foreach ($stringValues as $string) {
          $results[] = $this->platform->invoke($this->model, $string, $options); // N sequential calls
      }
  }
```

liuzemei:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #2127
| License       | MIT

OpenAI embedding models accept an array input and the existing model client forwards array payloads correctly. This adds `Capability::INPUT_MULTIPLE` to the OpenAI embedding catalog entries so vectorization can use a single batched request for multiple strings.

Tests:

```
./vendor/bin/phpunit Tests/ModelCatalogTest.php
./vendor/bin/phpunit Tests/Embeddings/ModelClientTest.php
./vendor/bin/phpunit
./vendor/bin/phpstan analyse --no-progress --memory-limit=1G
```

### DIFF
--- a/src/platform/src/Bridge/OpenAi/ModelCatalog.php
+++ b/src/platform/src/Bridge/OpenAi/ModelCatalog.php
@@ -312,15 +312,15 @@ final class ModelCatalog extends AbstractModelCatalog
             ],
             'text-embedding-ada-002' => [
                 'class' => Embeddings::class,
-                'capabilities' => [Capability::INPUT_TEXT, Capability::EMBEDDINGS],
+                'capabilities' => [Capability::INPUT_TEXT, Capability::INPUT_MULTIPLE, Capability::EMBEDDINGS],
             ],
             'text-embedding-3-large' => [
                 'class' => Embeddings::class,
-                'capabilities' => [Capability::INPUT_TEXT, Capability::EMBEDDINGS],
+                'capabilities' => [Capability::INPUT_TEXT, Capability::INPUT_MULTIPLE, Capability::EMBEDDINGS],
             ],
             'text-embedding-3-small' => [
                 'class' => Embeddings::class,
-                'capabilities' => [Capability::INPUT_TEXT, Capability::EMBEDDINGS],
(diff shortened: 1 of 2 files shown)


chr-hertel (approved):

chr-hertel (changes requested):
Even if the change itself in the catalog works and looks good, the issue we have here is that the model catalog is generated since #2126 and your change would get overwritten.

I don't think there's a good solution to this right now.

chr-hertel:
Closed in favor of #2230 - thanks for your contribution anyways @liuzemei!

PR #2147, opened 2026-06-08
[Platform] Streamline the models.dev bridge to a data-only layer

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | -
| License       | MIT

after #1923 we can fold a lot into the first-level provider/platform setup, so the bridge is basically down to the catalog based on models.dev

am i butchering it too much, could you please check @fabpot?

maybe more controversial changes would be
* removed the `provider::model` feature
* removed auto-wiring for GenericBridge via factore, but left the helper for the base url - see example


src/platform/src/Bridge/ModelsDev/ModelCatalog.php
```diff
@@ -24,12 +27,28 @@
...
+     */
+    private const MODEL_CLASS_OVERRIDES = [
+        '@ai-sdk/anthropic' => [
+            'completions' => Claude::class,
+        ],
+        '@ai-sdk/google' => [
+            'completions' => Gemini::class,
+            'embeddings' => Embeddings::class,
+        ],
+    ];
```
fabpot, line 57: I would also add `@ai-sdk/google-vertex` here.
Also, after having a closer look, we cannot make this work for Bedrock. So, we should probably fail early when trying to use it.

fabpot (approved):
As my favorite activity is to remove code, I like this PR a lot :)
More seriously, I like the overall direction a lot.

src/platform/src/Bridge/ModelsDev/ModelCatalog.php
chr-hertel: Thanks, yes :+1:

PR #2152, opened 2026-06-10
[Platform] Add `Together` platform bridge

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | --
| License       | MIT

Add a Together AI platform bridge covering the full inference API: chat completions, embeddings, image generation, text-to-speech, speech-to-text and reranking.

The architecture follows the Ollama-style approach: an endpoint-based factory wrapping the HTTP client in a ScopingHttpClient, a dynamic ModelCatalog backed by GET /v1/models (with a static overlay for audio models, which the endpoint does not expose), a single Together model class with capability-based request routing, URL-based result conversion, and a dedicated TogetherContract.

Includes the ai-bundle integration (config, service wiring, tests), documentation and runnable examples. PHPStan level 10 on the bridge package.

### DIFF
--- a/deptrac.yaml
+++ b/deptrac.yaml
@@ -245,6 +245,10 @@ deptrac:
       collectors:
         - type: classLike
           value: Symfony\\AI\\Platform\\Bridge\\Scaleway\\.*
+    - name: TogetherPlatform
+      collectors:
+        - type: classLike
+          value: Symfony\\AI\\Platform\\Bridge\\Together\\.*
     - name: TransformersPhpPlatform
       collectors:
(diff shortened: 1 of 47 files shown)


chr-hertel (approved):
Also rebased this here and adopted some of the conventions we picked up over the time - embeddings and reranking didn't work for me and would need a follow up. Thanks @Guikingone!

chr-hertel:
Thank you @Guikingone.

PR #2155, opened 2026-06-11
[Agent] Add SystemPromptInputProcessor::getSystemPrompt()

### LINKED ISSUE #2154: Add a public getter for the configured system prompt (SystemPromptInputProcessor)
## Problem

Tooling that lists or inspects agents wants to read an agent's **system prompt** — e.g. the bundle's `ai:chat` (which today shows only bare agent names) or a richer agent picker. But `SystemPromptInputProcessor::$systemPrompt` is `private readonly` with no getter, so the only way to read it is reflection.

## Correction to the original proposal

This issue originally also proposed adding `getModel()` to `AgentInterface`. That's the **wrong layer** and I've dropped it: `AgentInterface` is implemented by `MultiAgent`, `SpeechAgent`, `MockAgent`, etc., which have **no single model** (or several). A model is a property of the concrete `Agent` (which already exposes `getModel()`), not of the agent abstraction — an agent is just "something you can `call($messages)` on". Forcing `getModel()` onto the interface would be an abstraction leak.

So this is narrowed to the system prompt, which genuinely lives on the processor.

## Proposal

Add a public accessor to `SystemPromptInputProcessor`:

```php
public function getSystemPrompt(): \Stringable|TranslatableInterface|string|File
{
    return $this->systemPrompt;
}
```

Pure addition, no BC break. Lets consumers display/inspect the configured prompt without reflection. PR incoming.

tacman:
| Q             | A
| ------------- | ---
| Branch?       | main
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | Fixes #2154
| License       | MIT

The configured system prompt lives in a `private readonly` property on `SystemPromptInputProcessor` with no accessor, so inspecting or displaying an agent's prompt — e.g. in a CLI agent picker, or to enrich the bundle's own `ai:chat` listing — currently requires reflection.

This adds a public getter that returns the prompt exactly as configured:

```php
public function getSystemPrompt(): \Stringable|TranslatableInterface|string|File
{
    return $this->systemPrompt;
}
```

Pure addition, no BC break.

Per the discussion in #2154, this is intentionally scoped to the system prompt only. The original issue also floated `getModel()` on `AgentInterface`, but that's the wrong layer: `MultiAgent`, `SpeechAgent` and `MockAgent` have no single model, so a model belongs on the concrete `Agent` (which already exposes `getModel()`), not on the agent abstraction.

### DIFF
--- a/src/agent/src/InputProcessor/SystemPromptInputProcessor.php
+++ b/src/agent/src/InputProcessor/SystemPromptInputProcessor.php
@@ -43,6 +43,11 @@ final class SystemPromptInputProcessor implements InputProcessorInterface
         }
     }
 
+    public function getSystemPrompt(): \Stringable|TranslatableInterface|string|File
+    {
+        return $this->systemPrompt;
+    }
+
     public function processInput(Input $input): void
     {
         $messages = $input->getMessageBag();
--- a/src/agent/CHANGELOG.md
+++ b/src/agent/CHANGELOG.md
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+0.10
+----
+
+ * Add `SystemPromptInputProcessor::getSystemPrompt()` to read the configured system prompt without reflection
+
 0.8
 ---
 
--- a/src/agent/tests/InputProcessor/SystemPromptInputProcessorTest.php
+++ b/src/agent/tests/InputProcessor/SystemPromptInputProcessorTest.php
@@ -31,6 +31,21 @@ use Symfony\Contracts\Translation\TranslatorInterface;
 
 final class SystemPromptInputProcessorTest extends TestCase
 {
+    public function testGetSystemPromptReturnsTheConfiguredPrompt()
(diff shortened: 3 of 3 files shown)


src/agent/src/InputProcessor/SystemPromptInputProcessor.php
```diff
@@ -43,6 +43,17 @@ public function __construct(
         }
     }
 
+    /**
+     * Returns the configured system prompt as provided to the constructor.
+     *
+     * Useful for inspecting or displaying an agent's prompt (e.g. in a CLI agent
+     * picker) without resorting to reflection.
+     */
```
chr-hertel: i think that getter is straight forward 👍 
```suggestion
```

chr-hertel (approved):

chr-hertel:
Thank you @tacman.

PR #2157, opened 2026-06-11
[AI Bundle] Make agent tools opt-in instead of opt-out

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | -
| License       | MIT

[BC BREAK] Agents no longer receive all `ai.tool`-tagged services implicitly when the `tools` option is not configured.

- No `tools` option (or `null`, `false`, `[]`) → no toolbox registered
- `tools: true` → explicit opt-in to inject all tagged tools (previous default)
- Explicit tool lists keep working unchanged
- `prompt.include_tools` without configured tools now throws a clear exception
- Updated bundle reference docs, fast-track cookbook (+ regenerated website artifacts), CHANGELOG, and UPGRADE.md for 0.10

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -13,6 +13,25 @@ Agent
    +$agent->call(clone $messages);
    ```
 
+AI Bundle
+---------
+
+ * Agent tools are now opt-in. Previously, when an agent did not configure the `tools` option, all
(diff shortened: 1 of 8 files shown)


chr-hertel:
@OskarStark can you please have a look if the bundle config is what you'd expect?

OskarStark:
Looks good, thanks 🙏

PR #2158, opened 2026-06-12
[Demo] Add TUI agent chat command

tacman:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | -
| License       | MIT

Adds a terminal UI chat command to the demo application so the configured agents can be explored directly from the console.

The command provides an agent picker, streaming chat output, a verbose debug pane for tool calls and token usage, and returns to the picker after leaving a chat session. The demo dependencies are also updated to Symfony 8.1 packages so it can use the new TUI component.

No separate documentation was added because this is a demo-only interactive command and is discoverable via `bin/console list app`.

Validation:

- `composer validate --no-check-publish`
- `php -l demo/src/Tui/Command/ChatTuiCommand.php`
- `git diff --check origin/main..HEAD`
- `php bin/console lint:yaml config --parse-tags`
- `vendor/bin/php-cs-fixer fix --config=.php-cs-fixer.dist.php --dry-run --diff demo/src/Tui/Command/ChatTuiCommand.php`
- `cd demo && vendor/bin/phpstan`
- `cd demo && vendor/bin/phpunit`
- `cd demo && php bin/console list app --format=txt`


tacman:
I kept forgetting which agent did what, so I wrote the tui version of chat to give a lot more details to help select the agent.

image

The TUI component rocks.

The chat is similar to https://github.com/GromNaN/symfony-tui-games in that it prompts you for a chat agent and returns to it when you're done, like the games menu does.

OskarStark:
Nice one 😊

GromNaN (commented):
Very nice. 

Since it's requiring symfony/tui, you can assume symfony/console 8.1 is used and use the fancy features like invokable commands with validation.

demo/src/Tui/Command/ChatTuiCommand.php
```diff
@@ -0,0 +1,579 @@
...
+            $platform = $this->readProperty($agent, 'platform');
+            if (!$platform instanceof PlatformInterface) {
+                return '';
+            }
+
+            $capabilities = $platform->getModelCatalog()->getModel($model)->getCapabilities();
+            $names = array_map(static fn (object $c): string => strtolower(str_replace('_', '-', $c->name)), $capabilities);
+
+            return implode(', ', $names);
+        } catch (\Throwable) {
```
chr-hertel: which exception are you expecting here?

demo/src/Tui/Command/ChatTuiCommand.php
```diff
@@ -0,0 +1,579 @@
...
+     */
+    private function tokenUsageLine(object $result): ?string
+    {
+        if (!method_exists($result, 'getMetadata')) {
+            return null;
+        }
+
+        $metadata = $result->getMetadata();
+        $usage = $metadata->get('token_usage');
+        if (!$usage instanceof TokenUsage) {
```
chr-hertel, line 529: we should test for the interface instead

demo/src/Tui/Command/ChatTuiCommand.php
tacman: ModelNotFound -- I'm fixing the code to reflect that.

demo/src/Tui/Command/ChatTuiCommand.php
tacman: yep!  Fixing now.

tacman:
Thanks for the review — those were dev shortcuts, now fixed:

- `catch (\Throwable)` → `catch (ModelNotFoundException)` (the expected case: the agent's configured model isn't in its platform's catalog), so anything else fails loud.
- `method_exists($result, 'getMetadata')` → a typed `ResultInterface $result` param (it extends `MetadataAwareInterface`); and the `method_exists($agent, 'getModel')` checks → `$agent instanceof Agent`.

More broadly, the whole `unwrapAgent()` + reflection in here is a deliberate stopgap. `AgentInterface` exposes only `call()`/`getName()`, and `TraceableAgent` doesn't forward `getModel()`, so to show the model/capabilities the demo has to peel decorators and read private properties — not something the official demo should ship long-term.

Would you be open to `AgentInterface::getModel(): ?Model` (nullable for multi-agent setups with no single model), with the decorators delegating? That would let consumers drop the unwrapping and reflection entirely. Happy to open a small focused PR for it — until then the unwrap stays as a temporary workaround.

chr-hertel:
looks like this branch got mixed up with the #2161

chr-hertel (changes requested):
Needs a rebase on `main` - merged #2161 and the git history is somehow twisted.

Issue #2072, opened 2026-05-11
[Platform][Mistral] Add Mistral OCR (/v1/ocr) support to symfony/ai-mistral-platform


tacman:
## Summary

The `mistral-ocr-latest` model uses a dedicated `/v1/ocr` endpoint that is structurally different from chat completions. The current `symfony/ai-mistral-platform` bridge only wraps `/v1/chat/completions` and `/v1/embeddings`, so applications that need document OCR must bypass the agent abstraction entirely and call the Mistral API via raw `HttpClient`.

## The gap

The Mistral OCR endpoint accepts a `document` payload (URL or base64-encoded image/PDF) and returns structured output with `pages[]`, markdown text, layout blocks, and embedded images. It is not a chat call — you cannot route it through `AgentInterface::call()` or `MessageBag`. There is currently no `Platform`, `Model`, or `Response` class in `symfony/ai-mistral-platform` that covers it.

Example request shape:

```json
POST https://api.mistral.ai/v1/ocr
{
  "model": "mistral-ocr-latest",
  "document": {
    "type": "image_url",
    "image_url": "https://..."
  },
  "include_image_base64": true,
  "document_annotation_format": { ... }
}
```

Example response shape:

```json
{
  "pages": [
    {
      "index": 0,
      "markdown": "...",
      "dimensions": { "width": 2480, "height": 3508 },
      "images": [ { "id": "img-0", "image_base64": "...", ... } ],
      "document_annotation": "{ \"blocks\": [...] }"
    }
  ]
}
```

## Proposed addition

A `MistralOcrPlatform` (or an `OcrModel` within the existing Mistral platform) that:

- Accepts a document URL or binary payload
- POSTs to `https://api.mistral.ai/v1/ocr`
- Returns a typed `OcrResponse` with `pages`, `layout_blocks`, and `image_blocks`
- Integrates with the existing `MISTRAL_API_KEY` environment variable wiring

This would allow applications to use Mistral OCR through the same DI-configured service layer as all other AI tasks, rather than managing a raw HTTP client alongside the agent abstraction.

## Context

Mistral positions OCR as a first-class document intelligence product (see [docs.mistral.ai/api/endpoint/ocr](https://docs.mistral.ai/api/endpoint/ocr)). It supports PDFs with multiple pages, layout detection, bounding boxes, and per-page annotation — well beyond what vision models produce via chat completions. Given that `symfony/ai-mistral-platform` already exists, this feels like a natural addition to the package.

[...]

PR #2160, opened 2026-06-12
[Platform][Cache] Cache content-object inputs via CacheKeyGenerator strategy

### LINKED ISSUE #2159: [Platform][Cache] CachePlatform cannot cache content-object inputs (Whisper audio, Mistral OCR)
## Summary

`CachePlatform` (the `ai.platform.cache.*` decorator, `symfony/ai-cache-platform`) provides "free idempotency" for re-running platform invocations. But it can only build a cache key for `string`, `array`, and `MessageBag` inputs — it throws on any other input type, including the first-party `ContentInterface` content objects (`Audio`, `Document`, `DocumentUrl`, `Image`, `ImageUrl`, …).

That means it cannot decorate any platform task whose **top-level input is a single content object**, which is the normal shape for audio transcription, OCR, and single-image vision — exactly the slow/expensive calls most worth caching.

## The gap

`CachePlatform::invoke()` (`src/platform/src/Bridge/Cache/CachePlatform.php`):

```php
    \is_array($input) => json_encode($input),
    $input instanceof MessageBag => $input->getId()->toString(),
    default => throw new InvalidArgumentException(\sprintf('Unsupported input type: %s', get_debug_type($input))),
};
```

A `ContentInterface` instance falls through to `default` and throws (only when `prompt_cache_key` is set; otherwise the call silently passes through uncached).

## Reproduction

Both of these already exist in the codebase today:

**Whisper (audio transcription)** — `examples/openai/audio-transcript.php` invokes with an `Audio` content object directly:

```php
$file = Audio::fromFile(__DIR__.'/audio.mp3');
$platform->invoke('whisper-1', $file); // <-- $file is a ContentInterface
```

Wrap that platform in `CachePlatform` and pass `prompt_cache_key` → `InvalidArgumentException: Unsupported input type: Symfony\AI\Platform\Message\Content\Audio`.

**Mistral OCR** (the bridge proposed in #2072) — same shape with `DocumentUrl`/`Document`:

```php
$platform->invoke('mistral-ocr-latest', new DocumentUrl('https://…/document.pdf'), ['prompt_cache_key' => 'ocr']);
// InvalidArgumentException: Unsupported input type: …\Content\DocumentUrl
```

Confirmed by direct test against `CachePlatform` (string input caches fine; a `DocumentUrl` input throws).

## Proposed fix (surgical)

Content objects currently have no identity — `ContentInterface` is an empty marker. Introduce a small opt-in interface so an input can advertise a stable cache key, and have `CachePlatform` honor it:

```php
interface CacheableInputInterface
{
    public function getCacheKey(): string;
}
```

```php
// CachePlatform::invoke()
    \is_array($input) => json_encode($input),
    $input instanceof MessageBag => $input->getId()->toString(),
    $input instanceof CacheableInputInterface => $input->getCacheKey(),
    default => throw new InvalidArgumentException(...),
};
```

Then implement it on the URL/file content classes:

- `DocumentUrl` / `ImageUrl` → return the URL (natural content identity).
- `Document` / `File` → `hash('xxh128', $this->asBinary())` (content hash; `File` already exposes `asBinary()` and `__serialize()`).

This generalizes to every content-object task rather than special-casing OCR, and is additive (no BC break — `CacheableInputInterface` is opt-in).

### Note on identity semantics

`MessageBag` keys off `getId()`, which is a per-instance random `Uuid::v7()` — i.e. *instance* identity (same content, new instance → cache miss). The content classes above would instead use *content* identity (same URL/bytes → hit), which is what you want for idempotent re-runs of a pipeline over a document/audio archive. Worth a sentence in the docs so the two behaviors aren't surprising; happy to align with whatever the maintainers prefer.

## Scope

Surgical and standalone — independent of the OCR bridge in #2072 (the gap already affects Whisper). I'm happy to send a PR if the direction is accepted.

tacman:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no 
| Issues        | Fix #2159
| License       | MIT

`CachePlatform` (the `ai.platform.cache.*` decorator) could only build a cache
key for `string`, `array` and `MessageBag` inputs — it threw
`Unsupported input type` on anything else. That means it could not decorate any
platform task whose top-level input is a content object, which is the normal
shape for the slow/expensive calls most worth caching:

- audio transcription — `$platform->invoke('whisper-1', Audio::fromFile(...))`
- OCR — `$platform->invoke('mistral-ocr-latest', new DocumentUrl(...))`
- single-image vision — `$platform->invoke($model, new ImageUrl(...))`

### Approach

Per @chr-hertel's review (keying inputs is the bridge's job — it shouldn't leak
into the Platform component or its content classes), this keeps the knowledge of
*how* to key an input **inside the Cache bridge** via a small strategy
interface:

```php
namespace Symfony\AI\Platform\Bridge\Cache;

interface CacheKeyGenerator
{
    public function supports(object $input): bool;
    public function generate(object $input): string;
}
```

`CachePlatform` holds an ordered list of generators and the first one that
`supports()` an input keys it:

```php
    \is_array($input) => json_encode($input),
    default => $this->generateInputCacheKey($input), // tries each CacheKeyGenerator
};
```

The default set covers `MessageBag`, `DocumentUrl`, `ImageUrl` and
`File` (and its `Audio`/`Image`/`Video`/`Document` subclasses). `DocumentUrl`
and `ImageUrl` key on the URL; `File` on a hash of its bytes. `getCacheKey()`
values are used verbatim in a cache key, so the URL/byte arms hash their value
(`xxh128`) the same way the existing `string` arm uses `md5()`.

`MessageBag` now routes through the same mechanism
(`MessageBagCacheKeyGenerator`), so **all** object inputs go through one
consistent path instead of a special-cased inline arm.

```php
$cached = new CachePlatform($platform, cache: $pool);
$cached->invoke('mistral-ocr-latest', new DocumentUrl('https://…/doc.pdf'), [
    'prompt_cache_key' => 'ocr',
]); // now cached instead of throwing
```

A userland content type opts in by registering its own generator — no change to
`CachePlatform` or to core content classes:

```php
new CachePlatform($platform, cache: $pool, cacheKeyGenerators: [
    new MyContentCacheKeyGenerator(),
    new MessageBagCacheKeyGenerator(),
    // …
]);
```

### Scope

Fully isolated in the bridge: **no changes to the Platform component or its
content classes**, so the dependency points bridge → component only. This is
also additive — behavior for `string`/`array`/`MessageBag` inputs is unchanged.

Tests added to `CachePlatformTest` (content-object cached once; different inputs
not shared; unsupported object still throws; custom generator honored). Surfaced
while adding a Mistral OCR bridge (#2072), but independent of it — the gap
already affects Whisper today.

> Supersedes the earlier `CacheableInputInterface`-in-Platform approach from
> this PR's first revision, which leaked the bridge's caching concern into the
> core content classes.

### DIFF
--- a/src/platform/src/Bridge/Cache/CHANGELOG.md
+++ b/src/platform/src/Bridge/Cache/CHANGELOG.md
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+0.11
+----
+
+ * Add `CacheKeyGenerator` strategy and a default set (`MessageBag`, `DocumentUrl`, `ImageUrl`, `File`/`Audio`), so `CachePlatform` can cache document, OCR and audio-transcription tasks, not just `string`/`array`/`MessageBag` inputs; custom input types can be supported by registering an additional generator
+
 0.3
 ---
 
--- a/src/platform/src/Bridge/Cache/CacheKeyGenerator.php
+++ b/src/platform/src/Bridge/Cache/CacheKeyGenerator.php
@@ -1,0 +1,37 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Platform\Bridge\Cache;
+
+/**
+ * Builds a stable, cache-key-safe identifier for a non-scalar platform input.
+ *
+ * Implementations are registered with {@see CachePlatform} and tried in order;
+ * the first one whose {@see self::supports()} returns true keys the input. This
+ * keeps the knowledge of how to cache a given input type inside the Cache bridge
+ * instead of leaking it into the Platform component or its content classes.
+ *
+ * @author Tac Tacelosky <tacman@gmail.com>
+ * @author Christopher Hertel <mail@christopher-hertel.de>
+ */
+interface CacheKeyGenerator
+{
+    public function supports(object $input): bool;
+
+    /**
+     * Returns a stable, cache-key-safe identifier for the given input.
+     *
+     * The value is used verbatim as part of a cache key, so it must not contain
+     * the PSR-6 reserved characters ("{}()/\@:"). Hash any value that might
+     * (e.g. a URL or raw bytes).
+     */
+    public function generate(object $input): string;
+}
--- a/src/platform/src/Bridge/Cache/CachePlatform.php
+++ b/src/platform/src/Bridge/Cache/CachePlatform.php
@@ -12,7 +12,6 @@
 namespace Symfony\AI\Platform\Bridge\Cache;
 
 use Symfony\AI\Platform\Exception\InvalidArgumentException;
-use Symfony\AI\Platform\Message\MessageBag;
 use Symfony\AI\Platform\Model;
 use Symfony\AI\Platform\ModelCatalog\ModelCatalogInterface;
 use Symfony\AI\Platform\PlainConverter;
@@ -44,6 +43,9 @@ use Symfony\Contracts\Cache\ItemInterface;
  */
 final class CachePlatform implements PlatformInterface
 {
+    /**
+     * @param iterable<CacheKeyGenerator> $cacheKeyGenerators Tried in order to key non-scalar inputs (objects)
+     */
     public function __construct(
         private readonly PlatformInterface $platform,
         private readonly ClockInterface $clock = new MonotonicClock(),
@@ -56,6 +58,12 @@ final class CachePlatform implements PlatformInterface
         ], [new JsonEncoder()]),
         private readonly ?string $cacheKey = null,
         private readonly ?int $cacheTtl = null,
+        private iterable $cacheKeyGenerators = [
+            new MessageBagCacheKeyGenerator(),
+            new DocumentUrlCacheKeyGenerator(),
+            new ImageUrlCacheKeyGenerator(),
+            new FileCacheKeyGenerator(),
+        ],
     ) {
     }
 
@@ -70,8 +78,7 @@ final class CachePlatform implements PlatformInterface
         $normalizedInput = match (true) {
             \is_string($input) => md5($input),
             \is_array($input) => json_encode($input),
-            $input instanceof MessageBag => $input->getId()->toString(),
-            default => throw new InvalidArgumentException(\sprintf('Unsupported input type: %s', get_debug_type($input))),
+            default => $this->generateInputCacheKey($input),
         };
 
         $cacheKey = (new UnicodeString())->join([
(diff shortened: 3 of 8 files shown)


chr-hertel (changes requested):
From architecture point of view I'm not a real fan here since requirements of that cache bridge are leaking into the main Platform component - you avoided that it gets obvious with moving the `CacheableInputInterface` into the Platform namespace, but semantically it's still leaky. can we isolate this in the bridge itself?

tacman:
Yeah, you're right — keying inputs is the bridge's job, not something core content classes should advertise. The fix I have in mind keeps an interface (so it stays extensible) but moves it into the bridge as a strategy:

```php
namespace Symfony\AI\Platform\Bridge\Cache;

interface CacheKeyGenerator
{
    public function supports(object $input): bool;
    public function generate(object $input): string;
}
```

`CachePlatform` holds an ordered list of these (default set covers `DocumentUrl`/`ImageUrl`/`File`), so the dependency points bridge → component, content classes go back to pure, and a userland content type can register its own generator instead of implementing a core interface. `CacheableInputInterface` and the `getCacheKey()` methods get dropped.

One thing I want your call on before I push, since it shapes the implementation: **should `MessageBag` route through the same generator mechanism, or stay inline?** Today it's keyed inline (`$input->getId()`). I'd lean toward folding it into a `MessageBagCacheKeyGenerator` so *all* object inputs go through one consistent path — otherwise we've got two parallel mechanisms for objects, which is a bit of the same smell. But if you'd rather keep `MessageBag` special-cased inline and only route the content objects through generators, that's an easy variant. Which do you prefer?

tacman:
@chr-hertel thoughts on this?  I'd love to get it in before 0.11  -- the mistral calls are relatively expensive, so I'd love to cache them.

chr-hertel:
that pattern looks like a good fit to me 👍

tacman:
CacheKeyGenerator, with MessageBag going through it as well? I'll submit a PR shortly.

tacman:
Pushed the `CacheKeyGenerator` strategy as discussed — it lives entirely in the Cache bridge now, so there are **no changes to the Platform component or its content classes** (`CacheableInputInterface` and the `getCacheKey()` methods are gone). `MessageBag` routes through the same mechanism (`MessageBagCacheKeyGenerator`), so all object inputs share one consistent path, and custom input types opt in by registering their own generator. Default set: `MessageBag`, `DocumentUrl`, `ImageUrl`, `File`/`Audio`. CHANGELOG is under 0.11.

tacman:
Green now — the last red was a Fabbot nit (exception value needed double-quoting in `CachePlatform`), fixed. Recap of where this landed vs your review: the caching concern is fully isolated in the bridge via the `CacheKeyGenerator` strategy you sketched — nothing leaks into the Platform component or its content classes, and `MessageBag` routes through the same path. I credited you as co-author on the `CacheKeyGenerator` interface since the shape was your call. Mind giving it another look when you have a sec? 🙏

chr-hertel (approved):

chr-hertel:
Thank you @tacman.

PR #2161, opened 2026-06-12
[Platform][Mistral] Add OCR support via the /v1/ocr endpoint

### LINKED ISSUE #2072: [Platform][Mistral] Add Mistral OCR (/v1/ocr) support to symfony/ai-mistral-platform

tacman:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | Fix #2072
| License       | MIT

Adds OCR support to the Mistral bridge via the dedicated `/v1/ocr` endpoint, as
discussed in #2072.

Mistral's `mistral-ocr-latest` model is not a chat completion — it uses a
separate endpoint and returns structured per-page output, so it can't be routed
through `MessageBag`/chat. Following the existing Whisper bridge pattern, this
adds a dedicated `Ocr` model with its own `ModelClient` and `ResultConverter`:

```php
use Symfony\AI\Platform\Bridge\Mistral\Ocr\Result\OcrResult;
use Symfony\AI\Platform\Message\Content\DocumentUrl;

$result = $platform->invoke('mistral-ocr-latest', new DocumentUrl('https://example.com/document.pdf'));

$ocr = $result->asObject();
\assert($ocr instanceof OcrResult);

echo $ocr->getMarkdown();
```

The result is a typed `OcrResult` (pages with markdown, layout images + bounding
boxes, per-page annotations, usage info) via `ObjectResult`/`asObject()`, not a
text blob. Document URL, binary PDF and image inputs all work — the existing
`Document`, `DocumentUrl` and `ImageUrl` contract normalizers were widened to
also accept the `Ocr` model. Catalog entries (`mistral-ocr-latest`,
`mistral-ocr-2505`) plus a generator rule keep OCR models out of the chat class.

Verified end-to-end against the live API (a 29-page arXiv PDF by URL and a
binary PDF by base64), plus unit tests for the client, converter and catalog.

Also includes a `Document OCR` demo (`demo/`) showing the feature as a chat that
extracts a document's text and answers questions about it.

Docs: `docs/components/platform.rst`, the bridge `README.md`, examples under
`examples/mistral/`, and a `CHANGELOG.md` entry.

Companion PR #2160 (`CacheableInputInterface`) makes the `ai.platform.cache.*`
decorator able to cache content-object inputs like `DocumentUrl`, so OCR results
can be cached at the platform layer — independent of this PR.

### DIFF
--- a/src/platform/src/Bridge/Mistral/Ocr/ResultConverter.php
+++ b/src/platform/src/Bridge/Mistral/Ocr/ResultConverter.php
@@ -1,0 +1,84 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Platform\Bridge\Mistral\Ocr;
+
+use Symfony\AI\Platform\Bridge\Mistral\Ocr;
+use Symfony\AI\Platform\Bridge\Mistral\Ocr\Result\Image;
+use Symfony\AI\Platform\Bridge\Mistral\Ocr\Result\OcrResult;
+use Symfony\AI\Platform\Bridge\Mistral\Ocr\Result\Page;
+use Symfony\AI\Platform\Exception\RuntimeException;
+use Symfony\AI\Platform\Model;
+use Symfony\AI\Platform\Result\HttpStatusErrorHandlingTrait;
+use Symfony\AI\Platform\Result\ObjectResult;
+use Symfony\AI\Platform\Result\RawResultInterface;
+use Symfony\AI\Platform\ResultConverterInterface;
+use Symfony\AI\Platform\TokenUsage\TokenUsageExtractorInterface;
+
+/**
+ * @author Tac Tacelosky <tacman@gmail.com>
+ */
+final class ResultConverter implements ResultConverterInterface
+{
+    use HttpStatusErrorHandlingTrait;
+
+    public function supports(Model $model): bool
+    {
+        return $model instanceof Ocr;
+    }
+
+    public function convert(RawResultInterface $result, array $options = []): ObjectResult
+    {
+        $httpResponse = $result->getObject();
+
+        $this->throwOnHttpError($httpResponse);
+
+        if (200 !== $httpResponse->getStatusCode()) {
+            throw new RuntimeException(\sprintf('Unexpected response code %d: "%s"', $httpResponse->getStatusCode(), $httpResponse->getContent(false)));
+        }
+
+        $data = $result->getData();
+
+        if (!isset($data['pages']) || !\is_array($data['pages'])) {
+            throw new RuntimeException('Response does not contain pages.');
+        }
+
+        $pages = array_map(static function (array $page): Page {
+            $images = array_map(static fn (array $image): Image => new Image(
+                $image['id'] ?? '',
+                $image['top_left_x'] ?? null,
+                $image['top_left_y'] ?? null,
+                $image['bottom_right_x'] ?? null,
+                $image['bottom_right_y'] ?? null,
+                $image['image_base64'] ?? null,
+            ), $page['images'] ?? []);
+
+            return new Page(
+                $page['index'] ?? 0,
+                $page['markdown'] ?? '',
+                $images,
+                $page['dimensions'] ?? null,
+            );
+        }, $data['pages']);
+
+        return new ObjectResult(new OcrResult(
+            $pages,
+            $data['model'] ?? '',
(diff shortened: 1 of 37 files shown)


src/platform/src/Bridge/Mistral/README.md
```diff
@@ -8,6 +8,29 @@ Mistral Documentation
...
+    use Symfony\AI\Platform\Message\Content\DocumentUrl;
+
+    $platform = Factory::createPlatform($apiKey);
+
+    $result = $platform->invoke('mistral-ocr-latest', new DocumentUrl('https://example.com/document.pdf'));
+
+    $ocr = $result->asObject();
+    \assert($ocr instanceof OcrResult);
+
+    echo $ocr->getMarkdown();
```
chr-hertel: rather belongs in the docs folder, but is already present - we keep the readme slim


```suggestion
```

chr-hertel (changes requested):
Thanks for working on this @tacman - including that in the demo is great, but UI needs some love there: gap between the rows and no bg color/gradient (also for detail page) is inconsistent

Image

Thanks already!

tacman:
Yes. I'll tweak the UI and make getting the OCR layout a #[AsTool] -- then the agent can fetch the layout and respond.

src/platform/src/Bridge/Mistral/Ocr/ResultConverter.php
```diff
@@ -0,0 +1,84 @@
...
+                $image['bottom_right_y'] ?? null,
+                $image['image_base64'] ?? null,
+            ), $page['images'] ?? []);
+
+            return new Page(
+                $page['index'] ?? 0,
+                $page['markdown'] ?? '',
+                $images,
+                $page['dimensions'] ?? null,
+                $page['document_annotation'] ?? null,
```
chr-hertel: the `document_annotation` is not a page level field, but on `$data` - can you please double check?

Image

tacman: Good catch, you're right — `document_annotation` is a document-wide field on the response root (next to `model`/`usage_info`), not per-page. Moved it from `Page` to `OcrResult` (`getDocumentAnnotation()`), read it from `$data` in the converter, and added test coverage. Fixed in f968fde9.

chr-hertel:
Cherry-picked f5e8a487832de8f9ad072621fdf12db9fe4cf447 from #2158 and made some UI patches in the demo.

chr-hertel:
Thank you @tacman.

chr-hertel (approved):

PR #2165, opened 2026-06-12
[Platform] Throw ExceedContextSizeException when a request exceeds the model's context window

fabpot:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Docs?         | no 
| Issues        | Fix #... 
| License       | MIT

As ExceedContextSizeException already exists, let's use it when appropriate.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -70,6 +70,22 @@ Platform
    +public function supports(string|Model $model): bool
    ```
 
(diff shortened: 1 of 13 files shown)


chr-hertel (approved):
Works well, thanks!

The only thing i wonder if the `ExceedContextSizeException` rather should extend the `BadRequestException` instead - but off scope here.

chr-hertel:
Thank you @fabpot.

PR #2167, opened 2026-06-13
[Platform] Throw ExceedContextSizeException across remaining LLM bridges

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | -
| License       | MIT

Follow-up to #2165: extends `ExceedContextSizeException` to the remaining LLM bridges where the provider's context-overflow error is documented and detectable from the HTTP response.

Detection is centralized in a shared `ExceedContextSizeException::matches()` predicate, applied via `HttpStatusErrorHandlingTrait` (covers Cerebras, DeepSeek, Mistral, Cohere, Gemini, Perplexity) plus inline handling in Scaleway and VertexAI.

Skipped where no easy/documented signal exists: Bedrock, Ollama, Codex, ClaudeCode, Replicate, Azure/Meta, HuggingFace, TransformersPhp.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -89,9 +89,11 @@ Platform
    +public function supports(string|Model $model): bool
    ```
 
- * The Anthropic, Generic, OpenAI, and OpenResponses bridges now throw `ExceedContextSizeException`
-   instead of `BadRequestException` when a 400 response reports a context overflow. As
-   `ExceedContextSizeException` does not extend `BadRequestException`, catch it explicitly:
+ * The Anthropic, Generic, OpenAI, OpenResponses, Cerebras, DeepSeek, Mistral, Cohere, Gemini,
+   Perplexity, Scaleway, and VertexAI bridges now throw `ExceedContextSizeException` when a 400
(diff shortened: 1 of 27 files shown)


src/platform/src/Exception/ExceedContextSizeException.php
```diff
@@ -16,4 +16,40 @@
...
+     */
+    private const OVERFLOW_MESSAGE_NEEDLES = [
+        'context length',           // OpenAI-compatible, Mistral ("maximum context length")
+        'context_length_exceeded',  // snake_case code echoed in the message
+        'context window',           // OpenAI Responses ("exceeds the context window")
+        'prompt is too long',       // Anthropic
+        'too many tokens',          // Cohere
+        'maximum number of tokens', // Gemini / VertexAI
+        'input token count',        // Gemini / VertexAI
+    ];
```
fabpot: I haven't done it because I thought we would have these in each corresponding bridge instead of them being centralized here.

chr-hertel: feels odd, yes, agree

PR #2170, opened 2026-06-13
[Platform] Parse SSE streams that omit the text/event-stream content type

fabpot:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no 
| Docs?         | no 
| Issues        | n/a
| License       | MIT

SseStream only parsed `Server-Sent` Events when the response advertised `Content-Type: text/event-stream`. Some backends (e.g. the ChatGPT Codex backend) stream raw `event:/data:` framing without that header, so the framing reached `json_decode` verbatim and blew up with a syntax error.

The passthru branch now detects the payload format from the first bytes and parses accordingly:

* starts with `[/{` → streamed JSON array (Gemini/Vertex AI without alt=sse), unchanged behavior
* otherwise → SSE framing, buffered and split on blank lines

Properly-detected SSE keeps its existing code path untouched.

### DIFF
--- a/src/platform/src/Result/Stream/SseStream.php
+++ b/src/platform/src/Result/Stream/SseStream.php
@@ -16,7 +16,10 @@ use Symfony\Component\HttpClient\EventSourceHttpClient;
 use Symfony\Contracts\HttpClient\ResponseInterface;
 
 /**
- * Handles SSE (Server-Sent Events) streaming responses.
+ * Handles Server-Sent Events streaming responses that advertise the
+ * "text/event-stream" content type, so the HTTP client assembles each event
+ * into a {@see ServerSentEvent} chunk before it reaches this decoder. Backends
+ * that omit that header are handled by {@see RawSseStream}.
  *
  * @author Johannes Wachter <johannes@sulu.io>
  */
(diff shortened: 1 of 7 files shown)


src/platform/src/Result/Stream/SseStream.php
```diff
@@ -22,34 +22,114 @@
...
+    private const MODE_JSON = 'json';
+    private const MODE_SSE = 'sse';
+
     public function stream(ResponseInterface $response): iterable
     {
+        $mode = null;
+        $buffer = '';
+
         foreach ((new EventSourceHttpClient())->stream($response) as $chunk) {
             if ($chunk->isFirst() || $chunk->isLast() || ($chunk instanceof ServerSentEvent && '[DONE]' === $chunk->getData())) {
```
chr-hertel: we could move that `trim()` from `decode()` to here i think

chr-hertel:
can't really test that - at least i don't know how :D

the only thing i'm wondering is if we pull this into a separate `HttpStreamInterface` implementation, does that make sense?

fabpot:
> can't really test that - at least i don't know how :D
> 
> the only thing i'm wondering is if we pull this into a separate `HttpStreamInterface` implementation, does that make sense?

Refactored... and then saw #2186
This one is now a blend of your proposal and what's needed to make it work IRL :)

chr-hertel (approved):

chr-hertel:
Thank you @fabpot.

PR #2171, opened 2026-06-13
[Platform] Add CachedModelCatalog and smoother model catalog DX

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | -
| License       | MIT

Three small DX improvements around model catalogs:

- **`CachedModelCatalog`** — a decorator that memoizes any `ModelCatalogInterface` behind a PSR-6 cache pool (without caching `ModelNotFoundException`), so API-based catalogs (Ollama, amazee.ai, ...) can persist resolutions across requests instead of hitting the API on every lookup.
- **`CompositeModelCatalog`** — softened the docblock so it reads as a public composition tool ("bundled catalog + my own models"), not an internal Platform detail.
- **Base `Model` pitfall** — passing a bare `new Model(...)` (instead of a bridge subclass like `Gpt`/`Claude`) now fails with a message that names the required subclasses, instead of an opaque "No provider found".

```php
$catalog = new CachedModelCatalog(new OllamaCatalog($http), new FilesystemAdapter(), ttl: 3600);
```

Docs: new sections in `model-catalogs.rst` (composition, custom catalogs, caching, capability reference).

### DIFF
--- a/docs/components/platform/model-catalogs.rst
+++ b/docs/components/platform/model-catalogs.rst
@@ -188,6 +188,86 @@ directly::
 A model id resolves to the first provider (in array order) whose catalog knows it. When several
 providers expose the *same* id, order the array so the preferred one comes first.
 
+Combining Catalogs Explicitly
+-----------------------------
+
+A provider holds one catalog, but that catalog can be a composition.
+:class:`Symfony\\AI\\Platform\\ModelCatalog\\CompositeModelCatalog` merges several catalogs and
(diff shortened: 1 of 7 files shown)


chr-hertel:
Moving this to my fork (chr-hertel/ai) for the initial self-review round, to keep the noise out of the main repo. I'll re-open it here once it's ready for review.

PR #2172, opened 2026-06-13
[AI Bundle] Support model_catalog override on every catalog-bearing platform

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | -
| License       | MIT

The `model_catalog` configuration key — which replaces a platform's bundled catalog with a custom `ModelCatalogInterface` service — was only honored by the `generic`, `openresponses`, and `bedrock` platforms. First-class bridges (`openai`, `anthropic`, `gemini`, `mistral`, ...) silently ignored it.

This wires it up consistently via a shared `resolveModelCatalog()` helper, and adds the `model_catalog` node to every platform config that ships a catalog.

```yaml
ai:
    platform:
        anthropic:
            api_key: "%env(ANTHROPIC_API_KEY)%"
            model_catalog: "app.model_catalog.anthropic" # any ModelCatalogInterface service
```

`ollama` and `elevenlabs` build their catalog from the API at runtime and remain the exceptions (the Ollama one is lifted in a follow-up).

PR #2173, opened 2026-06-13
[AI Bundle][Platform] Allow a custom model catalog on the Ollama platform

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | -
| License       | MIT

> [!NOTE]
> Stacked on #2172 — please review/merge that first. This PR's own change is the last commit (the Ollama support); the diff shows #2172 underneath until it merges.

`Ollama\Factory::createProvider()` and `createPlatform()` gain an optional `$modelCatalog` parameter (after `$httpClient`, matching the other bridge factories), defaulting to the runtime API-based `ModelCatalog`. The bundle's `ollama` platform now accepts a `model_catalog` override too, so it is no longer an exception — handy for wrapping the API catalog in `CachedModelCatalog`. Only `elevenlabs` remains.

The new parameter shifts later positional arguments by one; all in-repo callers already use named args, and `UPGRADE.md` documents the change.


chr-hertel:
Moving this to my fork (chr-hertel/ai) for the initial self-review round, to keep the noise out of the main repo. I'll re-open it here once it's ready for review.

PR #2174, opened 2026-06-13
[Platform][Anthropic] Move unsupported schema features to `description`

### LINKED ISSUE #2119: Improve StructuredOutput options
Hello,

  I'm not sure if the issue is how I'm using the system or if there needs to be some improvements when trying to use structured output. We're trying to get some objects out from anthropic's API. 

  What it supports (in terms of json_schema) seems to be somewhat fickle, because of this I don't think we can pass a object or class name to the response_format. When we do, Anthropic complains. Because of this we did the following:

```
        if ($agentMessageConfig->getOutputSchema()) {
            $options[PlatformSubscriber::RESPONSE_FORMAT] = [
                'json_schema' => [
                    'schema' => json_decode($agentMessageConfig->getOutputSchema(), true, 512, JSON_THROW_ON_ERROR),
                ],
            ];
        }
...
        return $this->agent->call($messages, $options);
```

This works fairly well, except then we get an array returned instead of the object. This happens because `src/StructuredOutput/PlatformSubscriber.php`'s logic is

```
 $responseFormat = $options[self::RESPONSE_FORMAT];

        if (\is_object($responseFormat)) {
            $this->objectToPopulate = $responseFormat;
            $className = $responseFormat::class;
        } elseif (\is_string($responseFormat) && class_exists($responseFormat)) {
            $this->objectToPopulate = null;
            $className = $responseFormat;
        } else {
            return;
        }
```

Because we've provided the schema as an array already, no `$className` is set. So later on in the `processResult()` method we have

```
public function processResult(ResultEvent $event): void
{
        $options = $event->getOptions();

        if (!isset($options[self::RESPONSE_FORMAT])) {
            return;
        }

        $deferred = $event->getDeferredResult();
        $converter = new ResultConverter(
            $deferred->getResultConverter(),
            $this->serializer,
            $this->outputType ?? null,
            $this->objectToPopulate
        );

        $event->setDeferredResult(new DeferredResult($converter, $deferred->getRawResult(), $options));

        // Reset object to populate for next invocation
        $this->objectToPopulate = null;
}
```

Where the self::RESPONSE_FORMAT is set, but outputType is null, and so we just get back an array because the ResultConverter is instantiated without an outputType.

I'm not sure what the correct solution is.

I could create an event subscriber with the correct priority to replace the auto generated schema with the desired one but keeping the class. However, I'm not sure if that's the desired solution. I'm happy to code things up and provide a PR but need a bit of guidance on how you expect something like this to work.

valtzu:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #2119
| License       | MIT

Move unsupported schema properties in Anthropic structured output to `description`, like done in official Anthropic PHP SDK.

Related docs: https://platform.claude.com/docs/en/build-with-claude/structured-outputs#json-schema-limitations

### DIFF
--- a/examples/anthropic/structured-output-constrained.php
+++ b/examples/anthropic/structured-output-constrained.php
@@ -1,0 +1,75 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+// Demonstrates that JSON Schema constraints unsupported by Anthropic's API
+// (minimum, maximum, minLength, maxItems, etc.) are moved to the `description`
+// field by JsonSchemaSanitizerTrait so the model still reads and respects them.
+
+use Symfony\AI\Platform\Bridge\Anthropic\Factory;
+use Symfony\AI\Platform\Message\Message;
+use Symfony\AI\Platform\Message\MessageBag;
+use Symfony\AI\Platform\StructuredOutput\PlatformSubscriber;
+use Symfony\Component\EventDispatcher\EventDispatcher;
+
+require_once dirname(__DIR__).'/bootstrap.php';
+
+$dispatcher = new EventDispatcher();
+$dispatcher->addSubscriber(new PlatformSubscriber());
+
+$platform = Factory::createPlatform(env('ANTHROPIC_API_KEY'), http_client(), eventDispatcher: $dispatcher);
+
+$messages = new MessageBag(
+    Message::forSystem('You are a product catalog assistant. Generate realistic product data that strictly respects all constraints in the schema descriptions.'),
+    Message::ofUser('Create a product entry for a mid-range wireless headphone.'),
+);
+
+$result = $platform->invoke('claude-sonnet-4-5-20250929', $messages, ['response_format' => [
+    'type' => 'json_schema',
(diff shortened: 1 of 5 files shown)


chr-hertel:
Thanks @valtzu - messy topic - and I'd assume it's quite a gap across other bridges as well.

having second thoughts here about silently dropping this - like the idea of trying to recover under the hood - couldn't find the solution at first glance tho in https://github.com/anthropics/anthropic-sdk-php

valtzu:
> couldn't find the solution at first glance tho in https://github.com/anthropics/anthropic-sdk-php

Here I think https://github.com/anthropics/anthropic-sdk-php/blob/main/src/Lib/Helpers/SchemaInference.php#L101-L222

So they just put the unsupported properties `json_encode`d to `description`

valtzu:
> So they just put the unsupported properties json_encoded to description

Did the same here + added `examples/structured-output-constrained.php` to demonstrate it's working (and failing without the fix).

chr-hertel (approved):
Thanks @valtzu - will register the `Symfony\AI\Platform\StructuredOutput\PlatformSubscriber` while merging on the example - works well 👍

chr-hertel:
Thank you @valtzu.

aszenz:
@valtzu Anthropic also doesn't support oneOf, maybe we can replace it with anyOf, although that's a semantic change

PR #2178, opened 2026-06-15
[Platform] Fix double-counting of Anthropic streaming prompt and cache tokens

fabpot:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | -
| License       | MIT

When streaming, Anthropic reports token usage in **two** events:

- `message_start.usage` carries the prompt and cache token counts plus a provisional `output_tokens`.
- `message_delta.usage` repeats the **same cumulative** `input_tokens` / `cache_read_input_tokens` / `cache_creation_input_tokens`, together with the final `output_tokens`.

The Anthropic `ResultConverter` yielded a full `TokenUsage` for both events, and the stream aggregation (`TokenUsageAggregation`) **sums every field across all yielded usage events**. As a result, prompt and cache tokens (which dominate a cached conversation) were **counted twice**, roughly doubling the reported context usage.

This was observed in a coding agent where the context-window gauge climbed past 120% on a 1M-token model while the real usage was about 63%. A live replay of a stored request confirmed both events repeat the cumulative counts:

```
message_start.usage  → input_tokens:2, cache_creation:47119, cache_read:0, output_tokens:4
message_delta.usage  → input_tokens:2, cache_creation:47119, cache_read:0, output_tokens:64
```

The fix keeps the summing aggregation correct by emitting each number once: prompt and cache counts from `message_start` (dropping its provisional `output_tokens`), and the final `output_tokens` from `message_delta`.

A regression test asserts that the aggregated usage is not double-counted.

### DIFF
--- a/src/platform/src/Bridge/Anthropic/ResultConverter.php
+++ b/src/platform/src/Bridge/Anthropic/ResultConverter.php
@@ -155,13 +155,24 @@ class ResultConverter implements ResultConverterInterface
                 $inMessage = true;
             }
 
-            // Handle usage from message_start and message_delta
+            // Anthropic reports usage in both message_start and message_delta:
+            // message_start carries the prompt and cache token counts plus a
+            // provisional output_tokens, and message_delta repeats the same
+            // cumulative prompt/cache counts with the final output_tokens. As
+            // the stream aggregation sums every yielded usage, emitting the full
+            // payload from both events would double-count input and cache tokens.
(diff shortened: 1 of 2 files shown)


chr-hertel (approved):
oof 👍

chr-hertel:
Thank you @fabpot.

PR #2188, opened 2026-06-15
[Platform] Fix OpenResponses null tool-call id and refresh deprecated example models

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | -
| License       | MIT

Fixes a `TypeError` in the OpenResponses `ResultConverter` when a provider (e.g. Scaleway) returns a function call with a null `id` — now prefers `call_id` (which the Responses API uses to address tool results) and falls back to `id`. Also hardens the Generic completions converter against a null array offset on 401 responses, and refreshes deprecated models in three examples (`gemini-3-pro-preview` → `gemini-3.1-pro-preview`, `gemini-embedding-exp-03-07` → `gemini-embedding-001`).


OskarStark:
Thank you @chr-hertel.

PR #1937, opened 2026-04-12
[Platform][Anthropic] Return `MultiPartResult` for multi-part responses

valtzu:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| License       | MIT

Same as #1900 but for Anthropic.

I included code execution server tool just to be able to test the feature.

PR #2196, opened 2026-06-18
[Platform][Anthropic] Support server-side tools via server_tools option

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | -
| License       | MIT

Reworked from the original raw-list version of this PR.

Adds a `server_tools` option to the Anthropic `ModelClient`, using the `name => params` map shape the Gemini and Vertex AI bridges already use:

```php
$platform->invoke('claude-sonnet-5', $messages, [
    'server_tools' => ['web_search' => ['max_uses' => 3]],
]);
```

**Why the shape changed.** Raw pass-through already works today: `request()` merges `$options` into the body, and the docs already showed that for code execution. So the first version of this option added a name, not a capability - and it used Gemini's name with a different shape, which reads portable without being portable.

Only `web_search` and `code_execution` are mapped, to their versioned Anthropic `type`. An unrecognized name throws `InvalidArgumentException` rather than being silently forwarded, since Anthropic needs a versioned `type` the bridge cannot guess; unmapped tools stay reachable through the raw `tools` option, now documented as exactly that. Params are merged over the default spec, so a caller on Opus 4.6+/Sonnet 4.6+ can opt into `web_search_20260209` per call.

Server tools are appended *after* `injectToolsCacheControl()` runs, so `cache_control` lands on the last function tool and never on a server tool - the previous version put it on a server tool and asserted that, without it having been verified against the live API.

`ResultConverter` now merges the `server_tool_use` call and its matching `web_search_tool_result` into a single `Result\WebSearchResult` carrying query, id and status - matching `OpenResponses`, which emits one per search. Previously both blocks were dropped, and a response carrying only web search blocks threw `Response content does not contain any supported content.`

Known limitations, documented in the RST rather than papered over: `WebSearchResult` has no field for individual search hits, and server tool blocks are not converted on streaming requests. Follow-ups: search hits, `citations` on text blocks, `encrypted_content` round-tripping, a `Content\WebSearch` branch in `AssistantMessageNormalizer` (without it a search is dropped when re-serializing an assistant turn), and streaming.

### DIFF
--- a/docs/components/platform/anthropic-server-tools.rst
+++ b/docs/components/platform/anthropic-server-tools.rst
@@ -1,15 +1,65 @@
 Anthropic Server Tools
 ======================
 
-Anthropic provides built-in server-side tools that allow the model to perform specific actions like executing code or editing files in a sandboxed environment.
+Anthropic provides built-in server-side tools that allow the model to perform specific actions like searching the
+web or executing code in a sandboxed environment.
 
 Overview
 --------
 
-Anthropic's server tools can be enabled when calling the model. These tools are executed on Anthropic's infrastructure.
+Anthropic's server tools can be enabled when calling the model. These tools are executed on Anthropic's
+infrastructure. Enable them with the ``server_tools`` option, a map of tool name to parameters (``true`` enables a
+tool with no extra parameters)::
+
+    $result = $platform->invoke('claude-sonnet-4-5-20250929', $messages, [
+        'server_tools' => [
+            'web_search' => true,
+            'code_execution' => true,
+        ],
+    ]);
+
+The bridge maps each recognized name to Anthropic's versioned tool ``type``. Only ``web_search`` and
+``code_execution`` are mapped, because those are the only server tools whose result blocks
+:class:`Symfony\\AI\\Platform\\Bridge\\Anthropic\\ResultConverter` currently understands. An unrecognized name
+throws :class:`Symfony\\AI\\Platform\\Exception\\InvalidArgumentException`.
 
 Available Server Tools
 ----------------------
+
+Web Search
+~~~~~~~~~~
+
+The Web Search tool (``web_search``) lets the model search the web and ground its response in the results::
+
+    $result = $platform->invoke('claude-sonnet-4-5-20250929', $messages, [
+        'server_tools' => [
+            'web_search' => ['max_uses' => 3],
+        ],
+    ]);
+
+Any parameters given (e.g. ``max_uses``, ``allowed_domains``, ``blocked_domains``, ``user_location``) are merged
+into the tool definition Anthropic sends, and this also lets a caller override the versioned ``type`` the bridge
+defaults to::
+
+    $result = $platform->invoke('claude-sonnet-5', $messages, [
+        'server_tools' => [
+            'web_search' => ['type' => 'web_search_20260209', 'max_uses' => 3],
+        ],
+    ]);
+
+The bridge defaults to ``web_search_20250305``, the basic variant every Claude model supports. Anthropic also
+offers ``web_search_20260209`` (dynamic filtering, Claude 4.6 and later) and ``web_search_20260318`` (adds
+``response_inclusion``). A caller can opt into either per call, without falling back to the raw ``tools``
+escape hatch.
+
+.. caution::
+
+    From ``web_search_20260209`` on, Anthropic runs the search from inside code execution to filter the
+    results, and provisions that itself, so ``code_execution`` is not declared alongside it. Two consequences:
+    the response then also carries code execution blocks that
+    :class:`Symfony\\AI\\Platform\\Bridge\\Anthropic\\ResultConverter` does not convert, and ``allowed_callers``
+    defaults to code execution, so a model without programmatic tool calling needs
+    ``'allowed_callers' => ['direct']`` or the API answers 400.
 
 Code Execution
 ~~~~~~~~~~~~~~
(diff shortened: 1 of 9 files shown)


chr-hertel:
Hey @soyuka, thanks for that - let's use to sort the situation a bit - thinking out loud here:

* with #1937 we brought in an example and docs for folding those server tools for anthropic into the existing `tools` option - what you under the hood do here as well
* with [Gemini/Vertex Server Tools](https://symfony.com/doc/current/ai/components/platform/gemini-server-tools.html) we use the `server_tools` option

on the one hand i feel like adopting `server_tools` feels more consistent and i bet every dev also differentiates between local and server tools - but on the other hand, if it ends up in the same option - a bit weird :thinking:

i think i would appreciate to streamline docs for promoting `server_tools` - we don't need to block the usage via `tools`

cc @valtzu

soyuka:
I can document that if needed lmk.

valtzu:
It would be nice to have some abstraction so that the same would work for any platform that supports the feature. Then it would be very similar to capabilities, or maybe we could even use capabilities?

E.g. a model would have `Capability::WebSearchServerTool`

Then you'd have 
```yaml
ai:
  platform:
    antropic:
      # ..
      server_tools:
        web_search: true
    gemini:
      # ..
      server_tools:
        web_search: true
    failover:
      anthropic_to_gemini:
        platforms:
          - 'ai.platform.anthorpic'
          - 'ai.platform.gemini'
```

and then do the routing & constructing the request instead of current pass-through as-is.

For simple things like web search and code execution having the platform-specific config in yaml is not an issue, but some server tools take input (e.g. location-based ones) so for those its usually not feasible to have them within platform config, and by putting them into code you make the code platform-specific, which may be unacceptable in failover setups.

And of course it must still be possible to define platform-specific tools (in a pass-thru manner), not to handcuff people from using the latest features that have not yet landed on `symfony/ai`.

---

I understand this may be a stretch.

src/platform/src/Bridge/Anthropic/ModelClient.php
```diff
@@ -66,6 +66,11 @@ public function request(Model $model, array|string $payload, array $options = []
         $payload = $this->injectCacheControl($payload);
         $payload = $this->injectSystemCacheControl($payload);
 
+        if (isset($options['server_tools'])) {
+            $options['tools'] = [...$options['tools'] ?? [], ...$options['server_tools']];
+            unset($options['server_tools']);
```
wachterjohannes, line 71: The naming question is worth settling on shape rather than taste. Gemini and VertexAI already have `server_tools`, but as a **map of tool name to params**:

```php
'server_tools' => ['google_search' => true, 'url_context' => true]
```

This PR uses the same option name for a **list of raw Anthropic tool definitions**. That gets the worst of both: it reads as consistent while staying unportable, and it forecloses @valtzu's cross-platform idea instead of enabling it. The Gemini-shaped equivalent would be:

```php
'server_tools' => ['web_search' => ['max_uses' => 3]]
```

with the bridge mapping `web_search` to Anthropic's `['type' => 'web_search_20250305', ...]`. Raw pass-through stays valuable for tools we haven't mapped yet — but as an escape hatch via `tools`, which already works today, not as the primary shape.

---

Separately from the naming: `web_search` doesn't work end-to-end yet, even though it's the example in the changelog entry and both tests.

`ResultConverter` handles `server_tool_use` only for `bash_code_execution` and `text_editor_code_execution`, and has no `web_search_tool_result` branch at all. So the request goes out, Claude runs the search, and every search block is silently dropped — or, if the reply carries no text block, it throws `Response content does not contain any supported content.`

Platform already has `WebSearchResult`/`WebSearch` (the OpenResponses bridge uses them), so the mapping target exists; it just isn't wired for Anthropic. Whichever option shape we land on, the converter side needs to come with it.

Also worth confirming against the live API: `injectToolsCacheControl()` puts `cache_control` on the last tool, which after this merge is a server tool — does Anthropic accept that on a `web_search` definition? I have no key to check.

src/platform/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ CHANGELOG
 0.10
 ----
 
+ * Add support for server-side Anthropic tools (e.g. `web_search`) via the `server_tools` option, appended verbatim to the request tools
```
wachterjohannes, line 7: This section is long released — latest tag is `v0.12.0` and the unreleased section at the top of this file is `0.13`. Entry needs to move up there (it merges cleanly into `0.10`, which is what makes it easy to miss).

wachterjohannes (changes requested):
Sorry for the long silence here @soyuka. Picking up @chr-hertel's open question — I think there's a concrete argument for `server_tools`, but not in this shape. Details inline.

@chr-hertel what do you think? My reasoning is that Gemini/VertexAI's `server_tools` is a name→params map while this one is a raw pass-through list, so adopting the name without the shape looks consistent without actually being portable. If you'd rather keep the pass-through and just streamline the docs as you suggested, I'm happy to go that way instead — you opened the question, so it's your call.

Two smaller things while you're in here @soyuka: the branch is ~290 commits behind main, and yes please to your docs offer — `docs/components/platform/anthropic-server-tools.rst` currently documents code execution only.

soyuka:
Thanks @wachterjohannes - you were right on both counts, and I've reworked this rather than defend the original shape.

@chr-hertel you asked the shape question back in June and it's been open since; rather than leave it sitting I've gone with the Gemini/VertexAI map shape, since that's where both of you leaned. Happy to revisit if you'd rather have pass-through plus docs.

What changed:

- `server_tools` is now a `name => params` map, mapping to Anthropic's versioned `type`, matching Gemini/VertexAI. Raw `tools` stays the escape hatch for unmapped tools and is now documented as such - including for `web_fetch`, which isn't mapped because its results aren't converted yet.
- `web_search` results are wired up. You were right that `ResultConverter` had no `web_search_tool_result` branch - a search was silently dropped, or threw when the reply had no text block. Both are now covered by tests. The call and result blocks merge into one `WebSearchResult` (query + id + status), matching how OpenResponses represents a search.
- On `cache_control`: good catch, and I didn't verify it against the live API either - so I removed the question instead of answering it. Server tools are appended after cache-control injection, so it always lands on the last function tool. There's a regression test asserting it is *not* on the server tool.
- CHANGELOG entry moved to `0.13`.
- Docs rewritten, `examples/anthropic/server-tools-web-search.php` added, and the two code-execution examples migrated to `server_tools` so they match the docs.
- Rebased fresh onto main.

One limitation I want to flag rather than bury: `WebSearchResult` has no field for individual search hits, and carrying them means changing a Platform-wide class OpenResponses also populates. Same for streaming - `convertStream()` handles text, thinking and tool calls only, so server tool blocks don't surface there. Both are noted in the RST, with follow-ups for search hits, `citations`, `encrypted_content` round-tripping, and the missing `Content\WebSearch` branch in `AssistantMessageNormalizer`.

The two red `Unit / Store / S3Vectors` checks are a pre-existing async-aws pagination failure, unrelated to this branch.

---

*This comment was drafted by Claude Code (AI) and reviewed by me before posting.*

docs/components/platform/anthropic-server-tools.rst
```diff
@@ -1,16 +1,64 @@
...
+``web_search_20260209`` adds dynamic filtering but requires Opus 4.6+ or Sonnet 4.6+, so it is opt-in per call
+rather than the default - a caller on a newer model can request it without falling back to the raw ``tools``
+escape hatch.
+
+.. caution::
+
+    ``web_search_20260209`` runs code execution internally to perform its dynamic filtering. Declaring
+    ``code_execution`` alongside it (``'server_tools' => ['web_search' => ['type' => 'web_search_20260209'],
+    'code_execution' => true]``) gives the model two separate execution environments; avoid combining them unless
+    that is intentional.
```
wachterjohannes: The caution isn't what Anthropic documents. They say the API provisions the code execution for dynamic filtering itself, so you don't declare `code_execution` for it; there's nothing about two execution environments.

Two things that do belong here, and one is the same argument the allow-list makes above ("only server tools whose result blocks the converter can round-trip"): a `20260209` response carries code execution blocks this converter doesn't handle, and `allowed_callers` defaults to code execution from that version on, so a model without programmatic tool calling gets a 400 without `'allowed_callers' => ['direct']`. There are also three versions now, not two.

```suggestion
The bridge defaults to ``web_search_20250305``, the basic variant every Claude model supports. Anthropic also
offers ``web_search_20260209`` (dynamic filtering, Claude 4.6 and later) and ``web_search_20260318`` (adds
``response_inclusion``). A caller can opt into either per call, without falling back to the raw ``tools``
escape hatch.

.. caution::

    From ``web_search_20260209`` on, Anthropic runs the search from inside code execution to filter the
    results, and provisions that itself, so ``code_execution`` is not declared alongside it. Two consequences:
    the response then also carries code execution blocks that
    :class:`Symfony\\AI\\Platform\\Bridge\\Anthropic\\ResultConverter` does not convert, and ``allowed_callers``
    defaults to code execution, so a model without programmatic tool calling needs
    ``'allowed_callers' => ['direct']`` or the API answers 400.
```

wachterjohannes (commented):
Thanks for the rework, @soyuka. The map shape is the right call, and I checked both halves rather than taking them on trust: without the converter change a web-search-only response still throws `Response content does not contain any supported content.`, with it you get a `WebSearchResult`; `cache_control` stays on the function tool; `server_tools` never reaches the body. Verified locally, all green. One ask inline, on the docs.

Two notes, nothing to do here:

- A long search can come back as `stop_reason: pause_turn`, which maps to `FinishReasonCase::OTHER` with no resume path, so the new example would print the search line and stop. Pre-existing, but web search is the feature that triggers it.
- The shape now matches Gemini and VertexAI, the names still don't (`web_search` vs `google_search`), so @valtzu's failover example needs one more step. Worth its own issue.

@chr-hertel you opened the shape question in June and I handed it back to you in August, so this shouldn't merge on silence. Two things for your call: this picks the Gemini/VertexAI map shape, and it diverges from those two on unknown names, which they forward verbatim and this one rejects. I think rejecting is right here, because Anthropic needs a versioned `type` the bridge can't guess, but that's a deliberate inconsistency across three bridges sharing one option name.

wachterjohannes (approved):
Both open points are resolved: `server_tools` now maps names to Anthropic's versioned type like Gemini/VertexAI, and `web_search` converts into `WebSearchResult` instead of dropping. Changelog moved to 0.14. Verified live against the Anthropic API: `web_search` returns real results, and `cache_control` stays on the function tool and never lands on `web_search`, which the API accepts.

wachterjohannes:
Thank you @soyuka.

PR #2209, opened 2026-06-21
[Platform][Anthropic] Add a configurable base URL to target compatible endpoints

fabpot:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no 
| Issues        | Fix #... 
| License       | MIT

`ModelClient` currently hardcodes `https://api.anthropic.com`. But in practice, most production setups put an Anthropic-compatible gateway in front of the API (LiteLLM, Cloudflare AI Gateway, ...) for observability, caching, and key management. 

The OpenAI and Generic bridges already accept a base URL for exactly this.

### DIFF
--- a/src/platform/src/Bridge/Anthropic/CHANGELOG.md
+++ b/src/platform/src/Bridge/Anthropic/CHANGELOG.md
@@ -4,6 +4,7 @@ CHANGELOG
 0.11
 ----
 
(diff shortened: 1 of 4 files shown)


chr-hertel (approved):
Thanks, will roll that out as follow up to streamline behavior across bridges 👍

chr-hertel:
Thank you @fabpot.

PR #2210, opened 2026-06-21
[Platform] Skip built-in tool-call output items in OpenAI result converters

### LINKED ISSUE #2149: [Platform] Convert `web_search_call` (and similar remote tools) in `Gpt\ResultConverter` to typed result objects
I am currently trying to enable websearch natively on serverside (directly via OpenAI) instead of adding brave search API (or anything similar) and using its Tool.

Therefore, i added a new InputProcessor, that just adds a new option (web_search) so that the model knows that it's supposed to use websearch:

```php
final readonly class WebSearchInputProcessor implements InputProcessorInterface
{
    public function __construct(private bool $webSearchEnabled) {}

    {
        if (!$this->webSearchEnabled) {
            return;
        }

        $options['tools'] = array_merge($options['tools'] ?? [], [['type' => 'web_search']]);
        $input->setOptions($options);
    }
}
```

Using curl, its working fine. Also checking my logs shows that websearch was enabled and actually being used.

**Issue:** `Gpt\ResultConverter::convertOutputArray()` strips `function_call` items via `extractFunctionCalls()`, then passes the remaining items to `processOutputItem()`. The `web_search_call` item is not a `function_call`, so it is not stripped and reaches `processOutputItem()`:

```php
private function processOutputItem(array $item): iterable
{
    $type = $item['type'] ?? null;

    return match ($type) {
        'message'   => $this->convertOutputMessage($item),
        default     => throw new RuntimeException(\sprintf('Unsupported output type "%s".', $type)),
    };
}
```

In there, the `RuntimeException` is being thrown.

Same issue in `symfony/ai-open-responses-platform` and `symfony/ai-open-ai-platform`

Imo, a fix could be either just skipping items of type "web_search_call", or generating a "WebSearchResult" for these cases.
Of course, doing the search on my own with my own Search-API-Tool also works. But i want to try it natively on OpenAI-Side.

If the current behavior is intended and i'm just missing something, please let me know.

OskarStark:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | -
| License       | MIT

When a built-in server-side tool such as `web_search` runs, the OpenAI Responses API returns a `web_search_call` item in the response `output[]` array alongside the assistant `message`. Both the OpenAI (`Gpt`) and OpenResponses result converters only handled the `message` and `reasoning` output types and threw `RuntimeException: Unsupported output type "web_search_call"` on anything else, aborting conversion before the `message` item carrying the answer was read. As a result, any agent enabling a built-in tool failed.

This skips the built-in server-side tool-call output items (`web_search_call`, `file_search_call`, `code_interpreter_call`, `image_generation_call`, `computer_call`, `local_shell_call`, `mcp_call`, `mcp_list_tools`, `mcp_approval_request`) — they carry no assistant-facing result — so the assistant message is still converted. Tests added to both bridges' `ResultConverterTest`.

### DIFF
--- a/examples/gemini/server-tools.php
+++ b/examples/gemini/server-tools.php
@@ -10,9 +10,6 @@
  */
 
 use Symfony\AI\Agent\Agent;
-use Symfony\AI\Agent\Bridge\Clock\Clock;
-use Symfony\AI\Agent\Toolbox\AgentProcessor;
-use Symfony\AI\Agent\Toolbox\Toolbox;
 use Symfony\AI\Platform\Bridge\Gemini\Factory;
 use Symfony\AI\Platform\Message\Message;
 use Symfony\AI\Platform\Message\MessageBag;
--- a/src/platform/src/Bridge/OpenAi/CHANGELOG.md
+++ b/src/platform/src/Bridge/OpenAi/CHANGELOG.md
@@ -4,6 +4,8 @@ CHANGELOG
 0.11
 ----
 
+ * Throw `ServerException` on server errors (HTTP 5xx) instead of a generic `RuntimeException`
+ * Throw typed exceptions on rate limit and server error events mid-stream
  * Raise a `RuntimeException` on unhandled HTTP error statuses before streaming, instead of returning an empty stream
 
 0.10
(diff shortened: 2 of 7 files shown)


chr-hertel (approved):
Agreed, let's treat this as bug fix first - will revert CHANGELOG entires while merging.

chr-hertel:
Thank you @OskarStark.

PR #2211, opened 2026-06-21
[Platform][OpenAI] Skip responses built-in tool output items instead of throwing

OskarStark:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | --
| License       | MIT

When a built-in server-side tool (web_search, file_search, MCP, …) runs, the OpenAI Responses API returns its own `*_call` item in the `output[]` array alongside the assistant message. The OpenAI and OpenResponses result converters only handled `message`, `reasoning` and `function_call` and threw `Unsupported output type` on everything else, aborting the whole response before the message that carries the answer was read.

Skip the known server-side tool-call markers so the actual message item is still converted, keeping the `default` throw for genuinely unknown types. Adds regression tests for both converters and runnable web-search examples for the OpenAI and OpenResponses bridges.

### DIFF
--- a/src/platform/src/Bridge/OpenAi/Gpt/ResultConverter.php
+++ b/src/platform/src/Bridge/OpenAi/Gpt/ResultConverter.php
@@ -152,6 +152,13 @@ final class ResultConverter implements ResultConverterInterface
         return match ($type) {
             'message' => $this->convertOutputMessage($item),
             'reasoning' => $this->convertReasoning($item),
+            // Built-in server-side tool calls (web_search, file_search, code_interpreter,
+            // image generation, computer use, local shell, MCP, …) appear as their own
+            // output items but carry no assistant-facing result. Skip them so the actual
+            // message item is still converted instead of aborting the whole response.
+            'web_search_call', 'file_search_call', 'code_interpreter_call',
+            'image_generation_call', 'computer_call', 'local_shell_call',
+            'mcp_call', 'mcp_list_tools', 'mcp_approval_request' => [],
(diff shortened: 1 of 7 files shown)


src/platform/src/Bridge/OpenAi/Gpt/ResultConverter.php
```diff
@@ -152,6 +152,13 @@ private function processOutputItem(array $item): iterable
         return match ($type) {
             'message' => $this->convertOutputMessage($item),
             'reasoning' => $this->convertReasoning($item),
+            // Built-in server-side tool calls (web_search, file_search, code_interpreter,
+            // image generation, computer use, local shell, MCP, …) appear as their own
+            // output items but carry no assistant-facing result. Skip them so the actual
+            // message item is still converted instead of aborting the whole response.
+            'web_search_call', 'file_search_call', 'code_interpreter_call',
+            'image_generation_call', 'computer_call', 'local_shell_call',
+            'mcp_call', 'mcp_list_tools', 'mcp_approval_request' => [],
```
chr-hertel, line 161: i'm wondering what we could return instead of `[]` - i mean it's unlikely that we'd hit the `if ([] === $results)` in the `convert(...)` method, raising a `RuntimeException` then, but with Gemini and Anthropic we convert the results of remote tool calls to a object tho, e.g. `CodeExecutionResult`

short coming here, but fair as first step i'd say

OskarStark:
Duplicate of #2210

PR #1718, opened 2026-03-03
[Agent] Add support for `AgentSkills`

Guikingone:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | --
| License       | MIT

This PR adds comprehensive https://agentskills.io/specification support to the `Agent`, covering skill loading, validation, evaluation, and bundle integration.

# Skill Loading

  - `GithubSkillLoader`: loads skills from GitHub repositories via the Contents API. Supports public/private
  repos (token auth), configurable branch and subdirectory path.
  - `ChainSkillLoader`: composes multiple `SkillLoaderInterface` implementations with first-match-wins
  semantics for `loadSkill()` and aggregation for `loadSkills()`/`discoverMetadata()`.
  - `SkillParserInterface::parseFromContent()` / `parseMetadataFromContent()`: source-agnostic parsing methods
  enabling skills to be loaded from any source (GitHub, database, etc.) without filesystem access.

# Skill Evaluation System

  A complete evaluation framework for measuring agent quality improvement when using a skill:

  - `EvalSuiteLoader`: loads evals/evals.json test suites from skill directories.
  - `EvalRunner`: runs eval cases against an agent, collecting output, timing, and token usage.
  - `LlmGrader`: grades agent output against natural-language assertions using an LLM.
  - `BenchmarkAggregator`: compares with-skill vs without-skill runs (pass rate, time, tokens).
  - `WorkspaceManager`: persists evaluation results (outputs, timing, grading, benchmarks) to disk.

# Bundle Integration

  - YAML configuration for github_repositories (repository, path, branch, token) and evaluation (workspace,
   grading_model, grading_platform) under ai.agent.<name>.skills.
  - Automatic wiring: when github_repositories is configured alongside the default filesystem loader, a
  `ChainSkillLoader` is transparently registered.
  - `ai:agent:validate-skills command`: validates skills against the spec, supports --skill for single skill
  or all-skills mode with summary.
  - `ai:agent:eval-skill command`: runs evaluation suites with --agent, --baseline-agent, --iteration,
  --skip-grading options and outputs benchmark comparison tables.
  - `TraceableSkillLoader` / `TraceableSkillValidator`: profiler decorators for debug toolbar integration.

# Documentation & Examples

  - RST documentation in docs/components/agent.rst and docs/bundles/ai-bundle.rst covering skill structure,
   all loaders, evaluation system, bundle configuration, and commands.
[...]

PR #2213, opened 2026-06-21
[Mate] Add skills:install to install extension skills for coding agents

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | -
| License       | MIT

Agent Skills (`SKILL.md`) shipped by Mate extensions are now installed onto the
filesystem where coding agents read them — a polyfill until skills can be served
over MCP directly.

### Background — MCP Skills (SEP-2640)

This is the **consumer/polyfill side** of the MCP Skills effort:

- **SEP-2640** proposes serving Agent Skills over MCP's Resources primitive as
  `skill://` resources — modelcontextprotocol/modelcontextprotocol#2640.
- **#2132** already added skill *content* to the Mate Symfony bridge, anticipating
  the extension (agentskills.io + SEP-2640 layout) — #2132.
- **modelcontextprotocol/php-sdk#372** implements the *serving* half in the PHP
  SDK (`skill://` resources + discovery index).

The missing piece is **getting those skills onto the filesystem**: Claude Code and
other agents do not consume MCP-served skills yet (June 2026) — they only read
skills from disk. This PR bridges that gap and can be retired once agents consume
MCP-served skills end-to-end.

### How it works

Extensions declare one or more skills directories via a new `extra.ai-mate.skills` key:

```json
{
    "extra": {
        "ai-mate": {
            "scan-dirs": ["src"],
            "skills": ["skills"]
        }
    }
}
```

Each immediate subdirectory holding a `SKILL.md` is one skill. Mate **symlinks** each one
— under a `mate-` prefixed directory name (e.g. `mate-system-information`) to avoid
clashing with skills you maintain from other sources — into `.agents/skills/` (read by
Codex, OpenCode and GitHub Copilot) and mirrors it into `.claude/skills/` for Claude Code,
which only reads its own directory. The links point into the gitignored `vendor/`, so
skills auto-update on `composer update`.

### Usage

Skills install automatically as part of `discover`, which already runs after
`composer require`, so there's usually nothing to run:

```terminal
$ vendor/bin/mate discover
```

For an explicit re-sync:

```terminal
$ vendor/bin/mate skills:install
```

The core package ships a built-in `system-information` skill to demonstrate the
format and dogfood the mechanism.

### Notes

- Documentation added under `docs/components/mate.rst` and
  `docs/components/mate/creating-extensions.rst`; `CHANGELOG.md` updated.
- Covered by tests (installer, command, discovery, and the `discover` auto-sync);
  PHPStan clean.

### DIFF
--- a/docs/components/mate.rst
+++ b/docs/components/mate.rst
@@ -419,6 +419,31 @@ The core package provides basic system information tools:
 
 * ``server-info`` - Get PHP runtime environment details: version, OS, OS family, and loaded extensions
 
+Skills
+------
+
+`Agent Skills <https://agentskills.io>`_ are ``SKILL.md`` files that give your coding agent
+structured, multi-step "how-to" knowledge for a task. Extensions can ship skills alongside their MCP tools, and Mate
+installs them onto the filesystem where coding agents read them — a polyfill until skills can be
+served over MCP directly.
+
+You usually do not run anything: ``mate discover`` (which also runs automatically after
+``composer require``) installs the skills of every enabled extension. To sync them manually, use
+``mate skills:install``.
+
+Each skill is installed under a ``mate-`` prefixed directory name (e.g. ``mate-demo-skill``) to
+avoid clashing with skills you maintain from other sources, in two locations:
+
+* ``.agents/skills/`` is the source of truth, read directly by Codex, OpenCode and GitHub Copilot.
+* ``.claude/skills/`` is symlinked to ``.agents/skills/`` for Claude Code, which only reads its own
+  directory.
+
+Skills are symlinked into ``.agents/skills/`` so they auto-update on ``composer update`` (the link
+points into the gitignored ``vendor/`` directory, so this requires symlink privileges on Windows).
+
+The core package itself ships a ``system-information`` skill describing how to inspect the PHP
+runtime and installed package versions via the ``server-info`` tool.
+
 Commands
 --------
 
@@ -432,6 +457,12 @@ Commands
     - Generate or update ``mate/extensions.php`` with discovered extensions
     - Preserve existing enabled/disabled states for known extensions
     - Default new extensions to enabled
+    - Install Agent Skills shipped by enabled extensions (see `Skills`_)
+
+``mate skills:install``
+    Install the Agent Skills shipped by your enabled extensions so your coding agent can use
+    them. This runs automatically as part of ``mate discover``; use it for an explicit re-sync.
+    See `Skills`_.
 
 ``mate serve``
     Start the MCP server with stdio transport.
--- a/src/mate/skills/system-information/SKILL.md
+++ b/src/mate/skills/system-information/SKILL.md
@@ -1,0 +1,72 @@
+---
+name: system-information
+description: >-
+  Inspect the runtime and dependency context of the PHP application Mate is attached to —
+  PHP version, OS, and loaded extensions via the server-info tool, plus installed package
+  versions from composer.lock — when diagnosing environment- or version-specific behavior
+  such as a missing extension or a version mismatch (an API that does not exist in the
+  release that is actually installed).
+---
+
+# System Information
+
+Use this skill when a problem might depend on the **environment or the installed
+dependency versions** rather than the code: a missing or misconfigured extension,
+behavior that only reproduces on a specific PHP/OS version, or a bug that looks like a
+version mismatch — a method, argument, or class that does not exist in the release that
+is actually installed.
+
+## Capabilities
+
+| Need | Use |
+|---|---|
+| PHP version, OS, OS family, loaded extensions | `server-info` MCP tool (one call) |
+| Installed version of a package | `composer show <vendor/package>` |
+| Installed versions of a family of packages | `composer show '<vendor>/<prefix>*'` |
+| Authoritative record of every installed version | `composer.lock` (`packages` / `packages-dev`) |
+
+`server-info` reports the runtime only — it does **not** return package versions. Use
+Composer for those.
+
+## Runtime checks
+
+1. **Start with the `server-info` tool** for PHP version, OS, OS family, and loaded
+   extensions — prefer it over `php -v`, `php -m`, or `uname`.
+2. **Confirm a required extension is loaded** before blaming the code. If it is absent
+   from `server-info`, the fix is an environment/configuration change, not a code change.
+
+## Resolving package versions
+
+`composer.json` declares *constraints* (what is allowed); `composer.lock` records what is
+*actually installed*. For "what is really running", trust the lock, not the constraint.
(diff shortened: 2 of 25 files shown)


Guikingone:
What about using https://packagist.org/packages/guikingone/agent-skills as the retrieval layer for skills? Already implement the specifications and allows to add new loader if needed 🤔 

This could also trigger discussion about #1718

wachterjohannes:
Thanks @Guikingone — looked at both `agent-skills` and #1718. I think we're solving two different layers, so I don't think the package replaces this PR:

- **`agent-skills` / #1718** load skills *into* a PHP agent runtime (system-prompt injection, skill-as-tool, eval/grading).
- **This PR** goes the other way: it *materializes* skills onto disk (`.agents/skills/` + symlinks into `.claude/skills/`) so external coding-agent CLIs read them — the polyfill side of SEP-2640 until they consume `skill://` directly.

The only real overlap is the thin "parse `SKILL.md`" slice. That's worth sharing — but I'd rather see a common spec/parse abstraction land **first-party in the Agent component via #1718** than have Mate core take a third-party dependency. Happy to align this PR with whatever #1718 settles on.

wachterjohannes:
@kbond i have striped down the scope of this PR i have a few ideas for further PRs but that is not important for the first tests here.

@chr-hertel you can review but i would wait for @kbond to test this feature with maker skills

kbond:
@wachterjohannes Ok, I finally got around to testing this.

I used zenstruck/foundry as an example package to add skills. Here's [the branch](https://github.com/zenstruck/foundry/compare/2.x...kbond:foundry:ai-skills).

It worked as expected but I had to run `vendor/bin/mate discover` explicitly to sync. I have the plugin installed/enabled. Maybe something to do with my local vendor setup?

I love the simplicity of this! Let's discuss how to move forward!

wachterjohannes:
@kbond thanks for the feedback - this first version is indeed a very simple version that just proves that it works and make sense.

As discussed in slack one-by-one i think your developer setup with the link binary do not trigger the composer plugin and thats the reason why it does not run discover automatically.

The next steps should be following:

1. En/Disable Skills - so that they will not be symlinked or removed from the "blacklist"
2. Overwrite Skills- that will copy the skill into the directory instead of symlink and not overwrite it in any later on run of skills install
3. Make the init command more interactive 
  - ask for agents (codex, claude, ...) so that the symlinks, wrappers, ... can be build ontop of that information
  - ask for the purpose - mcp or skills or both - so that the developer can decide what he actually needs
  - maybe other questions to individualize that
5. Some kind of debugging command

docs/components/mate.rst
```diff
@@ -413,6 +413,31 @@ The core package provides basic system information tools:
...
+structured, multi-step "how-to" knowledge for a task. Extensions can ship skills alongside their MCP tools, and Mate
+installs them onto the filesystem where coding agents read them — a polyfill until skills can be
+served over MCP directly.
+
+You usually do not run anything: ``mate discover`` (which also runs automatically after
+``composer require``) installs the skills of every enabled extension. To sync them manually, use
+``mate skills:install``.
+
+Each skill is installed under a ``mate-`` prefixed directory name (e.g. ``mate-demo-skill``) to
+avoid clashing with skills you maintain from other sources, in two locations:
```
chr-hertel, line 435: avoiding vendor clashes would be vendor responsibility?
e.g. vendor a and vendor b would both ship a `create-table` skill

docs/components/mate.rst
```diff
@@ -413,6 +413,31 @@ The core package provides basic system information tools:
...
+
+* ``.agents/skills/`` is the source of truth, read directly by Codex, OpenCode and GitHub Copilot.
+* ``.claude/skills/`` is symlinked to ``.agents/skills/`` for Claude Code, which only reads its own
+  directory.
+
+Skills are symlinked into ``.agents/skills/`` so they auto-update on ``composer update`` (the link
+points into the gitignored ``vendor/`` directory, so this requires symlink privileges on Windows).
```
chr-hertel, line 442: does that also apply when `vendor/abc:1.5` added a new skill that 1.4 didn't have? we'd need another discover/install?

src/mate/skills/system-information/SKILL.md
```diff
@@ -0,0 +1,72 @@
+---
+name: system-information
```
chr-hertel, line 2: when testing, i would expect that to be available in demo app

chr-hertel (approved):
conceptual questions to scope out to a follow-up only - nothing blocking

demo/.gitignore
```diff
@@ -21,4 +21,9 @@ phpstan.neon
 .phpunit.cache
 ###< phpunit/phpunit ###
 
+###> symfony/ai-mate ###
+/.agents/skills/
+/.claude/skills/
+###< symfony/ai-mate ###
```
chr-hertel: was this "manually" added?

docs/components/mate.rst
kbond: I think so, this is the case with the maker-bundle (at least for most 3rd party makers I've seen). I'd be fine leaving clash resolution/warning to a future iteration.

docs/components/mate.rst
wachterjohannes: Good catch — two different things here:

The symlink only keeps the content of an already-installed skill fresh. A newly added skill directory in abc:1.5 isn't covered by that — but you don't need a manual step either: composer update fires POST_UPDATE_CMD, the Mate plugin runs mate discover, and that runs skills:install, which picks up the new skill folder. So on a normal Composer-installed setup it's automatic.

The one exception is a locally linked dev setup (like kbond's), where the Composer plugin never fires — there you re-run vendor/bin/mate discover yourself.

docs/components/mate.rst
wachterjohannes: added an issue to track that https://github.com/symfony/ai/issues/2355

src/mate/skills/system-information/SKILL.md
chr-hertel: So with this, it should work:
```bash
gh pr checkout 2213
cd demo
composer update
../link
vendor/bin/mate discover
claude
/reload-skills
/system-information
```
(only on my mobile, but to verify)

chr-hertel:
Btw, rebase should fix the pipeline

src/mate/skills/system-information/SKILL.md
wachterjohannes: yes it works - but i also have just removed the gitignore block and commited it - that avoids this issue for existing skills

chr-hertel:
Thank you @wachterjohannes.

PR #2214, opened 2026-06-21
[McpBundle] OAuth 2.1 authorization server via league/oauth2-server-bundle

wachterjohannes:
## Summary

Adds an opt-in OAuth 2.1 **authorization server** to the MCP bundle, backed by [`league/oauth2-server-bundle`](https://github.com/thephpleague/oauth2-server-bundle).

This follows `mcp/sdk` ADR 0001 (modelcontextprotocol/php-sdk#382), which scopes the SDK to **resource-server only** and recommends running `league/oauth2-server` in the application/bundle for the authorization-server role. So the AS lives entirely in this bundle + the consuming app — the SDK is never asked to mint tokens.

league owns the heavy lifting (`/authorize` + `/token`, the `oauth2` firewall authenticator, PKCE S256, RS256 JWT issuance, refresh rotation, client/token persistence). The bundle adds only the **four MCP-specific pieces league does not ship**.

## What it does

- **`McpBundle::prependExtension()`** derives the full `league_oauth2_server` config from the `mcp.oauth` options: private/public keys, encryption key, TTLs (→ `DateInterval`), auth-code + refresh grants only, PKCE required, and `doctrine`/`in_memory` persistence.
- **Slim `OAuthController`** serves the gaps: RFC 7591 dynamic client registration (builds a league `Client` via `ClientManagerInterface`), RFC 8414 authorization-server metadata, RFC 9728 protected-resource metadata, and the JWKS document.
- **`AuthorizationRequestResolveListener`** binds the authenticated firewall user as the OAuth subject and auto-approves, with an optional `ConsentInterface` hook.
- `/authorize` and `/token` point at league's controllers; bearer protection uses league's `oauth2: true` authenticator.

## Notes

- `league/oauth2-server-bundle` is **optional** (`require-dev` + `suggest`); `configureOAuth()` is gated on its presence, and the consuming app registers the bundle in `bundles.php`.
- The bundle no longer references any `Mcp\Server\Transport\Http\OAuth\*` classes.

## Tests

54 bundle tests pass (incl. rewritten OAuth DI tests covering the league prepend, route wiring, the resolve listener and config validation); PHPStan clean.

## Status

**Draft** — opened for direction feedback; validated end-to-end against a consuming app (Sulu MCP) at the container level (`lint:container`, derived config, routes), with a full token-flow run still pending. Supersedes the native-authorization-server approach previously proposed.


chr-hertel:
still my argument is similar: if users have that case, the bundle is not in their way. the extension of the config alone demonstrates the sheer size of this feature, and i don't think that's sustainable to maintain here at this point. I can easily believe that this is working implementation of DCR, but also sure there are quite variations and change scenarios to it.

Regarding AUTH and MCP in general:
With https://github.com/modelcontextprotocol/php-sdk/pull/391 i'm bringing in a rework that hopefully helps to expose that RS wellknown easily, but that's the scope then:
* scope of SDK defined by spec
* the bundle integrates the SDK => RS wellknown endpoint
* symfony/security recommended as a start for more
* more sophisticated is third party lib or application code

sorry for opting out on this one, but that's my call

PR #2215, opened 2026-06-21
[Platform][Store] Streamline base URL handling across bridges

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | -
| License       | MIT

Generalizes the trailing-slash robustness from #2209 to every HTTP platform and store bridge, and makes base-URL/endpoint handling consistent.

- Canonical pattern everywhere: accept a base URL, store it `rtrim($baseUrl, '/')`, concatenate the path.
- Make previously-hardcoded platform bridges configurable via a `baseUrl` arg (Mistral, DeepSeek, Cerebras, Perplexity, Cohere, Voyage, Cartesia, Scaleway, Gemini, Replicate, HuggingFace, OpenRouter).
- Azure now accepts a full base URL *with* scheme (bare host still works, https assumed) instead of rejecting any protocol; shared `BaseUrlNormalizerTrait`.
- Standardize naming: platform `$hostUrl` → `$baseUrl` (Decart, DockerModelRunner); HTTP store endpoints → `$endpoint` (ManticoreSearch, Cloudflare, Milvus, Supabase, SurrealDb).
- Albert/ElevenLabs and ScopingHttpClient store bridges now tolerate a missing/extra trailing slash.

BC breaks documented in `UPGRADE.md`; per-bridge CHANGELOG entries added; affected RAG examples updated.

Deliberate exceptions (cloud-provider-specific endpoints): VertexAi, Bedrock, OpenAi (region-based), Anthropic (#2209), and HuggingFace's metadata host.

cc @Guikingone WDYT?

### DIFF
--- a/src/platform/src/Bridge/Albert/CHANGELOG.md
+++ b/src/platform/src/Bridge/Albert/CHANGELOG.md
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+0.11
+----
+
+ * Tolerate a trailing slash on the base URL instead of rejecting it
+
 0.8
 ---
 
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -51,6 +51,31 @@ Platform
    `$model->supports(Capability::INPUT_MULTIPLE)` for an embedding model will now receive `false`; pass
    the full array of texts to `Platform::invoke()` directly instead — the bridge handles batching.
 
+ * Base URLs are now normalized across all bridges: a trailing slash on a configured base URL is
+   tolerated and no longer produces a double slash in request URLs.
+
+ * The Azure bridge now accepts a full base URL *with* scheme and no longer rejects it. A bare host
+   keeps working (the `https` scheme is assumed when none is given), so existing configurations are
+   unaffected, but you can now point it at an `http://` gateway:
+
+   ```diff
+   -// previously a scheme was rejected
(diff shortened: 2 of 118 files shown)


src/platform/src/Bridge/Albert/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ CHANGELOG
 0.8
```
OskarStark, line 9: 0.11 for all

Guikingone:
IMHO, on stores, that's not necessary as we tend to use `HttpClientInterface` without endpoint in the store class, the client being built in the factory, the store use the client without the endpoint.

The "end goal" would be to do to the same thing for platforms to prevent this kind of "in-class" magic.

IMHO, we should focus on bringing support for "scoped client" (like Ollama, ElevenLabs, etc) instead of standardizing endpoint usage 🤔

chr-hertel:
@Guikingone true, wherever we use the `ScopingHttpClient` already we can drop that change already, yes :+1: will shrink it down, but for some use cases it will already help - know for a fact that some are running into this while adopting it :angel: :laughing: 

edit, wait no, tested with the TypeSense, which is already ScopingHttpClient, and the added tests wouldn't run without the `rtrim` addition

PR #2220, opened 2026-06-22
[Platform][Generic] Treat empty-string streamed tool-call id as a continuation

soyuka:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | -
| License       | MIT

OpenAI sends a streamed tool call's `id` and `function.name` only on the first delta and omits `id` on the continuation (arguments) deltas. Some OpenAI-compatible providers (Alibaba Cloud Qwen / DashScope) instead repeat the id as an **empty string** on every continuation delta.

The stream converter keyed "is this a new tool-call start?" on `isset($toolCall['id'])`, which is `true` for `""`, so each continuation was misread as a new start: it read `function.name` from a delta that has none (`Undefined array key "name"`) and clobbered the accumulated arguments.

Fix: start a tool call only on a **non-empty** id, in both `convertStreamToToolCalls()` and `yieldToolCallDeltas()`.

A regression test (`testStreamingToolCallsWithEmptyStringIdOnContinuationChunks`) reproduces the DashScope-style stream and asserts a single `ToolCallStart`/`ToolCallComplete` with intact name and accumulated arguments.

### DIFF
--- a/src/platform/src/Bridge/Generic/Completions/CompletionsConversionTrait.php
+++ b/src/platform/src/Bridge/Generic/Completions/CompletionsConversionTrait.php
@@ -132,7 +132,11 @@ trait CompletionsConversionTrait
         }
 
         foreach ($data['choices'][0]['delta']['tool_calls'] as $i => $toolCall) {
-            if (isset($toolCall['id'])) {
+            // A new tool call starts only on a NON-EMPTY id. OpenAI omits the id on continuation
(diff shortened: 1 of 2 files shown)


chr-hertel (approved):

chr-hertel:
Thank you @soyuka.

PR #2231, opened 2026-06-24
[Store] Add DirectoryLoader with per-extension sub-loaders

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | 
| License       | MIT

Adds a built-in `DirectoryLoader` to the Store component that scans a directory and delegates each file to a per-extension sub-loader (e.g. `['md' => new MarkdownLoader(), 'txt' => new TextFileLoader()]`), with an optional `recursive` flag. Files with no registered loader are skipped. Generalizes the directory-scanning pattern previously written ad hoc in userland.

### DIFF
--- a/src/store/src/Document/Loader/DirectoryLoader.php
+++ b/src/store/src/Document/Loader/DirectoryLoader.php
@@ -1,0 +1,83 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Store\Document\Loader;
+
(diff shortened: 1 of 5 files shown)


src/store/src/Document/Loader/DirectoryLoader.php
```diff
@@ -0,0 +1,83 @@
...
+            throw new InvalidArgumentException(\sprintf('DirectoryLoader requires an existing directory as source, "%s" given.', $source ?? 'null'));
+        }
+
+        $finder = (new Finder())
+            ->files()
+            ->in($source)
+            ->sortByName();
+
+        if (!$this->recursive) {
+            $finder->depth('== 0');
```
chr-hertel: we could just go with

```suggestion
            $finder->depth(0);
```
right?

PR #2116, opened 2026-05-23
[Platform] Add streaming support for structured output

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | -
| License       | MIT

Combining `stream: true` with `response_format: SomeClass::class` no longer throws — it now yields **progressively populated typed objects** as the model emits tokens.

```php
$result = $platform->invoke('gpt-4o-mini', $messages, [
    'stream' => true,
    'response_format' => Recipe::class,
]);

foreach ($result->asStreamedObject() as $recipe) {
    render($recipe);          // partially populated Recipe (name, then ingredients, then steps)
}

$final = $result->asObject(); // fully materialized, validated Recipe
```

### How it works

`PartialObjectStreamListener` buffers the `TextDelta`s, recovers the largest valid JSON via `PartialJsonParser`, denormalizes it with the existing `StructuredOutput\Serializer`, and emits a `PartialObjectDelta` whenever the parsed shape changes (deduped by hash). `asStreamedObject()` unwraps those deltas and yields the typed object directly; the original `TextDelta`s still flow, so `asTextStream()` is unaffected. `asObject()` drains the stream once and returns the cached final object. If a `ValidatorInterface` is present, only the **final** object is validated — partials are incomplete by definition.

### Validation

- `vendor/bin/phpunit` — green
- `vendor/bin/phpstan analyse` — clean
- Manual: `OPENAI_API_KEY=… php examples/platform/streaming-structured-output.php`

PR #2232, opened 2026-06-24
[Platform] Fix asObject() after partial consumption of streamed structured output

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | -
| License       | MIT

Follow-up to #2116. When `stream: true` is combined with `response_format: SomeClass::class`, calling `asObject()` after iterating `asStreamedObject()` only **partially** — an early `break`, or an exception thrown inside the loop — threw a misleading `RuntimeException: Cannot json decode the streamed content.`

`asObject()` re-drained the stream via `asStream()`, restarting the one-shot generator and reprocessing the boundary delta, which corrupted the listener buffer into invalid JSON.

The fix caches the stream generator on the `DeferredResult` so it is driven exactly once, and has `asObject()` pump the remainder via `valid()/next()` (which resumes without reprocessing) instead of re-iterating. The stream is now finished to completion and the complete, validated object is returned — whether iteration never started or stopped halfway.

Two regression tests cover the early-`break` and exception-during-iteration cases.

cc @wachterjohannes

### DIFF
--- a/src/platform/src/Result/DeferredResult.php
+++ b/src/platform/src/Result/DeferredResult.php
@@ -36,6 +36,12 @@ final class DeferredResult
     private ?\Throwable $conversionFailure = null;
 
     /**
+     * Shared stream generator so the one-shot stream is driven exactly once
+     * across asStream(), asStreamedObject() and asObject().
+     */
+    private ?\Generator $stream = null;
+
+    /**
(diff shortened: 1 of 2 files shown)


wachterjohannes:
@chr-hertel Thank you 😄

PR #2029, opened 2026-05-04
[Platform] Refactor dispatch around EndpointClientInterface

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | -
| License       | MIT

> **Draft / RFC.** This branch is a single sweeping refactor of the Platform dispatch layer; opening it as draft to share the shape and gather feedback. **Do not merge as-is** — see "How to break this up" below.

## Summary

Replaces the per-bridge `ModelClient` + `ResultConverter` pair with a single `EndpointClientInterface` that owns request building, transport invocation, and response parsing for one endpoint contract. `Provider` routes by endpoint identifier (default = first endpoint declared on the `Model`, override via `$options['endpoint']`).

`ModelDescriptor` (intermediate concept introduced earlier in this branch) is folded into `Model`. `ContractHandlerInterface`, `ContractDispatcher`, and `HandlerBoundRawResult` are deleted. `TransportInterface` is kept only where it earns its keep (cross-bridge sharing of a client class) and inlined elsewhere.

Net diff: **216 files, +7535 / -8950**.

## Concerns tackled (each one is a follow-up PR)

This branch tangles together five separable refactors. Each could land independently:

### 1. Endpoint as a first-class concept on `Model`
- New `Endpoint` value object (contract identifier + defaults).
- `Model` gains `getEndpoints()`, `getDefaultEndpoint()`, `getEndpoint(string)`, `supportsEndpoint(string)`.
- Catalogs declare endpoints via the new `AbstractModelCatalog::endpointsForModel()` hook.
- *Minimum viable PR:* just the new types + the hook, no consumer changes. Backward-compatible; later PRs migrate consumers.

### 2. `EndpointClientInterface` replaces `ModelClient` + `ResultConverter` + `ContractHandler`
- Single interface with `endpoint()`, `supports(Model)`, `request()`, `convert()`, `getTokenUsageExtractor()`.
- Extends `ModelClientInterface` + `ResultConverterInterface` so the same instance plugs into `Provider`'s existing two-loop dispatch.
- `Provider::doInvoke()` reads `$options['endpoint']` once, routes by name when set, else falls back to `supports()`. Binds the chosen client to convert the result.
- *Minimum viable PR:* add the interface + Provider routing; migrate **one** bridge (e.g. Voyage) end-to-end as proof. Leave the rest on the legacy interfaces.

### 3. Bridge migration sweep (~28 bridges)
[...]

PR #2235, opened 2026-06-24
[Platform] Add Endpoint value object and expose it on Model

tacman:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no 
| Issues        | -
| License       | MIT

This carves out **slice 1** of @chr-hertel's endpoint refactor (#2029) as a small, self-contained, backward-compatible PR — the piece that PR explicitly marks as landable on its own:

> ### 1. Endpoint as a first-class concept on `Model`
> - New `Endpoint` value object (contract identifier + defaults).
> - `Model` gains `getEndpoints()`, `getDefaultEndpoint()`, `getEndpoint(string)`, `supportsEndpoint(string)`.
> - Catalogs declare endpoints via the new `AbstractModelCatalog::endpointsForModel()` hook.
> - *Minimum viable PR:* just the new types + the hook, no consumer changes. Backward-compatible; later PRs migrate consumers.

The intent is to let the `Endpoint` shape land and be iterated on now (it's also what batch routing — #1802 — and multi-contract models like OpenAI chat-completions vs. responses would hang off of), independently of the larger dispatch rework.

## What's here

**`Endpoint`** — a value object identifying one contract a model speaks, plus optional contract-specific defaults:

```php
namespace Symfony\AI\Platform;

final class Endpoint
{
    public function __construct(private readonly string $contract, private readonly array $defaults = []) { /* … */ }
    public function getContract(): string;
    public function getDefaults(): array;
}
```

**`Model`** — a new optional, last constructor parameter `array $endpoints = []`, and five accessors:

```php
$model->getEndpoints();         // Endpoint[]
$model->hasEndpoints();         // bool
$model->getDefaultEndpoint();   // ?Endpoint — the first declared
$model->getEndpoint('openai.responses');     // Endpoint, throws if unsupported
$model->supportsEndpoint('openai.responses'); // bool
```

**`AbstractModelCatalog`** — a new `endpointsForModel(array $modelConfig): array` hook (defaults to `[]`), called in `getModel()` and forwarded to the `Model` constructor, so a catalog can declare a model's endpoints by overriding one method.

## Backward compatibility

Fully additive:
- the new `Model` constructor parameter is optional and last;
- the catalog hook defaults to no endpoints, so existing catalogs produce identical models;
- **no consumer or bridge is changed** — models without endpoints behave exactly as today.

(Bridge `Model` subclasses with custom constructors — `Anthropic\Claude`, the OpenAI/Scaleway `Embeddings` — keep working unchanged; they'll opt into forwarding `$endpoints` if and when a catalog starts declaring endpoints for them, in a follow-up.)

## Tests

`EndpointTest` (construction, defaults, empty-contract guard), `ModelTest` (all five accessors incl. default-is-first and the unsupported-contract exception), and `AbstractModelCatalogTest` (default = no endpoints; the hook populates a model's endpoints). Full Platform suite green; PHPStan and CS-Fixer clean.

The `Endpoint` API here matches #2029 verbatim so it drops straight into that branch's later slices.

### DIFF
--- a/src/platform/src/Endpoint.php
+++ b/src/platform/src/Endpoint.php
@@ -1,0 +1,57 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Platform;
+
+use Symfony\AI\Platform\Exception\InvalidArgumentException;
+
+/**
+ * Identifies one contract a model speaks (e.g. "anthropic.messages",
+ * "openai.chat_completions", "openai.responses") together with any
+ * contract-specific defaults that should be folded in at invocation time.
+ *
+ * The contract identifier is a free-form string by design: bridges and
+ * dynamic catalogs can register new ones without core changes.
+ *
+ * @author Christopher Hertel <mail@christopher-hertel.de>
+ * @author Tac Tacelosky <tacman@gmail.com>
+ */
+final class Endpoint
+{
+    /**
+     * @param non-empty-string     $contract
+     * @param array<string, mixed> $defaults
+     */
+    public function __construct(
+        private readonly string $contract,
+        private readonly array $defaults = [],
+    ) {
+        if ('' === trim($contract)) {
+            throw new InvalidArgumentException('Endpoint contract cannot be empty.');
+        }
+    }
+
+    /**
+     * @return non-empty-string
+     */
+    public function getContract(): string
+    {
+        return $this->contract;
+    }
+
+    /**
+     * @return array<string, mixed>
+     */
+    public function getDefaults(): array
+    {
+        return $this->defaults;
+    }
+}
--- a/src/platform/CHANGELOG.md
+++ b/src/platform/CHANGELOG.md
@@ -5,6 +5,7 @@ CHANGELOG
 ----
 
  * [BC BREAK] Add `ListenerInterface::onError()` and `Result\Stream\ErrorEvent`, dispatched when draining a `StreamResult` throws, so a listener can finalize on a failed stream where `onComplete()` never fires; `AbstractStreamListener` provides a no-op default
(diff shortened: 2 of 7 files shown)


src/platform/src/Endpoint.php
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * Identifies one contract a model speaks (e.g. "anthropic.messages",
+ * "openai.chat_completions", "openai.responses") together with any
+ * contract-specific defaults that should be folded in at invocation time.
+ *
+ * The contract identifier is a free-form string by design: bridges and
+ * dynamic catalogs can register new ones without core changes.
+ *
+ * @author Christopher Hertel <mail@christopher-hertel.de>
```
OskarStark, line 24: should be you

tacman: actually, @chr-hertel wrote it in a comment, I copied it from there.

tacman: happy to take credit for it though!

tacman:
Rebased onto \`main\` to resolve the \`CHANGELOG.md\` conflict (both sides just added a bullet to the unreleased 0.11 section — kept both) and to drop the merge commit that Fabbot rejected. This PR is now a single Platform-only commit and all Platform checks are green.

Heads-up on the remaining red **PHPStan / Demo** job: it fails on \`demo/src/Crop/Image/Resampler.php:81\` (\`false === $bytes\` flagged as always-false), which is pre-existing code on \`main\` and untouched by this PR (the diff is 7 Platform files only). It passes locally with the committed lockfile and only trips CI under freshly-resolved dependencies, so it looks like a latent issue on \`main\` rather than something this PR introduces — probably best fixed separately on \`main\`.

tacman:
@chr-hertel quick nudge on this one: #2235 is the small, additive Endpoint slice from #2029, all checks are green, and it gives us a stable hook for the batch routing work in #1802 without waiting for the larger dispatch refactor.

Could you review/merge it, or say if you would rather keep Endpoint bundled with your forked refactor?

tacman:
@chr-hertel this is now rebased onto current `main`, with the changelog entry moved to the unreleased 0.13 section. All 51 checks are green and GitHub reports a clean merge state.

Since you’re planning a release next week, could you take another look when you have a chance?

chr-hertel:
Thanks @tacman for the support here - we figured to take a different route and will not adopt this Endpoint thing in that manner.

PR #2242, opened 2026-06-25
[Platform] Add conversion to a dedicated websearch result in open-responses platform

### LINKED ISSUE #2149: [Platform] Convert `web_search_call` (and similar remote tools) in `Gpt\ResultConverter` to typed result objects

push-mann:
| Q             | A
| ------------- | ---
| Bug fix?      |no
| New feature?  | yes 
| Docs?         | yes 
| Issues        | Fix #2149 
| License       | MIT

Added WebSearchResult — a new result type for the OpenResponses bridge that converts web_search_call output items into a structured object exposing the search id, status, query, and queries fields. Previously these items were
silently skipped; now they're surfaced to the caller.

### DIFF
--- a/src/platform/src/Bridge/OpenResponses/ResultConverter.php
+++ b/src/platform/src/Bridge/OpenResponses/ResultConverter.php
@@ -34,6 +34,7 @@ use Symfony\AI\Platform\Result\TextResult;
 use Symfony\AI\Platform\Result\ThinkingResult;
 use Symfony\AI\Platform\Result\ToolCall;
 use Symfony\AI\Platform\Result\ToolCallResult;
+use Symfony\AI\Platform\Result\WebSearchResult;
 use Symfony\AI\Platform\ResultConverterInterface;
 use Symfony\Contracts\HttpClient\ResponseInterface;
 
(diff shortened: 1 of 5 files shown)


src/platform/src/Bridge/OpenResponses/ResultConverter.php
```diff
@@ -179,18 +182,29 @@ private function processOutputItem(array $item): iterable
         return match ($type) {
             'message' => $this->convertOutputMessage($item),
             'reasoning' => $this->convertReasoning($item),
-            // Built-in server-side tool calls (web search, file search, code
-            // interpreter, image generation, computer use, MCP, …) are reported
-            // as their own output items but carry no assistant-facing result.
-            // Skip them so the actual message item is still converted instead of
-            // aborting the whole response.
-            'web_search_call', 'file_search_call', 'code_interpreter_call',
+            'web_search_call' => $this->convertWebSearch($item),
```
push-mann, line 185: Could also try to take care of the other types. Then it might also be worth it adding contracts for converters for specific types instead doing all of the converting-part in here.

OskarStark:
Closing as duplicate of https://github.com/symfony/ai/pull/2248

PR #2243, opened 2026-06-25
[Platform][Albert] Align bridge `base_url` convention with Generic bridge

### LINKED ISSUE #2240: [Platform] Inconsistent base_url convention between Albert and Generic bridges
When configuring both Albert and Generic bridges, the `base_url` convention
is inconsistent, which is confusing, especially when using both platforms
side by side (e.g. Albert + a vLLM instance via Generic).

**Albert** requires the API version in `base_url` and validates it:
```yaml
ai:
    platform:
        albert:
            base_url: 'https://albert.example.com/v1'  # /v1 required here
```
Internally calls `/chat/completions` (no version in path).

**Generic** expects the version in the path (default `/v1/chat/completions`):
```yaml
ai:
    platform:
        generic:
            default:
                base_url: 'https://vllm.example.com'  # no /v1 here
```
Internally calls `/v1/chat/completions`.

Both produce the same final URL, but the configuration expectation is
opposite. There is no error when Generic's `base_url` accidentally includes
`/v1`, it silently produces `/v1/v1/chat/completions`.

## Expected behavior

Both bridges should follow the same convention: either the version always
belongs in `base_url`, or always in the path.

## References

- `Albert/Factory.php`: validates `/v\d+` at end of `base_url`, passes
  `completionsPath: '/chat/completions'`
- `Generic/Factory.php`: defaults to `completionsPath: '/v1/chat/completions'`

jvancoillie:
The Albert bridge was requiring the API version in the base URL (e.g. https://example.com/v1) and overriding completionsPath/embeddingsPath to compensate, while Generic and all other bridges expect a bare base URL and include the version in the path (/v1/chat/completions).

Drop the version validation and the path overrides so Albert delegates entirely to Generic defaults, aligning both conventions.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no 
| Issues        | Fix #2240
| License       | MIT

### DIFF
--- a/src/platform/src/Bridge/Albert/CHANGELOG.md
+++ b/src/platform/src/Bridge/Albert/CHANGELOG.md
@@ -1,6 +1,12 @@
 CHANGELOG
 =========
 
+0.12
+----
+
+ * [BC BREAK] The `baseUrl` must no longer include the API version (e.g. use `https://albert.api.etalab.gouv.fr`
+   instead of `https://albert.api.etalab.gouv.fr/v1`), aligning with the Generic bridge convention
+
 0.11
 ----
 
--- a/src/platform/src/Bridge/Albert/Factory.php
+++ b/src/platform/src/Bridge/Albert/Factory.php
@@ -38,13 +38,8 @@ final class Factory
         ?EventDispatcherInterface $eventDispatcher = null,
         string $name = 'albert',
     ): ProviderInterface {
-        $baseUrl = rtrim($baseUrl, '/');
-
         if (!str_starts_with($baseUrl, 'https://')) {
             throw new InvalidArgumentException('The Albert URL must start with "https://".');
-        }
-        if (!preg_match('/\/v\d+$/', $baseUrl)) {
-            throw new InvalidArgumentException('The Albert URL must include an API version (e.g., /v1, /v2).');
         }
         if ('' === $apiKey) {
             throw new InvalidArgumentException('The API key must not be empty.');
@@ -58,8 +53,6 @@ final class Factory
             httpClient: $httpClient,
             modelCatalog: $modelCatalog,
             eventDispatcher: $eventDispatcher,
-            completionsPath: '/chat/completions',
-            embeddingsPath: '/embeddings',
             name: $name,
         );
     }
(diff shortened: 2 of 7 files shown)


src/platform/src/Bridge/Albert/CHANGELOG.md
```diff
@@ -4,7 +4,8 @@ CHANGELOG
 0.11
 ----
 
- * Tolerate a trailing slash on the base URL instead of rejecting it
```
OskarStark, line 7: this shouldn't be removed and also the code should stay in place, no?

jvancoillie: `rtrim($baseUrl, '/')` is already handled in `Generic\Completions\ModelClient`, so I removed it from the Albert factory to avoid duplication.

Do you want me to reintroduce it in Albert along with the CHANGELOG entry?

OskarStark: Ok makes sense

src/platform/src/Bridge/Albert/Factory.php
```diff
@@ -58,8 +53,6 @@ public static function createProvider(
             httpClient: $httpClient,
             modelCatalog: $modelCatalog,
             eventDispatcher: $eventDispatcher,
-            completionsPath: '/chat/completions',
-            embeddingsPath: '/embeddings',
```
OskarStark, line 62: why removing this part here, is it the default of the generic bridge?

jvancoillie: Yes, that's correct. It's the Generic bridge's default convention (`/v1/chat/completions` and `/v1/embeddings`, see `src/platform/src/Bridge/Generic/Factory.php`). Albert used to require the API version in the base URL
  and override completionsPath/embeddingsPath to compensate for that. This PR drops those overrides so Albert now inherits the Generic defaults directly, aligning both bridges on the same convention.

OskarStark:
Please check the failing CI step, thanks

jvancoillie:
Email fixed, the remaining Resampler.php failure is unrelated to my changes.

chr-hertel (approved):
Thanks @jvancoillie - will patch CHANGELOG/UPGRADE and tests on merge 👍

chr-hertel:
Thank you @jvancoillie.

PR #2245, opened 2026-06-25
[Mate] Guard extension paths against directory traversal

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | -
| License       | MIT

Mate discovers extensions from installed Composer packages and reads paths they declare under `extra.ai-mate` (`scan-dirs`, `includes`, `instructions`). Those package-relative paths are joined onto the project's `vendor/<package>/` directory and then read or included, so a malicious/compromised package using a parent (`..`) segment could reference files outside its own directory.

`ComposerExtensionDiscovery` already had a coarse `str_contains($path, '..')` check, but `AgentInstructionsAggregator` — which feeds the path read for the MCP `instructions` handshake — had **none**.

This PR:

- adds a shared, segment-based `PathGuard::hasTraversal()` and applies it across `ComposerExtensionDiscovery` (scan-dirs / includes / instructions) and `AgentInstructionsAggregator` (extension + root-project instructions);
- removes the duplicated `containsPathTraversal()` helper that previously lived in both classes;
- uses segment matching so legitimate names like `my..notes.md` are no longer rejected, while real `..` segments and null bytes still are.

`realpath()`-based containment was intentionally not used: the `./link` development workflow symlinks `vendor/symfony/ai-*` outside the project, which a `realpath` containment check would wrongly reject.

Tests are included for `PathGuard` (traversal vs. safe paths) and `AgentInstructionsAggregator` (traversing extension/root instructions are rejected, valid ones are aggregated).

### DIFF
--- a/src/mate/src/Agent/AgentInstructionsAggregator.php
+++ b/src/mate/src/Agent/AgentInstructionsAggregator.php
@@ -13,6 +13,7 @@ namespace Symfony\AI\Mate\Agent;
 
 use Psr\Log\LoggerInterface;
 use Symfony\AI\Mate\Discovery\ComposerExtensionDiscovery;
+use Symfony\AI\Mate\Discovery\PathGuard;
 
 /**
  * Aggregates agent instructions from all installed extensions.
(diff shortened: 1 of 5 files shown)


OskarStark (approved):

OskarStark:
Thank you Johannes.

PR #2246, opened 2026-06-25
[Mate] Escape package names when writing extensions.php

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | -
| License       | MIT

`ExtensionConfigSynchronizer::writeExtensionsFile()` interpolated the package name raw into the generated `mate/extensions.php`:

```php
$content .= "    '$packageName' => ['enabled' => $enabled],\n";
```

That file is later `include`d. Package names come from third-party `composer.json` files, so a package whose name contains a single quote could break out of the string literal and inject arbitrary PHP, e.g. a name like `evil', 'injected' => ['enabled' => true], 'x`.

The fix emits the name through `var_export($packageName, true)`, which produces a correctly escaped PHP string literal for any input:

```php
$content .= \sprintf("    %s => ['enabled' => %s],\n", var_export($packageName, true), $enabled);
```

This fully closes the sink — no regex allow-listing or manual quote-escaping needed. A test feeds a crafted injecting name and asserts it round-trips as a single array key with no extra entry, and that the generated file stays valid PHP.

### DIFF
--- a/src/mate/src/Service/ExtensionConfigSynchronizer.php
+++ b/src/mate/src/Service/ExtensionConfigSynchronizer.php
@@ -125,7 +125,9 @@ final class ExtensionConfigSynchronizer
 
         foreach ($extensions as $packageName => $config) {
             $enabled = $config['enabled'] ? 'true' : 'false';
-            $content .= "    '$packageName' => ['enabled' => $enabled],\n";
+            // Package names originate from third-party composer.json files and are written into a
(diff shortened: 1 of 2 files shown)


OskarStark (approved):

OskarStark:
Thank you Johannes.

PR #2247, opened 2026-06-25
[Mate] Create scaffolded secrets and log files with restrictive permissions

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | -
| License       | MIT

Files written by Mate that can hold secrets or local configuration were created with the process default umask (typically world-readable `0644`). On a shared/multi-user host, other users could read `mate/.env` (secrets), `mate/config.php`, and the debug log.

This PR makes those files owner/group-only:

- `mate init` creates `mate/.env` and `mate/config.php` with `0640` and the `mate/` (and `mate/src/`) directories with `0750`;
- `Logger` creates its log file with `0640` (only on first creation) and its log directory with `0750`.

The permission modes live in a single `Symfony\AI\Mate\Service\FilePermissions` holder (`FILE`, `DIRECTORY`, `EXECUTABLE`) used by both classes, so there are no duplicated constants or magic mode literals. Non-sensitive scaffolded files (`mcp.json`, `AGENTS.md`, `AGENT_INSTRUCTIONS.md`, `.gitignore`) are left at the default — they are typically committed and contain no secrets; `bin/codex` stays executable (`0755`).

Hardening is best-effort (`@chmod`), so it is a no-op on platforms without POSIX permissions rather than a hard failure. Tests assert the resulting permissions for the scaffolded files/directories and the log file/directory (skipped on Windows).

### DIFF
--- a/src/mate/src/Command/InitCommand.php
+++ b/src/mate/src/Command/InitCommand.php
@@ -13,6 +13,7 @@ namespace Symfony\AI\Mate\Command;
 
 use HelgeSverre\Toon\Toon;
 use Symfony\AI\Mate\Agent\AgentInstructionsMaterializer;
+use Symfony\AI\Mate\Service\FilePermissions;
 use Symfony\Component\Console\Attribute\AsCommand;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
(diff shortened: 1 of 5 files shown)


src/mate/src/Command/InitCommand.php
```diff
@@ -58,7 +70,8 @@ protected function execute(InputInterface $input, OutputInterface $output): int
 
         $mateDir = $this->rootDir.'/mate';
         if (!is_dir($mateDir)) {
-            mkdir($mateDir, 0755, true);
+            mkdir($mateDir, FilePermissions::DIRECTORY, true);
+            @chmod($mateDir, FilePermissions::DIRECTORY);
```
chr-hertel: why is `@chmod` needed here, but not for the other `mkdir` calls? wouldn't expect it is needed at all tbh

wachterjohannes:
@chr-hertel You're right, dropped it. mkdir($dir, 0750, true) already sets the mode at creation, and umask can only remove bits — so the dir is never more permissive than 0750 and the chmod was redundant (under a strict umask it'd even loosen 0700 back to 0750). Removed it from both InitCommand and Logger. The chmod on the files stays — copy()/file_put_contents() ignore the mode and create 0644, so that's the bit that actually matters.

chr-hertel (approved):

chr-hertel:
Thank you @wachterjohannes.

PR #2250, opened 2026-06-28
[MCP Bundle] Allow configuring DNS rebinding protection allowed hosts

### LINKED ISSUE #2205: [MCP Bundle] Configure MCP server allowed hosts
By default, the MCP SDK has a [`DnsRebindingProtectionMiddleware`](https://github.com/modelcontextprotocol/php-sdk/blob/ca849aa75b9b46f47b39077eb8549406c192874b/docs/transports.md#dns-rebinding-protection) rejecting all external requests (only localhost allowed).

It would be nice to be able to configure the `allowedHosts` so we can easily create public MCP servers.

Amoifr:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | Fix #2205
| License       | MIT

The MCP SDK installs a `DnsRebindingProtectionMiddleware` by default that only accepts requests targeting `localhost`, which makes it impossible to expose a public MCP server through the HTTP transport.

This PR adds a `http.dns_rebinding_protection` configuration to restrict the protection to custom hosts, or to disable it entirely (e.g. when the server is behind a reverse proxy that already validates the `Host` header).

```yaml
mcp:
    http:
        dns_rebinding_protection:
            allowed_hosts: ['example.com', 'mcp.example.com'] # replaces the default localhost allowlist
            # enabled: false # or disable the protection entirely
```

The `McpController` keeps the SDK secure defaults untouched (`null` middleware) when no customization is requested, so there is no behavior change by default. When customization is needed, it rebuilds the stack from `StreamableHttpTransport::defaultMiddleware()` and only swaps/removes the `DnsRebindingProtectionMiddleware`, leaving the other secure defaults (CORS, protocol version) in place.

Tests cover the DI wiring (default / custom hosts / disabled) and the controller's middleware resolution. `nyholm/psr7` was added to `require-dev` because the SDK middlewares perform PSR-17 discovery when instantiated.


freema:
We hit this exact problem in production, and this PR is the right fix.

Setup: a self-hosted MCP server exposed over HTTPS behind a reverse proxy (Cloudflare), with OAuth 2.1 Bearer auth. The `DnsRebindingProtectionMiddleware` default allowlist (`localhost` / `127.0.0.1` / `[::1]`) rejects the public host, so every authenticated `POST /_mcp` returns `403 "Forbidden: Invalid Host header."`.

What made this hard to diagnose: the host check runs **after** authentication, so clients that connect without a token first (MCP Inspector, CLI probes) fail earlier with `401` and never reach it — only fully-authenticated web connectors (e.g. the Claude/ChatGPT server-side proxies, which send `Host: <public-domain>`) trip the 403. That makes it look like an auth/CORS problem rather than a host-allowlist one.

We had to work around it by overriding the `mcp.server.controller` service to rebuild the transport without the middleware. A first-class `http.dns_rebinding_protection.allowed_hosts` / `enabled` option (exactly what this PR adds, keeping the other secure defaults) is the proper solution. Would be great to get this reviewed and merged. Thanks @Amoifr!

Amoifr:
Thanks a lot for the detailed report, @freema — this is gold. 🙏

You nailed the trickiest part: the host check running *after* auth is exactly why this shows up as a phantom "auth/CORS" issue. Only fully-authenticated web connectors (the Claude/ChatGPT server-side proxies sending `Host: <public-domain>`) ever reach the 403, while unauthenticated probes bail out at `401` first — so the one signal that would point you at the host allowlist never fires. That mismatch cost me a fair bit of head-scratching too. 😅

Glad the PR matches the workaround you had to hand-roll. It keeps the secure defaults and just lets you opt into the hosts you actually trust. Fingers crossed for a review! 🤞

chr-hertel (changes requested):
Thanks, this was indeed annoying - let's slim down the config, decoupling a bit from the class names of the or purpose.

what do you think about:

* default would be SDK behavior
* override with explicit hosts
  ```yaml
  mcp:
      http:
          allowed_hosts: ['example.com', 'mcp.example.com']
  ```
* disable allow lists
  ```yaml
  mcp:
      http:
          allowed_hosts: false
  ```

Amoifr:
Great call, much cleaner! 🙏 Applied exactly your proposal: dropped the nested `dns_rebinding_protection` node for a single `http.allowed_hosts` — unset keeps the SDK default (localhost), an array restricts to those hosts, and `false` disables the protection. Config, controller, docs and tests updated accordingly. Thanks for steering it toward something decoupled from the middleware internals!

chr-hertel (approved):
Thanks @Amoifr - will take care about the pipeline failure while merging - potentially also a follow up about the middleware stack instantiation

PR #2251, opened 2026-06-28
[Platform] Default tool_choice to "auto" in Generic completions when tools are provided

### LINKED ISSUE #2228: [Platform] Generic completions ModelClient does not set `tool_choice` when tools are provided
## Description

When using tools with the **Albert** bridge (which delegates to
`Generic\Completions\ModelClient` via `GenericFactory`), the `tool_choice` option
is never injected into the request body.

Several OpenAI-compatible providers do **not** apply a default of `auto` server-side,
so the model falls back to generating a text response that literally contains
`[tool_call]` in the content instead of returning a proper tool-call object.

## Anthropic already handles this correctly

`Bridge/Anthropic/ModelClient.php` (line 76-78):

```php
    $options['tool_choice'] ??= ['type' => 'auto'];
}
```

`Generic/Completions/ModelClient.php` has no equivalent.

## Proposed fix

Add the same guard in `Generic/Completions/ModelClient.php::request()`, using
the OpenAI-compatible string form:

```php
    $options['tool_choice'] ??= 'auto';
}
```

## Steps to reproduce

1. Configure the Albert bridge.
2. Register a tool and call `platform->invoke(...)` with it.
3. Observe that the response is a `TextResult` containing `[tool_call]`
   instead of a `ToolCallResult`.

## Note

This was observed on the **Albert** bridge. Other bridges using the same
`Generic\Completions\ModelClient` may be affected too, though I am not sure
whether this fix would have an impact on them.

Amoifr:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #2228
| License       | MIT

`Generic\Completions\ModelClient` (used by OpenAI-compatible bridges such as Albert via `GenericFactory`) never injects `tool_choice`. Several OpenAI-compatible providers do not apply a default of `auto` server-side, so the model returns a `TextResult` whose content literally contains `[tool_call]` instead of a proper `ToolCallResult`.

The Anthropic bridge already guards against this (`ModelClient.php`):

```php
    $options['tool_choice'] ??= ['type' => 'auto'];
}
```

This PR adds the equivalent guard to `Generic\Completions\ModelClient::request()` using the OpenAI-compatible string form:

```php
    $options['tool_choice'] ??= 'auto';
}
```

An explicit `tool_choice` is left untouched, and nothing changes when no tools are provided. Added `ModelClientTest` covering the three cases (default `auto`, explicit value preserved, no tools → no `tool_choice`).

### DIFF
--- a/src/platform/src/Bridge/Generic/Completions/ModelClient.php
+++ b/src/platform/src/Bridge/Generic/Completions/ModelClient.php
@@ -65,6 +65,12 @@ class ModelClient implements ModelClientInterface
             }
         }
 
+        // Some OpenAI-compatible providers do not default "tool_choice" to "auto" server-side,
+        // which makes the model answer with a text response instead of a proper tool call.
+        if ([] !== ($options['tools'] ?? [])) {
+            $options['tool_choice'] ??= 'auto';
+        }
+
         // cacheRetention is an internal Symfony AI option consumed by PromptCacheNormalizer
         // (Anthropic-only).  Strip it here so it is never forwarded to OpenAI-compatible
         // endpoints, which reject unknown request body fields with a 400 error.
--- a/src/platform/src/Bridge/Generic/Tests/Completions/ModelClientTest.php
+++ b/src/platform/src/Bridge/Generic/Tests/Completions/ModelClientTest.php
@@ -159,4 +159,69 @@ final class ModelClientTest extends TestCase
         $modelClient = new ModelClient($httpClient, 'http://localhost:8000', 'sk-valid-api-key');
         $modelClient->request(new CompletionsModel('gpt-4o'), ['messages' => [['role' => 'user', 'content' => "tool output \xB1 here"]]]);
     }
+
+    public function testToolChoiceDefaultsToAutoWhenToolsAreProvided()
+    {
+        $capturedBody = $this->captureRequestBody(['tools' => $this->tools()]);
+
+        $this->assertSame('auto', $capturedBody['tool_choice']);
+    }
+
+    public function testToolChoiceIsNotOverriddenWhenExplicitlySet()
+    {
+        $capturedBody = $this->captureRequestBody(['tools' => $this->tools(), 'tool_choice' => 'required']);
+
+        $this->assertSame('required', $capturedBody['tool_choice']);
+    }
+
+    public function testToolChoiceIsNotSetWhenNoToolsAreProvided()
+    {
+        $capturedBody = $this->captureRequestBody([]);
+
+        $this->assertArrayNotHasKey('tool_choice', $capturedBody);
+    }
+
+    public function testToolChoiceIsNotSetWhenToolsAreEmpty()
+    {
+        $capturedBody = $this->captureRequestBody(['tools' => []]);
+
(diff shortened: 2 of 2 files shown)


chr-hertel (approved):

src/platform/tests/Bridge/Generic/Completions/ModelClientTest.php
chr-hertel, line 1: this landed in the wrong folder, tests of a bridge belong into the bridge package. in this case: src/platform/src/Bridge/Generic/Tests/Completions/ModelClientTest.php

src/platform/src/Bridge/Generic/Completions/ModelClient.php
```diff
@@ -65,6 +65,12 @@ public function request(Model $model, array|string $payload, array $options = []
             }
         }
 
+        // Some OpenAI-compatible providers do not default "tool_choice" to "auto" server-side,
+        // which makes the model answer with a text response instead of a proper tool call.
+        if (isset($options['tools'])) {
```
chr-hertel: could be an empty array tho, let's go with:

```suggestion
        if ([] !== ($options['tools'] ?? [])) {
```

chr-hertel (changes requested):
Thanks @Amoifr - only small items, but worth patching

src/platform/tests/Bridge/Generic/Completions/ModelClientTest.php
Amoifr: Good catch, thanks! Moved to `src/platform/src/Bridge/Generic/Tests/Completions/`.

One thing I hit on the way: a `ModelClientTest` already lived there, so rather than moving the file I merged my three tests into it and adopted its conventions (`gpt-4o`, `http://localhost:8000`, `sk-valid-api-key`). Cleaner that way — and it taught me where bridge tests actually belong. 🙂

src/platform/src/Bridge/Generic/Completions/ModelClient.php
Amoifr: You're right, `isset()` happily lets an empty array through and we would have sent `tool_choice: auto` with no tools at all. Applied your suggestion.

I also added `testToolChoiceIsNotSetWhenToolsAreEmpty` to lock it down — it does fail against the old `isset()` guard, so the case is now actually covered rather than just fixed. 👍

Amoifr:
Thanks a lot for the review @chr-hertel — both remarks were spot on, and the second one was a genuine bug in my patch rather than a nitpick. Amended and force-pushed:

- test moved into the bridge package, merged into the existing `Bridge/Generic/Tests/Completions/ModelClientTest.php`;
- guard changed to `[] !== ($options['tools'] ?? [])`, plus a new `testToolChoiceIsNotSetWhenToolsAreEmpty` that fails against the previous `isset()` version.

`vendor/bin/phpunit src/Bridge/Generic` is green (75 tests, 247 assertions), no PHP 8.5 deprecation.

Thanks for maintaining this — and for teaching me the bridge test layout the gentle way. 😄

chr-hertel (approved):

chr-hertel:
Thank you @Amoifr.

Issue #2257, opened 2026-06-30
[Agent] Improve toolbox exception hierarchy


makomweb:
I find the current exception hierarchy of the Toolbox very confusing. So I wrapped my head around it. At the same time I don't wanna introduce breaking changes. That's why I am proposing a cautious refactoring. The fundamental change is to remove `ToolExecutionExceptionInterface` and use `ToolExecutionException` directly. I don't see any benefit of having this interface. If I am missing something please enlighten me. Here is the proposed hierarchy: **current implementation vs. without the interface**

<img width="1011" height="604" alt="Image" src="https://github.com/user-attachments/assets/59a5c508-b5de-43b4-ac25-8be4dbbf6ffe" />

<img width="1029" height="463" alt="Image" src="https://github.com/user-attachments/assets/b8d17da0-34b5-4a72-8aa6-6e6c2ecfd728" />

PR #2258, opened 2026-06-30
[Agent] increase coverage of Toolbox

makomweb:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | -
| License       | MIT

This PR is to increase coverage of `Toolbox` code to 100%. 

~Only change in non-test code was in `FaultTolerantToolbox` to use the exceptions `getToolCall` method for full coverage. Side-note: Noticing this brought me to the idea to simplify the exception hierarchy of the `Toolbox` (see https://github.com/symfony/ai/issues/2257 for details).~

---
**Before**
grafik

---
**After**

grafik

### DIFF
--- a/src/agent/tests/Toolbox/ToolboxTest.php
+++ b/src/agent/tests/Toolbox/ToolboxTest.php
@@ -31,6 +31,7 @@ use Symfony\AI\Agent\Toolbox\Exception\ToolNotFoundException;
 use Symfony\AI\Agent\Toolbox\Source\Source;
 use Symfony\AI\Agent\Toolbox\Tool\Subagent;
 use Symfony\AI\Agent\Toolbox\Toolbox;
+use Symfony\AI\Agent\Toolbox\ToolboxInterface;
 use Symfony\AI\Agent\Toolbox\ToolFactory\ChainFactory;
 use Symfony\AI\Agent\Toolbox\ToolFactory\MemoryToolFactory;
 use Symfony\AI\Agent\Toolbox\ToolFactory\ReflectionToolFactory;
@@ -170,10 +171,15 @@ final class ToolboxTest extends TestCase
 
     public function testExecuteWithException()
     {
-        $this->expectException(ToolExecutionException::class);
-        $this->expectExceptionMessage('Execution of tool "tool_exception" failed with error: Tool error.');
-
-        $this->toolbox->execute(new ToolCall('call_1234', 'tool_exception'));
+        try {
+            $this->toolbox->execute(new ToolCall('call_1234', 'tool_exception'));
+            $this->fail('Should have thrown before!');
+        } catch (ToolExecutionException $ex) {
+            $this->assertSame('Execution of tool "tool_exception" failed with error: Tool error.', $ex->getMessage());
+            $toolCall = $ex->getToolCall();
+            $this->assertInstanceOf(ToolCall::class, $toolCall);
+            $this->assertSame('call_1234', $toolCall->getId());
+        }
     }
 
     public function testExecuteWithCustomException()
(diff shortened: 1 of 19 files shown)


src/agent/src/Toolbox/FaultTolerantToolbox.php
```diff
@@ -37,13 +38,15 @@ public function execute(ToolCall $toolCall): ToolResult
     {
         try {
             return $this->innerToolbox->execute($toolCall);
+        } catch (ToolExecutionException $e) {
+            return new ToolResult($e->getToolCall(), $e->getToolCallResult());
```
chr-hertel: this is a duplicate of line 43+44 since `ToolExecutionException` implements the `ToolExecutionExceptionInterface`

src/agent/src/Toolbox/FaultTolerantToolbox.php
```diff
@@ -37,13 +38,15 @@ public function execute(ToolCall $toolCall): ToolResult
...
         } catch (ToolExecutionExceptionInterface $e) {
             return new ToolResult($toolCall, $e->getToolCallResult());
-        } catch (ToolNotFoundException) {
+        } catch (ToolNotFoundException $e) {
             $names = array_map(static fn (Tool $metadata) => $metadata->getName(), $this->getTools());
 
             return new ToolResult(
-                $toolCall,
+                $e->getToolCall() ?? $toolCall,
```
chr-hertel: why is this needed?

chr-hertel (changes requested):
Hey Martin, thanks for working on the test coverage here 👍
just have some questions about the changes in the prod code

src/agent/src/Toolbox/FaultTolerantToolbox.php
makomweb: You're right. This is from the idea of tossing the `ToolExecutionExceptionInterface` - but I didn't want to change this without discussing first - hence, this approach.
Reason: the interface does not provide `getToolCall()` in order to increase coverage.

I see 2 options to proceed here:\
A) lift `getToolCall()` into the interface in order to use it in  line 44\
B) ditching the interface entirely. This is the only place where it is used.\
Honestly, I don't see any benefits of having it. Getting rid of it would simplify things a little. And this is what I wanted to propose [here](https://github.com/symfony/ai/issues/2257).

src/agent/src/Toolbox/FaultTolerantToolbox.php
makomweb: `$e->getToolCall()` is not covered otherwise and this method returns a nullable.

makomweb:
I have lost sight of how to continue here. @chr-hertel can you give some advice?

src/agent/src/Toolbox/FaultTolerantToolbox.php
chr-hertel: There's the benefit of userland implementations - people could implement that interface and hook into the FaultTolerantToolbox, see #478

makomweb: Thanks for pointing me there. 

I enhanced `ToolboxTest::testExecuteWithException()` to cover the tool call during an exception. This way I was able to revert my changes for the 2 lines above.

src/agent/src/Toolbox/FaultTolerantToolbox.php
makomweb: I've added `ToolboxTest::testThrowsWhenToolNotFound()` to cover the `ToolNotFoundException::getToolCall()` and reverted the lines above.

src/agent/tests/Toolbox/ToolboxTest.php
```diff
@@ -417,4 +423,46 @@ public function testToolCallViaMetaDataReflection()
 
         $this->assertSame('Hello says "3".', $result->getResult());
     }
+
+    public function testThrowsWhenToolNotFound()
```
chr-hertel: this basically only tests the mock stub of the test class itself, right?

makomweb: @chr-hertel No. It tests if the exception provides the expected tool call.

makomweb: I have simplified this in 88d0907.

OskarStark (approved):

chr-hertel:
Thank you Martin.

PR #2263, opened 2026-07-01
[Platform][BedrockMantle] Add OpenAI-compatible Bedrock Mantle platform

### LINKED ISSUE #2256: [Platform] Support AWS Bedrock Mantle Endpoints
Bedrock now supports OpenAI-compatible formats (Completions/ Responses format), see:

https://docs.aws.amazon.com/bedrock/latest/userguide/bedrock-mantle.html

This is needed to run OpenAI models via Bedrock

aszenz:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | yes 
| Docs?         | yes 
| Issues        | Fix #2256 
| License       | MIT

Add a dedicated symfony/ai-bedrock-mantle-platform bridge for the AWS Bedrock Mantle endpoint, which exposes OpenAI-compatible APIs (Chat Completions and Responses) for models served through Bedrock. It shares no code with the SigV4/SDK-based Bedrock bridge (Nova/Claude/Llama), so it lives in its own package and reuses the Generic and OpenResponses bridges over the plain OpenAI wire protocol; the base URL is derived from the AWS region.

 * Factory: Chat Completions route (Generic completions client and converter)
 * Responses\Factory: Responses API route (OpenResponses wire protocol)
 * Two authentication modes, with the API key taking precedence: a Bedrock API key sent as a bearer token, and AWS SigV4 signing via the standard AWS credential chain when no key is provided
 * Dedicated ModelCatalogs registering the verified models for each route
 * Runnable examples under examples/bedrock/ for both auth modes


aszenz:
@OskarStark @chr-hertel 

Could I get a review for this ?

src/ai-bundle/composer.json
chr-hertel, line 1: `symfony/ai-bedrock-mantle-platform:^0.12` does not exist yet, please have a look here how to deal with that:
https://github.com/symfony/ai/blob/main/CONTRIBUTING.md#adding-a-new-package-or-bridge

examples/composer.json
chr-hertel, line 1: `symfony/ai-bedrock-mantle-platform:^0.12` does not exist yet, please have a look here how to deal with that:
https://github.com/symfony/ai/blob/main/CONTRIBUTING.md#adding-a-new-package-or-bridge

src/platform/src/Bridge/BedrockMantle/README.md
chr-hertel, line 1: we keep the README files of the bridges mostly short without code snippets, if needed this can go in `docs/components/platform/bedrock.rst`

chr-hertel:
didn't have time to look into that AWS feature yet to also set it up for myself:
does bedrock treat this as separate product? is the model catalog different from the main bedrock setup?
i wonder if we just bring it to the `symfony/ai-bedrock-platform` and have the dependency to open-responses optional

aszenz:
> does bedrock treat this as separate product? 

they recommend mantle going forward since it works with any openai compatible model

> is the model catalog different from the main bedrock setup? 

yes, it supports openai models too and various open one's in addition to also support anthropic one's

> i wonder if we just bring it to the `symfony/ai-bedrock-platform` and have the dependency to open-responses optional

not sure about this, the way i see it's a new endpoint for bedrock with a different catalog + api key support + dashboard console.

aszenz:
@chr-hertel @wachterjohannes 

Can i get a review here ?

Our team is using aws bedrock, and I have some more PRs targeting this platform.

chr-hertel (changes requested):
Hey @aszenz, just double checked with the AWS docs and mantle is in my understanding a different inference engine of bedrock, therefore please merge this into the existing bridge. However, we should be explicit in the factory path + open-responses/generic would be optional dependencies only required if you use the mantle path.

on top, please double check the added examples, do the all run on your end with:
```
cd examples/
composer u && ../link
./runner bedrock --filter=mantle
```

for me this breaks for three of them:
Image

aszenz:
> Hey @aszenz, just double checked with the AWS docs and mantle is in my understanding a different inference engine of bedrock, therefore please merge this into the existing bridge. However, we should be explicit in the factory path + open-responses/generic would be optional dependencies only required if you use the mantle path.

Thanks @chr-hertel 

Done, merged into existing bedrock platform. Also added support for invoking anthropic models using the mantle engine and added the new cassette tests.


> on top, please double check the added examples, do the all run on your end with:
> ```
> cd examples/
> composer u && ../link
> ./runner bedrock --filter=mantle
> ``` 

These failures appear to have been caused by regional model availability. I standardized the examples on models available in us-east-1, and all examples now pass on my end.

Note: Bedrock API keys are region-scoped, so you must select us-east-1 before generating the key and set AWS_DEFAULT_REGION=us-east-1. Standard AWS credentials via SSO also work, given the right IAM permissions.

chr-hertel (approved):
Rebased and reworked bundle and provider wiring a bit - if it's not working on your end anymore, let me know! Thanks for bringing this up @aszenz!

chr-hertel:
Thank you @aszenz.

PR #2265, opened 2026-07-02
[Chat] Normalize non-text AssistantMessage content in MessageNormalizer

### LINKED ISSUE #2198: Inconsistent data serialization for images in MessageSerializer
If an image is attached to a UserPrompt using `Image::fromFile($localPath)`, stored in a MariaDB via the `MessageSerializer`, and subsequently deserialized, an "invalid data format" error occurs.

If you modify the MessageSerializer as follows, everything works smoothly.

``` php

    public function denormalize(mixed $data, string $type, ?string $format = null, array $context = []): mixed
    {
        ... 

        $message = match ($dataType) {
            SystemMessage::class => new SystemMessage($content),
            AssistantMessage::class => new AssistantMessage(...self::denormalizeAssistantParts($data)),
            UserMessage::class => new UserMessage(...array_map(
                static fn(array $part): ContentInterface => match ($part['type']) {
                    File::class, Document::class, Image::class, Audio::class => $part['type']::fromDataUrl($part['content']),
                    Text::class => new Text($part['content']),
                    ImageUrl::class => new ImageUrl($part['content']),
                    DocumentUrl::class => new DocumentUrl($part['content']),
                    default => throw new LogicException(\sprintf('Unknown content type "%s".', $part['type'])),
                },
                $contentAsBase64,
            )),

            ...
    }

     public function normalize(mixed $data, ?string $format = null, array $context = []): array
    {
       ...

        return [
                $context['identifier'] ?? 'id' => $data->getId()->toRfc4122(),
            'type' => $data::class,
            'content' => $content,
            'contentAsBase64' => ($data instanceof UserMessage && [] !== $data->getContent()) ? array_map(
                static fn(ContentInterface $content) => [
                    'type' => $content::class,
                    'content' => match ($content::class) {
                        Text::class => $content->getText(),
                        File::class, Document::class, Image::class, Audio::class => $content->asDataUrl(),
                        ImageUrl::class,
                        DocumentUrl::class => $content->getUrl(),
                        default => throw new LogicException(\sprintf('Unknown content type "%s".', $content::class)),
                    },
                ],
                $data->getContent(),
            ) : [],
            'toolsCalls' => $toolsCalls,
            'parts' => $parts,
            'metadata' => $data->getMetadata()->all(),
            'addedAt' => (new \DateTimeImmutable())->getTimestamp(),
        ];
    }

```

There must be an inconsistency here between the stored and read data formats.

Amoifr:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #2198
| License       | MIT

The original round-trip issue for `UserMessage` images (base64 vs. data URL mismatch) was already fixed by 8ca2ab09. This PR addresses the remaining case reported in the issue's follow-up: `AssistantMessage` content.

`MessageNormalizer::normalizeAssistantParts()` only handled `Text`, `Thinking` and `ToolCall` parts and **silently dropped** anything else. So an `AssistantMessage` carrying an `Image` (e.g. a generated image returned by Gemini) was lost on the way to the store and could not be read back.

`File`/`Document`/`Image`/`Audio` (as data URL) and `ImageUrl`/`DocumentUrl` parts are now handled on both the normalize and denormalize sides, mirroring what is already done for `UserMessage` content. Added a round-trip test for an `AssistantMessage` holding an `Image`.

### DIFF
--- a/src/chat/src/MessageNormalizer.php
+++ b/src/chat/src/MessageNormalizer.php
@@ -22,6 +22,7 @@ use Symfony\AI\Platform\Message\Content\Image;
 use Symfony\AI\Platform\Message\Content\ImageUrl;
 use Symfony\AI\Platform\Message\Content\Text;
 use Symfony\AI\Platform\Message\Content\Thinking;
+use Symfony\AI\Platform\Message\Content\Video;
 use Symfony\AI\Platform\Message\MessageInterface;
 use Symfony\AI\Platform\Message\SystemMessage;
 use Symfony\AI\Platform\Message\ToolCallMessage;
@@ -152,6 +153,8 @@ final class MessageNormalizer implements NormalizerInterface, DenormalizerInterf
                 $parts[] = ['type' => Thinking::class, 'content' => $part->getContent(), 'signature' => $part->getSignature()];
             } elseif ($part instanceof ToolCall) {
                 $parts[] = ['type' => ToolCall::class, 'toolCall' => $this->normalizer->normalize($part, $format, $context)];
+            } elseif ($part instanceof File || $part instanceof ImageUrl || $part instanceof DocumentUrl) {
+                $parts[] = self::normalizeContentParts([$part])[0];
             }
         }
 
@@ -173,7 +176,8 @@ final class MessageNormalizer implements NormalizerInterface, DenormalizerInterf
                     File::class,
                     Document::class,
                     Image::class,
-                    Audio::class => $content->asDataUrl(),
+                    Audio::class,
+                    Video::class => $content->asDataUrl(),
                     ImageUrl::class,
                     DocumentUrl::class => $content->getUrl(),
                     default => throw new LogicException(\sprintf('Unknown content type "%s".', $content::class)),
@@ -195,7 +199,8 @@ final class MessageNormalizer implements NormalizerInterface, DenormalizerInterf
                 File::class,
                 Document::class,
                 Image::class,
-                Audio::class => $part['type']::fromDataUrl($part['content']),
+                Audio::class,
+                Video::class => $part['type']::fromDataUrl($part['content']),
                 Text::class => new Text($part['content']),
                 ImageUrl::class => new ImageUrl($part['content']),
                 DocumentUrl::class => new DocumentUrl($part['content']),
@@ -223,7 +228,7 @@ final class MessageNormalizer implements NormalizerInterface, DenormalizerInterf
                         $part['toolCall']['function']['name'],
                         json_decode($part['toolCall']['function']['arguments'], true),
                     ),
-                    default => throw new LogicException(\sprintf('Unknown assistant part type "%s".', $part['type'])),
+                    default => self::denormalizeContentParts([$part])[0],
                 };
             }
 
--- a/src/chat/tests/MessageNormalizerTest.php
+++ b/src/chat/tests/MessageNormalizerTest.php
@@ -21,6 +21,7 @@ use Symfony\AI\Platform\Message\Content\Image;
 use Symfony\AI\Platform\Message\Content\ImageUrl;
 use Symfony\AI\Platform\Message\Content\Text;
 use Symfony\AI\Platform\Message\Content\Thinking;
+use Symfony\AI\Platform\Message\Content\WebSearch;
 use Symfony\AI\Platform\Message\Message;
 use Symfony\AI\Platform\Message\MessageInterface;
 use Symfony\AI\Platform\Message\Role;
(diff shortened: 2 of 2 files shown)


src/chat/src/MessageNormalizer.php
```diff
@@ -152,6 +152,19 @@ private function normalizeAssistantParts(AssistantMessage $message, ?string $for
...
+                        File::class,
+                        Document::class,
+                        Image::class,
+                        Audio::class => $part->asDataUrl(),
+                        ImageUrl::class,
+                        DocumentUrl::class => $part->getUrl(),
+                        default => throw new LogicException(\sprintf('Unknown content type "%s".', $part::class)),
+                    },
+                ];
             }
```
wachterjohannes: This throws for every content type `Message::toContent()` can produce that isn't listed — `WebSearch`, `FileSearch`, `McpCall`, `McpListTools`, `McpApprovalRequest`, `ComputerCall`, `LocalShellCall`, `ExecutableCode`, `CodeExecution`, `Collection`, and `Video`. Those come from server-side tools, and `Chat::submit()` persists them, so conversations that work today start failing on save. Verified against your branch:

```
main:     WebSearch -> 0 parts restored (silently dropped)
this PR:  WebSearch -> LogicException: Unknown content type "…\WebSearch"
```

Could you skip unknown types instead of throwing, and reuse the existing helper? `normalizeContentParts()` is already identical apart from `Text`:

```suggestion
            } elseif ($part instanceof File || $part instanceof ImageUrl || $part instanceof DocumentUrl) {
                $parts[] = self::normalizeContentParts([$part])[0];
            }
```

That needs `Video::class` added to `normalizeContentParts()` too — it `extends File` and is missing there as well.

src/chat/src/MessageNormalizer.php
```diff
@@ -223,6 +236,12 @@ private static function denormalizeAssistantParts(array $data): array
                         $part['toolCall']['function']['name'],
                         json_decode($part['toolCall']['function']['arguments'], true),
                     ),
+                    File::class,
+                    Document::class,
+                    Image::class,
+                    Audio::class => $part['type']::fromDataUrl($part['content']),
+                    ImageUrl::class => new ImageUrl($part['content']),
+                    DocumentUrl::class => new DocumentUrl($part['content']),
                     default => throw new LogicException(\sprintf('Unknown assistant part type "%s".', $part['type'])),
```
wachterjohannes: Same on this side — these six arms duplicate `denormalizeContentParts()` exactly. Delegating there instead keeps the assistant and user paths from drifting apart again, which is the bug this PR is fixing.

wachterjohannes (changes requested):
Thanks @Amoifr — real bug and the right instinct, `normalizeAssistantParts()` had no `else` branch at all so non-text content was silently dropped. But the `default => throw` turns that silent loss into a crash on a live path; details inline.

Also, could you extend the tests beyond `Image`? A mixed-ordering round-trip (`Text` + `Image` + `ToolCall`) and a case pinning down the unknown-type behaviour would cover the interesting parts.

Amoifr:
Thanks @wachterjohannes, especially for verifying the behaviour against the branch. You're right that the throw traded silent loss for a crash on live conversations, wrong trade. Pushed:

- both assistant paths now delegate to `normalizeContentParts()`/`denormalizeContentParts()`, so the user and assistant handling can't drift apart again;
- unknown types (`WebSearch`, `McpCall`, ...) are skipped like on `main` instead of throwing;
- `Video::class` added to both shared helpers (nice catch);
- new tests: a mixed `Text` + `Image` + `ToolCall` + `ImageUrl` ordering round-trip, and a `WebSearch` part being skipped while the surrounding texts survive.

src/chat/src/MessageNormalizer.php
```diff
@@ -223,7 +228,7 @@ private static function denormalizeAssistantParts(array $data): array
                         $part['toolCall']['function']['name'],
                         json_decode($part['toolCall']['function']['arguments'], true),
                     ),
-                    default => throw new LogicException(\sprintf('Unknown assistant part type "%s".', $part['type'])),
+                    default => self::denormalizeContentParts([$part])[0],
```
wachterjohannes, line 231: Small asymmetry worth a thought: normalize now *skips* unknown types, but this delegates to a helper whose own `default` throws — so writing is lenient and reading is strict. Unreachable with data this version writes, so fine by me.

It does drop the wording though: the failure now reads `Unknown content type "…"` instead of `Unknown assistant part type "…"`, which pointed at where the bad data actually sits. Worth keeping the specific message if it's cheap.

wachterjohannes (approved):
Verified the revision locally — crash is gone, `Video` round-trips, and a `WebSearch` in the middle of a message no longer takes the surrounding text with it. Chat suite green. Thanks for turning this around so quickly.

One non-blocking note inline.

Amoifr:
Thanks @wachterjohannes for the thorough double-check, running the revision locally before approving is above and beyond. The delegation to shared helpers was your call and it made the fix better: one code path, no drift. Much appreciated!

wachterjohannes:
Thank you @Amoifr.

PR #2267, opened 2026-07-02
[Mate] Redact sensitive context/extra in Monolog log entries

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | n/a
| License       | MIT

The `monolog-search` / `monolog-context-search` / `monolog-tail` tools returned
each parsed log entry's `context` and `extra` verbatim via `LogEntry::toArray()`.
The same data the profiler logger collector exposes, but for the **file-based**
log path — reset tokens, credentials, session ids and PII reached the AI
unredacted (only wrapped as untrusted, which guards against prompt injection but
not data exposure).

Apply recursive key-based redaction to `context` and `extra` in `toArray()` (the
AI-facing output only). The raw values stay intact internally, so term/regex and
context-key search still match on the real data; the message is left untouched.

### DIFF
--- a/src/mate/src/Bridge/Monolog/Model/LogEntry.php
+++ b/src/mate/src/Bridge/Monolog/Model/LogEntry.php
@@ -32,6 +32,45 @@ final class LogEntry
     private const REGEX_BACKTRACK_LIMIT = 10000;
 
     /**
+     * Key-name substrings (case-insensitive) used to redact values in the log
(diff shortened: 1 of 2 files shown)


chr-hertel (approved):

chr-hertel:
Thank you @wachterjohannes.

PR #2271, opened 2026-07-02
[Mate] Ask for the PHP binary in init and support containerized setups

### LINKED ISSUE #2264: [Mate] `mate init` generates an .mcp.json that fails under DDEV/Docker (no host PHP)
## Summary

`mate init` writes an `.mcp.json` whose command is `./vendor/bin/mate` (run on the
**host**). In a container-based setup like DDEV, PHP only exists inside the container —
the host has no `php` — so Claude Code (running on the host) can't launch Mate and shows
the server as **failed**. This makes Mate unusable out of the box for the large share of
Symfony developers on DDEV / Docker.

Found while running a PoC against PR #2213; 

## Environment

- Symfony AI Mate **0.10.0** (branch `mate-skills-install`, PR #2213)
- Host app: Symfony 7.4 webapp running under **DDEV** (Docker, PHP 8.3 in the container)
- Claude Code running on the host (macOS)

## Steps to reproduce

1. A Symfony app running under DDEV (no PHP installed on the host).
2. `ddev exec vendor/bin/mate init` — it generates `.mcp.json`:
   ```json
   {
     "mcpServers": {
       "symfony-ai-mate": {
         "command": "./vendor/bin/mate",
         "args": ["serve", "--force-keep-alive"]
       }
     }
   }
   ```
3. Open Claude Code on the host, `/mcp`.
4. → `symfony-ai-mate · ✘ failed`. Claude Code tried to run `./vendor/bin/mate` on the
   host, where there is no PHP.

## Cause

The generated command assumes PHP is on the same host as Claude Code. Under DDEV (and any
Docker-based dev environment) PHP lives only in the container, so a host-launched
`./vendor/bin/mate serve` can't start.

## Workaround (what I used, and it works)

Rewrite the command so the host launches Mate **inside** the container; `ddev exec`
forwards stdio transparently, which is exactly what the MCP stdio transport needs:

```json
{
  "mcpServers": {
    "symfony-ai-mate": {
      "command": "ddev",
      "args": ["exec", "vendor/bin/mate", "serve", "--force-keep-alive"]
    }
  }
}
```

After this, Claude Code connects and the agent can call the Mate tools (verified the full
host → container → Mate → Symfony profiler loop end-to-end).

## Suggested fix

`mate init` could detect a container-based environment and emit a container-aware command,
or document the wrapper. Options, roughly in order of effort:

- **Detect DDEV** (`.ddev/config.yaml` present) and emit
  `command: "ddev", args: ["exec", "vendor/bin/mate", "serve", ...]`.
- Optionally detect generic Docker Compose (a `docker-compose.yml` / `compose.yaml` with a
  PHP service) and emit `docker compose exec <service> vendor/bin/mate serve ...` — though
  the service name isn't guessable, so this may be better left to docs.
- At minimum, add a **README note** for DDEV/Docker users showing the `ddev exec` wrapper,
  and/or a `mate init --command=<...>` flag (or a post-generation prompt) so the launcher
  can be overridden without hand-editing.

fatonh:
| Q             | A
  | ------------- | ---
  | Bug fix?      | no
  | New feature?  | yes
  | Docs?         | yes
  | Issues        | Fix #2264
  | License       | MIT

  `mate init` generates an `mcp.json` that the coding agent launches on the **host**. In a
  containerized setup (DDEV, and Docker-based envs generally) PHP only exists inside the
  container, so the default `./vendor/bin/mate` command can't start and Claude Code shows the
  server as failed. Reported in #2264.

  This makes `init` ask which PHP binary the agent should use to launch Mate, and detects a
  sensible default from the environment:

  - if a `.ddev/` directory is present → default `ddev exec php`
  - otherwise → default `php`

  The prompt is offered interactively (accept the default or override); under
  `--no-interaction` the detected default is used, so CI/scripted runs don't block. The
  answer is written into the generated `mcp.json` — the first token becomes `command` and the
  rest are prepended to `args`.

  ### Example

  DDEV project (`.ddev/` detected):

      $ vendor/bin/mate init
       PHP binary to run Mate for your coding agent [ddev exec php]:
       >

  produces:

  ```json
  {
      "mcpServers": {
          "symfony-ai-mate": {
              "command": "ddev",
              "args": ["exec", "php", "vendor/bin/mate", "serve", "--force-keep-alive"]
          }
      }
  }

  A plain Docker Compose user can instead type e.g. docker compose exec php php.

  Note

  The non-container default now emits php vendor/bin/mate … rather than the previous
  ./vendor/bin/mate …. It's functionally equivalent (and a bit more portable), but happy to
  special-case plain php back to ./vendor/bin/mate if you'd prefer zero change for the
  common case.

  I don't use DDEV enough to be sure the detection covers every layout — the .ddev/ check is
  deliberately minimal; generic Docker/Sail/Lando are left to the manual override for now.

### DIFF
--- a/src/mate/src/Command/InitCommand.php
+++ b/src/mate/src/Command/InitCommand.php
@@ -100,8 +100,16 @@ class InitCommand extends Command
             }
         }
 
-        // Create symlink from .mcp.json to mcp.json for compatibility
+        // Only prompt when the freshly written mcp.json still carries the
+        // placeholders; a kept existing file has nothing left to configure.
         $mcpJsonPath = $this->rootDir.'/mcp.json';
+        if ($this->mcpJsonNeedsPhpBinary($mcpJsonPath)) {
+            $phpBinary = $this->resolvePhpBinary($io);
+            $this->applyPhpBinaryToMcpJson($mcpJsonPath, $phpBinary);
+            $actions[] = ['✓', 'Configured', \sprintf('mcp.json launch command ("%s")', $phpBinary)];
+        }
+
+        // Create symlink from .mcp.json to mcp.json for compatibility
         $mcpJsonSymlink = $this->rootDir.'/.mcp.json';
         if (file_exists($mcpJsonPath)) {
             if (is_link($mcpJsonSymlink)) {
@@ -186,6 +194,58 @@ class InitCommand extends Command
         if (\in_array($template, self::SENSITIVE_FILES, true)) {
             @chmod($destination, FilePermissions::FILE);
         }
+    }
+
+    /**
+     * PHP binary the agent uses to launch Mate, defaulted by environment
+     * detection (containers such as DDEV need a wrapper) and confirmed by the user.
+     */
+    private function resolvePhpBinary(SymfonyStyle $io): string
+    {
+        $default = is_dir($this->rootDir.'/.ddev') ? 'ddev exec php' : 'php';
+
+        return trim($io->ask('PHP binary to run Mate for your coding agent', $default) ?? $default);
+    }
+
+    /**
+     * Whether the mcp.json at the given path still holds the unresolved
+     * placeholders, i.e. it was just written from the template this run.
+     */
+    private function mcpJsonNeedsPhpBinary(string $mcpJsonPath): bool
+    {
+        $contents = @file_get_contents($mcpJsonPath);
+
+        return \is_string($contents) && str_contains($contents, '##PHP_BINARY##');
+    }
+
+    /**
+     * Replace the mcp.json placeholders with the PHP binary. A multi-word binary
+     * (e.g. "ddev exec php") is split into the command and its leading args.
+     */
+    private function applyPhpBinaryToMcpJson(string $mcpJsonPath, string $phpBinary): void
+    {
+        $parts = preg_split('/\s+/', trim($phpBinary), -1, \PREG_SPLIT_NO_EMPTY);
+        $contents = file_get_contents($mcpJsonPath);
+        if (false === $parts || [] === $parts || false === $contents) {
+            return;
+        }
+
+        $command = array_shift($parts);
+        $args = [...$parts, './vendor/bin/mate', 'serve', '--force-keep-alive'];
+        $encodedArgs = implode(', ', array_map(
+            static fn (string $arg): string => json_encode($arg, \JSON_UNESCAPED_SLASHES),
+            $args
+        ));
+
+        // The args placeholder is quoted in the template so it stays valid JSON;
+        // the encoded fragment brings its own quotes, so the quotes are replaced too.
+        $contents = str_replace(
+            ['##PHP_BINARY##', '"##MATE_ARGS##"'],
+            [$command, $encodedArgs],
+            $contents
+        );
+
+        file_put_contents($mcpJsonPath, $contents);
     }
 
     /**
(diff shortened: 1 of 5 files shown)


fatonh:
@wachterjohannes Heads up: the PHPStan / Demo check fails on a pre-existing error in demo/src/Crop/Image/Resampler.php, which this PR doesn't touch — looks unrelated to the Mate change.

wachterjohannes:
@fatonh already fixed in https://github.com/symfony/ai/pull/2270 😂

src/mate/src/Command/InitCommand.php
```diff
@@ -88,8 +88,15 @@ protected function execute(InputInterface $input, OutputInterface $output): int
             }
         }
 
-        // Create symlink from .mcp.json to mcp.json for compatibility
         $mcpJsonPath = $this->rootDir.'/mcp.json';
+        if (file_exists($mcpJsonPath)) {
+            $phpBinary = $this->resolvePhpBinary($io);
```
wachterjohannes, line 107: i like the idea of resolving the PHP-Binary here and put that into the mcp.json!

But i would add a replacer to the resource file for mcp.json ##PHP_BINARY## or whatever :) and replace that here.

src/mate/src/Command/InitCommand.php
```diff
@@ -164,6 +171,57 @@ private function postCopyTemplateAction(string $template, string $destination):
         chmod($destination, 0755);
     }
 
+    /**
+     * Determine the PHP binary the coding agent should use to launch Mate.
+     *
+     * The generated mcp.json is executed by the agent on the host. When the
```
wachterjohannes: to long comments here - make them shorter

wachterjohannes (changes requested):
Thanks for this contribution 🎉

fatonh:
@wachterjohannes thanks! Addressed both issues :)

src/mate/src/Command/InitCommand.php
```diff
@@ -164,6 +171,55 @@ private function postCopyTemplateAction(string $template, string $destination):
...
+        return trim($io->ask('PHP binary to run Mate for your coding agent', $default) ?? $default);
+    }
+
+    /**
+     * Replace the ##PHP_BINARY## placeholder in the generated mcp.json. A
+     * multi-word binary (e.g. "ddev exec php") is split into command and args.
+     *
+     * @return bool whether the file was modified
+     */
+    private function applyPhpBinaryToMcpJson(string $mcpJsonPath, string $phpBinary): bool
```
wachterjohannes: it does not realy replace ist parses the json - i thaught about file_get_contents & str_replace & file_put_contents

wachterjohannes (changes requested):

src/mate/src/Command/InitCommand.php
```diff
@@ -164,6 +171,62 @@ private function postCopyTemplateAction(string $template, string $destination):
...
+        $indent = '                ';
+        if ([] === $parts) {
+            $argsLine = '';
+        } else {
+            $encoded = array_map(
+                static fn (string $part): string => json_encode($part, \JSON_UNESCAPED_SLASHES),
+                $parts
+            );
+            $argsLine = $indent.implode(",\n".$indent, $encoded).",\n";
+        }
```
wachterjohannes: this looks a bit weird especially the $indent thing ... is this realy neccesary?

wachterjohannes (changes requested):

wachterjohannes (approved):
@fatonh perfect  - now i think that makes sense!

thanks for that 🎉

src/mate/src/Command/InitCommand.php
fatonh: Thanks for the catch - it was ugly :)

chr-hertel:
Thank you @fatonh.

PR #2272, opened 2026-07-03
[Platform][ElevenLabs] Implement speech-to-text functionality with support for additional formats and options

### LINKED ISSUE #2266: [ElevenLabs] Support SRT output for speech-to-text via additional_formats
### Description

The ElevenLabs speech-to-text API supports exporting transcripts in additional formats, including SRT subtitles, through the `additional_formats` multipart field.

Currently, the Symfony AI ElevenLabs bridge supports `Capability::SPEECH_TO_TEXT`, but it is not possible to request SRT output cleanly through `PlatformInterface::invoke()`. The standard `asText()` result only exposes the plain transcript text, while the SRT content returned by ElevenLabs under `additional_formats[0].content` is not available through a first-class API.

This makes subtitle generation difficult when using Symfony AI with ElevenLabs Scribe.

### Use case

I want to generate SubRip subtitles from an audio file using ElevenLabs Scribe v2 through Symfony AI:

```php
$result = $platform->invoke(
    model: new ElevenLabs(
        name: 'scribe_v2',
        capabilities: [
            Capability::SPEECH_TO_TEXT,
            Capability::INPUT_AUDIO,
            Capability::OUTPUT_TEXT,
        ],
    ),
    input: Audio::fromFile(__DIR__ . '/audio.mp3'),
    options: [
        'language_code' => 'pl',
        'tag_audio_events' => false,
        'num_speakers' => 1,
        'diarize' => true,
        'timestamps_granularity' => 'word',
        'additional_formats' => [
            [
                'format' => 'srt',
                'include_timestamps' => true,
            ],
        ],
    ],
);
```

Expected result: Symfony AI should send these speech-to-text options to ElevenLabs and make the returned SRT content accessible.

### Current workaround

The only reliable workaround is to bypass Symfony AI and call ElevenLabs directly with Symfony HttpClient:

```php
public function createSubRipText(string $language, string $audio): string
{
    $formData = new FormDataPart(fields: [
        'model_id' => 'scribe_v2',
        'language_code' => $language,
        'tag_audio_events' => 'false',
        'num_speakers' => '1',
        'diarize' => 'true',
        'timestamps_granularity' => 'word',
        'additional_formats' => json_encode(value: [
            [
                'format' => 'srt',
                'include_timestamps' => true,
            ],
        ], flags: JSON_THROW_ON_ERROR),
        'file' => new DataPart(body: $audio, filename: 'audio.mp3', contentType: self::AUDIO_MIME_TYPE),
    ]);

    $url = sprintf('%s/v1/speech-to-text', self::API_URL);

    $response = $this->httpClient->request(method: 'POST', url: $url, options: [
        'headers' => [
            'xi-api-key' => $this->apiKey,
            ...$formData->getPreparedHeaders()->toArray(),
        ],
        'body' => $formData->bodyToIterable(),
    ])->toArray();

    $result = $response['additional_formats'][0]['content'] ?? null;

    if (empty($result)) {
        throw new \RuntimeException(message: 'Failed to create audio SRT subtitles');
    }

    return $result;
}
```

This works, but it duplicates functionality that should ideally be available through the ElevenLabs bridge.

### Proposed solution

There are two related improvements that would solve this:

1. Forward speech-to-text `$options` to the ElevenLabs `/v1/speech-to-text` request body.

   For example, options such as these should be supported:

   ```php
   [
       'language_code' => 'pl',
       'tag_audio_events' => false,
       'num_speakers' => 1,
       'diarize' => true,
       'timestamps_granularity' => 'word',
       'additional_formats' => [
           [
               'format' => 'srt',
               'include_timestamps' => true,
           ],
       ],
   ]
   ```

2. Expose additional transcript formats in the result.

   Possible API options:

   ```php
   $result->asText(); // existing transcript text
   $result->getRawResult()->getData()['additional_formats'][0]['content']; // already possible if raw data is preserved
   ```

   Or, preferably, a first-class result API such as:

   ```php
   $result->asSubRipText();
   ```

   or:

   ```php
   $result->getAdditionalFormat('srt');
   ```

### Why this is useful

SRT generation is a common speech-to-text use case for video subtitles, social media videos, accessibility captions, and content workflows. ElevenLabs already supports this natively through `additional_formats`, so Symfony AI only needs to expose this functionality through the bridge.

### Related API behavior

ElevenLabs `/v1/speech-to-text` accepts multipart fields such as:

```text
model_id
file
language_code
tag_audio_events
num_speakers
timestamps_granularity
diarize
additional_formats
```

The SRT content is returned in the response under:

```php
$response['additional_formats'][0]['content']
```

### Environment

Symfony AI ElevenLabs bridge: current main / latest release
ElevenLabs model: `scribe_v2`
Use case: speech-to-text with SRT subtitle output

khaperets:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Docs?         | yes 
| Issues        | Fix #2266
| License       | MIT

[ElevenLabs] Support SRT output for speech-to-text via `additional_formats`
===========================================================================

This exposes ElevenLabs Scribe's speech-to-text options and additional
transcript formats (e.g. SRT subtitles) through `PlatformInterface::invoke()`,
so subtitle generation no longer requires bypassing the bridge and calling
ElevenLabs directly with the HTTP client.

Motivation
----------

The ElevenLabs `/v1/speech-to-text` endpoint accepts multipart fields such as
`language_code`, `tag_audio_events`, `num_speakers`, `diarize`,
`timestamps_granularity` and `additional_formats`, and returns the requested
export formats under `additional_formats[].content`. Until now the bridge sent
only `file` and `model_id`, and `asText()` exposed only the plain transcript,
so generating SRT required a hand-rolled HttpClient call (see the workaround in
the issue).

Changes
-------

### 1. Forwarding speech-to-text options — `ElevenLabsClient`

`ElevenLabsClient::doSpeechToTextRequest()` now receives the merged request +
model options (mirroring the text-to-speech branch) and forwards them as
multipart fields. Because Symfony HttpClient builds multipart bodies via
`http_build_query`, options are normalized to the wire format ElevenLabs expects:

 * arrays (e.g. `additional_formats`) are JSON-encoded into a single field;
 * booleans are sent as `"true"`/`"false"` (not `"1"`/`""`);
 * other scalars are cast to strings.

### 2. First-class result API — `Bridge\ElevenLabs\Result\TranscriptionResult`

A new bridge-scoped DTO carrying the transcript text alongside the returned
`additional_formats`. It transparently decodes base64-encoded content and
exposes:

 * `getText()`
 * `getAdditionalFormats()`
 * `getAdditionalFormat(string $format)` — resolved by `requested_format`
   (e.g. `srt`, `txt`, `html`)
 * `asSubRipText()` — convenience accessor for SRT

### 3. Converter — `ElevenLabsResultConverter`

Speech-to-text responses now return a `MultiPartResult` containing both a
`TextResult` (so `asText()` keeps working) and an `ObjectResult` wrapping the
new `TranscriptionResult` (so `asObject()` returns the structured result).
This mirrors the pattern used by the Mistral OCR bridge and avoids adding
bridge-specific methods to the core `DeferredResult`.

Usage
-----

```php
use Symfony\AI\Platform\Bridge\ElevenLabs\ElevenLabs;
use Symfony\AI\Platform\Bridge\ElevenLabs\Factory;
use Symfony\AI\Platform\Capability;

$platform = Factory::createPlatform(apiKey: $apiKey);

$result = $platform->invoke(
    model: new ElevenLabs('scribe_v2', [
        Capability::SPEECH_TO_TEXT,
        Capability::INPUT_AUDIO,
        Capability::OUTPUT_TEXT,
    ]),
    input: Audio::fromFile('/path/audio.mp3'),
    options: [
        'language_code' => 'en',
        'tag_audio_events' => false,
        'num_speakers' => 1,
        'diarize' => true,
        'timestamps_granularity' => 'word',
        'additional_formats' => [
            ['format' => 'srt', 'include_timestamps' => true],
        ],
    ],
);

$result->asText();                       // plain transcript
$result->asObject()->asSubRipText();     // SRT subtitles
$result->asObject()->getAdditionalFormat('srt');

// The raw payload is still available:
$result->getRawResult()->getData()['additional_formats'][0]['content'];
```

Backward compatibility
----------------------

`$result->asText()` is preserved unchanged for speech-to-text, so existing
consumers keep working. The converted result type changes from `TextResult` to
`MultiPartResult`, but the public entry points (`asText()`, `asObject()`,
`getRawResult()`) remain consistent: `asText()` behaves identically, and
`asObject()` now succeeds (returning a `TranscriptionResult`) where it
previously threw `UnexpectedResultTypeException`. No `UPGRADE.md` entry is
therefore required.

Tests
-----

 * Updated `ElevenLabsConverterTest` for the new `MultiPartResult` shape, with
   dedicated cases for plain transcripts, additional formats (both raw and
   base64-encoded content) and the absence of additional formats.
 * Added `ElevenLabsClientTest::testClientCanPerformSpeechToTextRequestWithOptions`
   asserting that the multipart body forwards `language_code`, `diarize`
   (`true`), `tag_audio_events` (`false`), `num_speakers` (`1`),
   `timestamps_granularity` and the JSON-encoded `additional_formats`.
 * Full platform suite passes (733 tests).

Documentation
-------------

 * New "Speech-to-Text" section in `docs/components/platform.rst` covering the
   basic transcript and the SRT export flow (validated with `doctor-rst`).
 * New runnable example `examples/elevenlabs/speech-to-text-srt.php`.
 * `CHANGELOG.md` entry under the unreleased `0.11` section.

### DIFF
--- a/docs/components/platform.rst
+++ b/docs/components/platform.rst
@@ -806,6 +806,54 @@ Code Examples
 
 * `Audio Output with GPT`_
 
+Speech-to-Text
+--------------
+
+Some models can transcribe audio into text. Pass an
+:class:`Symfony\\AI\\Platform\\Message\\Content\\Audio` as input and read the transcript
+through ``asText()``::
+
+    use Symfony\AI\Platform\Bridge\ElevenLabs\Factory;
+    use Symfony\AI\Platform\Message\Content\Audio;
+
+    // Initialize Platform
+
+    $result = $platform->invoke(
+        model: 'scribe_v2',
+        input: Audio::fromFile('/path/audio.mp3'),
+    );
+
+    echo $result->asText(); // "Hello there"
+
+ElevenLabs Scribe accepts provider-specific options that are forwarded to the
+``/v1/speech-to-text`` request, such as ``language_code``, ``diarize``,
+``num_speakers``, ``timestamps_granularity`` and ``additional_formats``. When
+``additional_formats`` is requested, the result is exposed as a structured
+:class:`Symfony\\AI\\Platform\\Bridge\\ElevenLabs\\Result\\Transcript` object through
+``asObject()`` instead of a plain text result::
+
+    $result = $platform->invoke(
+        model: 'scribe_v2',
+        input: Audio::fromFile('/path/audio.mp3'),
+        options: [
+            'language_code' => 'en',
+            'diarize' => true,
+            'timestamps_granularity' => 'word',
+            'additional_formats' => [
+                ['format' => 'srt', 'include_timestamps' => true],
+            ],
+        ],
+    );
+
+    echo $result->asObject()->getText(); // the plain transcript
+    echo $result->asObject()->asSubRipText(); // the SRT subtitles
+    echo $result->asObject()->getAdditionalFormat('srt'); // the SRT subtitles
+
+Code Examples
+~~~~~~~~~~~~~
+
+* `ElevenLabs Speech-to-Text with SRT`_
+
 Document OCR
 ------------
 
@@ -1442,6 +1490,7 @@ Code Examples
 .. _`Image URL Input with GPT`: https://github.com/symfony/ai/blob/main/examples/openai/image-input-url.php
 .. _`Audio Input with GPT`: https://github.com/symfony/ai/blob/main/examples/openai/audio-input.php
 .. _`Audio Output with GPT`: https://github.com/symfony/ai/blob/main/examples/openai/audio-output.php
+.. _`ElevenLabs Speech-to-Text with SRT`: https://github.com/symfony/ai/blob/main/examples/elevenlabs/speech-to-text-srt.php
 .. _`PDF Input with GPT`: https://github.com/symfony/ai/blob/main/examples/openai/pdf-input-binary.php
 .. _`PDF Input with Claude`: https://github.com/symfony/ai/blob/main/examples/anthropic/pdf-input-binary.php
 .. _`OCR with Mistral (URL)`: https://github.com/symfony/ai/blob/main/examples/mistral/ocr-url.php
--- a/examples/elevenlabs/speech-to-text-srt.php
+++ b/examples/elevenlabs/speech-to-text-srt.php
@@ -1,0 +1,34 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use Symfony\AI\Platform\Bridge\ElevenLabs\Factory;
+use Symfony\AI\Platform\Message\Content\Audio;
+
+require_once dirname(__DIR__).'/bootstrap.php';
+
+$platform = Factory::createPlatform(apiKey: env('ELEVEN_LABS_API_KEY'), httpClient: http_client());
+
+$result = $platform->invoke(
+    model: 'scribe_v2',
+    input: Audio::fromFile(dirname(__DIR__, 2).'/fixtures/audio.mp3'),
+    options: [
+        'language_code' => 'en',
+        'tag_audio_events' => false,
+        'num_speakers' => 1,
+        'diarize' => true,
+        'timestamps_granularity' => 'word',
+        'additional_formats' => [
+            ['format' => 'srt', 'include_timestamps' => true],
+        ],
+    ],
+);
+
+echo $result->asObject()->asSubRipText().\PHP_EOL;
--- a/src/platform/src/Bridge/ElevenLabs/ElevenLabsClient.php
+++ b/src/platform/src/Bridge/ElevenLabs/ElevenLabsClient.php
@@ -37,7 +37,10 @@ final class ElevenLabsClient implements ModelClientInterface
     public function request(Model $model, array|string $payload, array $options = []): RawResultInterface
     {
         return match (true) {
-            $model->supports(Capability::SPEECH_TO_TEXT) => $this->doSpeechToTextRequest($model, $payload),
+            $model->supports(Capability::SPEECH_TO_TEXT) => $this->doSpeechToTextRequest($model, $payload, [
+                ...$options,
+                ...$model->getOptions(),
+            ]),
             $model->supports(Capability::TEXT_TO_SPEECH) => $this->doTextToSpeechRequest($model, $payload, [
                 ...$options,
                 ...$model->getOptions(),
@@ -48,8 +51,9 @@ final class ElevenLabsClient implements ModelClientInterface
 
     /**
      * @param array<string|int, mixed> $payload
+     * @param array<string, mixed>     $options
      */
-    private function doSpeechToTextRequest(Model $model, array|string $payload): RawHttpResult
+    private function doSpeechToTextRequest(Model $model, array|string $payload, array $options = []): RawHttpResult
     {
         if (!\is_array($payload)) {
             throw new InvalidArgumentException(\sprintf('Payload must be an array for speech-to-text request, got "%s".', \gettype($payload)));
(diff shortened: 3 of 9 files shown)


examples/elevenlabs/speech-to-text-srt.php
```diff
@@ -0,0 +1,40 @@
...
+require_once dirname(__DIR__).'/bootstrap.php';
+
+$platform = Factory::createPlatform(apiKey: env('ELEVEN_LABS_API_KEY'), httpClient: http_client());
+
+$result = $platform->invoke(
+    model: new ElevenLabs('scribe_v2', [
+        Capability::SPEECH_TO_TEXT,
+        Capability::INPUT_AUDIO,
+        Capability::OUTPUT_TEXT,
+    ]),
```
chr-hertel: we could just use `$platform->invoke('scribe_v2', Audio::fromFile(...), [...])`

no need for instantiating the `ElevenLabs` model class i guess

src/platform/src/Bridge/ElevenLabs/Tests/ElevenLabsConverterTest.php
```diff
@@ -142,4 +222,17 @@ public function testConvertThrowsExceptionWithoutErrorMessage()
         $this->expectExceptionCode(0);
         $converter->convert($rawResult);
     }
+
+    private function extractTranscription(ResultInterface $result): TranscriptionResult
```
chr-hertel: let's go with "Transcript" instead of "Transcription" - in my understanding "Transcription" is what ElevenLabs does and "Transcript" is the result of it

```suggestion
    private function extractTranscript(ResultInterface $result): Transcript
```

src/platform/src/Bridge/ElevenLabs/Result/TranscriptionResult.php
```diff
@@ -0,0 +1,86 @@
...
+ *
+ * It carries the plain transcript text alongside the additional export formats
+ * (e.g. SRT subtitles) that ElevenLabs returns when requested through the
+ * `additional_formats` multipart field.
+ *
+ * @see https://elevenlabs.io/docs/api-reference/speech-to-text/convert
+ *
+ * @author Dmytro Khaperets <khaperets@gmail.com>
+ */
+final class TranscriptionResult
```
chr-hertel: Please rename to `Transcript` and drop the `Result` suffix - it is used quite heavily in the Platform component already in a different meaning

docs/components/platform.rst
```diff
@@ -806,6 +806,63 @@ Code Examples
...
+    use Symfony\AI\Platform\Message\Content\Audio;
+
+    // Initialize Platform
+
+    $result = $platform->invoke(
+        model: new ElevenLabs('scribe_v2', [
+            Capability::SPEECH_TO_TEXT,
+            Capability::INPUT_AUDIO,
+            Capability::OUTPUT_TEXT,
+        ]),
```
chr-hertel: i don't think we need to instantiate the class here, just use the model identifier `scribe_v2`

src/platform/src/Bridge/ElevenLabs/ElevenLabsClient.php
```diff
@@ -63,14 +67,50 @@ private function doSpeechToTextRequest(Model $model, array|string $payload): Raw
...
 
+    /**
+     * ElevenLabs expects the speech-to-text multipart fields to be sent as plain
+     * strings: booleans as `true`/`false`, integers as their string representation
+     * and array-shaped options (e.g. `additional_formats`) as a JSON-encoded string.
+     */
+    private function stringifySpeechToTextOption(int|string $key, mixed $value): string
+    {
+        if (\is_array($value)) {
+            return json_encode($value, \JSON_THROW_ON_ERROR);
```
chr-hertel: please catch that `JsonException` and do an upcast to `Symfony\AI\Platform\Exception\InvalidArgumentException` while forwarding the original exception

src/platform/src/Bridge/ElevenLabs/ElevenLabsResultConverter.php
```diff
@@ -65,6 +68,27 @@ public function getTokenUsageExtractor(): null
...
+        $data = $result->getData();
+        $text = $data['text'] ?? '';
+
+        /** @var list<array<string, mixed>> $additionalFormats */
+        $additionalFormats = \is_array($data['additional_formats'] ?? null) ? $data['additional_formats'] : [];
+
+        return new MultiPartResult([
+            new TextResult($text),
+            new ObjectResult(new TranscriptionResult($text, $additionalFormats)),
+        ]);
```
chr-hertel: i don't think we should be that ambiguous here, but
1. if the user doesn't provide any `additional_formats` option => return `TextResult`
2. if the user requests `additions_formats` => return `ObjectResult` with Transcript object

see the Whisper implementation in OpenAI bridge for reference

src/platform/src/Bridge/ElevenLabs/Result/TranscriptionResult.php
```diff
@@ -0,0 +1,86 @@
...
+    ) {
+    }
+
+    public function getText(): string
+    {
+        return $this->text;
+    }
+
+    /**
+     * @return list<array<string, mixed>>
```
chr-hertel: we can type this as class as well, no matter which format is chosen, we always have:
* requested_format: string
* file_extension: string
* content_type: string
* is_base64_encoded: boolean
* content: string

chr-hertel (changes requested):
Hi @khaperets - thanks for working on this! 🙏 

Added a first round of comments and tested the provided example - worked right away 😎👍 

Thanks already for the follow-up patch!

chr-hertel:
btw, PHPStan issue in the demo should be gone after rebase on latest `main`

examples/elevenlabs/speech-to-text-srt.php
khaperets: done

khaperets:
@chr-hertel Thanks for your feedback, PR updated

src/platform/src/Bridge/ElevenLabs/Result/AdditionalFormat.php
```diff
@@ -0,0 +1,101 @@
...
+     */
+    public function getDecodedContent(): ?string
+    {
+        if (!$this->isBase64Encoded) {
+            return $this->content;
+        }
+
+        $decoded = base64_decode($this->content, true);
+
+        return false === $decoded ? null : $decoded;
```
chr-hertel: let's go with an exception here instead please. i'm generally not a huge fan of decoding errors in the return type, but using exceptions to communicate to callees instead

chr-hertel (approved):

chr-hertel:
Thank you @khaperets.

PR #2286, opened 2026-07-08
[Platform][Bedrock] Fix ClaudeResultConverter crash on thinking content blocks

karpilin:
| Q             | A
  | ------------- | ---
  | Bug fix?      | yes
  | New feature?  | no
  | Docs?         | no
  | Issues        | 
  | License       | MIT

  `convert()` returned `content[0]['text']`, but thinking-enabled responses lead with `thinking` blocks that have no `text` key:

  ```json
  {"content": [
      {"type": "thinking", "thinking": "", "signature": "..."},
      {"type": "text", "text": "the actual answer"}
  ]}
  ```

On Claude Sonnet 5 adaptive thinking is the default, so every non-tool-call invocation crashed with an undefined-array-key error. The existing guard missed it because thinking blocks do have a type.

The fix adopts the per-type block mapping of Bridge\Anthropic\ResultConverter: text → TextResult, thinking → ThinkingResult, tool_use → ToolCallResult; unknown types skipped, single result returned bare, several as MultiPartResult, none throws. convert() widens to ResultInterface — a BC break, documented in UPGRADE.md. Since thinking-plus-text is the normal Sonnet 5 response, MultiPartResult becomes the common return; DeferredResult::asText() is unaffected, and mixed text/tool-call responses now preserve the text preamble instead of dropping it.

One deliberate divergence from Bridge\Anthropic: parallel tool calls stay aggregated in a single ToolCallResult rather than one per tool_use block - Agent\Toolbox\AgentProcessor picks only the first ToolCallResult part of a MultiPartResult, so per-block results would drop all parallel calls but the first.

The sibling looks affected by exactly that today; happy to align later together with a fix in the agent's MultiPartResult handling.

### DIFF
--- a/src/platform/src/Bridge/Bedrock/Anthropic/ClaudeResultConverter.php
+++ b/src/platform/src/Bridge/Bedrock/Anthropic/ClaudeResultConverter.php
@@ -15,8 +15,11 @@ use Symfony\AI\Platform\Bridge\Anthropic\Claude;
 use Symfony\AI\Platform\Bridge\Bedrock\RawBedrockResult;
 use Symfony\AI\Platform\Exception\RuntimeException;
 use Symfony\AI\Platform\Model;
+use Symfony\AI\Platform\Result\MultiPartResult;
 use Symfony\AI\Platform\Result\RawResultInterface;
+use Symfony\AI\Platform\Result\ResultInterface;
 use Symfony\AI\Platform\Result\TextResult;
+use Symfony\AI\Platform\Result\ThinkingResult;
 use Symfony\AI\Platform\Result\ToolCall;
 use Symfony\AI\Platform\Result\ToolCallResult;
 use Symfony\AI\Platform\ResultConverterInterface;
@@ -32,7 +35,7 @@ final class ClaudeResultConverter implements ResultConverterInterface
         return $model instanceof Claude;
     }
 
-    public function convert(RawResultInterface|RawBedrockResult $result, array $options = []): ToolCallResult|TextResult
+    public function convert(RawResultInterface|RawBedrockResult $result, array $options = []): ResultInterface
     {
         $data = $result->getData();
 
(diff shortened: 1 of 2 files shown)


UPGRADE.md
```diff
@@ -4,6 +4,20 @@ UPGRADE FROM 0.10 to 0.11
...
+   returned unwrapped (text-only responses still yield a plain `TextResult`), and several supported
+   blocks yield a `Result\MultiPartResult` — the common return value on models with adaptive
+   thinking enabled by default (Claude Sonnet 5 and later), where thinking-plus-text is the normal
+   response shape. `DeferredResult::asText()` keeps working transparently (it extracts the single
+   text part) and `MultiPartResult::asText()` flattens only text parts. Mixed text/tool-call
+   responses no longer collapse to a bare `ToolCallResult` dropping the text preamble (parallel
+   tool calls remain aggregated into a single `ToolCallResult` part), and a thinking-only response
+   (e.g. `max_tokens` exhausted mid-thinking) converts to a bare `ThinkingResult` instead of the
+   converter throwing; `DeferredResult::asText()` then throws `UnexpectedResultTypeException`.
+   Code type-hinting the old union must handle `MultiPartResult`.
```
chr-hertel: changes to UPGRADE.md are only needed for BC breaks, but we don't have one here i think - at least widening the return type isn't one.

if we only consider this a bugfix, we can just skip this, and revert the changes in this file

chr-hertel (changes requested):
Good catch, thanks @karpilin - I think #2289 can help here already with the other glitch, right?

(just one minor comment, we can merge after that)

karpilin:
> Good catch, thanks @karpilin - I think #2289 can help here already with the other glitch, right?
> 
> (just one minor comment, we can merge after that)

Thanks @chr-hertel! Yes, exactly - #2289 fixes the glitch that made me aggregate the parallel tool calls into a single ToolCallResult here. Since `MultiPartResult::asToolCallResult()` flattens both shapes, this PR is safe to merge in either order relative to #2289. 

Once #2289 is in, I'm happy to do a small follow-up (update this, or new PR) aligning this converter with Bridge\Anthropic (one ToolCallResult per tool_use block), so the two Claude converters stay in sync.

chr-hertel:
Thanks for the confirmation @karpilin - merged in #2289, please update here - thanks already :)

karpilin:
> Thanks for the confirmation @karpilin - merged in #2289, please update here - thanks already :)

@chr-hertel All done. Let me know if it needs further corrections.

chr-hertel (approved):

chr-hertel:
Thank you @karpilin.

PR #2288, opened 2026-07-08
[Mate] Improve cache isolation, recursive cleanup, and cross-platform compatibility

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | 
| License       | MIT

### Description

This Pull Request introduces several improvements and bug fixes for the `mate` component:

1. **Cache Directory Isolation:** Default cache directories are now isolated by system username and project path hash to avoid permissions conflicts and state/session leakage on multi-tenant or shared systems.
2. **Cross-Platform Compatibility & Portability:**
   - Modified generated `mcp.json` templates to run `./vendor/bin/mate` via `php` directly (making it more portable on Windows and non-executable filesystems).
   - Handled symlink creation failure gracefully on Windows (e.g. lack of developer mode/admin rights) during `init` by outputting a warning instead of raising a PHP warning and terminating.
   - Added native Windows support to `StopCommand` using `taskkill`.
3. **Recursive Cache Cleanup:** `ClearCacheCommand` now recursively deletes empty subdirectories (e.g., `sessions/`) inside the cache directory.
4. **Test Coverage:** Added new test suites for `ClearCacheCommand` and `StopCommand`.

### DIFF
--- a/src/mate/src/Command/StopCommand.php
+++ b/src/mate/src/Command/StopCommand.php
@@ -47,16 +47,10 @@ class StopCommand extends Command
     {
         $io = new SymfonyStyle($input, $output);
 
-        if (!\function_exists('posix_kill')) {
-            $io->error('The "stop" command require the posix php extension.');
+        if (!is_dir($this->cacheDir)) {
+            $io->note('No running servers found.');
 
-            return Command::FAILURE;
-        }
-
-        if (!\defined('SIGUSR1')) {
-            $io->error('The "stop" command require the pcntl php extension.');
-
-            return Command::FAILURE;
+            return Command::SUCCESS;
         }
 
         $finder = new Finder();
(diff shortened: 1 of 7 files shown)


src/mate/src/Command/StopCommand.php
```diff
@@ -47,6 +47,44 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
+            $io->note('No running servers found.');
+
+            return Command::SUCCESS;
+        }
+
+        if (\PHP_OS_FAMILY === 'Windows') {
+            $killedCount = 0;
+            foreach ($finder as $file) {
+                $pid = (int) file_get_contents($file->getRealPath());
+                exec(\sprintf('taskkill /F /PID %d 2>&1', $pid), $execOutput, $resultCode);
```
chr-hertel: somehow the combination of `$finder` => `file_get_contents` => `exec('taskkill ...')` gives me some shivers - i know it was the same before with `posix_kill` but that slipped my eyes - or my memory tricks me 😂

but it's safe enough, right? or do you have ideas to harden this?

src/mate/src/Command/StopCommand.php
chr-hertel, line 1: you're doubling here output messages, like "no running servers found" or "stopped x running servers", and the entire foreach. how about moving the process stopping to a private method, that takes care of the OS part and the main sequence is stable and does not care?

chr-hertel (changes requested):

wachterjohannes:
@chr-hertel Good call — pulled the OS-specific bit into a tiny stopProcess() and now execute() reads as one straight line, with the messages in one place each. Bonus: POSIX now counts real kills instead of trusting the file count.

Re the taskkill shivers 😄 — safe as-is: $pid is cast to (int), so worst case is taskkill /F /PID 0. Could swap to the Process component, but that'd pull symfony/process into mate just for this one call, so I'd rather not unless you feel strongly.

chr-hertel (approved):

chr-hertel:
Thank you @wachterjohannes.

PR #2291, opened 2026-07-09
[Platform] Add an exception is the `response_format` option looks like a class, but is not

lyrixx:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        |
| License       | MIT

I'm not 100% sure about this one since I don't know all bridge APIs, but it could help a lot developpers to spot a missing class import


chr-hertel:
> it could help a lot developpers to spot a missing class import

i would assume if an import is missing, there's also no `\` - but if it's there, it's most likely a FQCN and the `class_exists` would return true

but if you actually ran into it, fair enough

lyrixx:
```
>…ome/gregoire/dev/github.com/lyrixx/symfony(messenger-disable-rejected) cat test.php 
<?php

namespace Foo\Bar;

echo DoNotExit::class . "\n";
>…ome/gregoire/dev/github.com/lyrixx/symfony(messenger-disable-rejected) php test.php
Foo\Bar\DoNotExit
```

But in my case, I moved a class, and forgot to update the use statement. Anyway, It covers both situation

chr-hertel (approved):
Yes, make sense - thanks @lyrixx!

PR #2293, opened 2026-07-09
[Store][Postgres] Quote table and column name

lyrixx:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| Docs?         | no
| Issues        |
| License       | MIT

The default table name is `default` and it's a reserved keyword.
Let's fix that

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -101,6 +101,9 @@ Store
    | Supabase        | `$url`         | `$endpoint`  |
    | SurrealDb       | `$endpointUrl` | `$endpoint`  |
 
+ * Postgres Store now uses quoted identifiers for table and column names, so they are case-sensitive. If you relied on
+   implicit lowercasing of unquoted identifiers, adopt them so they match the case of your table and column names.
+
 UPGRADE FROM 0.9 to 0.10
 ========================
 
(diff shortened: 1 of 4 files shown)


OskarStark (approved):

chr-hertel (changes requested):
TIL: quoting in postgres makes the identifier case sensitive
(ofc i didn't know, but claude told me^^)

> Quoting an identifier also makes it case-sensitive, whereas unquoted names are always folded to lower case. For example, the identifiers FOO, foo, and "foo" are considered the same by PostgreSQL, but "Foo" and "FOO" are different from these three and each other. (The folding of unquoted names to lower case in PostgreSQL is incompatible with the SQL standard, which says that unquoted names should be folded to upper case. Thus, foo should be equivalent to "FOO" not "foo" according to the standard. If you want to write portable applications you are advised to always quote a particular name or never quote it.)

see: https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS

makes sense to note that as BC break here

lyrixx:
@chr-hertel Thanks for warning. I was not aware! I added a note in the CHANGELOG.md

chr-hertel (approved):

chr-hertel:
Thank you @lyrixx.

PR #2294, opened 2026-07-09
[Agent][Chat][Store] Make AGENTS.md the canonical AI-agent guidance, slim CLAUDE.md

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | yes
| Issues        | n/a
| License       | MIT

Follow-up to #2101, applying the same pattern to the remaining locations.

After #2101, `Platform` and `Mate` used `AGENTS.md` as the canonical AI-agent guidance with `CLAUDE.md` reduced to a header plus an `@AGENTS.md` import. Per the [discussion on #2101](https://github.com/symfony/ai/pull/2101#issuecomment-4763241093), this brings every other component into the same shape.

For each location, `AGENTS.md` now carries the canonical guidance and `CLAUDE.md` is reduced to a header plus an `@AGENTS.md` import. Any `CLAUDE.md`-only content was moved into the respective `AGENTS.md` first, so nothing is lost.

### Straight slims

`src/agent`, `src/chat`, `src/store`, `examples` — `AGENTS.md` already carried the equivalent content, so `CLAUDE.md` is now just the import stub.

### `[AiBundle]`

Moved the one `CLAUDE.md`-only fact — PHPStan runs at level 6 with custom extension rules — into `AGENTS.md`, then slimmed `CLAUDE.md`.

### `[Demo]`

The demo's `CLAUDE.md` had drifted ahead of its `AGENTS.md` and carried richer, more current guidance (the RAG pipeline, multi-agent routing, the speech/subagent pattern, the `Video` direct-platform exception, and the two coexisting MCP servers). `demo/AGENTS.md` now carries that full content, while preserving the machine-managed `AI_MATE_INSTRUCTIONS` block.

### `ai.symfony.com`

This site had no `AGENTS.md` — it is created from the existing `CLAUDE.md`, which is then slimmed to the import stub.

### Monorepo root

The root `CLAUDE.md` was a superset of the root `AGENTS.md` (doctor-rst validation, cookbook website artifacts, the "Documentation vs Cookbook" guidance, and the extra CHANGELOG/UPGRADE conventions). That content is moved into the root `AGENTS.md`; the two files were byte-identical apart from the header afterwards, so `CLAUDE.md` becomes the import stub.

Net change is almost entirely de-duplication (~760 deletions vs ~165 insertions). Docs-only, so no CHANGELOG/UPGRADE entries; the touched files are Markdown, outside doctor-rst and cookbook scope.


chr-hertel (approved):
Thanks Johannes!

PR #2297, opened 2026-07-10
[CI] Collapse the ~419-job pipeline into a shared-workspace design

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | -
| License       | MIT

CI installed every package on its own — ~419 jobs and ~600 `composer install`s per PR. PHPStan and PHPUnit only need an autoloader, so most of that install work was redundant. This reworks the pipeline around a single shared install per resolution axis.

### Changes

- **Workspace install for PHPStan & PHPUnit** — `.github/build-workspace.php` generates one root `composer.json` (all 93 packages via path repos + the union of their third-party `require`/`require-dev` + their `autoload-dev`). Both tools run against that single install.
- **Per-package vendor shim** — each package's `vendor/` is reduced to the two paths the configs resolve (`autoload.php`, `phpstan`), avoiding the symlink cycle `isolate-bridge` used to work around.
- **Parallel per-package runner** — `.github/run-in-packages.sh` runs every package in parallel with buffered, grouped output and per-package failure attribution; PHPStan failures annotate inline at file:line.
- **Grouped `--prefer-lowest` jobs** — the lowest matrix (which genuinely can't share an install) collapses from 93 jobs into 6 component groups via `.github/run-lowest.sh`; components install in place, bridges are copied to a depth-preserving scratch tree, installs run serially against the warm cache.
- **Isolated PHPStan tmp dir** — each parallel package gets its own `TMPDIR` (PHPStan's container cache is shared and 32 bridges have byte-identical configs).
- **Removed dead steps/scripts** — `sqlite-vec` installs that couldn't take effect (bridge loads it only on PHP ≥ 8.4; integration job's Sqlite never enters that matrix), the unreferenced `.github/workflows/.utils.sh`, and the unused `build-matrix.yaml` outputs (now only the store-integration list).

### Before / after

| Metric | Before | After |
| --- | --- | --- |
| Jobs | 419 | 46 |
| Compute (job-minutes) | ~197 | ~41 |
| `composer install`s | ~600 | ~15 |
| Code Quality wall clock | 185s | 97s |
| Tests wall clock | 301s | 183s |

### Kept per-package (on purpose)

- `--prefer-lowest`: a merged install resolves to the highest of all per-package floors (Mate allows `symfony/console ^5.4`, Platform requires `^7.3`), so it's grouped but never merged.
- Mate on Symfony 5.4/6.4, and the store integration / demo / examples jobs — each needs its own install.


chr-hertel:
this approach is 42% slower (according to a single run), but maybe parallelization within the jobs/steps might help

chr-hertel:
parallelization basically cuts it to 50% :muscle:

lyrixx (approved):
It looks much simpler, and cleaner. Thanks !

chr-hertel:
@OskarStark we're losing here granularity on first level, but still have it on step level in case a job fails:

image

hard feelings about this rework? not sure tho if all combinations of change scenarios will work.

---

claude ran some tests for change scenarios on my fork:

  - A — coordinated Platform→OpenAI (expect GREEN): bridge uses a new Capability::REASONING added to Platform in the same PR.
  https://github.com/chr-hertel/ai/pull/22 
  - B — new bridge + example (expect GREEN): brand-new symfony/ai-demo-platform bridge (not on Packagist) plus an example depending on it; proves auto-discovery.
  https://github.com/chr-hertel/ai/pull/25
  (additional failure here is only Claude not being able to build a new bridge in one shot)
  - C — negative control (expect RED): OpenAI bridge references Capability::REASONING that is not added to Platform; proves the pipeline catches cross-package breakage.
  https://github.com/chr-hertel/ai/pull/23
  - D — coordinated Platform→Agent (expect GREEN): Agent test uses a new Capability::isOutput() method added to Platform.
  https://github.com/chr-hertel/ai/pull/24

(_and nvm the fabbot failures_)

OskarStark:
Great job, I like it 🖖

PR #2305, opened 2026-07-12
[Store] Add StoreInterface::clear() and implement it in all bridges

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | -
| License       | MIT

Adds `StoreInterface::clear()` to remove all documents from a store while keeping the underlying table, index or collection intact - in contrast to `ManagedStoreInterface::drop()`, which removes the infrastructure itself.

```php
$store->clear();      // documents gone, index still there - re-index without setup()
$store->drop();       // infrastructure gone
```

`clear()` follows the PHP/Symfony idiom (PSR-6 `CacheItemPoolInterface::clear()`, Doctrine `Collection::clear()`) and is vendor-neutral, unlike `truncate` (SQL-table vocabulary, SQLite has no such keyword) or `purge` (means reclaiming already-deleted rows in InnoDB/Solr).

* Implemented in all 25 stores with the native mechanism of each backend: `TRUNCATE TABLE` (Postgres, MariaDB, ClickHouse, Manticore), `_delete_by_query` (Elasticsearch, OpenSearch), `deleteMany({})` (MongoDB), label-scoped `DETACH DELETE` (Neo4j), match-all filters (Qdrant, Weaviate, Typesense), `deleteAll` (Pinecone), `SCAN` + `DEL` (Redis), ...
* AzureSearch and S3Vectors have no delete-all API and enumerate-then-delete in batches. Cloudflare Vectorize supports neither listing nor bulk delete, so it recreates the index - documented.
* Adds `ai:store:clear` command next to `ai:store:setup` and `ai:store:drop`.

Verified with the docker compose runtimes and the store integration tests, which now cover `clear()` and that the store stays usable afterwards.

Two bugs surfaced while doing so:

* `testRemoveDocuments` depended on the optional text/hybrid query tests, and PHPUnit skips dependents of a skipped test - so `remove()` and `drop()` were never actually run for stores without `TextQuery` support. The optional query tests now pass through instead of skipping.
* That uncovered a broken `remove()` in the Manticore Search store: `uuid` was created as a `TEXT` field, but documents are removed by an `equals` filter, which Manticore rejects on stored text fields ("unsupported column 'uuid' (stored field, NOT attribute)"). It is now a `STRING` attribute - existing tables need to be recreated, see UPGRADE.md.

### DIFF
--- a/src/store/src/Bridge/AzureSearch/SearchStore.php
+++ b/src/store/src/Bridge/AzureSearch/SearchStore.php
@@ -15,6 +15,7 @@ use Symfony\AI\Platform\Vector\NullVector;
 use Symfony\AI\Platform\Vector\Vector;
 use Symfony\AI\Store\Document\Metadata;
 use Symfony\AI\Store\Document\VectorDocument;
+use Symfony\AI\Store\Exception\InvalidArgumentException;
 use Symfony\AI\Store\Exception\RuntimeException;
 use Symfony\AI\Store\Exception\UnsupportedQueryTypeException;
 use Symfony\AI\Store\Query\QueryInterface;
@@ -27,6 +28,10 @@ use Symfony\Contracts\HttpClient\HttpClientInterface;
  */
 final class SearchStore implements StoreInterface
 {
+    private const BATCH_SIZE = 1000;
+    private const MAX_UNCHANGED_BATCHES = 10;
+    private const INDEX_CATCH_UP_DELAY = 200000;
+
     /**
      * @param string $vectorFieldName The name of the field int the index that contains the vector
      */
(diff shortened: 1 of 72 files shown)


src/store/src/Bridge/AzureSearch/SearchStore.php
```diff
@@ -71,6 +73,32 @@ public function remove(string|array $ids, array $options = []): void
...
+        // Azure AI Search has no delete-all operation, so documents are looked up and deleted in batches.
+        // Already deleted ids are tracked, since they can still show up in the search result while the
+        // index catches up, which would otherwise turn this into an endless loop.
+        $deleted = [];
+
+        while (true) {
+            $result = $this->request('search', [
+                'search' => '*',
+                'select' => 'id',
+                'top' => self::BATCH_SIZE,
```
chr-hertel: make the batch size configurable via `$options`

PR #2307, opened 2026-07-12
[Mate] Redact Doctrine query parameter values in profiler db collector

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | n/a
| License       | MIT

The profiler `db` collector exposed `sample_params` — the bound parameter
values of each SQL query — to the AI with only a 100-character truncation. For a
statement like `INSERT INTO users (email, password_hash, reset_token) VALUES
(?, ?, ?)` that means real user PII and credentials were handed to the model
verbatim, against the bridge's "sensitive data stays out" principle.

Bound parameters have no key name to decide which are sensitive, so every leaf
value is redacted while the array shape and parameter count are preserved (so
the AI can still see the query was parameterized). The SQL text, counts and
timings — which carry the diagnostic value — are unchanged.

Note (out of scope): a non-parameterized query that inlines literals into the
SQL string itself (`WHERE email = 'a@b.com'`) is not scrubbed; Doctrine rarely
produces these and it is a separate, harder problem.


chr-hertel (approved):

chr-hertel:
Thank you @wachterjohannes.

PR #2308, opened 2026-07-12
[Mate] Redact request/query parameters, raw body and headers in profiler request collector

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | n/a
| License       | MIT

The Symfony profiler `request` collector is exposed to the AI via the
`symfony-profiler-*` tools and resources. It shipped with only cookies,
response/request headers, server/dotenv vars and session attributes redacted —
while the data that actually carries user secrets was passed through verbatim.
This bundles the two request-collector fixes together, since the header change
builds directly on the parameter/body redaction (fork PRs wachterjohannes#36 +
wachterjohannes#39).

### Redact request/query parameters and raw body

Symfony's own collector masks only `request_request['_password']`, so any other
field name (`password`, `pwd`, `api_key`, …), every query-string value and the
whole raw body (e.g. a JSON login payload) reached the model in clear text.

- Redacts values whose key matches a sensitive parameter pattern across
  `request_request` / `request_query` / `request_attributes`, recursively,
  while keeping non-sensitive keys so the AI can still reason over the request.
- Omits the raw request body (`content`), which cannot be reliably
  field-redacted; the parsed, sanitized parameters remain available.
- Closes two re-exposure vectors that re-embed the same data under
  non-sensitive key names:
  - `curlCommand` — reconstructs the request with `--data-raw <body>`, the full
    URL (query string included) and request headers (Authorization included).
  - `REQUEST_URI` / `QUERY_STRING` server vars carry the raw query string;
    `HTTP_COOKIE` is now covered by the env patterns too.

The added test asserts the secret values are absent from the entire formatted
output (not just the one field under test — which is how the `curlCommand` /
server-var re-exposure was found).

### Match request headers by substring

`sanitizeHeaders()` only redacted five hard-coded header names via an exact
`in_array()` match, so common sensitive headers such as `proxy-authorization`,
`www-authenticate`, `x-csrf-token`, `x-xsrf-token` and `x-amz-security-token`
were forwarded unredacted. Replaces the exact-match denylist with
case-insensitive substring matching against a broader pattern set, mirroring the
approach already used for environment variables. A denylist is the wrong
polarity for secrets: over-redacting an occasional benign header is preferable
to leaking one.

Note: detection is key-based, so a secret placed under an innocuous key (e.g. a
JWT in `?q=`) is not caught; this is a deliberate, documented limitation.

### DIFF
--- a/src/mate/src/Bridge/Symfony/Profiler/Service/Formatter/RequestCollectorFormatter.php
+++ b/src/mate/src/Bridge/Symfony/Profiler/Service/Formatter/RequestCollectorFormatter.php
@@ -22,7 +22,9 @@ use Symfony\Component\HttpKernel\DataCollector\RequestDataCollector;
  * - Session cookies and tokens
  * - API keys and secrets in environment variables
  * - Authorization headers
- * - Password and authentication data
+ * - Password and authentication data submitted as request/query parameters
+ * - The raw request body (which may carry credentials, e.g. a JSON login payload)
+ * - The reconstructed curl command (re-embeds body, URL query string and headers)
  *
  * @author Johannes Wachter <johannes@sulu.io>
  *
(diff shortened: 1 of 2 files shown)


chr-hertel (approved):

chr-hertel:
Thank you @wachterjohannes.

PR #2310, opened 2026-07-12
[AI Bundle] Add AiAssertionsTrait to assert on AI usage in functional tests

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | -
| License       | MIT

Answers of a model are not deterministic, which makes them a poor thing to assert on. What an agent
*did* is deterministic enough: that it called the model, that it reached for a tool, that it did not
call the platform twice when once is expected.

The profiler already collects all of that - `AiAssertionsTrait` turns it into assertions for a
`WebTestCase`:

```php
final class AgentTest extends WebTestCase
{
    use AiAssertionsTrait;

    public function testTheAgentSearchesTheBlog()
    {
        $client = static::createClient();
        $client->enableProfiler();

        $client->request('POST', '/chat', ['message' => 'What happened last week?']);

        self::assertPlatformCallCount(2);   // once to decide on the tool, once with its result
        self::assertModelCalled('gpt-4.1');
        self::assertToolCalled('similarity_search');
    }
}
```

* `assertPlatformCallCount()`, `assertModelCalled()`, `assertToolCalled()`, `assertToolCallCount()`
  and `assertToolRegistered()`.
* `assertToolRegistered()` is named apart from `assertToolCalled()` on purpose: the collector gathers
  the tools of *every* toolbox of the application, not only of the agent that handled the request, so
  it says a tool is configured - not that this request used it.
* `getAiDataCollector()` is the escape hatch for anything the assertions do not cover. The data is
  collected as objects, so the `input` of a platform call is the `MessageBag` that was sent and the
  `metadata` is a `Metadata` carrying the token usage - there is nothing to parse.

The demo uses it in `AiAssertionsTest` on its home page, which calls no platform and therefore needs
no API key, so it runs in CI.

Split out of #2300, which is rebased on top of this and where the trait came up.

### DIFF
--- a/src/ai-bundle/CHANGELOG.md
+++ b/src/ai-bundle/CHANGELOG.md
@@ -4,6 +4,7 @@ CHANGELOG
 0.11
 ----
 
+ * Add `AiAssertionsTrait` for functional tests based on Profiler data
  * Add `minimax` platform configuration for the MiniMax bridge
  * Autoconfigure `SchemaProviderInterface` implementations so they can be referenced from `#[Schema(provider: ...)]`, and validate those references on tools at container build time
  * Add support for `ScopingHttpClient` usage in `Typesense` store via `http_client` option
--- a/demo/tests/AiAssertionsTest.php
+++ b/demo/tests/AiAssertionsTest.php
@@ -1,0 +1,54 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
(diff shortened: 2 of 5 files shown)


src/ai-bundle/CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ CHANGELOG
 0.11
 ----
 
+ * Add `AiAssertionsTrait` to assert in functional tests what the AI components did during a request:
+   the platforms that were called, the tools that were invoked, and the tools that are registered
```
chr-hertel:
```suggestion
 * Add `AiAssertionsTrait` for functional tests based on Profiler data
```

PR #2314, opened 2026-07-14
[Platform] Allow scoping generated JSON Schema to serializer groups

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | -
| License       | MIT

`PropertyInfoDescriber` already supports serializer-group filtering, but only via its constructor — there was no way to pass groups per call.

This threads an optional `$context` from `Factory` down to the describers:

* `Factory::buildProperties()` / `buildParameters()` accept `array $context = []`
* `ObjectSubject` carries it and exposes `getContext()`
* `PropertyInfoDescriber` prefers a per-call `serializer_groups` context over its constructed groups
* `SerializerDescriber` propagates the context into discriminator `anyOf` sub-schemas

```php
$factory->buildProperties(GroupedDto::class, ['serializer_groups' => ['write']]);
// => only the properties in the "write" group
```

Fully backward compatible: the new arguments default to `[]` / `['*']`, so existing callers and emitted schemas are unchanged.

### DIFF
--- a/src/platform/CHANGELOG.md
+++ b/src/platform/CHANGELOG.md
@@ -4,6 +4,7 @@ CHANGELOG
 0.11
 ----
 
+ * Add optional `$context` argument to `Contract\JsonSchema\Factory` build methods to limit schema to subset of properties
  * Add `MessageBag::isLastMessageFrom()` to check whether the last message has a given role
  * Add `MaxOutputTokensException` for completed responses truncated by output token limits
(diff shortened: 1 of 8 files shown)


src/platform/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ CHANGELOG
 0.11
 ----
 
+ * Add optional `$context` argument to `Contract\JsonSchema\Factory::buildProperties()` and `buildParameters()`, carried to the describers via the new `ObjectSubject::getContext()`. Passing `['serializer_groups' => ['write']]` scopes the generated JSON Schema to the given serializer groups, so structured output and tools can expose a subset of a class (e.g. an editable view) instead of every property
```
chr-hertel:
```suggestion
 * Add optional `$context` argument to `Contract\JsonSchema\Factory` build methods to limit schema to subset of properties
```

PR #2315, opened 2026-07-14
[AI Bundle] Pass the serializer to TemplateRendererListener

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | -
| License       | MIT

## Problem

`TemplateRendererListener` accepts an optional `NormalizerInterface` as its second argument and uses it to normalize **object** template variables, honoring `template_options['normalizer_context']` (e.g. serializer `groups`):

```php
if (\is_object($value) && !$value instanceof \Stringable) {
    if (null === $this->normalizer) {
        throw new InvalidArgumentException(\sprintf('Template variable "%s" is an object but no normalizer is configured. ...', $key));
    }

    $normalized[$key] = $this->normalizer->normalize($value, null, $normalizerContext);
}
```

The bundle registers the listener with **only** the renderer registry:

```php
->set('ai.platform.template_renderer_listener', TemplateRendererListener::class)
    ->args([
        service('ai.platform.template_renderer_registry'),
    ])
    ->tag('kernel.event_subscriber')
```

so `$normalizer` is always `null`. Passing an object as a template variable therefore throws, and the entire object-normalization feature — together with `template_options['normalizer_context']` — is unreachable in a Symfony application. Only scalar template vars work today.

## Solution

Wire the `serializer` service as the second argument (`nullOnInvalid()`, since the argument is optional and the serializer may not be installed).

This makes objects usable as template variables, with the normalizer context applied:

```php
    'template_vars' => ['star' => $star],
    'template_options' => ['normalizer_context' => ['groups' => ['prompt']]],
]);
```

Combined with `StringTemplateRenderer`'s array flattening, the normalized object is addressable with dotted placeholders — `{star.name}`, `{star.yearOfBirth}` — and serializer groups control exactly which fields reach the prompt.

## BC

None. The argument was already optional; this only supplies it. Behavior for scalar template vars is unchanged.

## Tests

`AiBundleTest::testTemplateRendererListenerReceivesNormalizer()` asserts the listener is registered with both arguments and still tagged `kernel.event_subscriber`.

Full `src/ai-bundle` suite is green (404 tests) and PHPStan reports no errors.

### DIFF
--- a/src/ai-bundle/CHANGELOG.md
+++ b/src/ai-bundle/CHANGELOG.md
@@ -4,6 +4,7 @@ CHANGELOG
 0.11
 ----
 
+ * Pass the `serializer` to `TemplateRendererListener` so objects can be used as `template_vars`
  * Add `AiAssertionsTrait` for functional tests based on Profiler data
  * Add `minimax` platform configuration for the MiniMax bridge
  * Autoconfigure `SchemaProviderInterface` implementations so they can be referenced from `#[Schema(provider: ...)]`, and validate those references on tools at container build time
--- a/docs/components/platform.rst
+++ b/docs/components/platform.rst
@@ -392,6 +392,45 @@ Multiple messages can use the same variable set::
         ],
     ]);
 
+Object Variables
+................
+
+Template variables are not restricted to scalar values - objects can be passed as
+well. They are normalized into an array first, and since the string renderer
+flattens nested arrays into dot-paths, the object's properties are addressable
+with dotted placeholders::
+
+    $messages = new MessageBag(
+        Message::forSystem('You are a product copywriter.'),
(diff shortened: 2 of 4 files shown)


src/ai-bundle/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ CHANGELOG
 0.11
 ----
 
+ * Pass the `serializer` to `TemplateRendererListener` so objects can be used as `template_vars`. Without a normalizer the listener rejected object template variables, which made `template_options.normalizer_context` (e.g. serializer `groups`) unreachable
```
chr-hertel:
```suggestion
 * Pass the `serializer` to `TemplateRendererListener` so objects can be used as `template_vars`
```

PR #2317, opened 2026-07-14
[Platform] Let the model router select the model, not just the provider

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | -
| License       | MIT

Lets the Platform model router decide **which model** serves a request, not only **which provider** — turning the routing layer into a home for input-/rule-/cost-based model selection. This is the idea behind #1090, one layer down so it applies to every Platform consumer, not just `Agent::call()`.

- `ModelRouterInterface::resolve()` accepts `string|Model` and returns a `RoutingDecision` (provider + optional model + optional options + reason) instead of a bare `ProviderInterface`. `null` model/options keep the requested ones; rewriting the options lets a decision fit them to the selected provider or model.
- `CatalogBasedModelRouter` never replaces the model, so it's the terminal resolver that model-selecting routers delegate to for provider lookup.
- `Model` instances now route through `ModelRoutingEvent` + the router like names do (previously bypassed both); `ModelRoutingEvent::getModel()` can return `string|Model`.

**BC breaks** (see `UPGRADE.md`): `resolve()` signature widened — custom routers wrap their provider in `new RoutingDecision(...)`; a provider with no `ModelClient` for a model now throws `ModelNotFoundException` instead of `RuntimeException`.

This is just the routing primitive. Concrete routers (rule-based, cost-based, AI-powered) compose on top as separate `ModelRouterInterface` implementations and can live in user-land or land later. Also out of scope: input **transformation** during routing (better as a `ModelRoutingEvent` listener) and `providers:`/`model_router:` bundle config (follow-up).

### DIFF
--- a/src/platform/src/Platform.php
+++ b/src/platform/src/Platform.php
@@ -13,10 +13,10 @@ namespace Symfony\AI\Platform;
 
 use Symfony\AI\Platform\Event\ModelRoutingEvent;
 use Symfony\AI\Platform\Exception\InvalidArgumentException;
-use Symfony\AI\Platform\Exception\ModelNotFoundException;
 use Symfony\AI\Platform\ModelCatalog\CompositeModelCatalog;
 use Symfony\AI\Platform\ModelCatalog\ModelCatalogInterface;
 use Symfony\AI\Platform\ModelRouter\CatalogBasedModelRouter;
+use Symfony\AI\Platform\ModelRouter\RoutingDecision;
 use Symfony\AI\Platform\Result\DeferredResult;
 use Symfony\Contracts\EventDispatcher\EventDispatcherInterface;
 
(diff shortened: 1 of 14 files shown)


chr-hertel:
@wachterjohannes having some of your use-cases in mind here - i guess you solved it already differently or is this interesting for you?

wachterjohannes:
> @wachterjohannes having some of your use-cases in mind here - i guess you solved it already differently or is this interesting for you?

i have not rewritten this part of Sulu.ai with symfony/ai - but this PR solves a real blind spot for us 🎉

src/platform/src/Platform.php
```diff
@@ -47,17 +47,18 @@ public function __construct(
...
-            ?? $this->modelRouter->resolve($event->getModel(), $this->providers, $event->getInput(), $event->getOptions());
+        $decision = null !== $event->getProvider()
+            ? new RoutingDecision($event->getProvider())
+            : $this->modelRouter->resolve($event->getModel(), $this->providers, $event->getInput(), $event->getOptions());
 
-        return $provider->invoke($event->getModel(), $event->getInput(), $event->getOptions());
+        return $decision->getProvider()->invoke(
+            $decision->getModel() ?? $event->getModel(),
+            $event->getInput(),
+            $event->getOptions(),
```
wachterjohannes: on thing we should think about - different providers/models support the options differently! so maybe the decision could also rewrite them! we would just need to use:


```suggestion
            $decision->getOptions(),
```

What do you think?

wachterjohannes (changes requested):
I like the rest 👍

src/platform/src/Platform.php
chr-hertel: true 👍

wachterjohannes (approved):

PR #2318, opened 2026-07-15
[MCP Bundle][Mate] Support mcp/sdk 0.7

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | -
| License       | MIT

Widen the `mcp/sdk` constraint to `^0.6 || ^0.7` in the MCP Bundle and Mate. v0.7's changes are additive for both consumers (the new `Registry` `$loader` constructor arg is optional; mate loads eagerly via `$loader->load()` and the bundle uses `setRegistry()`, so the lazy-loading BC break doesn't reach them). PHPStan and the full test suites pass on both 0.6 and 0.7.

cc @wachterjohannes

### DIFF
--- a/src/mate/composer.json
+++ b/src/mate/composer.json
@@ -31,12 +31,12 @@
     ],
     "require": {
         "php": ">=8.2",
-        "mcp/sdk": "^0.6",
+        "mcp/sdk": "^0.6|^0.7",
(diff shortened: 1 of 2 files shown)


wachterjohannes:
@chr-hertel thanks 🎉

PR #2323, opened 2026-07-15
[Platform] Make Gemini and VertexAI tolerant of empty tool-call roundtrips

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | -
| License       | MIT

Surfaced while running the examples. Two robustness gaps in the Gemini and VertexAI result converters, both around tool-call roundtrips:

- Gemini can return a valid HTTP 200 completion with a terminal finish reason but no `content.parts` (e.g. an empty message after a tool result). The converters threw `Response does not contain any content.` on this valid response. They now return an empty text result carrying the finish reason, and only throw when no candidate is present at all. Fixes `gemini/toolcall-roundtrip`, `vertexai/toolcall`, `rag/mariadb-gemini`.
- Gemini emits an empty string for optional tool-call arguments it has no value for, whereas OpenAI/Anthropic send `null`. That empty string broke denormalization of nullable properties such as `DateTime`. Arguments are now normalized so `""` becomes `null`, consistent across bridges. Fixes `agent/toolcall-polymorphic-interface` on Gemini.

Tests added for both behaviors in each bridge.

### DIFF
--- a/src/platform/src/Bridge/Gemini/Gemini/ResultConverter.php
+++ b/src/platform/src/Bridge/Gemini/Gemini/ResultConverter.php
@@ -95,6 +95,16 @@ final class ResultConverter implements ResultConverterInterface
                 throw new RuntimeException(\sprintf('Error "%s" - "%s": "%s".', $data['error']['code'], $data['error']['status'], $data['error']['message']));
             }
 
+            // Gemini can return a well-formed completion with a terminal finish reason but no
+            // content parts (e.g. an empty message after a tool result). Treat it as empty text
+            // instead of crashing on an otherwise valid response.
+            if (isset($data['candidates'][0]['finishReason'])) {
+                return $this->withFinishReason(
+                    new TextResult(''),
+                    FinishReasonMapper::map($data['candidates'][0]['finishReason']),
+                );
+            }
+
             throw new RuntimeException('Response does not contain any content.');
         }
 
@@ -208,6 +218,30 @@ final class ResultConverter implements ResultConverterInterface
      */
     private function convertToolCall(array $toolCall, ?string $signature = null): ToolCall
     {
-        return new ToolCall($toolCall['id'] ?? '', $toolCall['name'], $toolCall['args'], $signature);
+        return new ToolCall($toolCall['id'] ?? '', $toolCall['name'], $this->normalizeArguments($toolCall['args']), $signature);
+    }
+
+    /**
+     * Gemini emits empty strings for optional object properties it has no value for, whereas other
+     * providers omit them or send null. Coerce those empty strings to null (recursing into nested
+     * structures) so downstream denormalization — e.g. of nullable DateTime properties — behaves
+     * consistently across bridges. List elements (integer-keyed) are left untouched, as an empty
+     * string can be a legitimate value inside a list argument.
+     *
+     * @param mixed[] $arguments
+     *
+     * @return mixed[]
+     */
+    private function normalizeArguments(array $arguments): array
+    {
+        foreach ($arguments as $key => $value) {
+            if (\is_array($value)) {
+                $arguments[$key] = $this->normalizeArguments($value);
+            } elseif ('' === $value && !\is_int($key)) {
+                $arguments[$key] = null;
+            }
+        }
+
+        return $arguments;
     }
 }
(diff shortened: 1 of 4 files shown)


src/platform/src/Bridge/Gemini/Gemini/ResultConverter.php
```diff
@@ -208,6 +218,28 @@ private function convertPart(array $contentPart): ToolCallResult|TextResult|Thin
...
+     */
+    private function normalizeArguments(array $arguments): array
+    {
+        foreach ($arguments as $key => $value) {
+            if ('' === $value) {
+                $arguments[$key] = null;
+            } elseif (\is_array($value)) {
+                $arguments[$key] = $this->normalizeArguments($value);
+            }
+        }
```
chr-hertel, line 243: The comment scopes this to *optional named arguments*, but the coercion is broader than that: it rewrites **every** `''` to `null`, including empty-string elements inside list arguments (`["a", "", "b"]` → `["a", null, "b"]`) and required string params the model legitimately set to `""`. For the motivating nullable-`DateTime` case that's fine, but the breadth is a bit surprising given the docstring.

Optional refinement — only null out object-shaped (string-keyed) entries, so list elements are left intact while nested objects still get normalized:

```suggestion
        foreach ($arguments as $key => $value) {
            if (\is_array($value)) {
                $arguments[$key] = $this->normalizeArguments($value);
            } elseif ('' === $value && !\is_int($key)) {
                $arguments[$key] = null;
            }
        }
```

(The existing test stays green — all its keys are string-keyed.) If you'd rather keep it simple, at least loosening the docstring to say "all empty-string values" would avoid the mismatch.

src/platform/src/Bridge/VertexAi/Gemini/ResultConverter.php
```diff
@@ -226,6 +236,28 @@ private function convertPart(array $contentPart): ToolCallResult|TextResult|Thin
...
+     */
+    private function normalizeArguments(array $arguments): array
+    {
+        foreach ($arguments as $key => $value) {
+            if ('' === $value) {
+                $arguments[$key] = null;
+            } elseif (\is_array($value)) {
+                $arguments[$key] = $this->normalizeArguments($value);
+            }
+        }
```
chr-hertel, line 261: Same note as the Gemini bridge — the `''` → `null` coercion also rewrites list elements and required strings, not just optional named args. Mirror whichever refinement you pick there:

```suggestion
        foreach ($arguments as $key => $value) {
            if (\is_array($value)) {
                $arguments[$key] = $this->normalizeArguments($value);
            } elseif ('' === $value && !\is_int($key)) {
                $arguments[$key] = null;
            }
        }
```

src/platform/src/Bridge/Gemini/Gemini/ResultConverter.php
```diff
@@ -95,6 +95,16 @@ public function convert(RawResultInterface|RawHttpResult $result, array $options
                 throw new RuntimeException(\sprintf('Error "%s" - "%s": "%s".', $data['error']['code'], $data['error']['status'], $data['error']['message']));
             }
 
+            // Gemini can return a well-formed completion with a terminal finish reason but no
+            // content parts (e.g. an empty message after a tool result). Treat it as empty text
+            // instead of crashing on an otherwise valid response.
+            if (isset($data['candidates'][0]['finishReason'])) {
```
chr-hertel, line 101: Edge case, likely out of scope — flagging for awareness. This branch keys entirely off `candidates[0]`. With `candidateCount > 1`, if `candidates[0]` has no parts but a later candidate *does* carry content, we now return an empty `TextResult` and silently drop the others, whereas the normal path a few lines down builds a `ChoiceResult` over all candidates. The old code threw here, so this is still an improvement — just noting the asymmetry in case multi-candidate responses are in play. No change requested.

chr-hertel (commented):
Cross-checked the two robustness claims against the Gemini/VertexAI docs — both hold up (content is legitimately omitted on terminal `finishReason`, and optional/`nullable` args are model-populated-or-skipped). The fix is clean and symmetric. A couple of inline notes with optional patches.

PR #2331, opened 2026-07-15
[Platform][Mistral] Add Voxtral speech-to-text (opt-in transcription endpoint)

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | -
| License       | MIT

Adds **Voxtral speech-to-text** support to the Mistral bridge (dedicated
`POST /v1/audio/transcriptions` endpoint), mirroring the OpenAI `Whisper` and Cohere
`SpeechToText` bridges: `SpeechToText` model class, `ModelClient`, `ResultConverter` and
`AudioNormalizer`.

### `voxtral-mini-latest` serves multiple purposes

Unlike `whisper-1` or `cohere-transcribe-03-2026` (transcription-only names),
`voxtral-mini-latest` is a fully capable chat model that Mistral *also* exposes on the
transcription endpoint — the same name has two meanings. Since the `ModelCatalog` maps one
name to one class, it stays the chat `Mistral` model (no BC break) and transcription is
opt-in via an explicit `SpeechToText` instance. That is why this example wraps the model
where the OpenAI/Cohere ones pass a plain string.

### Usage

```php
use Symfony\AI\Platform\Bridge\Mistral\SpeechToText;

$result = $platform->invoke(new SpeechToText('voxtral-mini-latest'), Audio::fromFile('note.mp3'), ['language' => 'en']);
echo $result->asText();
```

Chat / audio understanding keeps working with the plain name: `$platform->invoke('voxtral-mini-latest', $messages)`.

### DIFF
--- a/src/platform/src/Bridge/Mistral/SpeechToText/AudioNormalizer.php
+++ b/src/platform/src/Bridge/Mistral/SpeechToText/AudioNormalizer.php
@@ -1,0 +1,47 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Platform\Bridge\Mistral\SpeechToText;
+
+use Symfony\AI\Platform\Bridge\Mistral\SpeechToText;
+use Symfony\AI\Platform\Contract;
+use Symfony\AI\Platform\Contract\Normalizer\ModelContractNormalizer;
+use Symfony\AI\Platform\Message\Content\Audio;
+use Symfony\AI\Platform\Model;
+
+/**
+ * @author Johannes Wachter <johannes@sulu.io>
+ */
+final class AudioNormalizer extends ModelContractNormalizer
+{
+    /**
+     * @param Audio $data
+     *
+     * @return array{model: string, file: resource}
+     */
+    public function normalize(mixed $data, ?string $format = null, array $context = []): array
+    {
+        return [
+            'model' => $context[Contract::CONTEXT_MODEL]->getName(),
+            'file' => $data->asResource(),
+        ];
+    }
+
+    protected function supportedDataClass(): string
+    {
+        return Audio::class;
+    }
+
+    protected function supportsModel(Model $model): bool
+    {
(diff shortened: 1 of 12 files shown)


chr-hertel:
uh another weird new combination - one model + two endpoints = different capabilities :joy:

wachterjohannes:
> uh another weird new combination - one model + two endpoints = different capabilities :joy: 

yeah its so fucking uncosistent all the times 😅

wachterjohannes:
@chr-hertel could you take a look? `voxtral-mini-latest` is both a chat model and a transcription model at Mistral, so the catalog keeps it as the chat `Mistral` model and transcription is opt-in via an explicit `SpeechToText` instance. I verified both paths reach the right endpoint and the chat one is untouched.

One thing i'd rather decide with you: the converter returns `null` from `getTokenUsageExtractor()` and puts the raw provider array into the metadata instead, so the same model name yields `token_usage` on the chat path and a plain `usage` array on the transcription path. `Embeddings` and `Llm` both have an extractor, `Ocr` has none. Wire one up here too, or drop the metadata line and stay with the Cohere sibling, which has neither?

src/platform/src/Bridge/Mistral/SpeechToText/AudioNormalizer.php
```diff
@@ -0,0 +1,48 @@
...
+
+use Symfony\AI\Platform\Bridge\Mistral\SpeechToText;
+use Symfony\AI\Platform\Contract;
+use Symfony\AI\Platform\Message\Content\Audio;
+use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
+
+/**
+ * @author Johannes Wachter <johannes@sulu.io>
+ */
+final class AudioNormalizer implements NormalizerInterface
```
chr-hertel: could be extending the `Symfony\AI\Platform\Contract\Normalizer\ModelContractNormalizer`

wachterjohannes: Done, it now extends `ModelContractNormalizer`.

wachterjohannes:
@chr-hertel updated, ready for another look.

chr-hertel (approved):

chr-hertel:
Thank you Johannes.

PR #2338, opened 2026-07-16
 Fix Amazee AI subtree config

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | 
| License       | MIT

@fabpot is that the fix then?


fabpot (approved):
Yes

PR #1784, opened 2026-03-16
[Documentation] Add cookbook for context compression

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | yes
| Issues        | 
| License       | MIT

Related to #1549 and #1301

PR #2340, opened 2026-07-16
[Documentation] Add cookbook for a generator and reviewer loop

vinceAmstoutz:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | yes
| Issues        |
| License       | MIT

This PR has been discussed [here](https://x.com/chr_hertel/status/2069138819824189660?s=20).

This adds a cookbook recipe for a pattern that keeps coming up but isn't covered yet: pairing an **generating** agent with a separate **reviewing** agent in an iterative loop, so the pair reaches a higher precision than a single pass can.

The generator proposes candidate findings (as structured output), a skeptical reviewer confirms or rejects each one, and confirmed findings, plus the reviewer's reason for each rejection, are fed back into the next generator prompt so later rounds refine instead of re-reporting settled findings. A confidence floor filters candidates before the (expensive) review step, and the loop stops early once a round confirms nothing new.

The running example is a security auditor (generator hunts for vulnerabilities, reviewer culls false positives), distilled from the attacker/reviewer loop in [`vinceamstoutz/symfony-security-auditor`](https://github.com/vinceAmstoutz/symfony-security-auditor), but the recipe frames the pattern generically, it applies equally to drafting + fact-checking, code generation + critique, or extraction + validation.

It uses only the public Platform + Agent APIs (`Agent`, `SystemPromptInputProcessor`, structured output via `response_format`) and closes with a "Using the AI Bundle?" section and best practices. It slots into the *Agents & Tools* category, next to `multi-agent-orchestration` (routing) as its refinement-oriented counterpart.

Modeled on #1784. The `ai.symfony.com/` cookbook artifacts (`config/cookbook.json` + the content fragment) are regenerated from the RST via `bin/console app:cookbook:build` and committed alongside the source.

### DIFF
--- a/ai.symfony.com/config/cookbook.json
+++ b/ai.symfony.com/config/cookbook.json
@@ -47,6 +47,15 @@
         ]
     },
     {
+        "slug": "generator-reviewer-loop",
+        "title": "Generator and Reviewer Loop",
+        "description": "Pair a generating agent with a reviewing agent that validates its findings over several iterations.",
+        "icon": "shield-search",
+        "components": [
+            "Agent"
+        ]
+    },
+    {
         "slug": "runtime-driven-tool-parameters",
         "title": "Runtime-driven Tool Parameters",
         "description": "Constrain tool parameters and structured output with values from env, a database, or services.",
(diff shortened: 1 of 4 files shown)


vinceAmstoutz:
@chr-hertel I hesitated to mention [symfony-security-auditor](https://github.com/vinceAmstoutz/symfony-security-auditor) in the cookbook but I don't think this is the right place

chr-hertel (approved):
Thanks @vinceAmstoutz - that's a good read! :+1:

PR #2343, opened 2026-07-17
[Platform][Mistral] Extract cached tokens from prompt_tokens_details

### LINKED ISSUE #2236: [Platform][Mistral] TokenUsageExtractor drops prompt_tokens_details.cached_tokens (prompt caching not reported in usage)
Hi there,

I've just integrated Mistral AI and related models into my application and I realised that I did not have any information about cached token usage. Here is a detailed note about it. Do not hesitate to correct me if I'm wrong into my appreciation.

### Description

Mistral's Chat Completions API (`/v1/chat/completions`) supports **prompt caching** and
returns the number of cached prompt tokens in the response under
`usage.prompt_tokens_details.cached_tokens` (cached tokens are billed at 10% of the input
price). See the official docs:
https://docs.mistral.ai/studio-api/conversations/advanced/prompt-caching

It seems that the Mistral bridge's `TokenUsageExtractor` never reads that field, so the resulting
`TokenUsage` always has `cachedTokens === null`. Any application relying on `TokenUsage` for
cost accounting or cache observability cannot see Mistral cached tokens, which leads to an
input-cost overestimate (the cached portion is billed at full rate instead of 10%).

### Affected code

`Symfony\AI\Platform\Bridge\Mistral\Llm\TokenUsageExtractor`
(`src/platform/src/Bridge/Mistral/Llm/TokenUsageExtractor.php`)

Current implementation only maps `prompt_tokens`, `completion_tokens`, `total_tokens` and the
rate-limit headers:

```php
return new TokenUsage(
    promptTokens: $content['usage']['prompt_tokens'] ?? null,
    completionTokens: $content['usage']['completion_tokens'] ?? null,
    remainingTokensMinute: null !== $remainingTokensMinute ? (int) $remainingTokensMinute : null,
    remainingTokensMonth: null !== $remainingTokensMonth ? (int) $remainingTokensMonth : null,
    totalTokens: $content['usage']['total_tokens'] ?? null,
);
```

`cachedTokens` is missing. The `TokenUsage` value object already supports it
(`TokenUsage::__construct(..., ?int $cachedTokens = null, ...)`), and the OpenAI bridge
populates it from `input_tokens_details.cached_tokens` in
`Symfony\AI\Platform\Bridge\OpenAi\Gpt\TokenUsageExtractor`, so there is an established pattern.

### How to reproduce

1. Send two consecutive chat-completion requests that share a long common prefix (e.g. a large
   system prompt) to a Mistral model, with `prompt_cache_key` set on the request.
2. Inspect the returned `TokenUsage`.

**Expected:** `getCachedTokens()` returns the value of `usage.prompt_tokens_details.cached_tokens`
(a non-zero multiple of 64 once the cache is warm).

**Actual:** `getCachedTokens()` is always `null`.

### Versions

- `symfony/ai-platform` / Mistral bridge: 0.10 (also present on `main` at time of writing)
- PHP: 8.4

Hope it is clear.
Do not hesitate if I'm doing something wrong.
Thanks for your amazing work and support.

Best !🤘

codexofc:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #2236
| License       | MIT

Mistral's chat completions API reports cached prompt tokens under `usage.prompt_tokens_details.cached_tokens` when prompt caching kicks in ([docs](https://docs.mistral.ai/studio-api/conversations/advanced/prompt-caching)), but the Mistral `TokenUsageExtractor` never read that field, so `TokenUsage::getCachedTokens()` was always `null` — cost accounting overestimates the input cost since cached tokens are billed at 10%.

This maps the field the same way the DeepSeek, Gemini and Codex bridges already do for their equivalents.

Verified against the live Mistral API with two requests sharing a `prompt_cache_key`:

```
usage: { "prompt_tokens": 801, "completion_tokens": 2, "prompt_tokens_details": { "cached_tokens": 768 } }
TokenUsage->getCachedTokens(): 768   // was null before the fix
```

(768 is a multiple of 64, matching the documented cache block size.)

### DIFF
--- a/src/platform/src/Bridge/Mistral/Llm/TokenUsageExtractor.php
+++ b/src/platform/src/Bridge/Mistral/Llm/TokenUsageExtractor.php
@@ -43,6 +43,7 @@ final class TokenUsageExtractor implements TokenUsageExtractorInterface
         return new TokenUsage(
             promptTokens: $content['usage']['prompt_tokens'] ?? null,
             completionTokens: $content['usage']['completion_tokens'] ?? null,
+            cachedTokens: $content['usage']['prompt_tokens_details']['cached_tokens'] ?? null,
             remainingTokensMinute: null !== $remainingTokensMinute ? (int) $remainingTokensMinute : null,
             remainingTokensMonth: null !== $remainingTokensMonth ? (int) $remainingTokensMonth : null,
             totalTokens: $content['usage']['total_tokens'] ?? null,
--- a/src/platform/src/Bridge/Mistral/Tests/TokenUsageExtractorTest.php
+++ b/src/platform/src/Bridge/Mistral/Tests/TokenUsageExtractorTest.php
@@ -53,6 +53,28 @@ final class TokenUsageExtractorTest extends TestCase
         $this->assertSame(10, $tokenUsage->getPromptTokens());
         $this->assertSame(20, $tokenUsage->getCompletionTokens());
         $this->assertSame(30, $tokenUsage->getTotalTokens());
+        $this->assertNull($tokenUsage->getCachedTokens());
+    }
+
+    public function testItExtractsCachedTokens()
+    {
+        $extractor = new TokenUsageExtractor();
+        $result = new InMemoryRawResult([
+            'usage' => [
+                'prompt_tokens' => 128,
+                'completion_tokens' => 20,
+                'total_tokens' => 148,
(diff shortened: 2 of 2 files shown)


chr-hertel (approved):
Thanks @codexofc!

PR #2344, opened 2026-07-17
[Platform][Mistral] Drop stale conversion overrides rejecting truncated responses

codexofc:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | -
| License       | MIT

Any Mistral chat completion truncated by `max_tokens` currently throws instead of returning the truncated text:

```
Symfony\AI\Platform\Exception\RuntimeException: Unsupported finish reason "length".
```

Root cause: the Mistral `ResultConverter` uses `CompletionsConversionTrait` but still carries local copies of `convertChoice()` and `convertToolCall()` from before the trait extraction (f08cca25). The local `convertChoice()` only accepts `stop` and `tool_calls`, while its own phpdoc documents `finish_reason: 'stop'|'length'|'tool_calls'|'content_filter'`.

Removing the stale overrides lets the trait handle the conversion, exactly like the Generic bridge already behaves:

- `length` → `TextResult` with the truncated content and the `LENGTH` finish reason (verified against the live Mistral API: a `max_tokens: 30` request now returns the truncated text with `finish_reason.raw = "length"` instead of throwing)
- empty tool-call `arguments` → empty array instead of a `JsonException` (the trait guards, the local copy force-decoded)

Net diff: +26/−54.

### DIFF
--- a/src/platform/src/Bridge/Mistral/Llm/ResultConverter.php
+++ b/src/platform/src/Bridge/Mistral/Llm/ResultConverter.php
@@ -12,7 +12,6 @@
 namespace Symfony\AI\Platform\Bridge\Mistral\Llm;
 
 use Symfony\AI\Platform\Bridge\Generic\Completions\CompletionsConversionTrait;
-use Symfony\AI\Platform\Bridge\Generic\Completions\FinishReasonMapper;
 use Symfony\AI\Platform\Bridge\Mistral\Mistral;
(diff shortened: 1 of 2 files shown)


chr-hertel (approved):
Nice cleanup, thanks @codexofc!

PR #2345, opened 2026-07-17
[Chat] Fix TypeError in Chat::submit() on non-text results

### LINKED ISSUE #2146: [Chat] Chat::submit() throws TypeError on MultiPartResult (reasoning models)
## Describe the bug

  `Chat::submit()` assumes the agent always returns a `TextResult`:

  https://github.com/symfony/ai/blob/main/src/chat/src/Chat.php#L46-L48

  ```php
```

  When the agent returns a MultiPartResult (e.g. a reasoning/thinking part plus the
  answer text, which OpenAI Responses reasoning models produce), getContent() returns an
  array, and the variadic Message::ofAssistant() rejects it:

```
  Message::ofAssistant(): Argument #1 must be of type
  ContentInterface|ResultInterface|string, array given,
  called in vendor/symfony/ai-chat/src/Chat.php on line 48
```

  The \assert() is a no-op in production (zend.assertions=-1), so it does not guard against
  this. AgentProcessor::processOutput() only unwraps a MultiPartResult to look for a
  ToolCallResult, so a final multi-part answer reaches Chat untouched.

  ### Expected

  Chat::submit() handles any ResultInterface. Message::ofAssistant() already unwraps
  MultiPartResult recursively, so passing the result object instead of its content works:

```php
  $assistantMessage = Message::ofAssistant($result);
```

  ### Actual

  TypeError whenever the final result is a MultiPartResult (intermittent — only on turns
  where the model emits a multi-part response).

  Steps to reproduce

  1. Configure a Chat with an OpenAI Responses reasoning model (e.g. gpt-5.4-mini) and a tool.
  2. Submit a prompt that triggers a tool call followed by a multi-part final answer
  (reasoning/thinking part + text).
  3. Chat::submit() throws the TypeError above.

  Versions

  - symfony/ai-chat, symfony/ai-agent, symfony/ai-platform,
  symfony/ai-open-ai-platform: v0.9.0
  - Model: gpt-5.4-mini (OpenAI Responses API)
  - The offending line is still present on main.

  Suggested fix

  Pass the result object to Message::ofAssistant() (it already maps TextResult,
  ThinkingResult, MultiPartResult, etc.) and drop the TextResult-only assumption:

```
  -\assert($result instanceof TextResult);
  -$assistantMessage = Message::ofAssistant($result->getContent());
  +$assistantMessage = Message::ofAssistant($result);
```

  Happy to open a PR if that direction looks right. Thanks for the great work on this project! 🙏

janssensglenn:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #2146
| License       | MIT

`Chat::submit()` assumed the agent always returns a `TextResult`:

```php
```

The `\assert()` is a no-op in production (`zend.assertions=-1`), so it never guarded
anything. When the agent returns a `MultiPartResult` — for example a reasoning/thinking
part plus the answer text, as OpenAI Responses reasoning models produce — `getContent()`
returns an array, and the variadic `Message::ofAssistant()` rejects it:

```
Message::ofAssistant(): Argument #1 must be of type
ContentInterface|ResultInterface|string, array given
```

This surfaced as an intermittent `TypeError`, only on turns where the model emitted a
multi-part response.

### Fix

`Message::ofAssistant()` already accepts a `ResultInterface` and maps every result type
recursively via `Message::toContent()` — `TextResult`, `ThinkingResult`, `ToolCallResult`,
`MultiPartResult`, etc. So the fix is to pass the result object straight through and drop
the `TextResult`-only assumption:

```diff
-        \assert($result instanceof TextResult);
-
-        $assistantMessage = Message::ofAssistant($result->getContent());
+        $assistantMessage = Message::ofAssistant($result);
```

The now-unused `TextResult` import is removed. The metadata merge on the next line is
unaffected (`ResultInterface extends MetadataAwareInterface`), and the existing
`TextResult` path is preserved by `toContent()`.

No test changes: multi-part unwrapping is `Message::toContent()`'s responsibility (covered
in the Platform component), not `Chat`'s — `Chat` contains no multi-part logic. The
existing `ChatTest` coverage of `submit()`'s contract continues to pass.

### DIFF
--- a/src/chat/src/Chat.php
+++ b/src/chat/src/Chat.php
@@ -17,7 +17,6 @@ use Symfony\AI\Platform\Message\Message;
 use Symfony\AI\Platform\Message\MessageBag;
 use Symfony\AI\Platform\Message\UserMessage;
 use Symfony\AI\Platform\Result\StreamResult;
-use Symfony\AI\Platform\Result\TextResult;
 
 /**
  * @author Christopher Hertel <mail@christopher-hertel.de>
@@ -43,9 +42,7 @@ final class Chat implements ChatInterface
         $messages->add($message);
         $result = $this->agent->call($messages);
 
-        \assert($result instanceof TextResult);
-
-        $assistantMessage = Message::ofAssistant($result->getContent());
+        $assistantMessage = Message::ofAssistant($result);
         $assistantMessage->getMetadata()->merge($result->getMetadata());
         $messages->add($assistantMessage);
 


chr-hertel (approved):
Thank you @janssensglenn!

PR #2349, opened 2026-07-20
[MCP Bundle] Replace file-based discovery with compile-time container registration

### LINKED ISSUE #2328: [MCP Bundle] Why is tool discovery run at runtime on every request rather than cached/warmed for prod?
I'm trying to understand the intended production model for tool discovery, because as far as I can tell the bundle re scans the filesystem and re runs reflection on every request, and I can't find a way to make prod use a fully cached discovery with no runtime building.

`mcp.server` is a per-request service built via a factory:

```php
->set('mcp.server.builder', Builder::class)
    ->factory([Server::class, 'builder'])
    // ...
    ->call('setDiscovery', [
        param('kernel.project_dir'),
        param('mcp.discovery.scan_dirs'),
        param('mcp.discovery.exclude_dirs'),
    ])
    // ...

->set('mcp.server', Server::class)
    ->factory([service('mcp.server.builder'), 'build'])
;
```

Every `build()` runs the `DiscoveryLoader`, which calls `discoverer->discover()` and re-scans + re-reflects each time. Since the server is built per request, that scan happens on every request.

The SDK clearly anticipates caching this. `Builder::setDiscovery()` takes a PSR-16 cache and wraps the discoverer in `CachedDiscoverer` when one is provided:

```php
// mcp/sdk - Mcp\Server\Builder
public function setDiscovery(
    string $basePath,
    array $scanDirs = ['.', 'src'],
    array $excludeDirs = [],
    ?CacheInterface $cache = null,
): self { /* ... */ }

private function createDiscoverer(LoggerInterface $logger): DiscovererInterface
{
    $discoverer = new Discoverer($logger, null, $this->schemaGenerator);

    if (null !== $this->discoveryCache) {
        return new CachedDiscoverer($discoverer, $this->discoveryCache, $logger);
    }

    return $discoverer;
}
```

But the bundle always calls `setDiscovery()` with three arguments, so the cache is never used, and the config schema (`config/options.php`) exposes no cache option under `discovery` (only `scan_dirs` and `exclude_dirs`).

### Questions

1. Is running discovery at runtime on every request the intended behavior, or am I holding it wrong? Is there a supported way to avoid it in prod that I've missed?
2. Why does the bundle never wire the SDK's `CachedDiscoverer`, given the SDK provides it and `setDiscovery()` accepts a cache argument for exactly this purpose?
3. Why is there no cache warmer that performs discovery at container warmup / deploy time? In prod I'd expect discovery to be computed once and the server assembled from a cached result.

This seems to go against symfony best practises that things like this should be compute once in a cache warmer.

The only guidance I can find for discovery cost so far is "narrow the scanned directories" (`scan_dirs` / `exclude_dirs` from #801, documented in #1660). This also looks like it could be the underlying cause of #1744 ("Long response from Tool", profiler shows `controller.get_callable` ~6s, no change in prod, reporter suspects the bundle scans everything).

### Related

- #1744 - Long response from Tool
- #801 - Add options for tool discovery (added scan_dirs/exclude_dirs; notes the unused `$cache` parameter)
- #1660 - Docs: discovery configuration example

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | Fix #2328
| License       | MIT

Replaces the SDK's runtime file-based discovery with container-based registration at compile time:

* Attribute autoconfiguration now records the attributed method on the `mcp.tool`/`mcp.prompt`/`mcp.resource`/`mcp.resource_template` tags.
* `McpPass` reflects each tagged method once at container compile time and emits `addTool()`/`addPrompt()`/`addResource()`/`addResourceTemplate()` calls on the server builder — including the generated tool input schema — so all metadata is cached in the dumped container. No filesystem scan, no schema generation at request time.
* Handlers stay lazy: `[FQCN, method]` resolved per call via the SDK's `ReferenceHandler` from a class-keyed service locator.
* The `mcp.discovery` config (`scan_dirs`/`exclude_dirs`) is removed (see `UPGRADE.md`); custom sources remain possible via the autoconfigured `mcp.loader` tag.
* Requires `mcp/sdk` ^0.7 (`addTool()` with precomputed `inputSchema`, instance handlers).

Verified with the demo app over stdio in dev and prod: `tools/list`/`tools/call` work, MCP Apps tools unaffected, and the prod container contains the registration verbatim with zero `Discoverer` references.


src/mcp-bundle/CHANGELOG.md
```diff
@@ -1,6 +1,13 @@
 CHANGELOG
 =========
 
+0.12
+----
+
+ * Add compile-time, container-based registration of tools, prompts, resources, and resource
+   templates, replacing the SDK's runtime file-based discovery: attribute metadata and tool input
+   schemas are derived once when the container is compiled and cached in the dumped container
```
chr-hertel:
```suggestion
 * Register tools, prompts, resources, and resource templates via container instead of the SDK's file-based discovery
```

docs/bundles/mcp-bundle.rst
```diff
@@ -36,7 +36,10 @@ To use your application as an MCP server, exposing tools, prompts, resources, an
 Creating MCP Capabilities
 .........................
 
-MCP capabilities are automatically discovered using PHP attributes.
+MCP capabilities are registered using PHP attributes on services: every service carrying one of the
+MCP attributes is picked up automatically at container compile time — there is no runtime file
+scanning involved. In a default Symfony application (with autoconfiguration enabled and the classes
+in ``src/`` registered as services) it is enough to add the attribute to a class.
```
chr-hertel:
```suggestion
MCP capabilities are registered using PHP attributes on services: every service carrying one of the MCP attributes is
picked up automatically at container compile time. In a default Symfony application (with autoconfiguration
enabled and the classes in ``src/`` registered as services) it is enough to add the attribute to a class.
```

src/mcp-bundle/src/DependencyInjection/McpPass.php
```diff
@@ -11,49 +11,277 @@
...
     public function process(ContainerBuilder $container): void
     {
         if (!$container->hasDefinition('mcp.server.builder')) {
             return;
         }
 
-        $allMcpServices = [];
-        $mcpTags = ['mcp.tool', 'mcp.prompt', 'mcp.resource', 'mcp.resource_template'];
+        $builder = $container->getDefinition('mcp.server.builder');
+        $schemaGenerator = new SchemaGenerator(new DocBlockParser(logger: new NullLogger()));
```
chr-hertel: use service definitions here instead - and use a concrete logger instance

src/mcp-bundle/src/DependencyInjection/McpPass.php
```diff
@@ -11,49 +11,277 @@
...
+        }
+
+        return ([] !== $attributes) ? $attributes[0]->newInstance() : null;
+    }
+
+    private function registerTool(Definition $builder, string $class, string $method, McpTool $attribute, SchemaGenerator $schemaGenerator, string $serviceId): void
+    {
+        try {
+            $inputSchema = $schemaGenerator->generate(new \ReflectionMethod($class, $method));
+        } catch (\Throwable $e) {
```
chr-hertel: we could use a more concrete exception here instead of `Throwable`

src/mcp-bundle/src/DependencyInjection/McpPass.php
```diff
@@ -11,49 +11,277 @@
...
+     * @template T of McpTool|McpPrompt|McpResource|McpResourceTemplate
+     *
+     * @param class-string    $class
+     * @param class-string<T> $attributeClass
+     *
+     * @return T|null
+     */
+    private function readAttribute(string $class, string $method, string $attributeClass): ?object
+    {
+        if (!method_exists($class, $method)) {
```
chr-hertel: looks dangerous to swallow misconfigured

PR #2363, opened 2026-07-23
[Platform] Notify stream listeners on a mid-stream error via onError()

### LINKED ISSUE #2360: [Platform] Stream listeners never observe completion when a stream throws an exception, so token usage is lost
Hi there!

First off, thanks for the work on this component — 0.11 brought a lot of good things our way.

I ran into an edge case while upgrading a production Symfony app from 0.10 to 0.11.

My app has an AI chat assistant whose answers are streamed to the browser as they are generated. Every round is metered: we write a usage row, tokens, cost, and a quota debit against the user's plan, from a stream listener's `onComplete`. That felt like the natural place for it, since it is the point where the usage metadata is final.

The edge case is simply a chat answer long enough to reach our `max_tokens` cap. Since 0.11 that raises `MaxOutputTokensException`, and we silently stopped recording those rounds: the user was served an answer (truncated, but perfectly usable), the tokens were spent at the provider, and nothing was written on our side. Nobody was billed for them.

I tried to complete as much as I could the following description. Do not hesitate if I am doing something wrong.

## Description

`StreamResult::getStream()` dispatches `CompleteEvent` only after the `foreach` over the underlying
generator returns normally:

```php
// src/Result/StreamResult.php
foreach ($this->generator as $delta) {
    $event = new DeltaEvent($this, $delta);
    foreach ($this->listeners as $listener) {
        $listener->onDelta($event);
    }
    $this->getMetadata()->merge($event->getMetadata());
    // ...
    yield $delta;
}

$event = new CompleteEvent($this);          // <- never reached if the generator throws
foreach ($this->listeners as $listener) {
    $listener->onComplete($event);
}
```

There is no `try`/`finally`, and `ListenerInterface` exposes no error hook (`onStart`, `onDelta`, `onComplete` only). So when a bridge throws mid-stream, every `onComplete` side effect is silently skipped, even though the stream produced a complete, usable payload and reported its usage.

The built-in `TokenUsage\StreamListener` is unaffected, since it promotes usage from `onDelta`. The casualty is any *consumer* listener that finalizes on `onComplete`, recording cost, writing an audit row, debiting a quota. Those lose the round entirely: the tokens were spent at the provider, but the application ends up with no record of them.

As far as I could see, **the usage is already in the result**.
Actually, `MaxOutputTokensException` is the easiest case to observe, because it is explicitly *not* a failure:
its own docblock says "the response completed normally at the token ceiling". 

The Anthropic converter emits usage on `message_delta`, and only then throws on `message_stop`:

```php
// ai-anthropic-platform/ResultConverter.php
if ('message_delta' === $type) {
    $stopReason = $data['delta']['stop_reason'] ?? $stopReason;
    if (isset($data['usage'])) {
        $outputTokens = $data['usage']['output_tokens'] ?? $outputTokens;
        yield $this->getTokenUsageExtractor()->extractFromArray([...]);   // usage IS emitted
    }
}
// ...
if ('message_stop' === $type) {
    if ('max_tokens' === $stopReason) {
        throw new MaxOutputTokensException($message);                    // ... then this throws
    }
}
```

So the usage delta has already been merged onto the result metadata by the time the exception is raised. As far as I understood, the data is available in the result but it is only the completion notification that is lost.
It seems that  consumer can still read `$result->getMetadata()->get('token_usage')` by catching the exception itself, but a listener registered for accounting cannot: it is never told the stream ended.

The same gap applies to the other mid-stream throws (`ServerException`, `RateLimitExceededException`, `IncompleteStreamException`). From there,  the stream failed, but whatever usage was reported before the failure is lost with it.

## Steps to reproduce

```php
$result = $platform->invoke('claude-haiku-4-5', $messages, [
    'stream' => true,
    'max_tokens' => 16,          // force truncation
])->getResult();

$result->addListener(new class extends AbstractStreamListener {
    public function onComplete(CompleteEvent $event): void
    {
        echo "never printed\n";
    }
});

try {
    foreach ($result->getContent() as $delta) { /* ... */ }
} catch (MaxOutputTokensException) {
    // usage is on the metadata here, but onComplete never fired
    var_dump($result->getMetadata()->get('token_usage'));
}
```

## Possible fixes

**Add an error hook**:  an `ErrorEvent` plus `ListenerInterface::onError()`, dispatched from a
   `catch` before rethrowing. Listeners that only care about usage can then treat both terminal events alike, while listeners that care about success keep today's semantics. Adding the method to the interface does break direct implementers, but a no-op default on `AbstractStreamListener` covers everything extending it, which is the documented way to write a listener.

... or maybe I'm doing something wrong on my side, any advice would be appreciated ! 🙏 

That's it !

Thanks a lot (again) for your amazing work.

Cheers ! :)

ghisleouf:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| Docs?         | no
| Issues        | Fix #2360
| License       | MIT

Follow-up to #2360, which @chr-hertel greenlit there.

`StreamResult::getContent()` dispatches `CompleteEvent` only after the generator drains normally.
A stream that reports its usage on a delta and *then* throws, e.g. `MaxOutputTokensException` on a
response completed at the output-token ceiling, therefore never notifies a listener that finalizes
on `onComplete()`. The usage is sitting on the result metadata, but any accounting side effect
(cost, audit row, quota debit) is silently skipped: the provider billed the round and the
application ends up with no record of it.

This keeps the upcast to an exception, semantically "this is not a clean success" and adds a
terminal event for the failure path instead of blurring `CompleteEvent`:

 * new `Result\Stream\ErrorEvent`, carrying the `\Throwable`;
 * `ListenerInterface::onError()`, with a no-op default on `AbstractStreamListener`;
 * `StreamResult` wraps the drain loop in `try`/`catch`, dispatches `ErrorEvent` to every listener,
   then rethrows.

`onComplete()` and `onError()` are mutually exclusive terminal events, mirroring the
`ResultConvertedEvent` / `ResultErrorEvent` pair added one layer up in 0.11.

Usage: a listener that meters on either terminal event:

```php
final class MyAccountingListener extends AbstractStreamListener
{
    public function onComplete(CompleteEvent $event): void
    {
        $this->meter($event->getResult());
    }

    public function onError(ErrorEvent $event): void
    {
        $this->meter($event->getResult());
    }
}
```

Worth flagging for review: **abandoning** a stream (breaking out of the consuming loop) tears the
generator down *without* entering the `catch`. PHP runs `finally`, not `catch`, on generator
destruction, so an abandoned stream still fires neither event and stays unmetered, exactly as
before. That is precisely why this uses `catch` + rethrow rather than a `finally`.

**BC break:** adding a method to `ListenerInterface` breaks direct implementers (anything extending
`AbstractStreamListener` is covered by the no-op default). Tagged `[BC BREAK]` in
`src/platform/CHANGELOG.md`, with a matching entry in the root `UPGRADE.md`

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -21,6 +21,30 @@ AI Bundle
 
    The default (`false`, tool messages are preserved) is unchanged.
 
+Platform
+--------
+
+ * `Result\Stream\ListenerInterface` gained an `onError()` method, dispatched with the new
+   `Result\Stream\ErrorEvent`. Listeners not extending
+   `AbstractStreamListener` must add the method:
+
+   ```diff
+   +use Symfony\AI\Platform\Result\Stream\ErrorEvent;
+
+    final class MyStreamListener implements ListenerInterface
+    {
+        public function onComplete(CompleteEvent $event): void
+        {
+            // ...
+        }
+   +
+   +    public function onError(ErrorEvent $event): void
+   +    {
+   +        // ...
+   +    }
+    }
+   ```
+
 UPGRADE FROM 0.11 to 0.12
 =========================
 
--- a/src/platform/CHANGELOG.md
+++ b/src/platform/CHANGELOG.md
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+0.13
+----
+
+ * [BC BREAK] Add `ListenerInterface::onError()` and `Result\Stream\ErrorEvent`, dispatched when draining a `StreamResult` throws, so a listener can finalize on a failed stream where `onComplete()` never fires; `AbstractStreamListener` provides a no-op default
+
 0.12
 ----
 
--- a/src/platform/src/Result/StreamResult.php
+++ b/src/platform/src/Result/StreamResult.php
@@ -14,6 +14,7 @@ namespace Symfony\AI\Platform\Result;
 use Symfony\AI\Platform\Result\Stream\CompleteEvent;
 use Symfony\AI\Platform\Result\Stream\Delta\DeltaInterface;
 use Symfony\AI\Platform\Result\Stream\DeltaEvent;
+use Symfony\AI\Platform\Result\Stream\ErrorEvent;
 use Symfony\AI\Platform\Result\Stream\ListenerInterface;
 use Symfony\AI\Platform\Result\Stream\StartEvent;
 
@@ -56,24 +57,36 @@ final class StreamResult extends BaseResult
         }
         $this->getMetadata()->merge($event->getMetadata());
 
-        foreach ($this->generator as $delta) {
-            $event = new DeltaEvent($this, $delta);
+        try {
+            foreach ($this->generator as $delta) {
+                $event = new DeltaEvent($this, $delta);
+                foreach ($this->listeners as $listener) {
+                    $listener->onDelta($event);
+                }
+                $this->getMetadata()->merge($event->getMetadata());
+
+                if ($event->isDeltaSkipped()) {
+                    continue;
+                }
+
+                $delta = $event->getDelta();
+
+                if ($delta instanceof DeltaInterface) {
+                    yield $delta;
+                } else {
+                    yield from $delta;
+                }
+            }
+        } catch (\Throwable $e) {
+            // Notify listeners before rethrowing (see ErrorEvent); an abandoned stream breaks out
+            // and tears the generator down without entering this catch, so it stays unbilled.
+            $event = new ErrorEvent($this, $e);
             foreach ($this->listeners as $listener) {
-                $listener->onDelta($event);
+                $listener->onError($event);
             }
             $this->getMetadata()->merge($event->getMetadata());
 
-            if ($event->isDeltaSkipped()) {
-                continue;
-            }
-
-            $delta = $event->getDelta();
-
-            if ($delta instanceof DeltaInterface) {
-                yield $delta;
-            } else {
-                yield from $delta;
-            }
+            throw $e;
         }
 
         $event = new CompleteEvent($this);
--- a/src/platform/src/Result/Stream/AbstractStreamListener.php
+++ b/src/platform/src/Result/Stream/AbstractStreamListener.php
@@ -27,4 +27,8 @@ abstract class AbstractStreamListener implements ListenerInterface
     public function onComplete(CompleteEvent $event): void
     {
     }
+
+    public function onError(ErrorEvent $event): void
+    {
+    }
 }
--- a/src/platform/src/Result/Stream/ErrorEvent.php
+++ b/src/platform/src/Result/Stream/ErrorEvent.php
@@ -1,0 +1,30 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Platform\Result\Stream;
+
+use Symfony\AI\Platform\Result\StreamResult;
+
+/**
+ * @author Ghislain Flandin <ghislain.flandin@gmail.com>
+ */
+final class ErrorEvent extends Event
+{
+    public function __construct(StreamResult $result, private readonly \Throwable $error)
+    {
+        parent::__construct($result);
+    }
+
+    public function getError(): \Throwable
+    {
+        return $this->error;
+    }
+}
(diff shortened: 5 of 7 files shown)


ghisleouf:
Fixed the changelog/upgrade placement: `v0.12.0` is already tagged. They now live in a new upcoming `0.13` section in `src/platform/CHANGELOG.md`, and a new `UPGRADE FROM 0.12 to 0.13` block in the root `UPGRADE.md`.

src/platform/src/Result/StreamResult.php
```diff
@@ -56,24 +57,36 @@ public function getContent(): \Generator
...
-                yield $delta;
-            } else {
-                yield from $delta;
+                if ($delta instanceof DeltaInterface) {
+                    yield $delta;
+                } else {
+                    yield from $delta;
+                }
             }
+        } catch (\Throwable $e) {
```
chr-hertel, line 80: gets me thinking if really `Throwable` or rather `Symfony\AI\Platform\Exception\ExceptionInterface` 🤔

src/platform/CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+0.13
+----
+
+ * [BC BREAK] Add `ListenerInterface::onError()`, receiving the new `Result\Stream\ErrorEvent`, dispatched when draining a `StreamResult` throws (right before the exception is rethrown to the caller) so a listener can finalize on a stream that reported its token usage on a delta and then threw - e.g. `MaxOutputTokensException` on a response completed at the output-token ceiling - where `onComplete` never fires. `AbstractStreamListener` provides a no-op default, and `onComplete`/`onError` are mutually exclusive terminal events
```
chr-hertel: We can keep it shorter here, but add some docs to the actual interface method stubs please

UPGRADE.md
```diff
@@ -1,3 +1,37 @@
...
+   +
+   +    public function onError(ErrorEvent $event): void
+   +    {
+   +        // ...
+   +    }
+    }
+   ```
+
+   `onComplete()` and `onError()` are mutually exclusive: a drained stream reaches exactly one of
+   them, and a stream the consumer abandons reaches neither.
```
chr-hertel: can be removed

UPGRADE.md
```diff
@@ -1,3 +1,37 @@
...
+Platform
+--------
+
+ * `Result\Stream\ListenerInterface` gained an `onError()` method, dispatched with the new
+   `Result\Stream\ErrorEvent` when draining a `StreamResult` throws, right before the exception is
+   rethrown to the caller. This lets a listener finalize on a stream that reported its token usage on
+   a delta and then failed — e.g. `MaxOutputTokensException` on a response completed at the
+   output-token ceiling — where `onComplete()` never fires. Listeners extending
+   `AbstractStreamListener` inherit a no-op default and need no change; direct implementers of the
+   interface must add the method:
```
chr-hertel: let's keep it short
```suggestion
 * `Result\Stream\ListenerInterface` gained an `onError()` method, dispatched with the new
   `Result\Stream\ErrorEvent`. Listeners not extending
   `AbstractStreamListener` must add the method:
```

chr-hertel (changes requested):
Looks good already - some minor phrasing and on question I'd like your thoughts on - thanks already!

UPGRADE.md
ghisleouf: done

src/platform/CHANGELOG.md
ghisleouf: done

src/platform/src/Result/StreamResult.php
ghisleouf: Hi @chr-hertel ,

I would keep `\Throwable` because `Event\ResultErrorEvent` already carries one, and not everything thrown during a drain is ours.

A connection dropped mid-stream comes out of `EventSourceHttpClient` as a `TransportException`, and a listener's own `onDelta()` can throw anything, so narrowing would leave exactly those cases unnotified.

Anyway, happy to switch if needed!

ghisleouf:
Regarding the 2 red checks left:

- `Validate Changelog & Upgrade Entries` fails on the missing BC Break label, which I can't add myself. Could you please  set it?
- `Unit / PHP 8.2 / Symfony 7.4` failed in `src/chat/src/Bridge/MongoDb` which is untouched by this PR and identical to the previous push, which passed that same job.

Thanks for your help!

src/platform/src/Result/StreamResult.php
chr-hertel: yea - we're rethrowing anyways - thanks!

chr-hertel (approved):

chr-hertel:
Thank you @ghisleouf.

PR #2365, opened 2026-07-25
[Agent] Surface tool-argument resolution failures to the LLM

### LINKED ISSUE #2137: [Agent] Tool-argument resolution failures are not surfaced to the LLM in a correctable form
## Summary

When `ToolCallArgumentResolver::resolveArguments()` fails — because the LLM supplied an argument that cannot be denormalized (invalid backed-enum value, malformed object/discriminator structure, missing constructor argument) or omitted a mandatory parameter — the resulting exception is caught by `Toolbox::execute()` in its generic `\Throwable` branch and wrapped as `ToolExecutionException::executionFailed()`. Its `getToolCallResult()` returns only:

```
An error occurred while executing tool "X".
```

`FaultTolerantToolbox` relays exactly that string to the LLM. The actual cause (which argument was invalid, what the valid values are, or which parameter was missing) survives only in the exception *message* (i.e. in logs), never in the agent-facing result.

These failures are caused by the LLM's own tool-call input and are exactly the kind of feedback it could act on, but it receives none — so it cannot self-correct and typically repeats the same invalid call.

## Background — the failure flow

**1. The resolver throws.** A missing mandatory parameter throws `ToolException`:

https://github.com/symfony/ai/blob/v0.9.0/src/agent/src/Toolbox/ToolCallArgumentResolver.php#L78-L79

```php
if (!$reflectionParameter->isOptional()) {
    throw new ToolException(\sprintf('Parameter "%s" is mandatory for tool "%s".', $name, $toolCall->getName()));
}
```

An undeserializable value makes the serializer throw inside `denormalize()`:

https://github.com/symfony/ai/blob/v0.9.0/src/agent/src/Toolbox/ToolCallArgumentResolver.php#L104-L106

```php
if ($this->denormalizer->supportsDenormalization($value, $parameterType, 'json')) {
    $value = $this->denormalizer->denormalize($value, $parameterType, 'json');
}
```

(e.g. `NotNormalizableValueException` for an invalid backed-enum value, `MissingConstructorArgumentsException` for a malformed object argument).

**2. `Toolbox::execute()` wraps it generically.** `resolveArguments()` runs inside the try block; the catch arms are:

https://github.com/symfony/ai/blob/v0.9.0/src/agent/src/Toolbox/Toolbox.php#L114-L120

```php
} catch (ToolExecutionExceptionInterface $e) {
    throw $e;
} catch (\Throwable $e) {
    $this->logger->warning(\sprintf('Failed to execute tool "%s".', $toolCall->getName()), ['exception' => $e]);

    throw ToolExecutionException::executionFailed($toolCall, $e);
}
```

Neither failure implements `ToolExecutionExceptionInterface` — `ToolException` implements `ExceptionInterface` (not the execution interface), and the serializer exceptions are plain Symfony Serializer exceptions — so both bypass the first arm and hit the generic `\Throwable` arm.

**3. The cause is dropped from the agent-facing result.** `executionFailed()` keeps the cause in the exception *message*, but `getToolCallResult()` returns a generic string:

https://github.com/symfony/ai/blob/v0.9.0/src/agent/src/Toolbox/Exception/ToolExecutionException.php

```php
public static function executionFailed(ToolCall $toolCall, \Throwable $previous): self
{
    $exception = new self(\sprintf('Execution of tool "%s" failed with error: %s', $toolCall->getName(), $previous->getMessage()), previous: $previous);
    $exception->toolCall = $toolCall;

    return $exception;
}

public function getToolCallResult(): string
{
    return \sprintf('An error occurred while executing tool "%s".', $this->toolCall->getName());
}
```

**4. `FaultTolerantToolbox` relays only that generic string to the LLM:**

https://github.com/symfony/ai/blob/v0.9.0/src/agent/src/Toolbox/FaultTolerantToolbox.php#L36-L41

```php
public function execute(ToolCall $toolCall): ToolResult
{
    try {
        return $this->innerToolbox->execute($toolCall);
    } catch (ToolExecutionExceptionInterface $e) {
        return new ToolResult($toolCall, $e->getToolCallResult());
```

## Observed behavior

- **Invalid backed-enum value:** the LLM calls a tool with `{"color": "not_a_color"}` for a `Color $color` parameter. The serializer throws `The data must belong to a backed enumeration of type ...`. The LLM is told only: `An error occurred while executing tool "X".`
- **Missing mandatory parameter:** the LLM omits a required parameter. The resolver throws `Parameter "x" is mandatory for tool "Y".`. The LLM is told only: `An error occurred while executing tool "X".`

In both cases the underlying message names exactly what to fix, but the LLM never sees it.

## Minimal reproducer

The deserialization failure itself, showing the exception that reaches `Toolbox::execute()` is a plain serializer exception (not a `ToolExecutionExceptionInterface`), so it is wrapped into the generic result by the chain above:

```php
<?php

enum Color: string
{
    case Red = 'red';
    case Green = 'green';
    case Blue = 'blue';
}

final class SomeTool
{
    public function __invoke(Color $color): array
    {
        return [];
    }
}

$toolCall = new ToolCall('tc-1', 'some_tool', ['color' => 'not_a_color']);

// NotNormalizableValueException (a plain Serializer exception, not a ToolExecutionExceptionInterface)
//   -> Toolbox::execute() generic \Throwable arm -> ToolExecutionException::executionFailed()
//   -> FaultTolerantToolbox relays getToolCallResult(): 'An error occurred while executing tool "some_tool".'
```

## Note

Could argument-resolution failures be treated as a recoverable, agent-correctable category whose cause reaches the LLM — e.g. by surfacing the underlying `ToolException` / serializer message in the tool result for this class of error? The messages already name the missing parameter or the invalid value (for backed enums, the expected type), which is precisely the feedback the model needs to retry correctly. (If the generic `getToolCallResult()` is intentional to avoid leaking internal exception details to the model, note that these particular errors originate from the model's own input, not from internal tool state.)

## Environment

- `symfony/ai-agent`: **v0.9.0**
- `symfony/ai-bundle`: v0.9.0
- Symfony: 7.4
- PHP: 8.4

## Related

- #2011 — passing `'json'` to `denormalize()` (object-property float case).
- #2136 — top-level scalar `float` int→float widening (same `resolveArguments()` denormalize path).
- #2004 — nullable class-typed parameter denormalization (same method).

Berenger:
| Q            | A           |
| ------------ | ----------- |
| Bug fix?     | yes         |
| New feature? | no          |
| Docs?        | no          |
| Issues       | Fixes #2137 |
| License      | MIT         |

## Problem

When `ToolCallArgumentResolver::resolveArguments()` fails because of invalid LLM-generated tool-call input—such as a missing mandatory parameter or a value that cannot be denormalized (for example, an invalid backed-enum value)—the exception is caught by the generic `\Throwable` arm in `Toolbox::execute()`.

It is then wrapped in a `ToolExecutionException`, whose `getToolCallResult()` only returns:

> An error occurred while executing tool "X".

As a result, `FaultTolerantToolbox` relays no actionable feedback to the model. The model cannot self-correct and typically repeats the same invalid call.

## Solution

This PR treats argument-resolution failures as agent-correctable errors by reusing the existing `InvalidToolCallArgumentsException`.

This exception implements `ToolExecutionExceptionInterface`, allowing its message to reach the LLM as the tool result. This is the same mechanism already used by `ValidateToolCallArgumentsListener`.

- A missing mandatory parameter now throws `InvalidToolCallArgumentsException` with the existing message:

  ```text
  Parameter "number" is mandatory for tool "tool_required_params".
  ```

- A denormalization failure is caught and rethrown as `InvalidToolCallArgumentsException`:

  ```text
  Invalid value for parameter "mode" of tool "tool_with_backed_enums": The data must belong to a backed enumeration of type ...
  ```

  The original Serializer exception is preserved as the previous exception for debugging and logging purposes, but is not exposed directly to the LLM.

These messages are derived from the model's input and the tool's argument metadata. Exceptions raised during tool execution remain hidden and continue to produce the generic error message.

## Compatibility

Code catching the component's `ExceptionInterface` remains unaffected, since both the old and new exceptions implement it.

Consumers catching the concrete `ToolExecutionException` may need to handle `InvalidToolCallArgumentsException` as well.

## Technical Note

`Toolbox::execute()` now initializes `$arguments` before entering the `try` block, ensuring that it is safely available to the catch arms, as reported by PHPStan.

## Tests

This behavior is covered by:

- Unit tests for the argument resolver.
- End-to-end tests through `Toolbox` and `FaultTolerantToolbox`.
- Assertions on the exact error message received by the LLM.

### DIFF
--- a/src/agent/src/Toolbox/ToolCallArgumentResolver.php
+++ b/src/agent/src/Toolbox/ToolCallArgumentResolver.php
@@ -11,12 +11,13 @@
 
 namespace Symfony\AI\Agent\Toolbox;
 
-use Symfony\AI\Agent\Toolbox\Exception\ToolException;
+use Symfony\AI\Agent\Toolbox\Exception\InvalidToolCallArgumentsException;
 use Symfony\AI\Platform\Result\ToolCall;
 use Symfony\AI\Platform\Tool\Tool;
 use Symfony\Component\PropertyInfo\Extractor\PhpDocExtractor;
 use Symfony\Component\PropertyInfo\Extractor\ReflectionExtractor;
 use Symfony\Component\PropertyInfo\PropertyInfoExtractor;
+use Symfony\Component\Serializer\Exception\ExceptionInterface as SerializerExceptionInterface;
 use Symfony\Component\Serializer\Mapping\ClassDiscriminatorFromClassMetadata;
 use Symfony\Component\Serializer\Mapping\Factory\ClassMetadataFactory;
 use Symfony\Component\Serializer\Mapping\Loader\AttributeLoader;
@@ -63,7 +64,7 @@ final class ToolCallArgumentResolver implements ToolCallArgumentResolverInterfac
     /**
      * @return array<string, mixed>
      *
-     * @throws ToolException When a mandatory tool parameter is missing from the tool call arguments
+     * @throws InvalidToolCallArgumentsException When a mandatory tool parameter is missing from the tool call arguments or an argument cannot be denormalized to the expected parameter type
      */
     public function resolveArguments(Tool $metadata, ToolCall $toolCall): array
     {
@@ -76,7 +77,7 @@ final class ToolCallArgumentResolver implements ToolCallArgumentResolverInterfac
         foreach ($parameters as $name => $reflectionParameter) {
             if (!\array_key_exists($name, $toolCall->getArguments())) {
                 if (!$reflectionParameter->isOptional()) {
-                    throw new ToolException(\sprintf('Parameter "%s" is mandatory for tool "%s".', $name, $toolCall->getName()));
+                    throw new InvalidToolCallArgumentsException(\sprintf('Parameter "%s" is mandatory for tool "%s".', $name, $toolCall->getName()));
                 }
                 continue;
             }
@@ -102,7 +103,11 @@ final class ToolCallArgumentResolver implements ToolCallArgumentResolverInterfac
             $parameterType .= $dimensions;
 
             if ($this->denormalizer->supportsDenormalization($value, $parameterType, 'json')) {
-                $value = $this->denormalizer->denormalize($value, $parameterType, 'json');
+                try {
+                    $value = $this->denormalizer->denormalize($value, $parameterType, 'json');
+                } catch (SerializerExceptionInterface $e) {
+                    throw new InvalidToolCallArgumentsException(\sprintf('Invalid value for parameter "%s" of tool "%s": %s', $name, $toolCall->getName(), $e->getMessage()), previous: $e);
+                }
             }
 
             $arguments[$name] = $value;
--- a/src/agent/src/Toolbox/ToolCallArgumentResolver.php
+++ b/src/agent/src/Toolbox/ToolCallArgumentResolver.php
@@ -106,7 +106,7 @@ final class ToolCallArgumentResolver implements ToolCallArgumentResolverInterfac
                 try {
                     $value = $this->denormalizer->denormalize($value, $parameterType, 'json');
                 } catch (SerializerExceptionInterface $e) {
-                    throw new InvalidToolCallArgumentsException(\sprintf('Invalid value for parameter "%s" of tool "%s": %s', $name, $toolCall->getName(), $e->getMessage()), previous: $e);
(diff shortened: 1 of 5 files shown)


chr-hertel (approved):
Sorry for the delay @Berenger - somehow this slipped. I took care about the update and an UPGRADE.md note.

chr-hertel:
Thank you @Berenger.

PR #2368, opened 2026-07-28
[Platform] Make base URL configurable in OpenRouter ModelApiCatalog

jvelten:
Make the base URL in `OpenRouter\ModelApiCatalog` configurable via a new `$baseUrl` constructor parameter (defaults to `https://openrouter.ai/api`).

This allows using OpenRouter-compatible APIs hosted on a different domain, e.g. self-hosted proxies or alternative endpoints.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | 
| License       | MIT

### DIFF
--- a/src/platform/src/Bridge/OpenRouter/ModelApiCatalog.php
+++ b/src/platform/src/Bridge/OpenRouter/ModelApiCatalog.php
@@ -25,9 +25,14 @@ final class ModelApiCatalog extends AbstractOpenRouterModelCatalog
 {
     protected bool $modelsAreLoaded = false;
 
+    private readonly string $baseUrl;
+
     public function __construct(
         private readonly HttpClientInterface $httpClient,
+        string $baseUrl = 'https://openrouter.ai/api',
     ) {
+        $this->baseUrl = rtrim($baseUrl, '/');
+
         parent::__construct();
     }
 
@@ -63,7 +68,7 @@ final class ModelApiCatalog extends AbstractOpenRouterModelCatalog
      */
     protected function fetchRemoteModels(): iterable
     {
-        $responseModels = $this->httpClient->request('GET', 'https://openrouter.ai/api/v1/models');
+        $responseModels = $this->httpClient->request('GET', $this->baseUrl.'/v1/models');
         foreach ($responseModels->toArray()['data'] as $model) {
             $capabilities = [];
 
@@ -128,7 +133,7 @@ final class ModelApiCatalog extends AbstractOpenRouterModelCatalog
      */
     protected function fetchRemoteEmbeddings(): iterable
     {
-        $responseEmbeddings = $this->httpClient->request('GET', 'https://openrouter.ai/api/v1/embeddings/models');
+        $responseEmbeddings = $this->httpClient->request('GET', $this->baseUrl.'/v1/embeddings/models');
         foreach ($responseEmbeddings->toArray()['data'] as $embedding) {
             yield $embedding['id'] => [
                 'class' => EmbeddingsModel::class,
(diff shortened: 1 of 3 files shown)


src/platform/src/Bridge/OpenRouter/ModelApiCatalog.php
```diff
@@ -27,6 +27,7 @@ final class ModelApiCatalog extends AbstractOpenRouterModelCatalog
 
     public function __construct(
         private readonly HttpClientInterface $httpClient,
+        private readonly string $baseUrl = 'https://openrouter.ai/api',
     ) {
         parent::__construct();
     }
```
wachterjohannes, line 37: Could you normalize a trailing slash here? The very same `$baseUrl` value is also handed to `Generic\Completions\ModelClient`, which does:

```php
/** @param string $baseUrl Base URL of an OpenAI-compatible endpoint, with or without a trailing slash */
$this->baseUrl = rtrim($baseUrl, '/');
```

So `https://gateway.internal/api/` currently works for chat completions but produces `https://gateway.internal/api//v1/models` for the catalog. Most gateways tolerate that, some don't — and since both consumers read the same configured value, they should treat it the same way.

```suggestion
    private readonly string $baseUrl;

    public function __construct(
        private readonly HttpClientInterface $httpClient,
        string $baseUrl = 'https://openrouter.ai/api',
    ) {
        $this->baseUrl = rtrim($baseUrl, '/');

        parent::__construct();
    }
```

src/platform/src/Bridge/OpenRouter/ModelApiCatalog.php
```diff
@@ -63,7 +64,7 @@ protected function preloadRemoteModels(): void
      */
     protected function fetchRemoteModels(): iterable
     {
-        $responseModels = $this->httpClient->request('GET', 'https://openrouter.ai/api/v1/models');
+        $responseModels = $this->httpClient->request('GET', $this->baseUrl.'/v1/models');
```
wachterjohannes, line 71: Could you add a test covering this? `ModelApiCatalogTest` already builds a `MockHttpClient` with `JsonMockResponse` and asserts `getRequestsCount()`, so asserting the requested URL is a short addition — something like:

```php
public function testItUsesTheConfiguredBaseUrl()
{
    $httpClient = new MockHttpClient([
        new JsonMockResponse(['data' => [['id' => 'anthropic/claude-3-opus', 'architecture' => ['input_modalities' => ['text'], 'output_modalities' => ['text']]]]]),
        new JsonMockResponse(['data' => []]),
    ]);

    $catalog = new ModelApiCatalog($httpClient, 'https://gateway.internal/api/');
    $catalog->getModels();

    $this->assertSame(2, $httpClient->getRequestsCount());
    // and assert the recorded URLs are https://gateway.internal/api/v1/models
    // and https://gateway.internal/api/v1/embeddings/models
}
```

That pins down both the new parameter and the trailing-slash handling from my other comment.

wachterjohannes (commented):
Thanks @jvelten! Good catch — `Factory` already gained a `$baseUrl` in 0.11, but `ModelApiCatalog` still hardcoded `openrouter.ai`, so pointing the factory at a self-hosted gateway silently split traffic across two hosts. Default and shape match the existing precedent in `AmazeeAi\ModelApiCatalog`.

Two small things inline, then this is good to go from my side.

wachterjohannes (approved):
Both points addressed, thanks! Verified locally against a fake OpenRouter-compatible gateway: `http://…/api` and `http://…/api/` both produce clean `/v1/models` and `/v1/embeddings/models` paths, and a live call against the real API still resolves 449 models. 👍

wachterjohannes:
Thank you @jvelten.

PR #2370, opened 2026-07-28
[Mate] Add skill lifecycle: discovery, copy install and state in extensions.php

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | -
| License       | MIT

First of three PRs building the Mate skill lifecycle. Follow-Ups will add `skills:override`/`skills:reset` and `skills:enable`/`skills:disable`.

Skills are the workflow/knowledge layer above tools: inert `SKILL.md` procedures the agent reads, managed — never executed — by the CLI. The `extra.ai-mate.skills` config key already exists as a list of directories, so extensions that ship skills need no change; this PR builds the metadata, install and state layer on top of it.

### Install as a copy

A skill is installed as a real copy under `.agents/skills/mate-<name>/`, with its frontmatter `name` rewritten to the installed name so it cannot collide with a skill the user maintains.

Copies rather than symlinks into `vendor/` are the point: what the agent loads is a file you can open and diff, and a package update cannot change it underneath you. Only the `.claude/skills/` mirror is a relative symlink to that copy, so the two can never drift apart — falling back to a second copy where `symlink()` is unavailable (Windows without developer mode), which is isolated behind a `LinkerInterface` so the fallback is reachable in tests.

### One state file

Everything lives in `mate/extensions.php`, next to the intent it derives from:

```php
'vendor/package' => [
    'enabled' => true,
    'skills' => [
        'demo-skill' => [
            'enabled' => true,          // yours to edit
            'mode'    => 'managed',     // yours to edit: managed|override
            'state'   => 'managed',     // written by mate below this point
            'source'  => 'vendor/vendor/package/skills/demo-skill',
            'source_hash' => 'sha256:...',
            'hash'        => 'sha256:...',
            'targets' => [
                '.agents/skills/mate-demo-skill',
                '.claude/skills/mate-demo-skill',
            ],
        ],
    ],
],
```

`SkillStateRepository` is the sole reader and writer of that path. `ContainerFactory` includes it independently and treats a non-array result as *every extension disabled*, so writes go through a temp file + `rename()`, and reads are served from cache once written in the same process (a mutating command does write → install → read, and opcache can otherwise hand back the pre-write file).

Two hashes are recorded because the copy is never byte-identical to its source — the rewritten `name` line: `hash` detects hand-edited output, `source_hash` detects an upstream change that has not been installed yet.

### Commands

* `skills:install` — idempotent reconciler. Rebuilds from source or from the user's `mate/skills/` copy, prunes skills whose source or extension vanished, and never writes into `mate/skills/`. `discover` runs it automatically.
* `skills:list` — read-only overview: enabled, mode, state, status (`ok`, `disabled`, `not installed`, `stale`, `broken`). Table, JSON or TOON.
* `skills:validate [name] [--strict]` — checks the generated folders against the record. Errors on hand-edited content, missing folders, a mispointed mirror, an overridden skill with no copy, and a disabled skill that still has folders; warns on a moved-on source, a never-installed skill, a copied mirror and strays. Exits non-zero on errors, or on warnings with `--strict`.
* `skills:prune [--dry-run]` — removes leftover generated folders. Only `mate-` prefixed entries are ever touched.

### Deliberately not git-ignored

The generated folders are **not** added to `.gitignore`. Because they are plain copies, committing them turns an upstream skill change into a reviewable diff instead of something that lands silently. (An earlier iteration managed a `.gitignore` block that ignored `/.agents/skills/` and `/.claude/skills/` wholesale — that would also have hidden skills the user maintains, including the five tracked `.claude/skills/*` files in this monorepo.)

### Notes for review

This is a BC break for anyone who configured skills under `0.12` — the `UPGRADE.md` entry covers it, and the PR needs the **BC Break** label (I cannot add it myself).

It also touches the `0.13` section of `src/mate/CHANGELOG.md`, which #2132 touches as well; whichever lands second needs a trivial rebase. Worth noting that #2132's skills use `description: >-` block scalars, which the front matter parser in this PR handles — on `0.12` they parse to the literal `">-"`.


kbond:
I've added the BC Break label.

~To clarify, symlinking to the vendor dir is no longer an option, they're always copied now?~ I just read the UPGRADE. I think this is the right move - now it's closer to a Flex recipe and `mate/extensions.php` is like `symfony.lock`

docs/components/mate.rst
```diff
@@ -168,14 +168,17 @@ Extensions Configuration
...
-    // This file is managed by 'mate discover'
-    // You can manually edit to enable/disable extensions
+    // Managed by "mate discover" and the "mate skills:*" commands, which can change
+    // every setting in here — prefer them over editing by hand. If you do edit: "enabled"
+    // and "mode" are kept, every other key is rewritten on the next install.
 
     return [
         'vendor/package-name' => ['enabled' => true],
         'vendor/another-package' => ['enabled' => false],
     ];
```
chr-hertel: Maybe rather promote the commands for configuration instead of the file?

src/mate/resources/mate/extensions.php
```diff
@@ -1,7 +1,8 @@
 <?php
 
-// This file is managed by 'mate discover'
-// You can manually edit to enable/disable extensions
+// Managed by "mate discover" and the "mate skills:*" commands, which can change
+// every setting in here — prefer them over editing by hand. If you do edit: "enabled"
+// and "mode" are kept, every other key is rewritten on the next install.
```
chr-hertel: I think it is valuable to be very explicit here to avoid frustration:

```suggestion
// This file is managed by Mate - use `discover` or `skills:*` commands
// over manual editing. Only changes to `mode` or `enabled` are kept,
// every other key overwritten by Mate.
```

src/mate/src/Command/SkillsPruneCommand.php
```diff
@@ -0,0 +1,80 @@
...
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+
+/**
+ * Removes generated mate-* folders that no longer belong to any skill.
+ *
+ * "skills:install" already prunes what it knows about; this command exists for the leftovers it
+ * cannot see — a folder left behind by an interrupted run, or by hand-editing mate/extensions.php.
+ * Only "mate-" prefixed entries are ever touched, so skills you maintain yourself are left alone.
```
chr-hertel, line 27: feels a bit edge-casey - would be also fine to add that in a later iteration if the issue comes up

src/mate/src/Skill/SkillManager.php
```diff
@@ -0,0 +1,275 @@
...
+     * @return list<SkillIssue>
+     */
+    private function collectMirrorIssues(string $installedName, string $claudeTarget): array
+    {
+        if (is_link($claudeTarget)) {
+            $expected = '../../'.SkillInstaller::AGENTS_SKILLS_DIR.'/'.$installedName;
+            $actual = readlink($claudeTarget);
+            if ($actual !== $expected && realpath($claudeTarget) !== realpath(\dirname($claudeTarget).'/'.$expected)) {
+                return [['level' => 'error', 'message' => \sprintf('Mirror ".claude/skills/%s" points at "%s" instead of the .agents copy.', $installedName, false === $actual ? '?' : $actual)]];
+            }
```
chr-hertel: this wouldn't be fixed by `SkillInstaller::isUpToDate` that only checks for existence.

However, I really wonder if this feature depth is needed from the start - at least i would think it's fine to keep the footprint smaller in the beginning. your call tho :)

chr-hertel (approved):
Looks pretty good to me 👍

few things to consider:
* would filesystem component remove a bit of load?
* is the feature set in this depth needed already
* after all PRs land we should take a look at the docs, to make sure Mate is easy to understand - maybe even waiting on the MCP removal

Feel free to go ahead please 🙏

wachterjohannes:
@chr-hertel Thanks, all pushed.

Filesystem component: yes. SkillInstaller now uses mirror(), remove(), mkdir() and dumpFile(), about 90 lines lighter. One caveat worth knowing: mirror() recreates symlinks it finds inside a skill folder, so it gets an iterator that filters those out.

Mispointed mirror: good catch. isUpToDate() accepted any resolvable symlink, so skills:install never repaired what skills:validate kept reporting. There is now a shared mirrorState() that both use, plus two regression tests.

Feature depth / skills:prune: kept for now, because prune is the only remedy for the stray folder warning skills:validate raises, so dropping the command means dropping that check too. And there will be more commands to manage the state of the skills in the nearer future

Docs: agreed, one pass after the remaining PRs and the MCP removal beats tuning them per PR. This one only flips the config section to lead with the commands instead of the file.

Also rebased onto main, the UPGRADE.md conflict is gone.

chr-hertel:
Thank you @wachterjohannes.

PR #2373, opened 2026-07-29
[Agent] Drive tool calling from the Agent instead of an output processor

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | -
| License       | MIT

**My summary**

* Tool calling is _the_ essential feature of an agent - no reason to bring it in via "Processor", but make it a first level citizen
* Part of larger AgentInterface rebuild - incl. getting rid of Processors
* Review by commit => heart of rework quite small in src/agent/src/
* 95% of changes are examples, tests and docs

could be another foundation for #1829 by @DZunke 
cc @ineersa what do you think?
cc @fabpot is that [ToolExecutorInterface](https://github.com/chr-hertel/ai/blob/5379367e4db22c1507dd191fba974170f6a2d184/src/agent/src/Toolbox/ToolExecutorInterface.php) interesting for your use-case as well or still too limited?

**Claude's Summary**

The tool-calling loop lives in `Toolbox\AgentProcessor`, an *output* processor, which blocks replacing the processor pipeline. This moves the loop into an internal `Execution\Runner` owned by the `Agent`, and puts execution of the requested tool calls behind a new `Toolbox\ToolExecutorInterface` (default `SequentialToolExecutor`).

The executor's contract stays deliberately small — run the calls, return the `ToolResult` list in order:

```php
public function execute(array $toolCalls): array; // ToolCall[] -> ToolResult[]
```

Turning results into conversation messages stays in the `Runner`, so execution strategies (sequential, parallel, remote, …) don't each re-implement the message/protocol format.

Configuration moves from the processor to the `Agent`:

```diff
-$processor = new AgentProcessor($toolbox, includeSources: true);
-$agent = new Agent($platform, $model, [$processor], [$processor]);
+$agent = new Agent($platform, $model, toolbox: $toolbox, includeSources: true);
```

Behaviour is preserved: source collection, `excludeToolMessages`, the `ToolCallsExecuted` short-circuit, `maxToolCalls` and streamed tool calls. `AgentProcessorTest` is ported to `Execution\RunnerTest`.

**BC break** (`AgentProcessor` removed) — documented in `UPGRADE.md`; please add the `BC Break` label.

Verified live against OpenAI: `toolbox/clock.php`, `openai/toolcall-stream.php`, `openai/agent-stream-sources.php`.


mahono:
Is there still a way to inject logic before and after calling the LLM?

Because this is needed to integrate proper PII protection and also prompt injection detection.

update:
I just had a look. It is. All fine for me. 🙂

chr-hertel:
@mahono switching from InputProcessor to EventListener on the Platform?

ineersa:
Looks good to me!
Moving loop to agent is good, and `ToolExecutorInterface` is good as extension point.
I use [similar one](https://github.com/ineersa/agent-core/blob/main/src/AgentCore/Contract/Tool/ToolExecutorInterface.php) but for a single tool call.

Batching support in interface is useful, is a good fit for synchronous/in-process strategies, also can implement something with fibers.
For my case I'm using messenger for now for parallel execution, so I'm handling batching on other level.

chr-hertel:
Tested in demo and examples - looks good :+1:

OskarStark:
Love it! Thanks

PR #2374, opened 2026-07-30
[Platform][OpenResponses] Replay reasoning items across turns of store: false conversations

fabpot:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Docs?         | no 
| Issues        | n/a
| License       | MIT

When using OpenResponses with `store: false`, the provider does not keep the reasoning state. But it also means that encrypted reasoning items must be sent back with the next request, especially after a tool call.

We currently discard those items during result conversion. The next request then loses the context the model just produced. This PR keeps them as thinking signatures so Symfony AI can replay them from the conversation history.

### DIFF
--- a/src/platform/src/Bridge/OpenResponses/Contract/Message/AssistantMessageNormalizer.php
+++ b/src/platform/src/Bridge/OpenResponses/Contract/Message/AssistantMessageNormalizer.php
@@ -15,6 +15,7 @@ use Symfony\AI\Platform\Bridge\OpenResponses\ResponsesModel;
 use Symfony\AI\Platform\Contract\Normalizer\ModelContractNormalizer;
 use Symfony\AI\Platform\Message\AssistantMessage;
 use Symfony\AI\Platform\Message\Content\Text;
+use Symfony\AI\Platform\Message\Content\Thinking;
 use Symfony\AI\Platform\Model;
 use Symfony\Component\Serializer\Normalizer\NormalizerAwareInterface;
 use Symfony\Component\Serializer\Normalizer\NormalizerAwareTrait;
@@ -29,16 +30,17 @@ final class AssistantMessageNormalizer extends ModelContractNormalizer implement
     /**
      * @param AssistantMessage $data
      *
-     * @return array{
-     *     role: 'assistant',
-     *     type: 'message',
-     *     content: ?string
-     * }
+     * @return list<array<string, mixed>>
      */
     public function normalize(mixed $data, ?string $format = null, array $context = []): array
     {
+        $reasoningItems = $this->extractReasoningItems($data);
+
         if ($data->hasToolCalls()) {
-            return $this->normalizer->normalize($data->getToolCalls(), $format, $context);
+            /** @var list<array<string, mixed>> $toolCalls */
+            $toolCalls = $this->normalizer->normalize($data->getToolCalls(), $format, $context);
+
+            return array_merge($reasoningItems, $toolCalls);
         }
 
         $text = '';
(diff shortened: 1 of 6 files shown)


src/platform/src/Bridge/OpenResponses/Contract/Message/AssistantMessageNormalizer.php
```diff
@@ -15,12 +15,18 @@
...
 use Symfony\Component\Serializer\Normalizer\NormalizerAwareInterface;
 use Symfony\Component\Serializer\Normalizer\NormalizerAwareTrait;
 
 /**
  * @author Guillermo Lengemann <guillermo.lengemann@gmail.com>
+ *
+ * @phpstan-type MessageItem array{role: 'assistant', type: 'message', content: ?string}
+ * @phpstan-type FunctionCallItem array{arguments: string, call_id: string, name: string, type: 'function_call'}
+ * @phpstan-type ReasoningItem array{type: 'reasoning', id: string, summary: list<array{type: string, text?: string}>, encrypted_content?: string|null}
+ * @phpstan-type InputItem MessageItem|FunctionCallItem|ReasoningItem
```
fabpot: I don't like this very much :(

chr-hertel: aren't array shapes on normalizers rather pointless? there's no caller caring for that

src/platform/src/Bridge/OpenResponses/Contract/Message/AssistantMessageNormalizer.php
```diff
@@ -64,4 +74,37 @@ protected function supportsModel(Model $model): bool
...
+        foreach ($data->getContent() as $part) {
+            if (!$part instanceof Thinking || null === $part->getSignature()) {
+                continue;
+            }
+
+            try {
+                $item = json_decode($part->getSignature(), true, flags: \JSON_THROW_ON_ERROR);
+            } catch (\JsonException) {
+                continue;
+            }
```
chr-hertel, line 90: that's safe to continue and swallow? 🤔

src/platform/src/Bridge/OpenResponses/Contract/Message/AssistantMessageNormalizer.php
fabpot: Gone

src/platform/src/Bridge/OpenResponses/Contract/Message/AssistantMessageNormalizer.php
fabpot: Signatures generated by OpenResponses are valid JSON. Non-JSON signatures can come from other providers and must be ignored.

fabpot: Maybe I should explain more what I'm doing for more context:

First, and to be clear, in a normal OpenResponses-only conversation it should never happen.

But I'm sometimes reusing a conversation history produced by another provider. I start a convo with Claude (opaque signature) and switch to GPT later on. If we don't want to support this use case (which is fine), we should remove the `catch`.

chr-hertel: interesting :thinking:

i think that's a valid use case, but might bring some more issues - never had multi-turn across multiple providers on my list for testing, but ideally the data layer that we keep in user land should be more or less agnostic as well :thinking:

chr-hertel (approved):

chr-hertel:
Thank you @fabpot.

src/platform/src/Bridge/OpenResponses/Contract/Message/AssistantMessageNormalizer.php
fabpot: Unfortunately, this use case doesn’t work well anymore with frontier models as they hide more and more.

PR #2375, opened 2026-07-31
[MCP Bundle] Fix FrameworkSessionStore destroy without open()

omouren:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| License       | MIT

`FrameworkSessionStore::destroy()` calls `SessionHandlerInterface::destroy()` without calling `open()` first.

With Symfony's `AbstractSessionHandler` (e.g. `RedisSessionHandler`), it throws on HTTP DELETE /_mcp:
```
LogicException: Session name cannot be empty, did you forget to call "parent::open()" in "Symfony\Component\HttpFoundation\Session\Storage\Handler\RedisSessionHandler"?
  at AbstractSessionHandler.php:101
     Symfony\Component\HttpFoundation\Session\Storage\Handler\AbstractSessionHandler::destroy()
  at FrameworkSessionStore.php:64
     Symfony\AI\McpBundle\Session\FrameworkSessionStore::destroy()
  at SessionManager.php:61
     Mcp\Server\Session\SessionManager::destroy()
  at Protocol.php:671
     Mcp\Server\Protocol::destroySession()
  at BaseTransport.php:161
     Mcp\Server\Transport\BaseTransport::handleSessionEnd()
  at StreamableHttpTransport.php:154
     Mcp\Server\Transport\StreamableHttpTransport::handleDeleteRequest()
  at StreamableHttpTransport.php:365
     Mcp\Server\Transport\StreamableHttpTransport::handleRequest()
     'DELETE' => $this->handleDeleteRequest()
  at McpController.php:48
     Symfony\AI\McpBundle\Controller\McpController::handle()
```

This breaks MCP Streamable HTTP session termination (`DELETE` on the MCP endpoint) when `http.session.store` is `framework`.

### Fix
Call `$this->handler->open()` before `destroy()`, as required by the `SessionHandlerInterface` lifecycle.

### DIFF
--- a/src/mcp-bundle/src/Session/FrameworkSessionStore.php
+++ b/src/mcp-bundle/src/Session/FrameworkSessionStore.php
@@ -61,6 +61,8 @@ final class FrameworkSessionStore implements SessionStoreInterface
 
     public function destroy(Uuid $id): bool
     {
+        $this->handler->open('', 'mcp');
+
         return $this->handler->destroy($this->getKey($id));
     }
 


src/mcp-bundle/src/Session/FrameworkSessionStore.php
```diff
@@ -61,6 +61,8 @@ public function write(Uuid $id, string $data): bool
 
     public function destroy(Uuid $id): bool
     {
+        $this->handler->open('', 'mcp');
```
chr-hertel, line 64: can you share your thoughts about the `''` and `mcp` values you chose? is it rather random or more considerations to it?

chr-hertel:
Thanks for raising this @omouren - appreciated!

src/mcp-bundle/src/Session/FrameworkSessionStore.php
omouren: It's a little bit arbitrary, we mainly need to call `open()` so `AbstractSessionHandler::destroy()` doesn't throw, and both parameters are required by the interface.

The first argument ('') should not be used: handlers like `RedisSessionHandler` ignore the save path.
It's fine for handlers that ignore it (like Redis). For file based handlers the path is typically already configured via `session.save_path` or the handler constructor. Maybe we could pass `ini_get('session.save_path') ?: ''` instead if you prefer ?

The session name 'mcp' is arbitrary, but chosen to stay aligned with the store's default `mcp-` key prefix.

chr-hertel: alright, fair enough - was only wondering - but doesn't look problematic to me :+1:

chr-hertel (approved):

chr-hertel:
Thank you @omouren.

PR #2376, opened 2026-07-31
[AI Bundle] Show configured components and their calls in the profiler panel

### LINKED ISSUE #525: [AI Bundle] Take Component Slicing into account in Profiler Toolbar/Panel
Currently the profiler panel only shows Platform and Tool calls, but does not really carry information of other high-level features like configured instances and calls of `Indexer`, `Agent` or `Store`.

sanya-misharin:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | Fix #525
| License       | MIT

The profiler panel previously showed only platform calls, tools and tool calls, with the "Configured Platforms" metric hardcoded to `1`, while the collector already gathered agent, chat, message store and store calls without rendering them.

As proposed in [the issue discussion](https://github.com/symfony/ai/issues/525#issuecomment-5084631012), this PR surfaces everything that is already collected - scoped to `src/ai-bundle` only:

- panel sections for Agent Calls, Chat Calls, Message Store Calls and Store Calls (reusing a shared `message_bag` macro extracted from the platform calls section);
- real per-component metrics (configured instances + calls) instead of the hardcoded platform count, and a "Configured Components" table listing each component with the service ids it was registered under (via indexed tagged iterators);
- the toolbar now reflects all component slices and counts every collected call.

A follow-up PR will add `TraceableIndexer` to `symfony/ai-store` and wire it into the collector.

Screenshots (before / after / narrow viewport):

ai-panel-before-top
ai-panel-after-top
ai-panel-after-configured
ai-panel-narrow

### DIFF
--- a/docs/bundles/ai-bundle.rst
+++ b/docs/bundles/ai-bundle.rst
@@ -1247,7 +1247,8 @@ The profiler panel provides insights into the agent's execution:
 
 Next to the platform calls, the registered tools and the tool calls, the panel lists the calls
 collected for agents, chats, message stores and stores, as well as every configured component
-with the service id it was registered under.
+with the service id it was registered under. Components without a configured instance are
+omitted from both the toolbar and the panel.
 
 Testing agents
 ~~~~~~~~~~~~~~
--- a/docs/bundles/ai-bundle.rst
+++ b/docs/bundles/ai-bundle.rst
@@ -1245,6 +1245,10 @@ The profiler panel provides insights into the agent's execution:
 .. image:: images/profiler-ai.png
    :alt: Profiler Panel
 
+Next to the platform calls, the registered tools and the tool calls, the panel lists the calls
+collected for agents, chats, message stores and stores, as well as every configured component
+with the service id it was registered under.
+
 Testing agents
 ~~~~~~~~~~~~~~
 
--- a/src/ai-bundle/CHANGELOG.md
+++ b/src/ai-bundle/CHANGELOG.md
@@ -1,6 +1,12 @@
 CHANGELOG
 =========
 
+0.13
+----
+
+ * Show the calls collected for agents, chats, message stores and stores in the profiler panel
+ * Show the configured platforms, toolboxes, agents, chats, message stores and stores with their service ids in the profiler panel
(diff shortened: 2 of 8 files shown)


chr-hertel (approved):
Valid next step, but untested from my end for chat & message store component - thanks @sanya-misharin!

chr-hertel (changes requested):
Just noticed while testing that this list is quite long now even if no message stores/chats are configured:

Image

let's skip components that are not configured please.

same here:
Image

Image

sanya-misharin:
Thanks for testing, @chr-hertel! Fixed in f269f63.

Every slice is now gated on the component actually having a configured instance, in all the spots you pointed at, plus the "Configured Components" table that had the same zero rows:

* **toolbar**: a `Configured X` / `X Calls` pair is only rendered when that component has a configured instance;
* **metrics**: the component list is built once and filtered, so the metric groups and the dividers between them only cover configured components;
* **panel sections**: `Platform Calls`, `Tools` + `Tool Calls`, `Agent Calls`, `Chat Calls`, `Message Store Calls` and `Store Calls` are skipped entirely for unconfigured components. The inner "No ... calls were made." empty states stay, so a configured but unused component still tells you it did nothing during this request;
* if nothing at all is configured, the panel shows a single "No AI components are configured." note instead of an empty metrics bar.

The docs paragraph about the panel mentions this as well now.

All screenshots below come from one and the same profiler token, rendered once with the previous template and once with the new one. The demo app configures no chat and no message store, so it reproduces your setup.

**Toolbar** (before / after):

before-toolbar

after-toolbar

**Panel metrics** (before / after):

before-panel-top

after-panel-top

**Empty sections** (before / after):

before-configured

after-configured

**Configured Components table** (before / after):

before-configured-table

after-configured-table

wachterjohannes (approved):
Confirms unconfigured components are correctly skipped in both the toolbar and the panel, verified against a real compiled container. Rebased onto current main and moved the changelog entry into the correct unreleased 0.14 section. Verified locally, all green.

PR #2380, opened 2026-08-03
[Mate] Replace the MCP server with a native CLI, and make it discoverable

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | -
| License       | MIT

> **On the measurements below.** They come from a controlled comparison of three builds
> against the same seeded fixture, and they are what this PR is argued from rather than
> intuition. They are not a benchmark: one task family, one model per run, and the sample
> sizes are small enough that the individual percentages should be read as a direction, not
> as a figure to quote. What they do support is the claim the PR rests on, that the CLI on
> its own is not used and only becomes effective together with the discovery layer.

Mate currently ships as an MCP server. This PR turns it into a plain CLI that coding
agents invoke directly, and adds the discovery layer that makes those commands findable.

**The two parts belong together.** Shipping the CLI alone is a regression: in a controlled
measurement (30 runs, Haiku 4.5, an N+1 on a Symfony page) the bare CLI was invoked
**0 times out of 30**, installed, working, and invisible. Its success rate was
statistically indistinguishable from not having Mate at all (15/30 vs 12/30, p = 0.60).
With skills and the materialised agent instructions on top, the same CLI reaches 34/45
against a baseline of 12/30 (p = 0.0034).

### What changes

* `mate serve` / `mate stop` and the whole MCP server runtime are gone, along with the
  `mcp/sdk` dependency, `mcp.json`/`.mcp.json` generation and the Codex wrappers.
* Tool and resource commands are renamed (`mcp:tools:list` → `tools:list`, …) and
  `tools:call` takes long options instead of a positional JSON blob.
* Custom tools, resources and resource templates use native Mate attributes:
  `#[MateTool]`, `#[MateResource]` and `#[MateResourceTemplate]` in
  `Symfony\AI\Mate\Attribute`. The `Mate` prefix keeps them clear of the Agent
  component's `AsTool`.
* Prompts are dropped. There was no native equivalent of `#[McpPrompt]` left, so the
  always-empty `--type=prompt` surface goes with it.
* `mate init` and `mate discover` write CLI-oriented agent instructions and keep a managed
  block in `AGENTS.md`, plus a `CLAUDE.md` that imports it.
* Five new skills ship with the component (profiler debugging, log investigation, request
  triage, service inspection, PHP environment check), joining the existing
  `system-information`, installed via `mate skills:install`.
* `mate init` records how the agent must invoke Mate and which PHP it expects, and Mate
  refuses to run under a different one. See below.
* The documentation is rewritten for the CLI: `docs/components/mate.rst` and its three
  subpages plus the component line in `docs/index.rst`.

### The Codex wrapper is not lost, it becomes unnecessary

`bin/codex` existed because Codex does not read a project-local MCP configuration. Without
an MCP server there is nothing to configure: the agent runs `vendor/bin/mate …` like any
other command. The wrapper is removed because the problem it worked around no longer
exists.

### Why the transport does not decide the outcome

The same measurement compared the MCP server against the CLI-plus-skills build:
20/30 vs 34/45, **p = 0.44**. There is no measurable effectiveness difference. The case for the CLI is therefore not "it works better": it is client independence, no second process,
and no permanent tool descriptions in the context window. The measurement's role here is to
show that this costs nothing in results.

### Which PHP interpreter, and where

`resources/mcp.json` carried `"command": "##PHP_BINARY##"`, filled in by `mate init`. That
one line answered a question the CLI would otherwise leave open: which interpreter runs Mate,
and in which environment. It matters more here than for an ordinary console command, because
Mate reads the profiler cache and the compiled container of *this* project, and extensions
(`matesofmate/*` among them) can behave differently per runtime. Run under the wrong
interpreter, it does not fail, it reports on something that is not the application under test.

This is now **configured and enforced**, not left to documentation:

* `mate init` asks which command the coding agent should use, defaulting to
  `ddev exec vendor/bin/mate` when a `.ddev/` directory is present.
* The answer is recorded as `mate.invocation` in `mate/config.php`, together with the PHP
  version it was initialized under as `mate.php_version`.
* `mate.invocation` is materialized into `mate/AGENT_INSTRUCTIONS.md` and the managed
  `AGENTS.md` block, so the prefix reaches the agent that has to type it.
* Under a different PHP major.minor, Mate refuses to start and names the configured command:

  ```
   [ERROR] Mate is running under PHP 8.4.15 but this project expects PHP 8.3.
           Run it as "ddev exec vendor/bin/mate". ...
  ```

  `init`, `list`, `help` and `completion` stay callable, since `init` is what writes the
  configuration and the others never read the application. They print the same mismatch as a
  warning rather than passing silently, so a wrong interpreter is visible before it reaches a
  command that refuses. Set `mate.php_version` to `null` to disable the check entirely.

### Upgrade path

`UPGRADE.md` documents every rename and removal: the commands, the attribute names (including
that `name` is now required and that the attributes are method-only), the prompt removal, and the
two new `mate/config.php` parameters. The component is still `0.x`, and this is a breaking change
for anyone who scripted `mate serve` or the `mcp:*` commands.


src/mate/src/Attribute/AsTool.php
```diff
@@ -0,0 +1,34 @@
...
+ * The method's parameters (plus their `@param` PHPDoc) are reflected into a JSON
+ * input schema, and the method is invoked by the `tools:call` command.
+ *
+ * This attribute is intentionally standalone: it is not related to the Agent
+ * component's `Symfony\AI\Agent\Toolbox\Attribute\AsTool`.
+ *
+ * @author Johannes Wachter <johannes@sulu.io>
+ */
+#[\Attribute(\Attribute::TARGET_METHOD)]
+final class AsTool
```
chr-hertel: i struggle a bit with the collision with the symfony/ai-agent `AsTool` attribute - what do you think of `#[MateTool]` instead? (other attributes would adopt accordingly

chr-hertel:
what I'm missing here, and what should go to docs, if it still holds: why do we keep the concepts of tools, resources and resource templates over "just exposing commands" via skills. i know it's exactly what we discussed as next step - no worries - we move in iterations here anyways.

but maybe that's a bit why i'm slow with feedback here: i'm still flip flopping in my head - i'm still in on the idea of demoting MCP in favor of skills and a better agent support on CLI level.
i wonder tho, if we do want to focus more on skill management and think of adopting other - - already existing - symfony commands.

my argument would be: models partially tend already to call some debug commands due to pre/post training - or is that rather my anecdotal experience?

for example, we could have an agent detector in bin/console application or for some debug commands - but this should block this effort here . just ideas :D

wachterjohannes:
@chr-hertel i will be on vacation next week! let us talk afterwards about this.

lunetics:
Looked into this as well, and I land on the same side: skills plus materialized instructions feel like the better direction than MCP here. I tested the generated instruction files against a range of models (Claude, gpt-5.6 via codex, grok, deepseek, kimi, a couple of local qwens): without the files the CLI stays practically invisible to every model, and rewording them took invocation from hit-or-miss to nearly every run. Full run matrix and details are in wachterjohannes/symfony-ai#81.

wachterjohannes:
Answering the three parts separately.

**Why tools, resources and resource templates.** Tools carry an input schema generated from the signature plus `@param`. That is what `tools:inspect` prints, and the same reflection drives the argument casting, so a wrong type fails with a message instead of reaching your code. A skill pointing at a bare command has none of that. Resources exist for a different reason, navigation: `symfony-profiler://profile/{token}` returns the collectors that profile actually has, and only then do you read one of them. Without that shape you either push a whole profile into the context window or invent one command per collector.

**Why not adopt `bin/console debug:*`.** Mate never boots the kernel, it parses the compiled `*DebugContainer.xml` and reads the profiler storage from disk. So it still answers when the application does not boot, which is exactly the situation you are debugging in. And the output is shaped for an agent (summary vs data, duplicate queries grouped, secrets redacted, `--format=json|toon`) where `debug:*` prints tables for humans.

**On models calling debug commands on their own.** Not just anecdotal, I see it too. The materialized agent instructions and the skills steer it toward Mate instead, and that part works really well in practice. Bringing the two together stays open: Mate could wrap the debug commands, or the debug commands could point at Mate when it is installed. Both are worth looking at later.

You are right that this rationale is missing from the docs. It goes into the docs pass.

src/mate/src/Attribute/AsTool.php
wachterjohannes: Renamed to `#[MateTool]`, `#[MateResource]` and `#[MateResourceTemplate]`, including the `UPGRADE.md` migration diff.

wachterjohannes:
One more thing, from a side discussion with @chr-hertel that belongs here.

The fair version of his question is: would we ever have arrived at tools/resources/templates if we had not come through MCP first? Probably not. But the split turns out to earn its keep on its own terms, not just as an MCP leftover:

- It is a concept the models already know, so the interface needs no explaining.
- It keeps us free on the output side. Format is a property of the call (`--format=json|toon`), not baked into each command.
- The DX is good, and the surface an LLM has to hold in its head stays small: list, inspect, call, read.

On the debug commands: yes, they boot the kernel, and no, that is not a law of nature. It is on the lab list to test whether Mate can lean on them, or they on Mate, once this lands.

This is the sensible next step, definitely not the last one.

wachterjohannes:
One known gap, so it does not read as an oversight: the installed skills do not carry
`mate.invocation`. 11 snippets across 4 of the 6 shipped `SKILL.md` files still say
`vendor/bin/mate`, so in a containerized project an agent following a skill runs the host binary.
The interpreter guard catches that whenever the host PHP version differs, but not when it happens
to match. Tracked in #2445, together with the freshness-check detail that fix has to handle.

Rebased onto current main. The one real conflict was `ServiceTool`: #2406 restructured it for
multi-kernel contexts while this branch adds a result cap and a hard failure when no container has
been dumped. Resolved so the cap applies per context, each context carrying its own `services`,
`count` and `truncated`. That also removes the case where a missing container was reported as a
missing service.

src/mate/skills/symfony-log-investigation/SKILL.md
```diff
@@ -0,0 +1,39 @@
...
+- `monolog-list-channels`: distinct channel names (`app`, `security`, `doctrine`, ...). Reads every file, so it is the slow one; skip it if you already know the channel.
+- `monolog-tail` (`lines`, `level`, `environment`, `channel`): most recent entries. Reads ONLY the single newest file.
+- `monolog-search` (`term`, `regex`, `level`, `channel`, `environment`, `from`, `to`, `limit`): searches across all files. Empty `term` with filters set = filter-only.
+- `monolog-context-search` (`key`, `value`, `level`, `environment`, `limit`): matches a structured context field. No channel or date filter here.
+
+These commands accept `--format`: `json` to parse the result, `toon` (when `helgesverre/toon` is installed) for the smallest context footprint. A wide search returns many entries, so narrow with filters and a `limit` before widening the output format.
+
+## Workflow
+
+1. Orient: `vendor/bin/mate tools:call monolog-list-files`. Confirm the environment you care about is present and recently modified.
```
chr-hertel, line 20: soft one: at first claude complaint about `vendor/bin/mate` being in conflict with `mate.invocation`, but agents.md covers that already - one alternative could be to use `mate tools:call` or just `tools:call`

but on the other hand i doubt that codex & co would really struggle - just wanted to flag once

src/mate/src/Runtime/PhpVersionGuard.php
```diff
@@ -0,0 +1,92 @@
...
+ * @author Johannes Wachter <johannes@sulu.io>
+ */
+final class PhpVersionGuard
+{
+    /**
+     * Commands that must stay callable under any interpreter: `init` writes the very
+     * configuration this guard reads, and the rest never touch the application. They still warn,
+     * so a wrong interpreter is visible before it reaches a command that does refuse.
+     */
+    private const EXEMPT_COMMANDS = ['init', 'list', 'help', 'completion', '_complete'];
```
chr-hertel: what about `discover`?
```suggestion
    private const EXEMPT_COMMANDS = ['init', 'discover', 'list', 'help', 'completion', '_complete'];
```

src/mate/src/Discovery/SchemaGenerator.php
```diff
@@ -0,0 +1,395 @@
...
+     */
+    private function mapPhpTypeToJsonSchemaType(string $phpType): array
+    {
+        $normalized = strtolower(trim($phpType));
+
+        if (1 === preg_match('/^array\s*{/i', $normalized)) {
+            return ['object'];
+        }
+
+        if (str_contains($normalized, '[]') || 1 === preg_match('/^(array|list|iterable|collection)</i', $normalized)) {
```
chr-hertel, line 246: assoc arrays should in theory become an object, not array

src/mate/skills/symfony-log-investigation/SKILL.md
chr-hertel: oh, you flagged it as well :+1: :joy:

chr-hertel (approved):
Let's get it in - thanks @wachterjohannes 👍

PR #2381, opened 2026-08-05
[Platform][OpenResponses] Add citations metadata from `url_citation` annotations

valtzu:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| License       | MIT

`output_text` content's `annotations` field (part of the [OpenResponses spec](https://www.openresponses.org/reference#object-OutputTextContent-title)) was previously dropped entirely. Extract `url_citation` entries into a deduplicated citations metadata entry, matching the existing Perplexity bridge convention.

### DIFF
--- a/docs/components/platform.rst
+++ b/docs/components/platform.rst
@@ -714,6 +714,28 @@ the application is not expected to answer it::
     Like the other built-in server-side tool results, ``custom_tool_call`` items are only
     available on non-streamed responses.
 
+Citations
+~~~~~~~~~
+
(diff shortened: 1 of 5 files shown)


OskarStark (approved):
Can we add an example script?

valtzu:
> Can we add an example script?

Sure, added.

chr-hertel (approved):

chr-hertel:
Thank you @valtzu.

PR #2382, opened 2026-08-05
[Platform][OpenResponses] Add `CustomToolCallResult` for `custom_tool_call` items

valtzu:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| Docs?         | yes
| License       | MIT

Some providers (e.g. xAI's `x_search`) report their own built-in sub-tool calls as `custom_tool_call` output items, already resolved server-side by the time they're reported. Let's convert them into `CustomToolCallResult`, the same way the other built-in server-side tool calls (`web_search_call`, `mcp_call`, ...) are handled: an informational result alongside the assistant's message, not something the application is expected to execute and answer.

This fixes the usage of `x_search` remote tool since without these changes it throws a `RuntimeException`.

### DIFF
--- a/src/platform/CHANGELOG.md
+++ b/src/platform/CHANGELOG.md
@@ -7,6 +7,7 @@ CHANGELOG
  * [BC BREAK] Add `ListenerInterface::onError()` and `Result\Stream\ErrorEvent`, dispatched when draining a `StreamResult` throws, so a listener can finalize on a failed stream where `onComplete()` never fires; `AbstractStreamListener` provides a no-op default
  * Add `Test\Replay\CassetteHttpClient` and `Test\Replay\HttpCassette` to record real HTTP responses (when the cassette file is missing) and replay them offline through the real bridge pipeline (Contract, `ModelClient`, `ResultConverter`) in tests, including raw Server-Sent Event streams
  * Add `Test\Recording\RecordingProvider` (with `Cassette`/`ResultSerializer`) to record a real provider's result once (when the cassette file is missing) and replay it offline in tests
+ * Add `Result\CustomToolCallResult` for `custom_tool_call` output items reported by provider-specific server-side tools (e.g. xAI's `x_search`), converted the same way as the other built-in tool call results instead of being surfaced as a `ToolCall` the application is expected to execute
 
 0.12
 ----
--- a/src/platform/src/Result/CustomToolCallResult.php
+++ b/src/platform/src/Result/CustomToolCallResult.php
@@ -1,0 +1,63 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Platform\Result;
+
+/**
+ * Result of a custom tool invoked server-side by the model as part of a built-in tool
+ * (e.g. xAI's `x_search` reporting its own `custom_tool_call` sub-calls). Unlike
+ * {@see ToolCall}, the provider has already resolved the call by the time it is reported,
+ * so it is exposed as an informational result rather than something the application is
+ * expected to execute and answer.
+ *
+ * @author Valtteri R <valtzu@gmail.com>
+ */
+final class CustomToolCallResult extends BaseResult
+{
+    /**
+     * @param string      $name   Name of the custom tool that was invoked
+     * @param string      $input  Freeform text input the tool was called with, as reported by the provider
+     * @param string|null $id     Identifier of the custom tool call output item (e.g. "ctc_...")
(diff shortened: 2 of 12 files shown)


OskarStark (approved):
Can we add an example script?

valtzu:
> Can we add an example script?

Sure, added.

src/platform/src/Result/CustomToolCallResult.php
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * Result of a custom tool invoked server-side by the model as part of a built-in tool
+ * (e.g. xAI's `x_search` reporting its own `custom_tool_call` sub-calls). Unlike
+ * {@see ToolCall}, the provider has already resolved the call by the time it is reported,
+ * so it is exposed as an informational result rather than something the application is
+ * expected to execute and answer.
+ *
+ * @author Valtteri R <valtzu@gmail.com>
+ */
+final class CustomToolCallResult extends BaseResult
```
wachterjohannes, line 23: This is the only one of its family that `Message::toContent()` does not know about, and it has no `Message\Content\*` counterpart. Its siblings all have both:

| Result | `toContent()` | Content class |
| --- | --- | --- |
| `WebSearchResult` | yes | `WebSearch` |
| `McpCallResult` | yes | `McpCall` |
| `LocalShellCallResult` | yes | `LocalShellCall` |
| `CustomToolCallResult` | **missing** | **missing** |

So the moment the result reaches an assistant message it throws:

```
WebSearchResult (existing sibling)   -> WebSearch, Text
CustomToolCallResult (this PR)       -> InvalidArgumentException:
        Unsupported assistant message part of type "...\CustomToolCallResult".
```

`Chat::submit()` calls `Message::ofAssistant($result)`, so a chat using OpenResponses with `x_search` swaps the converter's `RuntimeException` for an `InvalidArgumentException` when the reply is persisted. The PR still improves things, without it the converter throws first, but "fixes the usage of `x_search`" does not hold end to end yet.

Could you add a `Content\CustomToolCall` and the matching `toContent()` arm, the way `McpCall` does it?

src/platform/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ CHANGELOG
  * [BC BREAK] Add `ListenerInterface::onError()` and `Result\Stream\ErrorEvent`, dispatched when draining a `StreamResult` throws, so a listener can finalize on a failed stream where `onComplete()` never fires; `AbstractStreamListener` provides a no-op default
  * Add `Test\Replay\CassetteHttpClient` and `Test\Replay\HttpCassette` to record real HTTP responses (when the cassette file is missing) and replay them offline through the real bridge pipeline (Contract, `ModelClient`, `ResultConverter`) in tests, including raw Server-Sent Event streams
  * Add `Test\Recording\RecordingProvider` (with `Cassette`/`ResultSerializer`) to record a real provider's result once (when the cassette file is missing) and replay it offline in tests
+ * Add `Result\CustomToolCallResult` for `custom_tool_call` output items reported by provider-specific server-side tools (e.g. xAI's `x_search`), converted the same way as the other built-in tool call results instead of being surfaced as a `ToolCall` the application is expected to execute.
```
wachterjohannes: Nit: trailing period, the other entries in this section have none.

wachterjohannes (commented):
Converter side looks right and the example answers @OskarStark's request. Verified locally, all green.

One gap: the result cannot become an assistant message yet, so the `x_search` case still fails, just later and with a different exception. Details inline.

src/platform/CHANGELOG.md
valtzu: Fixed

src/platform/src/Result/CustomToolCallResult.php
valtzu: Right, good catch, fixed.

wachterjohannes:
Opened #2430 for the underlying gap, since this is the third time it has come up. Not something this PR has to solve.

wachterjohannes (approved):
Both addressed, and the round trip works now:

```
before: CustomToolCallResult -> InvalidArgumentException: Unsupported assistant message part
after:  CustomToolCallResult -> CustomToolCall, Text
```

Verified against a real `custom_tool_call` item from a live Responses API rather than a synthetic one:

```
live output items: reasoning, custom_tool_call

  ThinkingResult
  CustomToolCallResult   name=code_exec status=NULL id='ctc_tmp_...'
  as assistant message:  Thinking, CustomToolCall
```

On `main` the same payload throws the `RuntimeException` from `convertOutputArray()`, so the reported failure reproduces and this fixes it end to end. Two notes from that capture: the real item carries no `status`, which your optional handling covers, and it has a `call_id` next to `id` that you correctly ignore, since the call is already resolved.

Re-verified locally, all green. The underlying gap, that nothing enforces this pairing, is tracked in #2430 and is not yours to solve.

wachterjohannes:
Thank you @valtzu.

PR #2385, opened 2026-08-06
[Mate] Redact email bodies, recipients and link tokens in mailer collector

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | n/a
| License       | MIT

The Mailer profiler collector handed the AI recipient addresses, up to 500 chars
of the raw text body, and — most dangerously — `extractLinks()` scraped every
URL out of the bodies, so password-reset / verification / magic-login links were
returned **with their one-time tokens**, handing the model working credentials.

- Omit the email body entirely (keep a `has_text_body` flag for triage).
- For links: drop userinfo, query string and fragment, and mask high-entropy
  path segments — the canonical reset flow `/reset-password/reset/{token}`
  carries the token in the **path** — while keeping descriptive slugs so the
  endpoint stays visible.
- Mask the local part of recipient addresses and drop display names (keep the
  domain).

Subjects are kept (diagnostic value outweighs the residual risk).

### DIFF
--- a/src/mate/src/Bridge/Symfony/Profiler/Service/Formatter/MailerCollectorFormatter.php
+++ b/src/mate/src/Bridge/Symfony/Profiler/Service/Formatter/MailerCollectorFormatter.php
@@ -16,13 +16,14 @@ use Symfony\Component\HttpKernel\DataCollector\DataCollectorInterface;
 use Symfony\Component\Mailer\DataCollector\MessageDataCollector;
 use Symfony\Component\Mime\Address;
 use Symfony\Component\Mime\Email;
-use Symfony\Component\String\UnicodeString;
 
 /**
  * Formats Mailer collector data for AI consumption.
  *
- * Extracts email message details including recipients, subject,
- * body preview, attachments, and transport information.
+ * Extracts email message metadata (subject, masked recipients, attachments,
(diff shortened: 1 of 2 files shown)


src/mate/src/Bridge/Symfony/Profiler/Service/Formatter/MailerCollectorFormatter.php
chr-hertel, line 1: i guess some of this would be easier when using `symfony/string` - not a real concern tho, just an idea

chr-hertel (approved):

chr-hertel:
Thank you @wachterjohannes.

src/mate/src/Bridge/Symfony/Profiler/Service/Formatter/MailerCollectorFormatter.php
wachterjohannes: Thanks! Looked at the candidates (`maskAddress`, `maskPathTokens`, `looksLikeToken`). They all work on ASCII-only data, so `symfony/string` wouldn't add much, but it would add a new dependency to this package. Prefer keeping it as is unless you had a specific spot in mind.

PR #2387, opened 2026-08-07
[Agent] Fix int to float widening for scalar tool parameters

### LINKED ISSUE #2136: [Agent] ToolCallArgumentResolver does not widen int→float for top-level scalar `float` parameters (remaining gap after #2011)
## Summary

#2011 fixed the JSON int→float case for `float` **properties of object/DTO tool parameters** by passing `'json'` as the format to `denormalize()`. That coercion lives in `AbstractObjectNormalizer` and only fires for the JSON format, so passing the format was enough — on v0.9.0 object properties now denormalize correctly.

A **top-level scalar `float` parameter** (`__invoke(float $amount)`) is still rejected. Such a parameter is never denormalized by `AbstractObjectNormalizer`; it goes through the scalar branch of `Serializer::denormalize()`, which performs no int→float coercion and ignores `$format` entirely. Passing `'json'` (the #2011 fix) therefore has no effect on this path.

## Background — the JSON-specific cast lives only in `AbstractObjectNormalizer`

https://github.com/symfony/serializer/blob/v7.4.0/Normalizer/AbstractObjectNormalizer.php#L587-L595

```php
// JSON only has a Number type corresponding to both int and float PHP types.
// PHP's json_encode, JavaScript's JSON.stringify, Go's json.Marshal as well as most other JSON encoders convert
// floating-point numbers like 12.0 to 12 (the decimal part is dropped when possible).
// PHP's json_decode automatically converts Numbers without a decimal part to integers.
// To circumvent this behavior, integers are converted to floats when denormalizing JSON based formats and when
// a float is expected.
if (LegacyType::BUILTIN_TYPE_FLOAT === $builtinType && \is_int($data) && null !== $format && str_contains($format, JsonEncoder::FORMAT)) {
    return (float) $data;
}
```

This widening is implemented **inside `AbstractObjectNormalizer`**, i.e. it only runs while denormalizing the properties of an object. A top-level scalar argument never reaches it.

## The call site already passes `'json'` (v0.9.0, the #2011 fix)

https://github.com/symfony/ai/blob/v0.9.0/src/agent/src/Toolbox/ToolCallArgumentResolver.php#L104-L106

```php
if ($this->denormalizer->supportsDenormalization($value, $parameterType, 'json')) {
    $value = $this->denormalizer->denormalize($value, $parameterType, 'json');
}
```

For an **object** parameter this is sufficient: `ObjectNormalizer`/`AbstractObjectNormalizer` widens int→float for its `float` properties. For a **top-level scalar** `float` parameter it is not, because of the next section.

## Why scalars still fail — the scalar branch in `Serializer::denormalize()`

https://github.com/symfony/serializer/blob/v7.4.10/Serializer.php#L197-L206

```php
$normalizer = $this->getDenormalizer($data, $type, $format, $context);

// Check for a denormalizer first, e.g. the data is wrapped
if (!$normalizer && isset(self::SCALAR_TYPES[$type])) {
    if (!('is_'.$type)($data)) {
        throw NotNormalizableValueException::createForUnexpectedDataType(\sprintf('Data expected to be of type "%s" ("%s" given).', $type, get_debug_type($data)), $data, [$type], $context['deserialization_path'] ?? null, true);
    }

    return $data;
}
```

For `denormalize(100, 'float', 'json')`: none of the registered normalizers supports the scalar type `float`, so `$normalizer` is `null` and this branch runs. It evaluates `is_float(100)` (`false`) and throws. `$format` is never consulted here, so passing `'json'` makes no difference. The JSON int↔float ambiguity is thus unhandled for top-level scalar `float` parameters.

## Observed behavior

An LLM tool call `{"amount": 100}` (a JSON number without a fractional part — `json_decode` returns PHP `int`) against `__invoke(float $amount)` fails with:

```
NotNormalizableValueException:
  Data expected to be of type "float" ("int" given).
```

The same value wrapped in an object — `{"foo": {"amount": 100}}` against `__invoke(Foo $foo)` where `Foo::$amount` is `float` — denormalizes successfully on v0.9.0 thanks to #2011. Only the top-level scalar case remains broken.

## Minimal reproducer

```php
<?php

final class SomeTool
{
    public function __invoke(float $amount): array
    {
        return [];
    }
}

// LLM tool call payload: {"amount": 100}
// json_decode produces: ['amount' => 100]   (int, not float)
$toolCall = new ToolCall('tc-1', 'some_tool', ['amount' => 100]);

// NotNormalizableValueException:
//   Data expected to be of type "float" ("int" given).
```

## Note

Could `ToolCallArgumentResolver` widen the `int` to `float` itself when the resolved top-level parameter type is `float` (JSON has no distinct int/float type)? That would mirror what `AbstractObjectNormalizer` already does for object properties.

## Environment

- `symfony/ai-agent`: **v0.9.0**
- `symfony/ai-bundle`: v0.9.0
- `symfony/serializer`: 7.4.10
- Symfony: 7.4
- PHP: 8.4

## Related

- #2011 — passed `'json'` to `denormalize()`; fixed the object-property `float` case. This issue is the remaining top-level scalar case the same fix did not cover.
- #2004 — sits in the same `ToolCallArgumentResolver::resolveArguments()` method, different root cause (`NullableType` stringification).

lazerg:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #2136
| License       | MIT

JSON has a single Number type, so `{"amount": 100}` decodes to a PHP `int` and a tool declared as `__invoke(float $amount)` blows up with `Data expected to be of type "float" ("int" given)`.

#2011 passed `'json'` to `denormalize()`, which fixed the same problem for `float` properties of object parameters, because the int to float widening lives in `AbstractObjectNormalizer`. A top-level scalar never reaches that normalizer: `Serializer::denormalize()` handles it in its own scalar branch, which checks `is_float()` and ignores the format entirely.

So the resolver now widens the value itself when the resolved parameter type is `float`, same as the normalizer does for object properties. Nullable `float` parameters go through the same path, since the wrapped type is already unwrapped at that point.

Parameters typed `float[]` reach that same scalar branch one level down through `ArrayDenormalizer` and still throw. I left that out here since the issue is about the top-level case, but I can fold it in if you want both in one go.

### DIFF
--- a/src/agent/src/Toolbox/ToolCallArgumentResolver.php
+++ b/src/agent/src/Toolbox/ToolCallArgumentResolver.php
@@ -101,6 +101,10 @@ final class ToolCallArgumentResolver implements ToolCallArgumentResolverInterfac
 
             $parameterType .= $dimensions;
 
+            if ('float' === $parameterType && \is_int($value)) {
+                $value = (float) $value;
+            }
+
             if ($this->denormalizer->supportsDenormalization($value, $parameterType, 'json')) {
                 $value = $this->denormalizer->denormalize($value, $parameterType, 'json');
             }
--- a/src/agent/tests/Fixtures/Tool/ToolScalarFloat.php
+++ b/src/agent/tests/Fixtures/Tool/ToolScalarFloat.php
@@ -1,0 +1,23 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Agent\Tests\Fixtures\Tool;
+
+use Symfony\AI\Agent\Toolbox\Attribute\AsTool;
+
+#[AsTool('tool_scalar_float', 'A tool with float parameters')]
+final class ToolScalarFloat
+{
+    public function __invoke(float $height, ?float $weight = null): string
+    {
+        return \sprintf('Height: %.2fm, Weight: %.2fkg', $height, $weight);
+    }
+}
--- a/src/agent/tests/Toolbox/ToolCallArgumentResolverTest.php
+++ b/src/agent/tests/Toolbox/ToolCallArgumentResolverTest.php
@@ -18,6 +18,7 @@ use Symfony\AI\Agent\Tests\Fixtures\Tool\ToolArrayMultidimensional;
 use Symfony\AI\Agent\Tests\Fixtures\Tool\ToolDate;
 use Symfony\AI\Agent\Tests\Fixtures\Tool\ToolNoParams;
 use Symfony\AI\Agent\Tests\Fixtures\Tool\ToolObjectFloat;
+use Symfony\AI\Agent\Tests\Fixtures\Tool\ToolScalarFloat;
 use Symfony\AI\Agent\Tests\Fixtures\Tool\ToolWithNullableClass;
 use Symfony\AI\Agent\Toolbox\ToolCallArgumentResolver;
 use Symfony\AI\Platform\Result\ToolCall;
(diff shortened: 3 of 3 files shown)


chr-hertel (approved):

chr-hertel:
Thank you @lazerg & welcome to symfony/ai repository! :)

PR #2388, opened 2026-08-07
[Platform][Cache] Key MessageBag inputs on their message content

### LINKED ISSUE #2192: [Platform][Cache] CachePlatform never hits cache when input is a MessageBag

lazerg:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #2192
| License       | MIT

`MessageBagCacheKeyGenerator` returned `$input->getId()->toString()`, and `MessageBag` generates a fresh `Uuid::v7()` in its constructor. Every request builds a new bag, so every request produced a new cache key and `CachePlatform` never returned a cached result for a conversation.

The generator now hashes the messages themselves: their class, their content and their order. The identifiers and the metadata of the bag and of its messages stay out of the hash, since they describe the instance rather than the conversation. Identifiers a provider assigns to a content part, a tool call id for instance, are kept, because they are sent along with it. Binary content is keyed on a hash of its bytes, the way `FileCacheKeyGenerator` already does, so a lazily read file does not leak its closure into the key.

`template_vars` now goes into the key as well. Templates are rendered inside the inner platform, after `CachePlatform` has built the key, so without that a bag of templates would serve the first render to every later set of variables.

The existing test for two separate bags with the same messages asserted the current behaviour (two calls to the inner platform), so it now asserts the cache hit, and further tests cover two bags with different messages and two calls with different template variables.

### DIFF
--- a/src/platform/src/Bridge/Cache/CachePlatform.php
+++ b/src/platform/src/Bridge/Cache/CachePlatform.php
@@ -81,6 +81,10 @@ final class CachePlatform implements PlatformInterface
             default => $this->generateInputCacheKey($input),
         };
 
+        if (isset($options['template_vars'])) {
+            $normalizedInput = hash('xxh128', $normalizedInput.serialize([$options['template_vars'], $options['template_options'] ?? []]));
+        }
+
         $cacheKey = (new UnicodeString())->join([
             $options['prompt_cache_key'] ?? $this->cacheKey,
             (new UnicodeString($modelName))->camel(),
--- a/src/platform/src/Bridge/Cache/MessageBagCacheKeyGenerator.php
+++ b/src/platform/src/Bridge/Cache/MessageBagCacheKeyGenerator.php
@@ -12,9 +12,18 @@
 namespace Symfony\AI\Platform\Bridge\Cache;
 
 use Symfony\AI\Platform\Exception\InvalidArgumentException;
+use Symfony\AI\Platform\Message\Content\File;
 use Symfony\AI\Platform\Message\MessageBag;
+use Symfony\AI\Platform\Message\MessageInterface;
 
 /**
+ * Keys a {@see MessageBag} on the content of its messages, in order.
+ *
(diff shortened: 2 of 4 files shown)


lazerg:
Hey @chr-hertel, could you review this fix that keys MessageBag inputs on their message content in the cache? Open for a month with no feedback.

PR #2389, opened 2026-08-07
[Chat][Platform] Notify stream listeners about deltas yielded from an injected generator

### LINKED ISSUE #2330: [Chat][Agent][Platform] Streamed tool-call answers are not persisted by ChatStreamListener
# [Chat] Streamed tool-call answer is not persisted to the store

I have an agent with a couple of tools and I'm streaming the response with `Chat::stream()`. When the agent replies directly (plain text) the streamed message is persisted to my store just fine, but when the reply comes *after* a tool call, the streamed follow-up text shows up in my frontend yet never lands in the store — an empty assistant message gets saved instead.

When looking at `StreamResult::getContent()` this kind of makes sense: a listener-injected sub-generator is forwarded with `yield from` and its inner deltas are never re-dispatched to the listeners' `onDelta()`, so `ChatStreamListener` never sees the follow-up text. It does make me wonder what the intended way to persist a streamed tool-call answer is — should `getContent()` re-dispatch the sub-generator deltas, or should `ChatStreamListener` accumulate differently?

Small reproducer (with `InMemoryStore`, so unrelated to the session-store streaming limitation):

```php
$store = new InMemoryStore();
$messages = new MessageBag();
$messages->add(Message::ofUser('question'));

$stream = new StreamResult((static function (): \Generator {
    yield new ToolCallComplete([new ToolCall('id', 'some_tool')]);
})());

// the toolbox injects the tool follow-up as a streamed sub-generator, like AgentProcessor does
$stream->addListener(new StreamListener(
    static fn (ToolCallResult $r, ?AssistantMessage $m = null): StreamResult => new StreamResult(
        (static function (): \Generator {
            yield new TextDelta('The ');
            yield new TextDelta('answer.');
        })()
    )
));
$stream->addListener(new ChatStreamListener($messages, $store));

iterator_to_array($stream->getContent()); // the frontend correctly receives "The answer."

$store->load()->getMessages()[1]->asText(); // '' — expected 'The answer.'
```

lazerg:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #2330
| License       | MIT

When a stream listener replaces a delta with a generator, `StreamResult::getContent()` forwarded it with `yield from`. PHP does not re-enter the outer generator for the values of a delegated one, so the deltas coming out of that generator were never sent to `onDelta()`. The toolbox `StreamListener` injects the tool-call follow-up exactly that way, which is why `ChatStreamListener` saw nothing after a tool call and stored an empty assistant message, while the frontend received the text.

The delta dispatch now recurses into an injected generator and notifies the listeners about each delta it yields. The listener that injected the generator is left out of that nested round, so it does not reprocess its own output: `PartialObjectStreamListener` would otherwise buffer the same text twice and re-inject forever, and the toolbox `StreamListener` would drop the follow-up it just produced. The order and the number of yielded deltas stay the same.

The forwarding also went through `yield from`, which keeps the inner generator's keys, so the outer stream repeated key `0` and `iterator_to_array()` on it silently lost deltas. Yielding one by one gives the whole stream sequential keys again.

### DIFF
--- a/src/chat/tests/ChatStreamListenerTest.php
+++ b/src/chat/tests/ChatStreamListenerTest.php
@@ -17,9 +17,13 @@ use Symfony\AI\Chat\InMemory\Store as InMemoryStore;
 use Symfony\AI\Platform\Message\AssistantMessage;
 use Symfony\AI\Platform\Message\Message;
 use Symfony\AI\Platform\Message\MessageBag;
+use Symfony\AI\Platform\Result\Stream\AbstractStreamListener;
 use Symfony\AI\Platform\Result\Stream\Delta\TextDelta;
 use Symfony\AI\Platform\Result\Stream\Delta\ThinkingDelta;
+use Symfony\AI\Platform\Result\Stream\Delta\ToolCallComplete;
+use Symfony\AI\Platform\Result\Stream\DeltaEvent;
 use Symfony\AI\Platform\Result\StreamResult;
+use Symfony\AI\Platform\Result\ToolCall;
 
 final class ChatStreamListenerTest extends TestCase
 {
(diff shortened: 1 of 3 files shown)


lazerg:
The Examples / PHP 8.5 job failed while pulling the service images, with a 502 from the registry

lazerg:
Hey @chr-hertel, could you take a look at this fix for notifying stream listeners about deltas yielded from an injected generator? The earlier CI failure was a registry 502, unrelated to the change.

PR #2390, opened 2026-08-09
[Agent] Preserve assistant content during tool calls

fabpot:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Docs?         | no 
| Issues        | n/a
| License       | MIT

Preserves the assistant’s complete response (text, thinking/signatures, and tool calls) when an agent enters the tool loop. Content order is retained for streaming and buffered responses, including OpenResponses, while empty replay blocks are discarded.

### DIFF
--- a/src/agent/src/Toolbox/StreamListener.php
+++ b/src/agent/src/Toolbox/StreamListener.php
@@ -11,12 +11,20 @@
 
 namespace Symfony\AI\Agent\Toolbox;
 
-use Symfony\AI\Platform\Message\Message;
+use Symfony\AI\Platform\Message\AssistantMessage;
+use Symfony\AI\Platform\Message\Content\ContentInterface;
+use Symfony\AI\Platform\Message\Content\Text;
+use Symfony\AI\Platform\Message\Content\Thinking;
 use Symfony\AI\Platform\Result\ResultInterface;
 use Symfony\AI\Platform\Result\Stream\AbstractStreamListener;
 use Symfony\AI\Platform\Result\Stream\CompleteEvent;
 use Symfony\AI\Platform\Result\Stream\Delta\TextDelta;
+use Symfony\AI\Platform\Result\Stream\Delta\ThinkingComplete;
+use Symfony\AI\Platform\Result\Stream\Delta\ThinkingDelta;
+use Symfony\AI\Platform\Result\Stream\Delta\ThinkingSignature;
+use Symfony\AI\Platform\Result\Stream\Delta\ThinkingStart;
 use Symfony\AI\Platform\Result\Stream\Delta\ToolCallComplete;
+use Symfony\AI\Platform\Result\Stream\Delta\ToolCallStart;
 use Symfony\AI\Platform\Result\Stream\DeltaEvent;
 use Symfony\AI\Platform\Result\Stream\StartEvent;
 use Symfony\AI\Platform\Result\ToolCallResult;
@@ -27,7 +35,18 @@ use Symfony\AI\Platform\Result\ToolCallResult;
  */
 final class StreamListener extends AbstractStreamListener
 {
-    private string $buffer = '';
+    /**
+     * Content is collected in provider order; strings reserve the positions announced by ToolCallStart.
+     *
+     * @var list<ContentInterface|string>
+     */
+    private array $assistantContent = [];
+
+    // Index of the thinking block currently receiving streamed chunks
+    private ?int $currentThinkingIndex = null;
+
+    // Last thinking block, for providers that emit its signature after completion
+    private ?int $lastThinkingIndex = null;
     private ?ResultInterface $result = null;
     private bool $toolHandled = false;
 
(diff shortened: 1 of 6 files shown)


src/agent/src/Toolbox/StreamListener.php
```diff
@@ -27,7 +35,18 @@
  */
 final class StreamListener extends AbstractStreamListener
 {
-    private string $buffer = '';
+    /**
+     * Content is collected in provider order; strings reserve the positions announced by ToolCallStart.
```
chr-hertel, line 39: i wonder if we need #2396 to make this more reliable ... 🤔

src/agent/src/Toolbox/StreamListener.php
chr-hertel, line 1: feels a bit like my StreamListener idea doesn't scale well :D

src/agent/src/Toolbox/StreamListener.php
fabpot: Yes! It will help tremendously.

src/agent/src/Toolbox/StreamListener.php
```diff
@@ -27,7 +35,18 @@
...
+     */
+    private array $assistantContent = [];
+
+    // Index of the thinking block currently receiving streamed chunks
+    private ?int $currentThinkingIndex = null;
+
+    // Last thinking block, for providers that emit its signature after completion
+    private ?int $lastThinkingIndex = null;
     private ?ResultInterface $result = null;
     private bool $toolHandled = false;
```
wachterjohannes, line 51: @chr-hertel on the "doesn't scale well" point, I think the shape is the problem rather than this code.

What this class does is stream reassembly: rebuilding the structured message the provider would have sent, from its deltas. That needs per-bridge delta vocabulary, and the vocabulary is not uniform:

```
ToolCallStart      Anthropic, ClaudeCode, Generic
ThinkingStart      Anthropic, Gemini, OpenResponses, VertexAi
ThinkingSignature  Anthropic, OpenResponses
ThinkingComplete   Anthropic, Gemini, Generic, OpenResponses, VertexAi
ToolInputDelta     Anthropic, ClaudeCode, Generic
```

Five types, each from a different subset. That is why this version needs four "the provider did not send X" compensations: `currentThinkingIndex ?? startThinking()` twice, the signature-without-an-open-block branch, and the trailing loop for bridges that never emit `ToolCallStart`. The class lives in Agent, reconstructs Platform semantics, and cannot know which bridge it is attached to, so every new delta type lands here. 84 to 193 lines for one feature.

The listener shape costs on top of that: five mutable fields reset in `onStart`/`onComplete` because an instance outlives one logical run, a state machine spread across callbacks, and `setDelta()` replacing one delta with the whole continuation, so it rewrites the stream it observes.

Two halves to fixing it, and you already built the first in #2436: consuming the stream with a plain `foreach` inside a generator makes the state local, drops the reset and the substitution entirely. The second half is that the reassembly itself belongs in Platform, next to the deltas it decodes. That knowledge is bridge knowledge, Agent cannot have it. And Platform already does exactly this once, `PartialObjectStreamListener` rebuilds the final `ObjectResult` from the deltas, which #2436 then asks for in `streamedObjectResult()`. Generalising that instead of keeping structured output as the special case would let `consumeStream()` collect `ContentInterface` parts in provider order and build `Message::ofAssistant(...$parts)`, and the four compensations above become each bridge's own business.

wachterjohannes (commented):
The problem is real and still unfixed. I ran the same tool round on three branches, with a model answering `Thinking` + `Text` + `ToolCall`, and looked at what reaches the message bag:

```
main (6e901317)   AssistantMessage  ToolCall(clock)                        <- thinking and text gone
PR 2436           AssistantMessage  ToolCall(clock)                        <- identical to main
this PR           AssistantMessage  Thinking(sig=...) + Text(...) + ToolCall(clock)
```

Streamed, `main` and #2436 both produce two consecutive assistant messages (text, then tool call) where this PR produces one. `Bridge/Anthropic/Contract/AssistantMessageNormalizer.php:67-92` already serializes all three part types out of a single assistant message, and `MessageBagNormalizer` does not merge anything, so the bridges are ready for this and simply never receive it.

Two things about where this now stands:

* **#2436 does not fix this**, which is easy to assume given it rewrites the same code. It moves the tool loop into an iterative generator but still does `Message::ofAssistant(...$toolCalls)`, so the measurement above comes out byte for byte identical to `main`. Worth saying out loud so it does not fall between the two PRs.
* **It does delete what this PR patches.** `Toolbox\StreamListener` is gone there and `handleToolCalls()` no longer exists, so both agent-side hunks need porting to the new loop. Rebasing before that lands means doing the work twice. After it lands this reduces to one place in `Runner::consumeStream()`, and the OpenResponses normalizer stands on its own either way.

Also: #2396 has been merged since this discussion, so the reliability concern you both raised is available now.

The red check is `Validate Changelog & Upgrade Entries`. Given content is being lost, `Bug fix? yes` / `New feature? no` is probably the more honest table and clears it without a changelog entry.

Verified locally, all green.

chr-hertel:
@fabpot cannot verify with your use-case of course, but I'd hope that the issue is gone after #2450 now with v0.13.0 - can you confirm or do we still need something here?
Thanks already! :pray:

fabpot:
Closing now. I'm updated my project and it almost works for me. Thank you.
I might have a couple of follow-ups :)

PR #2395, opened 2026-08-11
[Agent] Fix maxToolCalls never being reached across nested agent calls

### LINKED ISSUE #2393: [Agent] `maxToolCalls` never triggers: the tool-calling loop is unbounded in the standard wiring
# `AgentProcessor`: `maxToolCalls` never triggers — the tool-calling loop is unbounded in the standard wiring

**Package:** `symfony/ai-agent` v0.10.0 (reproduced with `symfony/ai-bundle` v0.10.0, `max_tool_calls` per agent)
**File:** `src/agent/src/Toolbox/AgentProcessor.php` (as installed: `symfony/ai-agent`, `src/Toolbox/AgentProcessor.php`)

The documented safety limit is unreachable, so a tool-calling loop has no upper bound other than the PHP stack. Reporting here rather than on `symfony/ai-agent`, whose issues are disabled as a split.

## What happens

`AgentProcessor::$maxToolCalls` (the bundle's per-agent `max_tool_calls`, default 50) is checked against a counter that is **local to one entry into `handleToolCallsCallback()`**:

```php
private function handleToolCallsCallback(Output $output, ToolCallResult $result, ?AssistantMessage $streamedAssistantResponse = null): ResultInterface
{
    ++$this->nestingLevel;
    // …
    try {
        $iterations = 0;                                     // <— per entry, not per conversation
        do {
            if (null !== $this->maxToolCalls && ++$iterations > $this->maxToolCalls) {
                throw new MaxIterationsExceededException($this->maxToolCalls);
            }
            // … execute the tool calls …
            $result = $event->hasResult() ? $event->getResult() : $this->agent->call($messages, $output->getOptions());
        } while ($result instanceof ToolCallResult);
    } finally {
        --$this->nestingLevel;
    }
    // …
}
```

But `$this->agent->call(...)` is a **full agent invocation**: `Agent::call()` runs every registered output processor on the new result, and this processor is one of them (`Agent::call()` even calls `setAgent($this)` on it first). So when the model asks for another tool, the nested `processOutput()` handles it by re-entering `handleToolCallsCallback()` — with a **fresh `$iterations = 0`**.

The nested entry returns a result that is no longer a `ToolCallResult`, so the parent's `do … while ($result instanceof ToolCallResult)` is false on its first check. **Every level therefore executes exactly one iteration, and `$iterations` never exceeds 1.** The check can only fire for `maxToolCalls <= 0`.

Net effect: a tool-calling conversation of N rounds becomes N levels of recursion, each with its own counter at 1, and the configured limit is never reached no matter how long the loop runs. The class is aware of the recursion elsewhere — `$nestingLevel` exists precisely to reset `$sources` only at the outermost level — but the iteration counter is written as if the loop were flat.

## Evidence from production

A Symfony 8 app on `symfony/ai` 0.10 with `max_tool_calls: 12` configured for one agent (verified as wired: the compiled container reads `new AgentProcessor(…, false, false, 12)`):

- one continuous conversation ran **17 tool-calling rounds** — 18 platform calls for that agent within a single request, 17 of which returned tool calls, with the message bag growing monotonically (so it is one conversation, not several invocations);
- nine further conversations sit between 12 and 16 rounds;
- across **3 394 runs**, `MaxIterationsExceededException` was never thrown once, and every one of those runs completed normally.

That is exactly what the code predicts: the limit is not "loose", it is unreachable.

## Why it matters

`max_tool_calls` is the only documented protection against a model that keeps calling tools — the CHANGELOG entry that introduced it says as much ("Cap an agent's tool-calling loop at `50` iterations by default … or set it to `null` to restore the previous unbounded behaviour"). As shipped, the default and the unbounded mode behave identically. An agent stuck in a tool loop bills the API provider for every round and is stopped only by whatever the model decides, by an HTTP timeout, or eventually by exhausting the PHP stack — a stack overflow rather than the intended `MaxIterationsExceededException`.

It also silently misleads operators: a value that is set, documented and visible in the container looks like a live safety net.

## Suggested fix

Count rounds across the whole conversation instead of per entry. The class already has the mechanism it needs in `$nestingLevel`:

```php
private int $iterations = 0;   // instance state, alongside $nestingLevel

// in handleToolCallsCallback(), on entry:
++$this->nestingLevel;

// inside the loop:
if (null !== $this->maxToolCalls && ++$this->iterations > $this->maxToolCalls) {
    throw new MaxIterationsExceededException($this->maxToolCalls);
}

// in the finally block:
--$this->nestingLevel;
if (0 === $this->nestingLevel) {
    $this->iterations = 0;      // reset only when the outermost call finishes, like $sources
}
```

Two details worth deciding deliberately rather than by accident:

1. **Reset semantics.** Resetting at nesting level 0 mirrors what the class already does for `$sources`, and makes the limit "per outermost agent call", which is what a reader of `max_tool_calls` expects. Note the processor is a shared service, so a counter that is never reset would leak across requests in a long-running worker (Messenger, RoadRunner, FrankenPHP).
2. **The streaming path.** `processOutput()` registers a `StreamListener` that calls the same callback, so it inherits the same behaviour and the same fix.

## Reproduction sketch

No mocking of the platform is strictly needed to see it — a unit test that drives `processOutput()` with a stub agent whose `call()` re-enters `processOutput()` (i.e. behaves like the real `Agent::call()`) shows the counter resetting. A test asserting `MaxIterationsExceededException` after `maxToolCalls + 1` rounds with a *flat* stub agent will pass today and still pass after the fix; the missing coverage is precisely the nested case, which is the only one the standard wiring produces.

lazerg:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #2393
| License       | MIT

The iteration counter in `Runner::handleToolCalls()` is a local variable, but every round of tool calling goes through `$agent->call()`, which lands back in the same method one level deeper with a fresh counter. Each level then runs exactly one iteration, so `maxToolCalls` is only ever compared against 1 and never triggers. The loop ends up bounded by the PHP stack rather than by the configured limit.

The counter now lives on the class next to `$nestingLevel` and gets reset once the outermost call finishes, the same way `$sources` already does.

The existing tests pass today because their stub agent returns straight into the same loop. The new test re-enters the runner the way the real `Agent::call()` does.

### DIFF
--- a/src/agent/src/Execution/Runner.php
+++ b/src/agent/src/Execution/Runner.php
@@ -51,6 +51,12 @@ final class Runner
      */
     private int $nestingLevel = 0;
 
+    /**
+     * Tool calling iterations get counted on class level to also cover the nested agent calls.
+     * They get reset when the outermost agent call is finished via nesting level.
+     */
+    private int $iterations = 0;
+
     public function __construct(
         private readonly PlatformInterface $platform,
         private readonly ?ToolboxInterface $toolbox = null,
@@ -143,9 +149,8 @@ final class Runner
         }
 
         try {
-            $iterations = 0;
             do {
-                if (null !== $this->maxToolCalls && ++$iterations > $this->maxToolCalls) {
+                if (null !== $this->maxToolCalls && ++$this->iterations > $this->maxToolCalls) {
                     throw new MaxIterationsExceededException($this->maxToolCalls);
                 }
 
@@ -172,6 +177,10 @@ final class Runner
             } while ($result instanceof ToolCallResult);
         } finally {
             --$this->nestingLevel;
+
+            if (0 === $this->nestingLevel) {
+                $this->iterations = 0;
+            }
 
             if ($this->includeSources && 0 === $this->nestingLevel && $result instanceof ToolCallResult) {
                 $this->sources = new SourceCollection();
--- a/src/agent/tests/Execution/RunnerTest.php
+++ b/src/agent/tests/Execution/RunnerTest.php
@@ -468,6 +468,35 @@ final class RunnerTest extends TestCase
         $runner->run($agent, 'gpt-4', new MessageBag(), []);
     }
 
+    public function testThrowsExceptionWhenMaxIterationsExceededAcrossNestedAgentCalls()
+    {
+        $toolCall = new ToolCall('id1', 'tool1', ['arg1' => 'value1']);
+        $toolbox = $this->createMock(ToolboxInterface::class);
+        $toolbox
+            ->method('execute')
+            ->willReturn(new ToolResult($toolCall, 'Test response'));
+
+        $platform = $this->platform(
+            new ToolCallResult([$toolCall]),
+            new ToolCallResult([$toolCall]),
+            new ToolCallResult([$toolCall]),
+            new ToolCallResult([$toolCall]),
+            new ToolCallResult([$toolCall]),
+            new TextResult('Final response'),
+        );
+        $runner = $this->createRunner($platform, $toolbox, maxToolCalls: 3);
+
+        $agent = $this->createMock(AgentInterface::class);
+        $agent
+            ->method('call')
+            ->willReturnCallback(static fn (MessageBag $messages, array $options): ResultInterface => $runner->run($agent, 'gpt-4', $messages, $options));
+
+        $this->expectException(MaxIterationsExceededException::class);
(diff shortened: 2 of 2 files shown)


lazerg:
Missed that #2394 was already open for this, and it covers the streaming path too. Closing in favour of it.

PR #2396, opened 2026-08-11
[Platform] Announce streamed tool calls across bridges

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | n/a
| License       | MIT

Makes the streaming tool call contract uniform across bridges, so consumers can rely on `ToolCallStart`: every call is announced at the position it appears in, its arguments follow as `ToolInputDelta` chunks when the provider streams them, and one terminal `ToolCallComplete` carries all calls.

* **OpenResponses**: `response.output_item.added` -> `ToolCallStart`, `response.function_call_arguments.delta` -> `ToolInputDelta`. Argument deltas address the output item, so announced items are mapped back onto the `call_id` the call is completed with. Reasoning items and tool calls now stream in the order the model generated them, which the Responses API requires when reasoning is replayed with `store => false`.
* **Cohere**: maps the native `tool-call-start` / `tool-call-delta` events.
* **Ollama**: announces calls at the chunk they arrive in. The per-chunk array index was used as the `ToolCall` id, so calls in separate chunks collided on `"0"` — now a stream-wide counter (the id is never sent back to Ollama).
* **Gemini / VertexAI**: announces calls at their `functionCall` part and batches them into a single terminal `ToolCallComplete` instead of one per part. Previously parallel calls produced several `ToolCallComplete` deltas, of which consumers stopping at the first only saw one. The Agent's own `Toolbox\StreamListener` is such a consumer, so parallel calls from Gemini and VertexAI ended up with only the first one executed. Calls without an id (Gemini < 3.0 leaves it empty) are batched but not announced, as the announcement could not be correlated.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -41,6 +41,26 @@ AI Bundle
 Platform
 --------
 
+ * The Gemini and VertexAI bridges now yield a single `ToolCallComplete` delta at the end of a stream,
+   carrying all function calls of the response, instead of one `ToolCallComplete` per streamed
+   `functionCall` part. Consumers collecting tool calls across several deltas can rely on the single
+   batch instead:
+
+   ```diff
+   -$toolCalls = [];
+    foreach ($result->asStream() as $delta) {
+        if ($delta instanceof ToolCallComplete) {
+   -        $toolCalls = array_merge($toolCalls, $delta->getToolCalls());
+   +        $toolCalls = $delta->getToolCalls();
+        }
+    }
+   ```
+
+   Multi-candidate responses (`candidateCount` > 1) are the exception and keep their previous shape: their
+   deltas are wrapped in a `ChoiceDelta`, carrying one `ToolCallComplete` per function call, without
+   `ToolCallStart` and without a terminal batch. A consumer walking `ChoiceDelta::getDeltas()` still has to
+   merge those.
+
  * `Result\Stream\ListenerInterface` gained an `onError()` method, dispatched with the new
    `Result\Stream\ErrorEvent`. Listeners not extending
    `AbstractStreamListener` must add the method:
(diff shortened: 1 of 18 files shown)


UPGRADE.md
```diff
@@ -41,6 +41,21 @@ AI Bundle
 Platform
 --------
 
+ * The Gemini and VertexAI bridges now yield a single `ToolCallComplete` delta at the end of a stream,
+   carrying all function calls of the response, instead of one `ToolCallComplete` per streamed
+   `functionCall` part. Consumers collecting tool calls across several deltas can rely on the single
```
wachterjohannes, line 46: Not true for multi-candidate responses: `Gemini/ResultConverter.php:152` keeps the old shape on purpose, one `ToolCallComplete` per `ToolCallResult` inside a `ChoiceDelta`, no `ToolCallStart`. Someone following this note and switching to `$toolCalls = $delta->getToolCalls()` would lose calls when `candidateCount > 1`. Could you mention the exception?

src/platform/src/Bridge/Gemini/Gemini/ResultConverter.php
```diff
@@ -254,6 +278,8 @@ private function resultToDeltas(ResultInterface $result): \Generator
                 yield new BinaryDelta($result->getContent(), $result->getMimeType());
 
                 return;
+                // Only reached through the multi-candidate path: the single-candidate stream batches
+                // tool calls into one terminal ToolCallComplete instead.
```
wachterjohannes, line 283: Reads as dead code after the `return;` above until you notice it belongs to the `case` below. Better under the `case` line.

wachterjohannes (commented):
This adds `ToolCallStart` to exactly the five bridges that had `ToolCallComplete` without it, so the contract is now complete.

Worth noting louder: our own `Toolbox\StreamListener` stops at the first `ToolCallComplete` and skips the rest, so on `main` Gemini and VertexAI silently drop all but the first parallel tool call. This is closer to a bug fix than the `Feature` label suggests.

Verified locally, all green. One accuracy issue inline.

wachterjohannes (approved):
Both points addressed, and the extra edge-case tests around missing ids are a good addition, the `call_id` without item id case in particular is the one that bites in production. Re-verified locally, all green.

wachterjohannes:
Thank you @chr-hertel.

PR #2398, opened 2026-08-11
[Platform][VertexAi] Normalize nullable fields in structured output schema

### LINKED ISSUE #1999: [Vertex/Gemini] Nullable fields in structured output throw exception
I'll be brief because I have to go, but I wanted to put this out here because I keep forgetting:

- There is a bug for Vertex/Gemini when using structured output
- If a property of the output object/DTO is nullable, the normalized schema will not comply with the Gemini-specific requirement: nullable fields should be explicitly marked `nullable` in the schema
- This is basically near identical to a pre-existing bug in the Gemini toolcall normalizer if I recall correctly

Error thrown:
```Symfony\\AI\\Platform\\Exception\\RuntimeException","message":"Error from Gemini API: \"Invalid JSON payload received. Unknown name \"type\" at 'generation_config.response_schema.properties[0].value.items.properties[5].value': Proto field is not repeating, cannot start list.\"","code":400,"...```

Hope this is enough, if not I'll update next week. 

Thanks and have a good weekend :-)

### PR DESCRIPTION
**Pull request by ousamabenyounes**
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #1999
| License       | MIT

### Problem

Using structured output with the Vertex/Gemini bridge throws as soon as a DTO property is nullable:

```
Error from Gemini API: "Invalid JSON payload received. Unknown name "type" at
'generation_config.response_schema.properties[0].value.items.properties[5].value':
Proto field is not repeating, cannot start list."
```

The JSON Schema factory represents a nullable field with an array-style type (`"type": ["string", "null"]`). VertexAI's OpenAPI-flavored schema parser does not accept an array there — it requires a single `type` plus `"nullable": true`.

`Contract\ToolNormalizer` already performed this conversion for **tool parameter** schemas, but the **structured-output** `responseSchema` in `Gemini\ModelClient` was assigned straight from the `response_format` payload without any normalization, so nullable fields broke.

### Fix

- Extract the nullable-array → `nullable: true` conversion (and the `additionalProperties` stripping) from `ToolNormalizer` into a shared `SchemaNormalizer`.
- `ToolNormalizer` now delegates to it (identical behavior, no duplicated logic).
- `Gemini\ModelClient` runs the structured-output `responseSchema` through the same normalizer, so tool and response schemas stay in sync.

New test `testItNormalizesNullableFieldsInStructuredOutputResponseSchema` asserts the wire `responseSchema` has a single `type` + `nullable: true` (top-level and nested `items`) and no `additionalProperties`.

On the unmodified branch (RED):

```
1) ...VertexAi\Tests\Gemini\ModelClientTest::testItNormalizesNullableFieldsInStructuredOutputResponseSchema
Failed asserting that Array &0 [
    0 => 'string',
    1 => 'null',
] is identical to 'string'.
FAILURES! Tests: 1, Assertions: 2, Failures: 1.
```

With the fix (GREEN), the whole VertexAi bridge suite (including the existing ToolNormalizer nullable tests, proving behavior parity after the extraction):

```
OK (109 tests, 457 assertions)
```

PHPStan and PHP-CS-Fixer are clean on the changed files.

### DIFF
--- a/src/platform/src/Bridge/VertexAi/SchemaNormalizer.php
+++ b/src/platform/src/Bridge/VertexAi/SchemaNormalizer.php
@@ -1,0 +1,63 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Platform\Bridge\VertexAi;
+
+/**
+ * Normalizes a JSON Schema for VertexAI compatibility.
+ *
+ * - Removes 'additionalProperties' (not supported by VertexAI).
+ * - Converts array-style nullable types ['string', 'null'] to
+ *   ['type' => 'string', 'nullable' => true], which VertexAI's OpenAPI-flavored
+ *   schema parser requires. Sent as-is, an array type triggers
+ *   "Proto field is not repeating, cannot start list".
+ *
+ * Used both for tool parameter schemas and for structured-output response
+ * schemas so the two stay in sync.
+ *
+ * @author Ben Younes <benyounes.ousama@gmail.com>
+ */
+final class SchemaNormalizer
+{
+    /**
+     * @template T of array
+     *
+     * @param T $schema
+     *
+     * @return T
+     */
+    public static function normalize(array $schema): array
+    {
+        unset($schema['additionalProperties']);
+
+        if (isset($schema['type']) && \is_array($schema['type'])) {
+            $nullIndex = array_search('null', $schema['type'], true);
+            if (false !== $nullIndex) {
+                $types = $schema['type'];
+                unset($types[$nullIndex]);
+                $types = array_values($types);
+
+                if (1 === \count($types)) {
+                    $schema['type'] = $types[0];
+                    $schema['nullable'] = true;
+                }
+            }
+        }
+
+        foreach ($schema as &$value) {
+            if (\is_array($value)) {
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/platform/src/Bridge/VertexAi/SchemaNormalizer.php [hunk 1]
```diff
@@ -0,0 +1,63 @@
...
+ * - Removes 'additionalProperties' (not supported by VertexAI).
+ * - Converts array-style nullable types ['string', 'null'] to
+ *   ['type' => 'string', 'nullable' => true], which VertexAI's OpenAPI-flavored
+ *   schema parser requires. Sent as-is, an array type triggers
+ *   "Proto field is not repeating, cannot start list".
+ *
+ * Used both for tool parameter schemas and for structured-output response
+ * schemas so the two stay in sync.
+ *
+ * @author Ben Younes <benyounes.ousama@gmail.com>
```
chr-hertel, line 26:
```suggestion
 *
 * @author Ben Younes <benyounes.ousama@gmail.com>
 * @internal
```

chr-hertel (approved):
Reproduced issue and confirmed as fix!

src/platform/src/Bridge/VertexAi/SchemaNormalizer.php
```diff
@@ -0,0 +1,64 @@
...
+ * @internal
```
chr-hertel, line 28:
```suggestion
 * @author Ben Younes <benyounes.ousama@gmail.com>
 *
 * @internal
```

chr-hertel:
Thank you @ousamabenyounes.

PR #2399, opened 2026-08-11
[Store][ChromaDb] Only support TextQuery when an embedding function is configured

### LINKED ISSUE #1601: [Store][ChromaDb] TextQuery requires embedding function configuration
## Description

The ChromaDB Store bridge reports that it supports `TextQuery` via the `supports()` method, but text queries fail at runtime with the error:

```
Error: Call to a member function generate() on null
```

This occurs because ChromaDB requires an embedding function to be configured on the collection to convert query text into vectors for similarity search. Without an embedding function, the `queryTexts` parameter cannot work.

## Expected Behavior

One of the following should happen:

1. **Option A**: The `Store::supports(TextQuery::class)` method should return `false` when no embedding function is configured
2. **Option B**: The Store should be enhanced to accept an optional embedding function parameter and configure it during collection creation

## Current Behavior

- `Store::supports(TextQuery::class)` returns `true` 
- Calling `query(new TextQuery('some text'))` throws: `Error: Call to a member function generate() on null`
- The error occurs in the chromadb-php library when it tries to call `$embeddingFunction->generate()` on a null embedding function

## Steps to Reproduce

```php
use Codewithkyrian\ChromaDB\ChromaDB;
use Symfony\AI\Store\Bridge\ChromaDb\Store;
use Symfony\AI\Store\Query\TextQuery;

$client = ChromaDB::factory()
    ->withHost('http://127.0.0.1')
    ->withPort(8000)
    ->connect();

$store = new Store($client, 'test_collection');
$store->setup(); // Creates collection without embedding function

// This check passes but shouldn't
if ($store->supports(TextQuery::class)) {
    // This fails with "Call to a member function generate() on null"
    $results = $store->query(new TextQuery('search text'));
}
```

## Environment

- **Component**: `symfony/ai-store-chromadb`
- **ChromaDB PHP Client**: `codewithkyrian/chromadb-php`
- **PHP Version**: 8.4

## Root Cause

The `Store::setup()` method creates collections without an embedding function:

```php
public function setup(array $options = []): void
{
    try {
        $this->client->createCollection($this->collectionName);
    } catch (ChromaException $e) {
        throw new RuntimeException(sprintf('Could not create collection "%s".', $this->collectionName), previous: $e);
    }
}
```

According to ChromaDB documentation, when querying with `queryTexts`, the collection's embedding function is used to convert text to vectors. Without it, text queries cannot work.

## Documentation References

- [ChromaDB Embedding Functions](https://docs.trychroma.com/docs/embeddings/embedding-functions)
- [ChromaDB PHP Client](https://github.com/CodeWithKyrian/chromadb-php)
- [ChromaDB Usage Guide](https://docs.trychroma.com/guides)

## Affected Tests

This issue is visible in the integration tests where `testQueryDocumentsWithTextQuery` is skipped with the message: "ChromaDb TextQuery requires an embedding function to be configured."

**File**: `src/store/src/Bridge/ChromaDb/Tests/IntegrationTest.php`

## Proposed Solution

**Option A - Quick Fix**: Update `supports()` to return `false` for `TextQuery::class`

**Option B - Full Solution**: 
1. Add embedding function support to the Store constructor
2. Pass the embedding function when creating collections
3. Update `supports()` to check if an embedding function is configured

Example for Option B:
```php
use Codewithkyrian\ChromaDB\Embeddings\EmbeddingFunction;

public function __construct(
    private readonly Client $client,
    private readonly string $collectionName,
    private readonly ?EmbeddingFunction $embeddingFunction = null,
) {
}

public function supports(string $queryClass): bool
{
    return match ($queryClass) {
        VectorQuery::class => true,
        TextQuery::class => null !== $this->embeddingFunction,
        default => false,
    };
}

public function setup(array $options = []): void
{
    $createOptions = ['name' => $this->collectionName];
    
    if (null !== $this->embeddingFunction) {
        $createOptions['embedding_function'] = $this->embeddingFunction;
    }
    
    try {
        $this->client->createCollection(...$createOptions);
    } catch (ChromaException $e) {
        throw new RuntimeException(sprintf('Could not create collection "%s".', $this->collectionName), previous: $e);
    }
}
```

## Related

This was discovered while adding TextQuery and HybridQuery integration tests to `AbstractStoreIntegrationTestCase` in PR #1570.

ousamabenyounes:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| Docs?         | no
| Issues        | Fix #1601
| License       | MIT

### Problem

`Store::supports(TextQuery::class)` returned `true` unconditionally, but running a text query then failed at runtime:

```
Error: Call to a member function generate() on null
```

ChromaDB embeds a text query client-side, in `prepareEmbeddings()`, through the collection's embedding function. The bridge never configured one, and `getOrCreateCollection()` returns a fresh collection whose embedding function is `null`, so every `TextQuery` fataled before a request was ever sent — while `supports()` claimed it was fine.

### Fix

- Accept an optional `?EmbeddingFunction $embeddingFunction` on `Store` (and `StoreFactory::create`), a backward-compatible trailing nullable argument.
- Forward it to `getOrCreateCollection()` on the text-query path so the query text can actually be embedded.
- `supports(TextQuery::class)` now returns `true` only when an embedding function is configured, so it no longer advertises a capability the store cannot fulfil. `VectorQuery` is unaffected (it carries its own vector).

The embedding function is only passed on the text-query path; `add`/`remove`/`clear`/vector queries provide their own ids/vectors and never call `generate()`, so they are intentionally left untouched.

Marked as a feature as well as a bug fix per review: the new public `$embeddingFunction` parameter makes `TextQuery` usable on this bridge for the first time, hence the `CHANGELOG.md` entry. Exposing it through the bundle's `config/store/chromadb.php` is tracked separately in #2418.

New unit tests assert both the flag and its consequence: `supports(TextQuery::class)` is `false` without an embedding function, and `query(new TextQuery(...))` throws `UnsupportedQueryTypeException` without ever reaching `getOrCreateCollection()`.

On the unmodified base branch, with only the test files of this PR applied (RED):

```
1) ...ChromaDb\Tests\StoreTest::testStoreDoesNotSupportTextQueryWithoutEmbeddingFunction
Failed asserting that true is false.

2) ...ChromaDb\Tests\StoreTest::testQueryWithTextQuery
Expectation failed for method name is "getOrCreateCollection" when invoked 1 time
Parameter 2 for invocation Client::getOrCreateCollection('test-collection', null, null): Collection does not match expected value.
null does not match expected type "object".

3) ...ChromaDb\Tests\StoreTest::testQueryWithTextQueryThrowsWithoutEmbeddingFunction
Failed asserting that exception of type "...\UnsupportedQueryTypeException" is thrown.

FAILURES! Tests: 35, Assertions: 220, Failures: 3.
```

With the fix (GREEN), the full ChromaDb bridge unit suite:

```
OK (35 tests, 233 assertions)
```

The `IntegrationTest::testQueryDocumentsWithTextQuery()` override was dropped: `AbstractStoreIntegrationTestCase` now carries the same "do not skip, dependents would be skipped too" guard, keyed on `supports(TextQuery::class)`, so the bridge no longer needs its own copy. The ChromaDb integration suite was run against a real `chromadb/chroma:0.6.3` container to confirm the removal keeps the dependency chain intact.

PHPStan and PHP-CS-Fixer are clean on the changed files.

### DIFF
--- a/src/store/src/Bridge/ChromaDb/Store.php
+++ b/src/store/src/Bridge/ChromaDb/Store.php
@@ -12,6 +12,7 @@
 namespace Symfony\AI\Store\Bridge\ChromaDb;
 
 use Codewithkyrian\ChromaDB\Client;
+use Codewithkyrian\ChromaDB\Embeddings\EmbeddingFunction;
 use Codewithkyrian\ChromaDB\Exceptions\ChromaException;
 use Codewithkyrian\ChromaDB\Responses\QueryItemsResponse;
 use Symfony\AI\Platform\Vector\Vector;
@@ -36,6 +37,7 @@ final class Store implements ManagedStoreInterface, StoreInterface
     public function __construct(
         private readonly Client $client,
         private readonly string $collectionName,
+        private readonly ?EmbeddingFunction $embeddingFunction = null,
     ) {
     }
 
@@ -119,10 +121,17 @@ final class Store implements ManagedStoreInterface, StoreInterface
 
     public function supports(string $queryClass): bool
     {
-        return \in_array($queryClass, [
-            VectorQuery::class,
-            TextQuery::class,
-        ], true);
+        if (VectorQuery::class === $queryClass) {
+            return true;
+        }
+
+        // A TextQuery is embedded client-side by the collection's embedding function; without one the
+        // ChromaDB client fatals on a null embedding function before a request is ever sent.
+        if (TextQuery::class === $queryClass) {
+            return null !== $this->embeddingFunction;
+        }
+
+        return false;
     }
 
     /**
@@ -189,7 +198,7 @@ final class Store implements ManagedStoreInterface, StoreInterface
     {
         $include = $this->buildInclude($options);
 
-        $collection = $this->client->getOrCreateCollection($this->collectionName);
+        $collection = $this->client->getOrCreateCollection($this->collectionName, embeddingFunction: $this->embeddingFunction);
         $queryResponse = $collection->query(
             queryTexts: $query->getTexts(),
             nResults: $options['limit'] ?? 4,
--- a/src/store/src/Bridge/ChromaDb/Tests/StoreTest.php
+++ b/src/store/src/Bridge/ChromaDb/Tests/StoreTest.php
@@ -12,6 +12,7 @@
 namespace Symfony\AI\Store\Bridge\ChromaDb\Tests;
 
 use Codewithkyrian\ChromaDB\Client;
+use Codewithkyrian\ChromaDB\Embeddings\EmbeddingFunction;
 use Codewithkyrian\ChromaDB\Models\Collection;
 use Codewithkyrian\ChromaDB\Responses\GetItemsResponse;
 use Codewithkyrian\ChromaDB\Responses\QueryItemsResponse;
@@ -22,6 +23,7 @@ use Symfony\AI\Store\Bridge\ChromaDb\StoreFactory;
 use Symfony\AI\Store\Document\Metadata;
 use Symfony\AI\Store\Document\VectorDocument;
 use Symfony\AI\Store\Exception\InvalidArgumentException;
+use Symfony\AI\Store\Exception\UnsupportedQueryTypeException;
 use Symfony\AI\Store\Query\HybridQuery;
 use Symfony\AI\Store\Query\TextQuery;
 use Symfony\AI\Store\Query\VectorQuery;
@@ -828,10 +830,11 @@ final class StoreTest extends TestCase
 
         $collection = $this->createMock(Collection::class);
         $client = $this->createMock(Client::class);
+        $embeddingFunction = $this->createMock(EmbeddingFunction::class);
 
         $client->expects($this->once())
             ->method('getOrCreateCollection')
-            ->with('test-collection')
+            ->with('test-collection', null, $embeddingFunction)
             ->willReturn($collection);
 
         $collection->expects($this->once())
@@ -846,7 +849,7 @@ final class StoreTest extends TestCase
             )
             ->willReturn($queryResponse);
 
-        $store = StoreFactory::create($client, 'test-collection');
+        $store = StoreFactory::create($client, 'test-collection', $embeddingFunction);
         $documents = iterator_to_array($store->query(new TextQuery('search for this text')));
 
         $this->assertCount(1, $documents);
@@ -866,9 +869,29 @@ final class StoreTest extends TestCase
     public function testStoreSupportsTextQuery()
     {
         $client = $this->createMock(Client::class);
-        $store = StoreFactory::create($client, 'test-collection');
+        $store = StoreFactory::create($client, 'test-collection', $this->createMock(EmbeddingFunction::class));
 
         $this->assertTrue($store->supports(TextQuery::class));
+    }
+
+    public function testStoreDoesNotSupportTextQueryWithoutEmbeddingFunction()
+    {
+        $client = $this->createMock(Client::class);
+        $store = StoreFactory::create($client, 'test-collection');
+
+        $this->assertFalse($store->supports(TextQuery::class));
+    }
+
+    public function testQueryWithTextQueryThrowsWithoutEmbeddingFunction()
+    {
+        $client = $this->createMock(Client::class);
+        $client->expects($this->never())->method('getOrCreateCollection');
+
+        $store = StoreFactory::create($client, 'test-collection');
+
+        $this->expectException(UnsupportedQueryTypeException::class);
+
+        $store->query(new TextQuery('search for this text'));
     }
 
     public function testStoreNotSupportsHybridQuery()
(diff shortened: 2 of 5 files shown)


src/store/src/Bridge/ChromaDb/Store.php
```diff
@@ -119,10 +121,17 @@ public function clear(array $options = []): void
...
-        return \in_array($queryClass, [
-            VectorQuery::class,
-            TextQuery::class,
-        ], true);
+        if (VectorQuery::class === $queryClass) {
+            return true;
+        }
+
+        // A TextQuery is embedded by the collection's embedding function; without one
+        // ChromaDB cannot turn the query text into a vector and the request fails.
```
wachterjohannes: Nit: it's not ChromaDB rejecting the request — the embedding happens client-side in `prepareEmbeddings()`, which fatals before anything is sent.

```suggestion
        // A TextQuery is embedded client-side by the collection's embedding function; without one the
        // ChromaDB client fatals on a null embedding function before a request is ever sent.
```

src/store/src/Bridge/ChromaDb/Tests/StoreTest.php
```diff
@@ -866,11 +868,19 @@ public function testStoreSupportsVectorQuery()
...
         $this->assertTrue($store->supports(TextQuery::class));
     }
 
+    public function testStoreDoesNotSupportTextQueryWithoutEmbeddingFunction()
+    {
+        $client = $this->createMock(Client::class);
+        $store = StoreFactory::create($client, 'test-collection');
+
+        $this->assertFalse($store->supports(TextQuery::class));
+    }
```
wachterjohannes, line 883: Could you also cover the consequence, not just the flag?

```php
public function testQueryWithTextQueryThrowsWithoutEmbeddingFunction()
{
    $client = $this->createMock(Client::class);
    $client->expects($this->never())->method('getOrCreateCollection');

    $store = StoreFactory::create($client, 'test-collection');

    $this->expectException(UnsupportedQueryTypeException::class);

    $store->query(new TextQuery('search for this text'));
}
```

wachterjohannes (commented):
Thanks @ousamabenyounes — correct fix. `TextQuery` is embedded client-side, so without an embedding function `prepareEmbeddings()` fatals on null; `supports()` was lying. Two comments inline, plus two asks:

- **Drop `Tests/IntegrationTest::testQueryDocumentsWithTextQuery()`** — that override only existed because `supports()` was wrong. `AbstractStoreIntegrationTestCase` has the same guard now.
- **Label + changelog:** this also adds a public `$embeddingFunction` parameter that makes `TextQuery` usable here for the first time — looks like `Feature` to me, so it needs a CHANGELOG entry.

Follow-up on my side: `config/store/chromadb.php` exposes no embedding function, so bundle users can't reach this yet.

wachterjohannes:
Tracked the bundle side in #2418 — not blocking this PR.

ousamabenyounes:
Thanks for the review @wachterjohannes — all four points are in, plus a rebase onto the current `main`:

- **Comment wording**: applied your suggestion verbatim, the embedding happens client-side in `prepareEmbeddings()` and fatals before anything is sent.
- **`IntegrationTest::testQueryDocumentsWithTextQuery()`**: dropped. `AbstractStoreIntegrationTestCase` carries the same "must not be skipped, dependents would be skipped too" guard keyed on `supports(TextQuery::class)`, so the override is redundant.
- **Consequence test**: added `testQueryWithTextQueryThrowsWithoutEmbeddingFunction()` as you wrote it — `getOrCreateCollection()` is never reached and `UnsupportedQueryTypeException` is thrown.
- **Changelog**: added a `0.13` entry to `src/store/src/Bridge/ChromaDb/CHANGELOG.md` and flipped `New feature?` to `yes` in the description so `validate-changelog-upgrade.php` accepts it. I can't set labels on this repo — could you add `Feature` when you get a chance?

Validation: on `main` with only this PR's test files applied → `FAILURES! Tests: 35, Assertions: 220, Failures: 3` (RED); with the fix → `OK (35 tests, 233 assertions)`, and the whole Store component at `OK (417 tests, 1335 assertions)`. The ChromaDb integration group was run against a real `chromadb/chroma:0.6.3` container and stays at `OK (10 tests, 14 assertions)` after removing the override. PHPStan and PHP-CS-Fixer are clean on the bridge.

wachterjohannes (approved):
All four points in, thanks. `Feature` label added.

Verified independently rather than taking the numbers on trust: test files alone on `main` fail as you describe, and the fix turns them green. I also pulled in `codewithkyrian/chromadb-php` and ran the integration group against a real `chromadb/chroma:0.6.3` container to confirm the removed override really was redundant, and it passes.

One note for anyone repeating that: `examples/compose.yaml` maps chromadb to port 8001 while `IntegrationTest` hardcodes 8000, so the compose service alone will not run that suite.

The bundle side is tracked in #2418.

wachterjohannes:
Thank you @ousamabenyounes.

PR #2404, opened 2026-08-14
[AI Bundle][Platform] Preserve periods in model option names

### LINKED ISSUE #1437: [AI Bundle][Platform] Problems with arrays / periods in model options
In my project I am trying to use a GPT-5 model with the `reasoning.effort` option.

There are currently 3 ways this could be defined but non of them seem work.

1.  
```yaml
ai:
    platform:
        openai:

    agent:
        default:
            platform: 'ai.platform.openai'
            model:
                name: 'gpt-5?reasoning.effort=medium'
```

This ends up turning the model name into `gpt-5?reasoning_effort=medium` changing the `.` into `_`. I've traced this down to https://github.com/symfony/ai/blob/v0.1.0/src/ai-bundle/config/options.php#L362

2.
```yaml
ai:
    platform:
        openai:

    agent:
        default:
            platform: 'ai.platform.openai'
            model:
                name: 'gpt-5' 
                options:
                    reasoning.effort: 'medium'
```
This options does get through the config layer but ends up breaking here:
https://github.com/symfony/ai/blob/v0.1.0/src/platform/src/ModelCatalog/AbstractModelCatalog.php#L87

With basically the same problem as point 1 but in a different place

3.
```yaml
ai:
    platform:
        openai:

    agent:
        default:
            platform: 'ai.platform.openai'
            model:
                name: 'gpt-5' 
                options:
                    reasoning:
                        effort: 'medium'
```
This version seems to break at line: https://github.com/symfony/ai/blob/v0.1.0/src/ai-bundle/config/options.php#L378 
Which turn the model name into: `gpt-5?reasoning%5Beffort%5D=medium`, and then causes the error: `You have requested a non-existent parameter "5Beffort".`

saifulferoz:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #1437
| License       | MIT

`parse_str()` replaces periods in parameter *names* with underscores, so an
option like `reasoning.effort` arrives as `reasoning_effort`. A model string is
parsed in two places, and both used it:

* `src/ai-bundle/config/options.php`, when the model name carries a query string
  (twice, once for agents and once for vectorizers);
* `AbstractModelCatalog::parseModelName()`, when the normalized model string is
  turned back into options.

That covers both ways of configuring such an option, which is why neither of the
first two cases in the issue works:

```yaml
# option 1 — mangled by the bundle, then again by the catalog
model: 'gpt-5?reasoning.effort=medium'

# option 2 — survives http_build_query(), then mangled by the catalog
model:
    name: 'gpt-5'
    options:
        reasoning.effort: 'medium'
```

`http_build_query()` is not part of the problem: it emits `reasoning.effort=medium`
unchanged, and only the `parse_str()` on the way back folds the period. So both
paths needed the same single fix rather than a change to how options are encoded.

The new `ModelOptionsParser` masks periods in the name of each pair, runs
`parse_str()`, and restores them. Masking rather than hand-rolling the parse keeps
everything else `parse_str()` does, in particular the nested array notation that
`test-model?options[metadata][version]=1` relies on, and it also covers `%2E`,
which `parse_str()` decodes before folding.

Per the discussion on #1438, this fixes options 1 and 2 only. Option 3 (a nested
`options` array in YAML) is left alone.

Reported by @TomvdPeet in #1437, who also traced both call sites.

### DIFF
--- a/src/ai-bundle/config/options.php
+++ b/src/ai-bundle/config/options.php
@@ -14,6 +14,7 @@ namespace Symfony\Component\Config\Definition\Configurator;
 use Symfony\AI\AiBundle\Exception\InvalidArgumentException;
 use Symfony\AI\Platform\Capability;
 use Symfony\AI\Platform\Model;
+use Symfony\AI\Platform\ModelOptionsParser;
 use Symfony\AI\Platform\PlatformInterface;
 use Symfony\AI\Store\Document\VectorizerInterface;
 use Symfony\AI\Store\StoreInterface;
@@ -158,8 +159,7 @@ return static function (DefinitionConfigurator $configurator): void {
                                             if ([] !== $options) {
                                                 throw new InvalidConfigurationException('Cannot use both query parameters in model name and options array.');
                                             }
-                                            parse_str($parsed['query'], $existingOptions);
-                                            $options = $existingOptions;
+                                            $options = ModelOptionsParser::parse($parsed['query']);
                                         }
                                     }
 
@@ -510,8 +510,7 @@ return static function (DefinitionConfigurator $configurator): void {
                                             if ([] !== $options) {
                                                 throw new InvalidConfigurationException('Cannot use both query parameters in model name and options array.');
                                             }
-                                            parse_str($parsed['query'], $existingOptions);
-                                            $options = $existingOptions;
+                                            $options = ModelOptionsParser::parse($parsed['query']);
                                         }
                                     }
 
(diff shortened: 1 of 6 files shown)


wachterjohannes (approved):

wachterjohannes:
Tested this locally and the fix works end to end: with `gpt-5?reasoning.effort=medium` the option now reaches the request payload unchanged, and both option 1 and option 2 from #1437 are covered. I also probed the parser against `parse_str()` for nested arrays, `%2E`, periods in values and repeated names, no behaviour changes beyond the intended one.

One note for later: `ModelOptionsParser` lands as new public API in a bug fix PR, and it really only exists for the two call sites. It should probably carry `@internal`. Nothing to change here, we will add that in a follow-up PR right after merge so this one stays a clean bug fix.

wachterjohannes:
Thank you @saifulferoz.

wachterjohannes:
Thank you @saifulferoz.

PR #2411, opened 2026-08-16
[Platform] Add `name` argument to `#[Schema]` to rename Structured Output JSON keys

### LINKED ISSUE #2386: [Platform] Allow renaming Structured Output JSON keys (e.g. Schema name / SerializedName)
### Description

Structured Output currently uses the **PHP property name** as the JSON Schema key, and the dedicated `StructuredOutput\Serializer` intentionally ignores Serializer naming metadata (`#[SerializedName]`, name converters) so schema generation and hydration stay aligned (see #937 / #940).

That is a solid default, but there is no first-class way to expose a **different JSON key** to the model while keeping idiomatic PHP property names.

Example use case: PHP / domain code wants `restBetweenRounds`, but persisted JSON or an external contract already uses `rest_between_rounds`. Today the only options are:
- rename the PHP property to match the JSON key (awkward in PHP), or
- maintain custom mapping outside Symfony AI (easy to drift from the generated schema).

`#[Schema]` also has no `name` (or equivalent) argument today.

### Proposed direction

One of (or a combination of):

1. **`#[Schema(name: 'rest_between_rounds')]`** — used by the JsonSchema Factory as the property key in `properties` / `required`, and respected by the Structured Output serializer when denormalizing.
2. **Honor `#[SerializedName]`** in both schema generation and the dedicated Structured Output serializer (same attribute as Symfony Serializer).
3. **Optional name converter / service** injected into the JsonSchema Factory + Structured Output serializer (opt-in, default = current behavior: PHP property name).

Whatever the API, schema keys and hydration must stay in sync (the #937 lesson).

### Benefits

- Idiomatic camelCase PHP DTOs
- Stable/snake_case (or legacy) JSON contracts for models / storage when needed
- No custom mapping layers that can diverge from the generated schema

### Example (illustrative)

```php
final readonly class CircuitMetadata
{
    public function __construct(
        #[Schema(
            name: 'rest_between_rounds',
            description: 'Rest between rounds in seconds',
            minimum: 0,
        )]
        public int $restBetweenRounds,
    ) {}
}
```

Generated schema / model JSON would use `rest_between_rounds`; PHP keeps `$restBetweenRounds`.

Happy to discuss whether `Schema(name: …)`, `SerializedName`, or an opt-in name converter is the preferred Symfony AI approach.

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | Fix #2386
| License       | MIT

Adds an optional `name` argument to `#[Schema]`, so a property or parameter can be exposed under a different JSON key than its PHP name.

```php
final class CircuitMetadata
{
    public function __construct(
        #[Schema(description: 'Rest between rounds in seconds', minimum: 0, name: 'rest_between_rounds')]
        public readonly int $restBetweenRounds,
    ) {
    }
}
```

The generated schema uses `rest_between_rounds` in `properties` and `required`, and hydration reads that same key back into `$restBetweenRounds`. Both sides resolve the key through one `SchemaNameResolver`, so schema generation and deserialization cannot drift apart, which was the lesson from #937. Tool calls behave the same way: `ToolCallArgumentResolver` looks the argument up by its schema key and passes it to the original PHP parameter.

Serializer metadata like `#[SerializedName]` stays ignored, as before. `name` is the only argument allowed next to `ref`, because it renames the key instead of describing the value.

This is option 1 from the issue; options 2 and 3 are deliberately left out.


wachterjohannes:
@chr-hertel mind taking this one? New public attribute API, `#[Schema(name:)]` renames a Structured Output key, and it now rejects colliding keys within a class, which silently dropped properties before.

One case i'd rather decide with you: two tool method parameters renamed to the same key still yield a single property, and `ToolCallArgumentResolver` feeds one value into both. `forClass()` cannot see it since method params are not class properties. `Describer::describeObject()` would catch it and the other three cases at once, so that might be the better home for the whole guard.

chr-hertel:
@wachterjohannes waiting on feedback by @BriceFab in the issue - have you been able to identify the use-case here?

wachterjohannes:
> @wachterjohannes waiting on feedback by @BriceFab in the issue - have you been able to identify the use-case here?

@chr-hertel Not a strong use-case, more a nice-to-have. If you already have existing processes, skills or naming conventions built around specific field names, being able to rename the model's output to match without touching your PHP naming is useful.

You're right that this is normally internal, user-land never sees the schema either way, neither in tool calls nor structured output. Still, I see it similar to Doctrine's `#[Column]`: you don't strictly need it most of the time, but it's there for exactly this kind of case. Nothing dramatic here, not blocking anything - so let us wait for the response of @BriceFab.

BriceFab:
> > @wachterjohannes waiting on feedback by @BriceFab in the issue - have you been able to identify the use-case here?
> 
> @chr-hertel Not a strong use-case, more a nice-to-have. If you already have existing processes, skills or naming conventions built around specific field names, being able to rename the model's output to match without touching your PHP naming is useful.
> 
> You're right that this is normally internal, user-land never sees the schema either way, neither in tool calls nor structured output. Still, I see it similar to Doctrine's `#[Column]`: you don't strictly need it most of the time, but it's there for exactly this kind of case. Nothing dramatic here, not blocking anything - so let us wait for the response of @BriceFab.

@wachterjohannes Yes — after checking our FitMetrics backend, this PR does address the problem we originally had.

We had persisted workout metadata using snake_case keys like `rest_between_rounds` / `rest_after_block`, while our PHP Structured Output DTO uses `$restBetweenRounds` / `$restAfterBlock`. Because Symfony AI could not rename schema keys, we migrated the persisted JSON to camelCase just to keep schema generation and hydration aligned.

With `#[Schema(name: ...)]`, we could have kept the existing JSON contract while preserving idiomatic PHP property names. So this PR fixes that use case for us.

Thank you!

wachterjohannes:
@BriceFab thanks for your message! thought already something like that. 

@chr-hertel what do you mean? have you already reviewed it?

chr-hertel:
@BriceFab So you're calling `['response_format' => $loadPersistedMetadata]` or sth?

and I don't agree on the doctrine comparison: the use case of having a potentially pre-existing schema with a real DB engine vs. something that we create and consume doesn't have the same ownership/change-scenarios.

and FTR, i'm just challenging this because i still can't follow the use-case yet, so please provide some user land code to reproduce the issue you're running into. Or different, how or why did you end up with a different schema as input - take it as me doing user research here 😆

BriceFab:
@chr-hertel 

No, we're not doing something like `['response_format' => $loadPersistedMetadata]`.

I checked the migration again and I think the missing context is that the JSON contract existed **before** we adopted Symfony AI Structured Output.

We previously had a manually maintained JSON Schema using snake_case, for example:

```json
{
    "duration_weeks": 4,
    "workouts": [
        {
            "blocks": [
                {
                    "metadata": {
                        "rounds": 3,
                        "rest_between_rounds": 60,
                        "rest_after_block": 90
                    },
                    "exercises_config": []
                }
            ]
        }
    ]
}
```

That contract was already used by our generator/import-export flow, and some of the metadata was persisted with those keys.

When we migrated the schema to `ResponseFormatFactory` / Structured Output DTOs, our PHP naturally looked like this:

```php
final readonly class CircuitMetadataPayload
{
    public function __construct(
        public int $rounds,
        public int $restBetweenRounds,
        public int $restAfterBlock,
    ) {}
}
```

Symfony AI therefore generated `restBetweenRounds` / `restAfterBlock`, and the Structured Output serializer expects those same keys.

Since there was no way to explicitly map the schema key, we ended up changing our AI JSON contract to camelCase and even migrating the persisted metadata:

```sql
rest_between_rounds -> restBetweenRounds
rest_after_block    -> restAfterBlock
```

We then kept our public/mobile API contract separate, where it still exposes snake_case using the regular Symfony Serializer:

```php
#[SerializedName('rest_between_rounds')]
public ?int $restBetweenRounds = null;
```

So the issue isn't really that persisted data is being passed as the response format. It's that an application may already have an established JSON contract when adopting Symfony AI, and currently the PHP property name effectively becomes part of that external/model-facing contract.

With the proposed API we could have kept the existing contract without migrating unrelated persistence/import-export code:

```php
#[Schema(name: 'rest_between_rounds')]
public int $restBetweenRounds;
```

That's the concrete FitMetrics use case that originally led me to open the issue.

I agree this is not required when Symfony AI owns the JSON contract from day one; in our case it mattered because we were migrating an existing contract to Structured Output.

PR #2412, opened 2026-08-16
[MCP Bundle] Support multiple servers and clients

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | -
| License       | MIT

Turns the MCP Bundle config terminology around and lifts the one-server limit.

`servers:` now configures the MCP servers the application **exposes**, `clients:` the MCP
clients it uses to **reach** remote servers. Both take named entries. Previously
`client_transports:` configured the server's own transports (despite the name), and the
documented `servers:` key was an unimplemented placeholder for the client side.

### Multiple servers

Each entry gets its own registry, builder, session store, route (`_mcp_endpoint_<name>`)
and identity. A server exposes only what its `registry:` names — service ids, class names,
namespace prefixes or `*` — given either as one list covering every kind of capability or
as a map narrowing each kind:

```yaml
mcp:
    servers:
        public:
            http: { path: /mcp }
            registry:
                tools: ['App\Mcp\Public\']
                resources: ['*']
        editors:
            http: { path: /mcp/editors }
            registry: '*'
        reporting:
            http: { path: /mcp/reporting }
            registry: ['App\Mcp\Reporting\']   # every kind, from one namespace
```

Access control stays plain Symfony security against those paths.

### Multiple clients

A client is a named group of remote server connections:

```yaml
mcp:
    clients:
        research:
            servers:
                github:     { transport: http, url: '…' }
                filesystem: { transport: stdio, command: ['npx', '-y', '@mcp/fs', '/tmp'] }
        simple:
            servers:
                github:     { transport: http, url: '…' }
```

Connections are `McpClientInterface` / `ServerConnectionInterface` services. The bundle owns
the lifecycle: a connection opens on first use and closes on `kernel.reset`, so callers never
touch `connect()` and a worker never carries a stale stdio child process between messages.

`debug:mcp` now covers both sides of the bundle — the configured servers and their
capabilities, plus `--clients` / `--client=<name>` for what the clients reach. `mcp:server`
gained a name argument.

### Notes

- **BC break** — every server option moved under `servers.<name>:`, documented in `UPGRADE.md`.
  Capabilities are no longer exposed implicitly; `registry: '*'` is the 1:1 migration.
- Session storage is isolated per server by default and a shared store is rejected at compile
  time: session ids are not namespaced by server, so a session minted on a public server would
  otherwise be accepted by a firewalled one.
- Fixes a pre-existing bug where icons reached clients as `mime_type` instead of `mimeType`.
- A pattern matching nothing **at all** on its server fails at compile time — practically always
  a typo. Matching only some kinds is fine, which is what makes `registry: ['App\Mcp\']` usable:
  few applications carry all five kinds under one namespace. An attributed service that no server
  lists is simply not exposed, and `debug:mcp` reports it.

171 tests / 514 assertions, PHPStan and CS clean. Verified end-to-end on the demo app with two
servers (distinct routes, identities and tool sets, cross-server session replay rejected) and a
client reaching both endpoints — plus, on a stacked branch, against the public third-party MCP
server at `subwayinfo.nyc`, which exercised the HTTP transport against SSE-framed responses.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -38,6 +38,104 @@ AI Bundle
 
    The default (`false`, tool messages are preserved) is unchanged.
 
+MCP Bundle
+----------
+
+ * The bundle now supports several MCP servers per application, so every server-related option moved
+   under a named entry in `mcp.servers`. **Capabilities are no longer exposed automatically**: each
+   server declares what it exposes, and `['*']` is the explicit "everything of that kind" — that is the
+   1:1 migration of the previous behavior:
+
+   ```diff
+    # config/packages/mcp.yaml
+    mcp:
+   -    app: 'my-app'
+   -    version: '1.0.0'
+   -    description: 'My MCP server'
+   -    instructions: 'Use this server for time calculations.'
+   -    pagination_limit: 50
+   -    client_transports:
+   -        stdio: true
+   -        http: true
+   -    http:
+   -        path: /_mcp
+   -        allowed_hosts: ['example.com']
+   -        session:
+   -            store: cache
+   -    apps:
+   -        enabled: true
+   +    servers:
+   +        default:
+   +            name: 'my-app'
+   +            version: '1.0.0'
+   +            description: 'My MCP server'
+   +            instructions: 'Use this server for time calculations.'
+   +            pagination_limit: 50
+   +            transports:
+   +                stdio: true
+   +                http: true
+   +            http:
+   +                path: /_mcp
+   +                allowed_hosts: ['example.com']
+   +            session:
+   +                store: cache
+   +            registry: '*'
+   ```
+
+   Option by option: `app` became `servers.<name>.name` (defaulting to the configuration key),
+   `client_transports` became `servers.<name>.transports` (its `http` option now defaults to `true`),
+   `http.session` moved up to `servers.<name>.session` (it was never HTTP-specific — a STDIO server
+   uses it too), and `apps.enabled` was replaced by the `servers.<name>.registry.apps` list
+   (`enabled: false` → omit it, the auto/`true` modes → `apps: ['*']`).
+
+   `http.path` now defaults to `/mcp/<name>` instead of `/_mcp`. The default is always derived from the
+   server name, so that adding a second server can never move an existing endpoint; set `http.path`
+   explicitly to keep the old URL.
+
+   The required `registry:` option decides what a server exposes. Give it one list to cover every kind
+   of capability, or a map to narrow each kind separately. Entries match a service id, a class name, or
+   a namespace prefix (written with a trailing backslash):
+
+   ```yaml
+   mcp:
+       servers:
+           public:
+               http: { path: /mcp }
+               registry:
+                   tools: ['App\Mcp\Public\']
+                   resources: ['*']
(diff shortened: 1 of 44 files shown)


OskarStark (approved):

Toflar:
Thanks for the work @chr-hertel! ❤️ Just my 2c: Maybe instead of class filtering for tools, it would be better to do that on some tool tagging? So one would do

```php
#[McpTool(name: 'current-time', tags: ['foo', 'bar'])]
```

```yaml
mcp:
    servers:
        public:
            http: { path: /mcp }
            tools: ['foo']
```

This would allow grouping tools coming from different PHP vendor namespaces (think extensibility) and also, devs could add a tag to a tool via compiler pass etc. 😊

OskarStark:
I like the idea, but no one guarantees they are unique from a vendor. I think we could do a follow up PR in this case. I would like to test this in a real world project with the current status quo

wachterjohannes:
Agreed with @OskarStark, and the current config shape backs it up: `tools:` is already a heterogeneous scalar list taking service ids, FQCNs, namespace prefixes and `*`. A tag form can be added later purely additively, say `#foo` or `tag:foo`, without changing what any existing value means and without a second BC break. So deferring costs nothing and locks nothing in.

The uniqueness point is the substantive one. DI tags are global strings, so `foo` from two vendors collide, while FQCN and namespace-prefix filtering is unambiguous by construction. That is exactly why a tag design needs a naming convention of its own, and why it should not be bolted onto a 45-file PR that already carries a BC break.

@Toflar the grouping-across-vendors use case is real though, worth an issue so it does not get lost.

wachterjohannes:
Tracked as #2426 so it does not get lost.

chr-hertel:
@Toflar that's not that straight forward unfortunately since we're using the `mcp/sdk`'s attributes here - which still makes sense to me, especially when mounting non-symfony mcp server packages into a kernel/container. there is a concept of a flexible `meta` collection - that's something we could think of reusing here. but def a follow up - this thing is too big anyways :D

chr-hertel:
Tested again, even with the latest spec version, and it looks good. the only thing i'm a bit annoyed about is this duplication:

image

i currently lean to simplifying this to:
```yaml
    registry: ['App\Mcp\']
```

opinions?

OskarStark:
Would be good if we can use the simple way you proposed @chr-hertel

wachterjohannes:
@chr-hertel easier is always better :) would be it be instead or additionally?

chr-hertel:
> @chr-hertel easier is always better :) would be it be instead or additionally? 

It's both with a nested syntax:
```yaml
mcp:
    servers:
        public:
            http: { path: /mcp }
            registry:
                tools: ['App\Mcp\Public\']
                resources: ['*']
        editors:
            http: { path: /mcp/editors }
            registry: '*'
        reporting:
            http: { path: /mcp/reporting }
            registry: ['App\Mcp\Reporting\']
```

wachterjohannes:
@chr-hertel thats awesome! lets do it like that

wachterjohannes:
@chr-hertel 🚀

PR #2416, opened 2026-08-17
[Platform] Mark ModelOptionsParser as internal

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | -
| License       | MIT

Follow-up to #2404. `ModelOptionsParser` was introduced there as a helper for the two places that parse a model query string, but it landed as public API. Marking it `@internal` keeps it out of the BC surface. It is not released yet, so this is not a BC break.

It is used from `src/ai-bundle/config/options.php` and therefore crosses a package boundary, same as Symfony does with `VarExporter\Internal`.


wachterjohannes:
@chr-hertel FYI ...

PR #2417, opened 2026-08-17
[Platform] Add support for `Eden AI` gateway

welcoMattic:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | -
| License       | MIT

[Eden AI](https://www.edenai.co/) is an AI gateway exposing hundreds of models from many
providers behind a single API, using `provider/model` identifiers. This adds a
`symfony/ai-eden-ai-platform` bridge covering both halves of its v3 API.

### OpenAI-compatible endpoints

Chat completions (`/v3/chat/completions`, including streaming and tool calling) and
embeddings (`/v3/embeddings`) reuse the Generic bridge:

```php
use Symfony\AI\Platform\Bridge\EdenAi\Factory;

$messages = new MessageBag(Message::ofUser('What is the Symfony framework?'));
echo $platform->invoke('openai/gpt-4o-mini', $messages)->asText();

$vectors = $platform->invoke('openai/text-embedding-3-small', 'Some text')->asVectors();
```

### Expert models

The `/v3/universal-ai` endpoints expose non-LLM features, addressed as
`feature/subfeature/provider[/model]`: OCR, document parsing (invoices, resumes, identity
documents), text-to-speech, speech-to-text, image analysis (object detection, explicit
content) and image generation.

```php
use Symfony\AI\Platform\Bridge\EdenAi\Ocr\Result\OcrResult;
use Symfony\AI\Platform\Message\Content\ImageUrl;

// OCR: raw text and bounding boxes
$result = $platform->invoke('ocr/ocr/google', new ImageUrl('https://example.com/scan.jpg'), [
    'language' => 'en',
]);

$ocr = $result->asObject();
\assert($ocr instanceof OcrResult);
echo $ocr->getText();

// Document parsing: structured data out of an invoice
$parsing = $platform->invoke('ocr/financial_parser/affinda', new DocumentUrl('https://example.com/invoice.pdf'), [
    'language' => 'en',
])->asObject();

// Text-to-speech
$platform->invoke('audio/tts/amazon/neural', 'Welcome to Symfony AI!')->asFile('welcome.mp3');

// Speech-to-text: the async job is polled transparently, and a local file is
// uploaded through /v3/upload first
echo $platform->invoke('audio/speech_to_text_async/deepgram', Audio::fromFile('./audio.mp3'))->asText();

// Image generation
$platform->invoke('image/generation/stabilityai', 'A red apple on a white table')->asFile('apple.png');
```

Expert models accept their input as a direct file URL, as a file ID, or as `Audio`,
`Document` and `Image` content, which is transparently uploaded through `/v3/upload`
beforehand. Speech-to-text runs on `/v3/universal-ai/async`, whose job is polled until it
reaches a terminal state. Options that are not root-level request fields (`fallbacks`,
`provider_params`, `show_original_response`, `webhook_receiver`,
`user_webhook_parameters`) are forwarded inside the `input` object, so per-feature
parameters like `language`, `document_type` or `voice` are passed as invocation options.

### Notes

The catalog ships a curated subset of the available models; any other one can be
registered through the `$additionalModels` constructor argument. Capabilities are derived
from the metadata `/v3/models` exposes per entry: `supports_response_schema` maps to
`OUTPUT_STRUCTURED`, `supports_function_calling` to `TOOL_CALLING`, and the
`input_modalities` to the matching `INPUT_*` cases.

Gateways like Eden AI answer authentication failures with `{"detail": "..."}` rather than
an OpenAI-style error payload. The Generic completions converter already threw an
`AuthenticationException` on those, it just could not find the reason in that shape and
fell back to its generic `Authentication failed.` message. It now reads both shapes, so
the gateway's own wording reaches the exception.

Endpoints, payload keys, the upload contract and every output shape were checked against
the live API, and the thirteen examples under `examples/edenai/` were all run against it
and recorded as replay cassettes under `examples/tests/fixtures/edenai/`, so every converter
is covered offline by `ExamplesReplayTest`.
One discrepancy is worth flagging for future readers: the `output_schema` advertised by
`GET /v3/info` for `image/generation` describes a single item without the enclosing
`items` list the endpoint really returns, which the converter documents.

Bridge documentation is in `docs/components/platform/edenai.rst`, and 115 tests cover the model
catalog, the contract normalizers, both model clients and every result converter.

cc @Guikingone


src/platform/src/Bridge/EdenAi/SpeechToText/ModelClient.php
```diff
@@ -0,0 +1,142 @@
...
+        $response = $this->httpClient->request('POST', $this->baseUrl.'/v3/universal-ai/async', [
+            'auth_bearer' => $this->apiKey,
+            'headers' => ['Content-Type' => 'application/json'],
+            'json' => $this->createRequestBody($model, $payload, $options),
+        ]);
+
+        $data = $this->decode($response);
+
+        for ($poll = 0; $this->isPending($data); ++$poll) {
+            if ($poll >= $this->maxPolls) {
```
wachterjohannes, line 92: `decode()` forces the response, so the whole poll loop runs inside `Platform::invoke()`, not when the result is read. Measured with a counting clock:

```
invoke() returned after 3 sleeps, 4 http calls
reading the result caused 0 further sleeps
```

With the defaults that is up to `120 × 1s` inside `invoke()`, and the returned `DeferredResult` is not deferred for this model type.

#2408 is open and adds exactly this as a shared concern (`Job\JobClientInterface`, `JobRunner`, `JobTimeoutException`), with resumability this loop gives up. Not to be solved here, but worth ordering: adopting it once #2408 lands would drop the loop entirely.

src/platform/src/Bridge/Generic/Completions/ResultConverter.php
```diff
@@ -48,7 +48,8 @@ public function convert(RawResultInterface|RawHttpResult $result, array $options
         $response = $result->getObject();
 
         if (401 === $response->getStatusCode()) {
-            $errorMessage = json_decode($response->getContent(false), true)['error']['message'] ?? 'Authentication failed.';
+            $error = json_decode($response->getContent(false), true);
+            $errorMessage = $error['error']['message'] ?? $error['detail'] ?? 'Authentication failed.';
```
wachterjohannes, line 52: The change and its test are right, but I could not reproduce the reason given in the description ("PHP warnings instead of throwing an `AuthenticationException`"). With a `detail`-only body, `error` as a string, a list, `null`, invalid JSON or an empty body, the old expression warns in none of them, `??` suppresses the offset notices, and it threw `AuthenticationException` every time, just with the generic message.

So this improves the message rather than fixing a warning. Worth rewording, since the description lands verbatim in the merge commit.

wachterjohannes (commented):
Structurally this holds up: scaffolding, deptrac, splitsh and the bundle wiring all match the sibling bridges, and the two things that looked like omissions (no `base_url` in the bundle config, two catalogs) are what the siblings do. Details inline.

* Record and replay is the direction we want to take bridges, so that a converter is pinned against the provider's real shapes rather than hand-written fixtures. This bridge is the strongest case for it so far, eight converters at once, and you already ran every example against the live API, so `examples/runner --record edenai` would come almost for free. `ExamplesReplayTest` iterates over the cassettes, so without one the bridge stays silently uncovered.

On scope I am deliberately not deciding: a gateway with eight expert-feature families and its own result types is a maintenance call, so I would like @chr-hertel to weigh in before this moves.

Verified locally, all green.

examples/composer.json
```diff
@@ -31,6 +31,7 @@
         "symfony/ai-deepgram-platform": "^0.12",
         "symfony/ai-docker-model-runner-platform": "^0.12",
         "symfony/ai-doctrine-message-store": "^0.12",
+        "symfony/ai-eden-ai-platform": "^0.12",
```
chr-hertel: We'd need `dev-main` plus a [path repo here for making](https://github.com/symfony/ai/blob/main/CONTRIBUTING.md#adding-a-new-package-or-bridge) it work

```suggestion
        "symfony/ai-eden-ai-platform": "dev-main",
```

docs/components/platform.rst
```diff
@@ -1135,11 +1138,120 @@ completions, it is invoked with a single document content object - a
 The result exposes every ``Page`` with its markdown, dimensions, extracted layout images
 (with bounding boxes) and optional annotations.
 
+Eden AI exposes OCR and document parsing (invoices, resumes, identity documents) from multiple
```
chr-hertel: please move the changes from here and below to a provider specific rst file like `docs/components/platform/edenai.rst`

src/platform/src/Bridge/EdenAi/SpeechToText/ModelClient.php
chr-hertel: i guess that's also the reason for failing on my end:

Image

so let me pivot reviewing to 2408 to get that in :)

welcoMattic: Rebased, should be resolved?

wachterjohannes: No, and nothing for you to do here. This one waits on #2408, which is still open and waiting on @chr-hertel's review.

chr-hertel:
@welcoMattic i rebased this and adopted the async job capabilities from #2408 - somehow still not all example run on my side?

can you please give it a final test before we bring it in?

welcoMattic:
@chr-hertel I've fixed the branch and re-run examples scripts against the real Eden AI API.
All good on my side.

chr-hertel (approved):

chr-hertel:
Thank you @welcoMattic.

PR #2419, opened 2026-08-17
[Skills] Add pr-triage and pr-review agent skills

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | no
| Issues        | -
| License       | MIT

Adds two agent skills covering the maintainer review loop. They complement the existing `run-examples` skill, which already describes itself as "a phase-two integration check **after** a code review (e.g. via `/pr-review`)" — that counterpart did not exist until now.

No component code is touched; this is `.claude/skills/` tooling only.

### `pr-triage`

Surveys the open PR backlog (currently ~90) and produces a ranked review queue instead of a flat dump.

- Buckets every PR by draft state, review decision, mergeability and authorship, so the output is what is actually reviewable right now — usually a handful.
- Fetches CI status per-PR for the reviewable subset only. Requesting `statusCheckRollup` inside a 100-PR `gh pr list` reliably 502s/504s on this repo, so the skill splits the calls.
- Distinguishes checks that go red for reasons unrelated to the diff (`Validate Changelog & Upgrade Entries`, `PHPStan / Demo`, `Integration / *`, `Unit / Store / S3Vectors`, `Fabbot`) from failures that implicate the change.
- Rates **review complexity** from blast radius, BC surface, semantic depth, verification cost and test signal rather than from diff size, with a calibration table — a 1000-line cookbook PR is Low, a 70-line serialization change can be High.

Read-only: it never comments, labels or merges.

### `pr-review`

Reviews a single PR locally, end to end.

- **Read beyond the diff.** Sibling implementations, the interface, every caller, and the vendor library when the change depends on third-party behaviour. Includes the questions that reliably find something here — does an existing helper already do this, does the PR leave behind a workaround it obsoletes, is a new component capability reachable from the AI Bundle.
- **Verify behavioural claims by running them.** Apply the patch, run a before/after reproduction, restore the tree. A finding that has not been verified is phrased as a question.
- **Check the repo conventions** the changelog workflow enforces: bug-fix-only PRs must not touch `CHANGELOG.md`/`UPGRADE.md`, features need an entry in the unreleased section, `BC Break` label ⇄ `UPGRADE.md` entry.
- **Draft, show, post.** Deep findings go in the chat answer; the posted review body is one to three sentences and each inline comment is a single ask plus its evidence. Nothing is posted without explicit approval.

It ships `references/inline-review.md` (anchoring rules, suggestion-block mechanics, the common 422 causes) and `scripts/build-review.sh`, which resolves the head SHA and validates every anchor against the PR's actual diff hunks before emitting a payload:

```
$ build-review.sh --pr 2368 --body body.md \
    --comment 'src/platform/src/Bridge/OpenRouter/ModelApiCatalog.php:200:c1.md'
error: anchor src/platform/src/Bridge/OpenRouter/ModelApiCatalog.php:200 is not inside a single diff hunk.
  valid ranges for this file:
    27 33
    64 70
    129 135
```

The helper was checked against two reviews posted from this workflow (#2368, #2265) and reproduces both payloads byte-identically.

Both skills carry `evals/evals.json` in the same shape as the existing skills.

### DIFF
--- a/.claude/skills/issue-triage/SKILL.md
+++ b/.claude/skills/issue-triage/SKILL.md
@@ -1,0 +1,164 @@
+---
+name: issue-triage
+description: >
+  Surveys the open GitHub issues on the Symfony AI monorepo and finds which
+  ones are answerable or fixable right now. Distinct from pr-triage/pr-review,
+  which cover pull requests. Buckets issues by whether the maintainer can
+  resolve them by reading current code: already fixed by a merged PR or an
+  independent rewrite, a factual question the code can answer directly, a
+  real unfixed bug, or a design decision needing the maintainer's opinion.
+  For issues already fixed elsewhere, drafts a short, friendly closing
+  comment linking the PR or commit that fixed it, and closes the issue after
+  explicit approval. Use when the user asks to triage issues, "check the
+  issues", "any issues we could close or answer", or "/issue-triage". Never
+  comments, closes or answers without showing the draft first and getting an
+  explicit yes, one issue at a time.
+---
+
+# Issue Triage
+
+You help a maintainer of the Symfony AI monorepo find which open issues can
+actually be moved right now, either closed because the code has already
+moved past them, or answered because the current code settles the question,
+or fixed with a small correct change. Most open issues are none of those:
+they're RFCs waiting on a contributor, or discussions waiting on the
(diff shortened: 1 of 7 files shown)


OskarStark:
Can we get something from the Symfony/Symfony skills what could help here?

PR #2423, opened 2026-08-20
[Platform][Mistral] Fix streaming/buffered crash on reasoning (thinking) content

### LINKED ISSUE #2422: [Platform][Mistral] Streaming with reasoning_effort crashes: TextDelta expects string, array given
### Describe the problem

Streaming a chat completion from a Mistral reasoning model with `reasoning_effort: 'high'` crashes. With reasoning enabled, Mistral returns `choices[].delta.content` as an **array** of thinking/text chunks (https://docs.mistral.ai/capabilities/reasoning/), but the Mistral bridge reuses `CompletionsConversionTrait`, which assumes the OpenAI schema where `content` is always a string. `TextDelta`/`TextResult` then receive an array.

The buffered (non-streamed) path breaks the same way for `TextResult`.
Setting `reasoning_effort: 'none'` (string content) avoids the crash.

### Expected vs actual

- Expected: the thinking trace is surfaced as thinking deltas and the answer as
  text deltas (as the Anthropic bridge does).
- Actual: a `TypeError` is thrown mid-stream.

### Error

```
[info] Request: "POST https://api.mistral.ai/v1/chat/completions"
[info] Response: "200 https://api.mistral.ai/v1/chat/completions" 0.533111 seconds
  {"http_code":200,"url":"https://api.mistral.ai/v1/chat/completions","total_time":0.533111}

Fatal error: Uncaught TypeError: Symfony\AI\Platform\Result\Stream\Delta\TextDelta::__construct(): Argument #1 ($text) must be of type string, array given, called in /app/src/platform/src/Bridge/Generic/Completions/CompletionsConversionTrait.php on line 109 and defined in /app/src/platform/src/Result/Stream/Delta/TextDelta.php:20
Stack trace:
#0 /app/src/platform/src/Bridge/Generic/Completions/CompletionsConversionTrait.php(109): Symfony\AI\Platform\Result\Stream\Delta\TextDelta->__construct(Array)
#1 /app/src/platform/src/Result/StreamResult.php(61): Symfony\AI\Platform\Bridge\Mistral\Llm\ResultConverter->convertStream(Object(Symfony\AI\Platform\Result\RawHttpResult))
#2 /app/src/platform/src/Result/DeferredResult.php(286): Symfony\AI\Platform\Result\StreamResult->getContent()
#3 /app/src/platform/src/Result/DeferredResult.php(299): Symfony\AI\Platform\Result\DeferredResult->asStream()
#4 /app/examples/mistral/bug.php(17): Symfony\AI\Platform\Result\DeferredResult->asTextStream()
#5 {main}
  thrown in /app/src/platform/src/Result/Stream/Delta/TextDelta.php on line 20
```

### Steps to reproduce

Package: `symfony/ai-platform` (monorepo `main`).
Minimal standalone reproducer under `examples/mistral/` (run with `-vvv`, key redacted):

```php
<?php

$platform = Factory::createPlatform(env('MISTRAL_API_KEY'), http_client());

$messages = new MessageBag(Message::ofUser('What is 23 * 17?'));
$result = $platform->invoke('mistral-medium-2604', $messages, [
    'stream' => true,
    'reasoning_effort' => 'high',
]);

foreach ($result->asTextStream() as $delta) {
    echo $delta;
}
```

Run: `php examples/mistral/reproducer.php -vvv`

### Environment

- Component: `symfony/ai-platform` @ `main`
- PHP: 8.x
- Model: `mistral-medium-2604` with `reasoning_effort: high`

benoitbauchet:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #2422
| License       | MIT

The Mistral `Llm\ResultConverter` reuses `CompletionsConversionTrait`, which assumes the OpenAI chat-completions schema where `choices[].delta.content` (and `message.content`) is always a string. With `reasoning_effort: high` (catalog thinking models such as `mistral-medium-2604` and `mistral-small-2603`), Mistral returns `content` as an array of thinking/text chunks (https://docs.mistral.ai/capabilities/reasoning/), so streaming throws `TextDelta::__construct(): Argument #1 ($text) must be of type string, array given` and the buffered path throws the equivalent for `TextResult`.

This extracts the per-delta content handling into a protected `yieldContentDeltas()` seam on the trait (behaviour-preserving for all OpenAI-compatible providers) and overrides it — plus `convertChoice()` — in the Mistral bridge to parse the chunks into `ThinkingDelta`/`ThinkingComplete`/ `TextDelta` (streaming) and `ThinkingResult` + `TextResult` in a `MultiPartResult` (buffered), matching the Anthropic bridge's thinking model.

An `examples/mistral/chat-stream-reasoning.php` example is added (`mistral-medium-2604`), and the generic completions regression tests confirm the OpenAI-compatible paths are unchanged.

### DIFF
--- a/src/platform/src/Bridge/Generic/Completions/CompletionsConversionTrait.php
+++ b/src/platform/src/Bridge/Generic/Completions/CompletionsConversionTrait.php
@@ -90,23 +90,7 @@ trait CompletionsConversionTrait
                 yield new ToolCallComplete(array_map($this->convertToolCall(...), $toolCalls));
             }
 
-            $reasoningContent = $data['choices'][0]['delta']['reasoning_content']
-                ?? $data['choices'][0]['delta']['reasoning'] ?? null;
-            if (null !== $reasoningContent && '' !== $reasoningContent) {
-                $reasoning .= $reasoningContent;
-                yield new ThinkingDelta($reasoningContent);
-            }
-
-            if ('' !== $reasoning && isset($data['choices'][0]['delta']['content']) && '' !== $data['choices'][0]['delta']['content']) {
-                yield new ThinkingComplete($reasoning);
-                $reasoning = '';
-            }
-
-            if (!isset($data['choices'][0]['delta']['content'])) {
-                continue;
-            }
-
-            yield new TextDelta($data['choices'][0]['delta']['content']);
+            $reasoning = yield from $this->yieldContentDeltas($data['choices'][0]['delta'] ?? [], $reasoning);
         }
 
         if ('' !== $reasoning) {
@@ -122,6 +106,32 @@ trait CompletionsConversionTrait
         if (null !== $finishReason) {
             yield new MetadataDelta('finish_reason', $finishReason);
         }
+    }
+
+    /**
+     * @param array<string, mixed> $delta     The `choices[0].delta` payload
+     * @param string               $reasoning Reasoning accumulated by the previous chunks
+     *
+     * @return \Generator<int, ThinkingDelta|ThinkingComplete|TextDelta, mixed, string> Yields the content deltas; returns the updated reasoning
+     */
+    protected function yieldContentDeltas(array $delta, string $reasoning): \Generator
+    {
+        $reasoningContent = $delta['reasoning_content'] ?? $delta['reasoning'] ?? null;
+        if (null !== $reasoningContent && '' !== $reasoningContent) {
+            $reasoning .= $reasoningContent;
+            yield new ThinkingDelta($reasoningContent);
+        }
+
+        if ('' !== $reasoning && isset($delta['content']) && '' !== $delta['content']) {
+            yield new ThinkingComplete($reasoning);
+            $reasoning = '';
+        }
+
+        if (isset($delta['content'])) {
+            yield new TextDelta($delta['content']);
+        }
+
+        return $reasoning;
     }
 
     /**
--- a/examples/mistral/chat-stream-reasoning.php
+++ b/examples/mistral/chat-stream-reasoning.php
@@ -1,0 +1,52 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use Symfony\AI\Platform\Bridge\Mistral\Factory;
+use Symfony\AI\Platform\Message\Message;
+use Symfony\AI\Platform\Message\MessageBag;
+use Symfony\AI\Platform\Result\Stream\Delta\TextDelta;
+use Symfony\AI\Platform\Result\Stream\Delta\ThinkingComplete;
+use Symfony\AI\Platform\Result\Stream\Delta\ThinkingDelta;
+
+require_once dirname(__DIR__).'/bootstrap.php';
+
+$platform = Factory::createPlatform(env('MISTRAL_API_KEY'), http_client());
+
+$messages = new MessageBag(
+    Message::forSystem('You are a helpful assistant.'),
+    Message::ofUser('What is 23 * 17?'),
+);
+
+// With `reasoning_effort: high`, Mistral returns the thinking trace as array-shaped content
+// chunks, which the bridge maps to ThinkingDelta/ThinkingComplete before the answer TextDeltas.
+$result = $platform->invoke('mistral-medium-2604', $messages, [
+    'stream' => true,
+    'reasoning_effort' => 'high',
+]);
+
+$thinking = false;
+foreach ($result->asStream() as $delta) {
+    if ($delta instanceof ThinkingDelta) {
+        if (!$thinking) {
+            output()->writeln('<info><thinking></info>');
+            $thinking = true;
+        }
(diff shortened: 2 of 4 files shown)


benoitbauchet:
Status: Needs Review

cc @chr-hertel @OskarStark

Tested against the live Mistral API with `mistral-medium-2604` and
`reasoning_effort: 'high'`: `main` throws `TextDelta … array given`, this branch
streams the thinking trace and answer with no error.

<details>
<summary>examples/mistral/chat-stream-reasoning.php -vvv</summary>

​```
[info] Request: "POST https://api.mistral.ai/v1/chat/completions"
[info] Response: "200 https://api.mistral.ai/v1/chat/completions" 0.320707 seconds
  {"http_code":200,"url":"https://api.mistral.ai/v1/chat/completions","total_time":0.320707}

\<thinking\>
Okay, the user is asking for the multiplication of 23 and 17. Let me think about how to approach this. I remember that multiplying two numbers can be done using the standard method or maybe breaking it down. Let me try the standard multiplication first.                                                                                                                                                                     
                                                                                                                                                                                                                 
23 times 17. So, 23 multiplied by 10 is 230, and 23 multiplied by 7 is 161. Then adding those together: 230 + 161 = 391. That seems straightforward. Alternatively, maybe using the distributive property. 23*17 is the same as (20+3)*17 = 20*17 + 3*17. 20*17 is 340, and 3*17 is 51. Adding those gives 340 + 51 = 391. Both methods give the same result. So I think the answer is 391.                                       
\</thinking\>

23 multiplied by 17 equals **391**.
​```

</details>

src/platform/src/Bridge/Generic/Completions/CompletionsConversionTrait.php
```diff
@@ -124,6 +108,32 @@ protected function convertStream(RawResultInterface $result): \Generator
         }
     }
 
+    /**
+     * @param array<string, mixed> $delta     The `choices[0].delta` payload
+     * @param string               $reasoning Reasoning accumulated so far, updated in place
+     *
+     * @return \Generator<ThinkingDelta|ThinkingComplete|TextDelta>
+     */
+    protected function yieldContentDeltas(array $delta, string &$reasoning): \Generator
```
wachterjohannes: Could the accumulator stay inside the trait rather than being passed by reference? It makes the seam awkward to override, since an implementer has to know they are expected to mutate the caller's state.

A property, or returning the updated reasoning alongside the deltas, would keep the extension point to "given a delta payload, yield the content deltas". Not blocking if there is no clean way, the behaviour is right either way.

wachterjohannes (commented):
Thanks @benoitbauchet, and a good first contribution: found, diagnosed down to the schema mismatch in the shared trait, fixed, tested, and verified against the live API.

I captured the raw payload with a real key and replayed it through this branch. It handles the tricky chunk correctly, the one where the thinking block closes and the text starts in the same delta with an empty thinking array. Details in #2422. Verified locally, all green.

Walking the chunks in order is the part I would keep: it preserves the provider's ordering rather than collapsing all thinking and all text per chunk, which matters if Mistral ever interleaves them differently.

One question inline.

@lazerg opened #2424 for the same bug seven minutes after this one. Same behaviour on the captured payload, so I am going with this one; more on that over there.

benoitbauchet:
> Thanks @benoitbauchet, and a good first contribution: found, diagnosed down to the schema mismatch in the shared trait, fixed, tested, and verified against the live API.
> 
> I captured the raw payload with a real key and replayed it through this branch. It handles the tricky chunk correctly, the one where the thinking block closes and the text starts in the same delta with an empty thinking array. Details in #2422. Verified locally, all green.
> 
> Walking the chunks in order is the part I would keep: it preserves the provider's ordering rather than collapsing all thinking and all text per chunk, which matters if Mistral ever interleaves them differently.
> 
> One question inline.
> 
> @lazerg opened #2424 for the same bug seven minutes after this one. Same behaviour on the captured payload, so I am going with this one; more on that over there.

Thanks for the thorough review !

Addressed the seam question: `yieldContentDeltas()` no longer takes the reasoning
accumulator by reference.

src/platform/src/Bridge/Generic/Completions/CompletionsConversionTrait.php
benoitbauchet: Done — switched from `string &$reasoning` to taking it by value and returning the updated reasoning:

    protected function yieldContentDeltas(array $delta, string $reasoning): \Generator
    {
        // ... yields ThinkingDelta / ThinkingComplete / TextDelta ...
        return $reasoning;
    }

Caller:

    $reasoning = yield from $this->yieldContentDeltas($delta, $reasoning);

wachterjohannes (approved):
Nice solution, `$reasoning = yield from ...` reads much better than mutating the caller's variable, and it keeps the seam to "given a delta payload, yield the content deltas".

Re-verified against the live API with a second, independent response, 190 chunks this time instead of 97, and it comes out right: the thinking trace streams as `ThinkingDelta`, closes once with `ThinkingComplete`, and the answer follows as text. Same for the buffered path. Verified locally, all green.

Thanks for a thorough first contribution here, and for the patience through two rounds.

wachterjohannes:
Thank you @benoitbauchet.

PR #2424, opened 2026-08-20
[Platform][Mistral] Handle chunked reasoning content in completions results

### LINKED ISSUE #2422: [Platform][Mistral] Streaming with reasoning_effort crashes: TextDelta expects string, array given

lazerg:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #2422
| License       | MIT

With `reasoning_effort` set to anything but `none`, Mistral no longer sends `content` as a string. It sends a list of thinking and text chunks, and only falls back to a plain string once the thinking block has closed. `CompletionsConversionTrait` assumes the OpenAI schema, so that list went straight into `TextDelta` and `TextResult` and blew up with a `TypeError` on both the streamed and the buffered path.

The Mistral converter now flattens those chunks before the shared conversion sees them. Streaming turns thinking chunks into `reasoning_content`, which the trait already surfaces as `ThinkingDelta`/`ThinkingComplete`, and text chunks into the string `content` it expects. The buffered path returns the trace as its own `ThinkingResult` next to the answer, the way the Anthropic and Gemini bridges do, so a response that is still mid-thinking when it gets truncated no longer collapses into an empty string.

Chunk reshaping happens in a new `normalizeStreamChunk()` hook that defaults to a no-op, so nothing changes for the other bridges using the trait. Happy to move the flattening elsewhere if you'd rather not widen that trait.

### DIFF
--- a/src/platform/src/Bridge/Generic/Completions/CompletionsConversionTrait.php
+++ b/src/platform/src/Bridge/Generic/Completions/CompletionsConversionTrait.php
@@ -50,6 +50,8 @@ trait CompletionsConversionTrait
         $finishReason = null;
 
         foreach ($result->getDataStream() as $data) {
+            $data = $this->normalizeStreamChunk($data);
+
             if (isset($data['error'])) {
                 $message = \is_array($data['error']) ? ($data['error']['message'] ?? 'Unknown error') : (string) $data['error'];
                 $code = \is_array($data['error']) ? ($data['error']['code'] ?? null) : null;
@@ -122,6 +124,18 @@ trait CompletionsConversionTrait
         if (null !== $finishReason) {
             yield new MetadataDelta('finish_reason', $finishReason);
         }
+    }
+
+    /**
+     * Reshapes a stream chunk for bridges whose delta payload deviates from the OpenAI schema.
+     *
+     * @param array<string, mixed> $data
+     *
+     * @return array<string, mixed>
+     */
+    protected function normalizeStreamChunk(array $data): array
+    {
+        return $data;
     }
 
     /**
--- a/src/platform/src/Bridge/Mistral/Llm/ResultConverter.php
+++ b/src/platform/src/Bridge/Mistral/Llm/ResultConverter.php
@@ -12,16 +12,19 @@
 namespace Symfony\AI\Platform\Bridge\Mistral\Llm;
 
 use Symfony\AI\Platform\Bridge\Generic\Completions\CompletionsConversionTrait;
+use Symfony\AI\Platform\Bridge\Generic\Completions\FinishReasonMapper;
 use Symfony\AI\Platform\Bridge\Mistral\Mistral;
 use Symfony\AI\Platform\Exception\ExceedContextSizeException;
 use Symfony\AI\Platform\Exception\RuntimeException;
 use Symfony\AI\Platform\Model;
 use Symfony\AI\Platform\Result\ChoiceResult;
 use Symfony\AI\Platform\Result\HttpStatusErrorHandlingTrait;
+use Symfony\AI\Platform\Result\MultiPartResult;
 use Symfony\AI\Platform\Result\RawHttpResult;
 use Symfony\AI\Platform\Result\RawResultInterface;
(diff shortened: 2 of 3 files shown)


wachterjohannes:
Thanks @lazerg, and sorry for the duplicate effort here.

I captured the raw Mistral payload with a live key and replayed it through both this and #2423: byte-identical output, same deltas in the same order, same buffered result. Details in #2422. So this came down to the seam on the shared trait rather than behaviour.

`normalizeStreamChunk()` is the tidier of the two, purely additive and reusing the `reasoning_content` path the trait already handles, and I would reach for it if a second bridge ever needs to reshape chunks. What tipped it to #2423 is that it walks the content chunks in order instead of flattening all thinking and all text per chunk, so it keeps the provider's ordering, and that it came from the reporter with an example and live verification attached.

Closing in favour of #2423. The hook idea is worth revisiting on its own if another bridge hits the same schema drift.

PR #2425, opened 2026-08-21
[Agent] Fix tool call budget being shared across concurrent streaming calls

### LINKED ISSUE #2421: [Agent] Tool call budget is shared across concurrent streaming calls
Found while reviewing #2394, which fixes the counter that made `maxToolCalls` unreachable. That fix is right, this is the case it leaves open.

`Runner::$iterations` lives on the `Runner`, which is a shared service, and is reset in `run()` when `nestingLevel` is 0. In the streaming path the loop starts much later, when the returned `StreamResult` is consumed. So two streaming calls started before either is consumed share one budget: the second call's reset happens while the first has not started counting, and the first then exhausts the counter that the second inherits.

Counting real tool rounds per stream with `maxToolCalls: 3`:

```
stream A -> capped after 3 tool rounds
stream B -> capped after 0 tool rounds
```

Stream B is refused before executing a single tool.

### Why the obvious fix does not work

Resetting inside `handleToolCalls()` when `nestingLevel` becomes 1 looks like the natural answer, but it breaks the streaming path: the parent's `handleToolCalls()` returns before the nested round runs, so nesting has already unwound by the time the next round starts and every round resets. Tried it, the loop then runs 29 rounds against a limit of 3 and `RunnerTest::testThrowsExceptionWhenMaxIterationsExceededWhileStreaming` fails.

### Shape of a real fix

The budget needs to belong to one top-level agent call rather than to the `Runner` instance, for example a small per-call state object created in `run()` at nesting 0 and carried into the stream listener closure.

`$sources` and `$nestingLevel` have the same shape and would presumably move with it, so this is worth doing as one deliberate change rather than three patches.

Not urgent: the situation today is strictly worse, since #2394 has not landed and the limit never fires at all.

ousamabenyounes:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #2421
| License       | MIT

`Runner::$iterations` lives on the `Runner`, which is a shared service, and is reset in `run()` when
`nestingLevel` is 0. In the streaming path the tool calling loop only starts when the returned
`StreamResult` is consumed, which happens much later. Two streaming calls started before either one is
consumed therefore share a single budget: the second call's reset runs while the first has not started
counting, and the first then exhausts the counter the second inherits.

With `maxToolCalls: 3`, counting real tool rounds per stream:

```
stream A -> capped after 3 tool rounds
stream B -> capped after 0 tool rounds
```

Stream B is refused before executing a single tool.

### The fix

The budget now belongs to one outermost agent call instead of to the `Runner` instance. A
`ToolCallBudget` is created in `run()` at nesting level 0, captured there, and passed into the stream
listener closure, so the loop always spends the budget of the call it belongs to — whenever the stream
is consumed.

`handleToolCalls()` installs the budget it was handed for the duration of the loop and restores the
previously active one when it unwinds, so nested agent calls keep continuing on the budget of the loop
they belong to rather than opening a fresh one. That is the part the obvious "reset when nesting
level becomes 1" fix gets wrong: in the streaming path the parent's `handleToolCalls()` returns before
the nested round runs, so nesting has already unwound by the time the next round starts.

`$sources` and `$nestingLevel` are deliberately left on the `Runner` for now: `$nestingLevel` correctly
tracks the synchronous call stack (`handleToolCalls()` has no suspension point, so it always runs to
completion once the listener fires), and I could not construct a failing case for `$sources` under
concurrent streams — each level already attaches and resets it before unwinding. Happy to move them
into the same per-call object if you prefer that as one deliberate change.

RED — new test on unmodified `main`, production code untouched (`git status --porcelain -- src/agent/src/` empty):

```
$ vendor/bin/phpunit -c src/agent/phpunit.xml.dist --filter testMaxIterationsLimitIsNotSharedBetweenConcurrentStreams
PHPUnit 11.5.56 by Sebastian Bergmann and contributors.

F                                                                   1 / 1 (100%)

There was 1 failure:

1) Symfony\AI\Agent\Tests\Execution\RunnerTest::testMaxIterationsLimitIsNotSharedBetweenConcurrentStreams
Failed asserting that 3 is identical to 6.

src/agent/tests/Execution/RunnerTest.php:527

FAILURES!
Tests: 1, Assertions: 2, Failures: 1.
```

The first assertion passes and the second fails at `3`, which is exactly the reported behaviour: the
first stream spends its 3 rounds, the second gets 0.

GREEN — same test with the fix:

```
$ vendor/bin/phpunit -c src/agent/phpunit.xml.dist --filter testMaxIterationsLimitIsNotSharedBetweenConcurrentStreams
.                                                                   1 / 1 (100%)

OK (1 test, 2 assertions)
```

No regressions in the component — `main` was `OK (276 tests, 656 assertions)`, the branch is:

```
$ vendor/bin/phpunit -c src/agent/phpunit.xml.dist
OK (277 tests, 658 assertions)
```

`php-cs-fixer` reports 0 fixable files and `phpstan analyse` reports no errors for `src/agent`.

### DIFF
--- a/src/agent/src/Execution/Runner.php
+++ b/src/agent/src/Execution/Runner.php
@@ -12,7 +12,6 @@
 namespace Symfony\AI\Agent\Execution;
 
 use Symfony\AI\Agent\AgentInterface;
-use Symfony\AI\Agent\Exception\MaxIterationsExceededException;
 use Symfony\AI\Agent\Toolbox\Event\ToolCallsExecuted;
 use Symfony\AI\Agent\Toolbox\Source\SourceCollection;
 use Symfony\AI\Agent\Toolbox\StreamListener;
@@ -52,9 +51,9 @@ final class Runner
     private int $nestingLevel = 0;
 
     /**
-     * Counts the tool calling rounds of one agent call, including the ones happening in nested agent calls.
+     * Budget of the agent call currently driving the tool calling loop, shared with its nested calls.
      */
-    private int $iterations = 0;
+    private ToolCallBudget $budget;
 
     public function __construct(
         private readonly PlatformInterface $platform,
@@ -67,6 +66,7 @@ final class Runner
         private readonly ToolResultConverter $resultConverter = new ToolResultConverter(),
     ) {
         $this->sources = new SourceCollection();
+        $this->budget = new ToolCallBudget($maxToolCalls);
     }
 
     /**
@@ -76,8 +76,12 @@ final class Runner
     {
         // nested calls continue the tool calling loop of the outermost call and share its budget
         if (0 === $this->nestingLevel) {
-            $this->iterations = 0;
+            $this->budget = new ToolCallBudget($this->maxToolCalls);
         }
+
+        // the streaming loop starts when the result is consumed, which can be after another call
+        // already replaced the current budget, so it is captured here and passed along explicitly
+        $budget = $this->budget;
 
         $options = $this->exposeTools($options);
 
@@ -89,7 +93,7 @@ final class Runner
 
         if ($result instanceof StreamResult) {
             $result->addListener(new StreamListener(
-                fn (ToolCallResult $toolCallResult, ?AssistantMessage $streamedAssistantResponse = null): ResultInterface => $this->handleToolCalls($agent, $messages, $options, $toolCallResult, $streamedAssistantResponse),
+                fn (ToolCallResult $toolCallResult, ?AssistantMessage $streamedAssistantResponse = null): ResultInterface => $this->handleToolCalls($budget, $agent, $messages, $options, $toolCallResult, $streamedAssistantResponse),
             ));
 
             return $result;
@@ -101,7 +105,7 @@ final class Runner
             return $result;
         }
 
-        return $this->handleToolCalls($agent, $messages, $options, $toolCallResult);
+        return $this->handleToolCalls($budget, $agent, $messages, $options, $toolCallResult);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


wachterjohannes (commented):
Thanks @ousamabenyounes, this is the right shape and it lands the fix where I could not: the budget now belongs to the call rather than to the shared `Runner`, and capturing it into a local before it reaches the listener closure is exactly what survives the streaming unwind.

Verified with the reproduction from #2421, unchanged:

```
before: stream A -> capped after 3 tool rounds
        stream B -> capped after 0 tool rounds
after:  stream A -> capped after 3 tool rounds
        stream B -> capped after 3 tool rounds
```

Your new test also fails on `main` exactly as you describe and passes with the fix. Verified locally, all green.

On `$sources`: I tried to break it with two concurrent streams collecting different sources and could not either, each result keeps its own. Same output before and after this change, so leaving it in place is right.

One comment inline.

src/agent/src/Execution/Runner.php
```diff
@@ -141,10 +145,13 @@ private function isFlatStringArray(array $tools): bool
...
      * @param array<string, mixed> $options
      */
-    private function handleToolCalls(AgentInterface $agent, MessageBag $messages, array $options, ToolCallResult $result, ?AssistantMessage $streamedAssistantResponse = null): ResultInterface
+    private function handleToolCalls(ToolCallBudget $budget, AgentInterface $agent, MessageBag $messages, array $options, ToolCallResult $result, ?AssistantMessage $streamedAssistantResponse = null): ResultInterface
     {
         \assert($this->toolExecutor instanceof ToolExecutorInterface);
 
+        // nested agent calls have to continue on the budget of the call this loop belongs to
+        $previousBudget = $this->budget;
+        $this->budget = $budget;
```
wachterjohannes: This save and restore is the load-bearing part of the fix, and it only works because `handleToolCalls()` has no suspension point: once the listener fires, the loop runs to completion, so no other call can install its budget in between. That reasoning is in the PR description but not in the code, and it is exactly what someone would break later by making this method lazy.

Worth stating here, something like: "safe because the loop runs to completion, so no other call can install a budget in between".

ousamabenyounes: Added in `fb43df6` — kept it right on the assignment:

```php
// nested agent calls have to continue on the budget of the call this loop belongs to.
// Safe because this loop runs to completion with no suspension point, so no other call
// can install its budget in between; making this method lazy would break that.
```

Comment only, no behaviour change, so no new RED to show — `src/agent` suite re-run on PHP 8.4 to confirm nothing moved: `OK (277 tests, 658 assertions)`.

Thanks for digging into `$sources` independently, that was the part I was least sure about.

wachterjohannes (approved):
Comment reads well, and putting it on the assignment rather than above the method is the right spot, that is where someone would change it. Re-verified, all green.

Thanks for taking this on from the issue and for working out the suspension-point argument independently, that was the part that made the naive fix look correct.

wachterjohannes:
Thank you @ousamabenyounes.

PR #2429, opened 2026-08-22
[AI Bundle][Store] Expose the ChromaDb embedding function in the configuration

### LINKED ISSUE #2418: [AI Bundle][Store] Expose ChromaDb `embedding_function` in the bundle configuration
#2399 adds an optional `EmbeddingFunction` to `Store` and `StoreFactory::create()` for the ChromaDb bridge. Without it `supports(TextQuery::class)` returns `false`, because the ChromaDB client embeds query texts client-side and fatals on a null embedding function.

The bundle has no way to configure it:

- `src/ai-bundle/config/store/chromadb.php` only defines `client` and `collection`
- `AiBundle.php` (~L1478) passes only those two arguments to `ChromaDbStoreFactory::create()`

So every bundle-configured ChromaDb store gets `null` and can never serve a `TextQuery`. Not a regression — text query never worked on this bridge — but the fix from #2399 is reachable only from plain PHP.

### Proposal

Add an `embedding_function` service-reference node, mirroring the existing `client` node:

```yaml
ai:
    store:
        chromadb:
            symfonycon:
                collection: 'symfony_blog'
                embedding_function: 'app.chromadb.embedding_function'
```

and thread it through as the third factory argument:

```php
->setArguments([
    new Reference($store['client']),
    $store['collection'] ?? $name,
    isset($store['embedding_function']) ? new Reference($store['embedding_function']) : null,
])
```

Should stay optional, so existing configurations keep working unchanged.

### Notes

- Blocked on #2399 being merged first.
- Worth adding a `testChromaDbStoreCanBeConfiguredWithEmbeddingFunction()` case next to the existing ChromaDb tests in `AiBundleTest`.
- The store docs (`docs/components/store.rst`) show a ChromaDb YAML example that could mention the new option.

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | Fix #2418
| License       | MIT

#2399 added an optional `EmbeddingFunction` to the ChromaDb `Store` and `StoreFactory::create()`, so a `TextQuery` can be embedded client-side. The bundle never passed it: `AiBundle` handed the factory only the client and the collection, so every store configured through YAML got `null` and reported `supports(TextQuery::class)` as `false`. The capability was reachable from plain PHP only.

```yaml
ai:
    store:
        chromadb:
            symfonycon:
                collection: 'symfony_blog'
                embedding_function: 'app.chromadb.embedding_function'
```

The node is optional and holds a service id, mirroring the existing `client` node, so existing configurations keep working unchanged and still get `null`.

### On the version constraint

`symfony/ai-chroma-db-store` moves from `^0.12` to `dev-main`. The third factory parameter only exists on `main`; the released `v0.12.0` still takes two, and passing a third makes `BridgeConfigCompilationTest` fail, since it reflects every store factory and compares the argument count:

```
Service "ai.store.chromadb.default" has 3 arguments but factory
"ChromaDbStoreFactory::create" accepts at most 2.
```

Same situation as 08abac0b, and it goes back to a stable constraint with the next release.

### Verification

`AiBundleTest` gets a `testChromaDbStoreCanBeConfiguredWithEmbeddingFunction()` case, and the three existing ChromaDb cases now assert the third argument is `null` when the option is absent.

The bundle suite has three pre-existing failures unrelated to this change, caused by the bundle resolving the released v0.12 platform and agent splits while `main` has moved on (`ModelOptionsParser` missing, `Agent` constructor changed). Baseline is 410 tests with 2 errors and 1 failure, this branch is 411 with the same three, so nothing new. PHPStan, PHP-CS-Fixer and `doctor-rst` are clean.


src/ai-bundle/tests/DependencyInjection/AiBundleTest.php
```diff
@@ -911,6 +914,28 @@ public function testChromaDbStoreCanBeConfigured()
     }
 
     #[TestDox('Configuring only one store type does not require packages for other store types')]
```
chr-hertel: this `TestDox` belongs to the test method below the new one

wachterjohannes: Good catch, thanks. My insertion landed between the attribute and its method. Moved back in `83e62b9`, and the neighbouring test is green again.

chr-hertel (approved):

chr-hertel:
Thank you @wachterjohannes.

PR #2431, opened 2026-08-22
[AI Bundle] Support template variables in file prompts

### LINKED ISSUE #1621: [Agent][AI Bundle] Allow template variables when using `agent.prompt.file` prompt
Hello,

I want to use the templating system from the Symfony AI library. I want to do so in combination with the `prompt.file` parameter (which you can supply when configuring an `Agent`).

However, I soon found out this is not possible: the file referenced in `prompt.file` is read and treated as a static prompt. Due to not being added to the `MessageBag` as `Template` object, it does not respect the `template_vars` option when invoking `AgentInterface::call()`.

I suspect this is a small change, but a welcome on nonetheless. It would be great if file-based prompts are converted to a `Template` object, allowing the use of `template_vars` to dynamically build prompts. Currently, the only way to do have it become a `Template` object is by defining the prompt in code, which beats the purpose of the `prompt.file` parameter: keeping prompt building logic separate from implementing services.

I know this can be circumvented rather easily by implementing a thin prompt resolving layer (or even a custom processor), but I feel like this is missing functionality and it seems illogical to have the whole template rendering setup, yet no way to combine it with `prompt.file`.

Thanks in advance for your time :-)

mikemikimike:
| Q | A |
| --- | --- |
| Bug fix? | yes |
| New feature? | yes |
| Docs? | yes |
| Issues | Fix #1621 |
| License | MIT |

`prompt.file` was read as a static `File`, so file-based prompts could not participate in the existing `template_vars` rendering flow.

This PR creates file-backed prompts as `Template` instances while preserving runtime file loading:

- Registers a container service using `FilePromptTemplateFactory`.
- Stores only the configured file path in the container definition.
- Reads the file when the prompt service is created and wraps it with `Template::string()`.
- Keeps unreadable-file validation through `File::fromFile()`.
- Preserves templates through `SystemPromptInputProcessor`, including toolbox prompts; expression templates append tool definitions as an escaped expression literal.
- Renders remaining templates with empty variables so file prompts without `template_vars` retain their raw text.

This avoids placing a `Template` object directly in the compiled container, so `cache:clear` and container dumping continue to work. It also avoids turning the prompt file into a compile-time snapshot; changes made before service creation are observed.

Tests run:

- `php -l src/ai-bundle/src/AiBundle.php`
- `php -l src/ai-bundle/src/DependencyInjection/FilePromptTemplateFactory.php`
- `php -l src/ai-bundle/tests/DependencyInjection/AiBundleTest.php`
- `php -l src/agent/src/InputProcessor/SystemPromptInputProcessor.php`
- `php -l src/agent/tests/InputProcessor/SystemPromptInputProcessorTest.php`
- `vendor/bin/phpunit tests/InputProcessor/SystemPromptInputProcessorTest.php`: `OK (15 tests, 50 assertions)`
- `vendor/bin/phpstan analyse -c phpstan.dist.neon src/InputProcessor/SystemPromptInputProcessor.php tests/InputProcessor/SystemPromptInputProcessorTest.php --no-progress`: no errors
- `vendor/bin/php-cs-fixer fix --dry-run --diff --config=.php-cs-fixer.dist.php` on both changed PHP files: no changes required
- Focused AI Bundle PHPUnit regression test: `OK (1 test, 6 assertions)`
- Docker Symfony container export smoke test: passed
- `git diff --check`: passed

The full local `AiBundleTest` suite still has unrelated baseline/dependency-linking failures; the focused regression tests pass. GitHub CI has been triggered for the updated commit.


src/ai-bundle/src/DependencyInjection/FilePromptTemplateFactory.php
```diff
@@ -0,0 +1,26 @@
...
+
+use Symfony\AI\Platform\Message\Content\File;
+use Symfony\AI\Platform\Message\Template;
+
+/**
+ * @author mikemikimike
+ */
+final class FilePromptTemplateFactory
+{
+    public static function create(string $path): Template
```
wachterjohannes, line 22: The `Template` this returns is flattened again one step later, so nothing changes for the user. `SystemPromptInputProcessor` falls into its `else` branch and casts:

```php
} else {
    $message = (string) $this->systemPrompt;
}
...
$messages->prepend(Message::forSystem($message));
```

I fed the processor exactly what this factory produces:

```
Template (what this PR passes)   -> SystemMessage content is string: 'Hello {name}!'
plain string (before this PR)    -> SystemMessage content is string: 'Hello {name}!'
```

Identical. And `TemplateRendererListener` only renders when the content really is a `Template`:

```php
if ($content instanceof Template) { ... }
```

So #1621 stays open. The container test passes because it asserts the definition, not what the prompt becomes once the service is used.

The fix belongs in `SystemPromptInputProcessor`, and there are two parts to it:

1. Accept `Template` in the constructor union and pass it through instead of casting, so `Message::forSystem()` receives it and keeps it.
2. The toolbox branch appends the tool descriptions by string interpolation, which flattens a `Template` just as effectively. With tools enabled it needs to produce a new `Template` carrying the appended text.

Worth an end-to-end test too: render with `template_vars` and assert the placeholder is actually substituted, rather than asserting on the container definition.

wachterjohannes (changes requested):
Thanks @mikemikimike, and welcome. The bundle half of this is right: building the prompt lazily through a factory instead of snapshotting the file at compile time is the correct call, and `ai.agent.prompt.<name>` is exactly the service pattern the translatable text prompt already uses.

The `Template` does not survive though, so `template_vars` still does not reach a file prompt. Details inline. Sorry to send you one layer deeper on a first contribution.

src/ai-bundle/src/DependencyInjection/FilePromptTemplateFactory.php
mikemikimike: Fixed in commit `0a2c4d8`. `SystemPromptInputProcessor` now accepts and preserves `Template` instances instead of casting them to strings. When tool definitions are appended, template content and type are preserved in a new `Template`. Added a regression test that verifies the system message retains the template object; the focused Docker test passes (`1 test, 2 assertions`).

src/ai-bundle/src/AiBundle.php
```diff
@@ -1314,11 +1316,14 @@ private function processAgentConfig(string $name, array $config, ContainerBuilde
             // Create prompt from file if configured, otherwise use text
             if (isset($config['prompt']['file'])) {
                 $filePath = $config['prompt']['file'];
-                // File::fromFile() handles validation, so no need to check here
-                // Use Definition with factory method because File objects cannot be serialized during container compilation
-                $prompt = (new Definition(File::class))
-                    ->setFactory([File::class, 'fromFile'])
+                // Keep the file path as a container argument so the prompt is read when the service is created.
+                $promptId = 'ai.agent.prompt.'.$name;
+                $templateDefinition = (new Definition(Template::class))
```
wachterjohannes, line 1321: Every `prompt.file` becomes a `Template` now, including the ones without placeholders. Those agents never pass `template_vars`, so `TemplateRendererListener` returns on its first line and the `Template` reaches the normalizer unrendered:

```
content type : Symfony\AI\Platform\Message\Template
payload      : {"role":"system","content":{}}
```

Before this PR the argument was a `File` and the prompt went out as a string, so existing file-prompt users would silently lose their system prompt.

Fixing it in `SystemMessageNormalizer` alone won't do, around eight bridge `MessageBagNormalizer`s read `getSystemMessage()->getContent()` straight into the payload. Flattening remaining templates in `TemplateRendererListener` even when `template_vars` is absent covers all of them, and `StringTemplateRenderer` with an empty array returns the raw text unchanged.

src/agent/tests/InputProcessor/SystemPromptInputProcessorTest.php
```diff
@@ -60,6 +61,18 @@ public function testProcessInputAddsSystemMessageWhenNoneExists()
...
+    public function testProcessInputPreservesTemplatePrompt()
+    {
+        $processor = new SystemPromptInputProcessor(Template::string('Hello {name}!'));
+
+        $input = new Input('gpt-4o', new MessageBag(Message::ofUser('This is a user message')));
+        $processor->processInput($input);
+
+        $systemMessage = $input->getMessageBag()->getSystemMessage();
+        $this->assertInstanceOf(Template::class, $systemMessage?->getContent());
+        $this->assertSame('Hello {name}!', $systemMessage?->getContent()->getTemplate());
```
wachterjohannes: PHPStan fails here: after the `assertInstanceOf` above, `$systemMessage` is no longer nullable.

```suggestion
        $this->assertSame('Hello {name}!', $systemMessage->getContent()->getTemplate());
```

wachterjohannes (changes requested):
The `Template` survives now: `template_vars` reaches a file prompt and the toolbox branch keeps it instead of interpolating it away. Verified end to end against main. One regression left, details inline.

- PHPStan is red on the new test, and it is a real error, not infra.
- The toolbox branch with a `Template` is untested; the new test only covers the case without a toolbox.
- `Template::expression()` plus `include_tools` concatenates the tool markdown into the expression and then fails to render it. Not a regression, but the new path makes the combination reachable.
- I am relabelling this as a feature, so it needs CHANGELOG entries: a bullet in the existing `0.13` section of `src/agent/CHANGELOG.md`, and a new `0.13` section in `src/ai-bundle/CHANGELOG.md`, which is still at `0.11`.

src/agent/src/InputProcessor/SystemPromptInputProcessor.php
```diff
@@ -79,15 +82,23 @@ public function processInput(Input $input): void
...
             ));
 
-            $message = <<<PROMPT
-                {$message}
+            if ($message instanceof Template) {
+                $toolsPrompt = \PHP_EOL.\PHP_EOL
+                    .'# Tools'.\PHP_EOL.\PHP_EOL
+                    ."The following tools are available to assist you in completing the user's request:".\PHP_EOL.\PHP_EOL
+                    .$tools;
+                $message = new Template($message->getTemplate().$toolsPrompt, $message->getType());
```
wachterjohannes: `Template::expression()` plus a toolbox still ends up here, and the concatenation puts markdown into the expression source:

```
Unexpected character "#" around position 19 for expression `"you are " ~ role  # Tools  The following tools are available ...`
```

The combination is new in this PR, the constructor did not accept a `Template` before, so nothing could reach it. Either append the tools as an escaped expression literal (`~ "...\n\n# Tools ..."` renders fine, the lexer turns `\n` into a real newline) or reject the combination with a clear message, so it does not fail deep inside the renderer.

wachterjohannes (changes requested):
The regression is gone, verified end to end: without `template_vars` a file prompt goes out as its raw text again, with them it substitutes. Verified locally, all green.

Two changelog entries still missing, and one thing inline.

- `src/agent/CHANGELOG.md`, `0.13`: `SystemPromptInputProcessor` now accepts and preserves a `Template`. That section already exists.
- `src/platform/CHANGELOG.md`, `0.13`: `TemplateRendererListener` now renders templates when `template_vars` is absent instead of returning early. Worth naming the visible consequence: `Template::expression()` without `template_vars` now throws `Variable "x" is not valid` where it previously reached the normalizer as `{}`.

src/agent/src/InputProcessor/SystemPromptInputProcessor.php
mikemikimike: Fixed in bda6f72d. Expression templates now append the tool section as a JSON-escaped expression string literal, preserving quotes, backslashes, and line breaks instead of injecting Markdown into the expression source. Added a regression test covering an expression template with a toolbox and a quoted tool description. The full processor test file passes (15 tests, 50 assertions), and focused PHPStan reports no errors.

src/ai-bundle/src/AiBundle.php
mikemikimike: Addressed in the current head `ad6fd1b9`: `TemplateRendererListener` now renders templates even when `template_vars` is absent, using an empty variable map. This flattens file-backed templates to their raw text before bridge normalizers, and the listener regression test covers the no-variables path.

src/agent/tests/InputProcessor/SystemPromptInputProcessorTest.php
mikemikimike: Addressed in the current head `ad6fd1b9`: after `assertInstanceOf(Template::class, $systemMessage->getContent())`, the test now accesses `$systemMessage->getContent()->getTemplate()` directly, satisfying PHPStan without the redundant nullable access.

src/agent/src/InputProcessor/SystemPromptInputProcessor.php
```diff
@@ -79,15 +82,26 @@ public function processInput(Input $input): void
...
-
-                {$tools}
-                PROMPT;
+            if ($message instanceof Template) {
+                $toolsPrompt = \PHP_EOL.\PHP_EOL
+                    .'# Tools'.\PHP_EOL.\PHP_EOL
+                    ."The following tools are available to assist you in completing the user's request:".\PHP_EOL.\PHP_EOL
+                    .$tools;
+                $template = 'expression' === $message->getType()
+                    ? $message->getTemplate().' ~ '.json_encode($toolsPrompt, \JSON_THROW_ON_ERROR | \JSON_UNESCAPED_SLASHES | \JSON_UNESCAPED_UNICODE)
```
wachterjohannes: The original expression is appended without parentheses, and `~` binds tighter than the ternary, so the tool section attaches to one branch only:

```
admin ? "ADMIN" : "USER"    admin=false -> USER\n\n# Tools\n\n...
                            admin=true  -> ADMIN
```

With `admin=true` the model is never told its tools exist. That is worse than the parse error it replaces, because it fails silently.

```suggestion
                    ? '('.$message->getTemplate().') ~ '.json_encode($toolsPrompt, \JSON_THROW_ON_ERROR | \JSON_UNESCAPED_SLASHES | \JSON_UNESCAPED_UNICODE)
```

Both branches carry the tools with that. Worth a test with an expression that is not a plain concatenation, the current one would not have caught this.

wachterjohannes (changes requested):
All three addressed, thanks, and the changelog entries read well. Verified locally, all green.

One thing the new expression branch introduced, inline.

src/agent/src/InputProcessor/SystemPromptInputProcessor.php
mikemikimike: Fixed in 3608162dd8b1a494036db8cbb98b7fb7fc06ade2. The expression template is now parenthesized before concatenating the JSON-encoded tools prompt, so both ternary branches retain the tool instructions. The regression test now uses a ternary expression to cover the precedence issue. PHP syntax checks and PHP CS Fixer dry-runs pass; PHPUnit could not be run because PHP/Composer are unavailable locally and the Composer container image could not be pulled completely.

wachterjohannes (commented):
The precedence fix is right, I checked it against a ternary, a nested ternary, a comparison and an `or`, and all of them keep the tool section on every branch. Removing the parentheses again makes the new test fail, so it guards what it should. I also ran the whole chain, file to platform input: placeholders get substituted, a prompt file without placeholders goes out unchanged, and a JSON prompt file is left alone.

One thing left, and it is the last one from me:

* The docs are missing. `docs/bundles/ai-bundle.rst` still tells people under "File-Based Prompts" that "The entire content of the file will be used as the system prompt text", which is exactly what this PR changes. A short paragraph there saying the file may contain `{placeholder}` markers and that `template_vars` fills them at call time would cover it, and the table at the top should say `Docs? yes`.

Ping me once that is in and I will approve.

mikemikimike:
Implemented the remaining documentation update in b893a9b5fb. The File-Based Prompts section now documents {placeholder} markers and call-time substitution through 	emplate_vars; the PR table is updated to Docs? yes. git diff --check passes. No additional test was run because this commit changes documentation only.

wachterjohannes (approved):
Docs are in and the table is updated, so this is good to go from my side. Thanks for the patience across the rounds, and for the ternary regression test in particular.

wachterjohannes:
It took time, but here we go, this is in now. Thank you very much @mikemikimike.

Issue #2413, opened 2026-08-16
[Platform] Share one result (de)normalizer between the cache bridge and recording


wachterjohannes:
Follow-up from @chr-hertel's review on #2129.

Two classes turn a finished `ResultInterface` into something storable and back:

- `Bridge\Cache\ResultNormalizer` covers `BinaryResult`, `ChoiceResult`, `MultiPartResult`, `ThinkingResult`, `ObjectResult`, `TextResult`, `ToolCallResult` and `VectorResult`, and deliberately throws on `StreamResult`. Shape: `['class' => FQCN, 'payload' => ...]`, built on the Symfony Serializer with an injected `ObjectNormalizer`.
- `Test\Recording\ResultSerializer` (#2129) covers text, object, vector, tool calls and text streams. Shape: `['type' => 'text'|'object'|...]`, plain static, no serializer.

Four of the five recording types are already covered by the cache one, so this should be a single implementation in the platform core that both use, gaining stream support on one side and the four missing result types on the other.

What makes it its own PR rather than a change to #2129: `ResultNormalizer` sits in the `symfony/ai-cache-platform` bridge while `Test\Recording` sits in the core package, so reuse today would point the core at a bridge. Moving the class down into `symfony/ai-platform` is a BC break for the bridge and needs a deprecation alias plus an `UPGRADE.md` entry.

Open questions for whoever picks this up: which serialized shape wins (the FQCN-keyed one is more extensible, a short type key diffs better in a committed cassette), and whether the merged class keeps the cache bridge's refusal to serialize a stream at its own call site.

PR #2432, opened 2026-08-23
[Platform] Preserve result metadata and token usage in provider cassettes

### LINKED ISSUE #2410: [Platform] Preserve result metadata and token usage in provider cassettes
Follow-up from @MikiBuilder's review in https://github.com/symfony/ai/pull/2129#issuecomment-5114825333.

`Test\Recording\ResultSerializer` (#2129) drops result metadata and token usage, so a replayed result cannot be asserted on for usage. The reporter's argument for keeping it: usage data is the raw material for noticing that a provider silently changed behaviour, for example a field going from `int` to `string`, which breaks a typed DTO in production while a mock frozen months ago keeps passing.

Scope note: this is not a drift detector on its own. Replay returns the already converted result and never runs the `ResultConverter`, so drift belongs at the HTTP boundary (#2128), where the raw bytes are kept. Preserving metadata here is for tests that want to assert on usage.

It was left out of the original PR because metadata holds arbitrary objects and needs its own serialization pass.

ousamabenyounes:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| Docs?         | yes
| Issues        | Fix #2410
| License       | MIT

`Test\Recording\ResultSerializer` writes only a result's *content* to a cassette, so everything a
provider reports alongside the answer is lost on the way in. A replayed result therefore has empty
metadata, and the usage data #2410 asks for cannot be asserted on at all.

This adds the metadata pass that was deferred out of #2129.

### What is preserved

Metadata is written per key, and only in forms that come back as the same type:

* scalars, `null`, and arrays nesting any of these to any depth — verbatim, including a float with
  no fractional part (`signature`, Whisper's `usage`, Perplexity's `citations`/`search_results`, …)
* `FinishReason\FinishReason` — its case plus the untouched provider wording
* `TokenUsage\TokenUsage` and `TokenUsage\TokenUsageAggregation` — every field, and for an
  aggregation its individual usages, so `count()` and the sums survive a round trip

Anything else throws while recording, matching how the class already treats an unsupported result or
delta type. Dropping it silently is the defect this issue is about, and recording is a developer-time
step in a test double: the throw surfaces once, locally, to the person who can act on it.

Non-scalar values are written as a small envelope tagged with a reserved `#type` key. A plain
metadata array using that key is rejected on the write path, so an envelope can never be confused
with provider data, and every envelope field is validated on the way back out: a malformed
`#type` envelope in a hand-edited cassette raises the component's `InvalidArgumentException` rather
than a `\ValueError` or a PHP warning. That applies to the metadata envelopes this PR adds — the
existing content path (`content`, `vectors`, `tool_calls`) is unchanged and still reacts to a
corrupt cassette the way it does on `main`.

```php
$provider = new RecordingProvider($realProvider, new Cassette(__DIR__.'/fixtures/weather.json'));

$result = $provider->invoke('claude-sonnet-4-5', $messages);

$result->asText();
$result->getMetadata()->get('token_usage')->getTotalTokens();   // 46, on record *and* on replay
$result->getMetadata()->get('finish_reason')->getCase();        // FinishReasonCase::STOP
```

Metadata is read after the content pass, because a `StreamResult` only fills its metadata while its
generator is drained — a stream whose usage is promoted by `TokenUsage\StreamListener` replays that
usage too.

`TokenUsageAggregation::getTokenUsages()` is added because an aggregation cannot be serialized
without reading the parts it sums up, and the class exposed no way to do that.

`Cassette::save()` is hardened in the same PR rather than separately, because this change widens
what reaches it: it called `json_encode()` without `JSON_THROW_ON_ERROR`, so a value JSON cannot
represent (`NAN`, `INF`, invalid UTF-8) made it return `false` and the file was overwritten with a
bare newline, destroying every interaction already recorded. It now raises, the write is checked
against the length actually written, the directory creation is checked, and it encodes with
`JSON_PRESERVE_ZERO_FRACTION` so a recorded `2.0` does not replay as `int(2)`. The truncation path
is reachable on `main` today through a result's content alone.

Happy to split the `Cassette` commit out into its own PR if you would rather review it separately.

Sharing one (de)normalizer with the cache bridge is #2413 and is deliberately left alone here.

28 tests are added. 26 of them fail against **unmodified `main`** (`6a95fac3`), with
`src/platform/src` bind-mounted read-only from a detached worktree so the production code is
pristine and only the new tests are applied on top:

```
1) Symfony\AI\Platform\Tests\TokenUsage\TokenUsageAggregationTest::testGetTokenUsagesReflectsUsagesAddedAfterConstruction
Error: Call to undefined method Symfony\AI\Platform\TokenUsage\TokenUsageAggregation::getTokenUsages()

/app/src/platform/tests/TokenUsage/TokenUsageAggregationTest.php:38

2) Symfony\AI\Platform\Tests\TokenUsage\TokenUsageAggregationTest::testGetTokenUsagesReturnsTheAggregatedUsagesInOrder
Error: Call to undefined method Symfony\AI\Platform\TokenUsage\TokenUsageAggregation::getTokenUsages()

/app/src/platform/tests/TokenUsage/TokenUsageAggregationTest.php:27


There were 24 failures:

2) Symfony\AI\Platform\Tests\Test\Recording\ResultSerializerTest::testAggregatedUsageThatIsNotATokenUsageThrows
Failed asserting that exception of type "Symfony\AI\Platform\Exception\InvalidArgumentException" is thrown.

3) Symfony\AI\Platform\Tests\Test\Recording\ResultSerializerTest::testRoundTripPreservesTokenUsageAggregation
Failed asserting that null is an instance of class Symfony\AI\Platform\TokenUsage\TokenUsageAggregation.

/app/src/platform/tests/Test/Recording/ResultSerializerTest.php:163

4) Symfony\AI\Platform\Tests\Test\Recording\ResultSerializerTest::testRoundTripThroughRealJsonPreservesScalarTypes
Failed asserting that null is identical to Array &0 [
    'seconds' => 2.0,
    'ratio' => 1.5,
    'score' => 0.0,
    'count' => 3,
].

/app/src/platform/tests/Test/Recording/ResultSerializerTest.php:370

Tests: 66, Assertions: 125, Errors: 2, Failures: 24.
```

(Excerpt of 5 of the 26; each test file numbers its own failures from 1.)

The two tests that pass on both sides are invariant guards rather than proof:
`testResultWithoutMetadataDoesNotWriteMetadataKey` — a result with no metadata still produces a
byte-identical cassette — and `testAbsentMetadataKeyOnFromArrayIsAccepted` — a cassette recorded
before this change still replays.

GREEN — same tests, same command, with the patch:

```
vendor/bin/phpunit -c src/platform/phpunit.xml.dist \
    --filter '/(ResultSerializerTest|RecordingProviderTest|TokenUsageAggregationTest|CassetteTest)/'

................................................................. 65 / 66 ( 98%)
.                                                                 66 / 66 (100%)

Time: 00:00.087, Memory: 22.00 MB

OK (66 tests, 165 assertions)
```

### Full local suite

Every package's suite plus the PHPStan shards, PHP-CS-Fixer, Deptrac, DOCtor-RST, the docs build and
the CHANGELOG validator, replayed on PHP 8.5 from the repo's own workflow commands:

```
find src -name phpunit.xml.dist -not -path '*/vendor/*' | sort \
    | .github/run-in-packages.sh vendor/bin/phpunit -c '{}' --exclude-group integration

Ran in 93 package(s) with parallelism 8, 1 failed.
```

`src/platform` goes from 826 to 854 tests, all green, no new skips. The one failing package is
`src/mate` (`LoggerTest::testFileWriteFailureThrowsException` and
`::testThrowsExceptionWhenDirectoryCreationFailsDueToPermissions`), identical before and after: both
assert that an unwritable path throws, which does not happen when the suite runs as root in a
container. Untouched by this PR, and present on `6a95fac3` too.

### DIFF
--- a/src/platform/src/Test/Recording/Cassette.php
+++ b/src/platform/src/Test/Recording/Cassette.php
@@ -26,6 +26,8 @@ use Symfony\AI\Platform\Exception\RuntimeException;
  */
 final class Cassette
 {
+    private const DIRECTORY_MODE = 0777;
+
     /**
      * @var list<Interaction>
      */
@@ -114,10 +116,25 @@ final class Cassette
     private function save(): void
     {
         $directory = \dirname($this->path);
-        if (!is_dir($directory)) {
-            mkdir($directory, 0777, true);
+        if (!is_dir($directory) && !@mkdir($directory, self::DIRECTORY_MODE, true) && !is_dir($directory)) {
+            throw new RuntimeException(\sprintf('Cannot create cassette directory "%s".', $directory));
         }
 
-        file_put_contents($this->path, json_encode(['interactions' => $this->interactions], \JSON_PRETTY_PRINT | \JSON_UNESCAPED_SLASHES | \JSON_UNESCAPED_UNICODE)."\n");
+        // JSON_THROW_ON_ERROR: without it an unencodable value (NAN, INF, invalid UTF-8) makes
+        // json_encode() return false, and the file would be overwritten with a bare newline,
+        // destroying every interaction recorded so far.
+        // JSON_PRESERVE_ZERO_FRACTION: without it a float with no fractional part is written as
+        // an integer, so a recorded 2.0 replays as int(2) and the cassette silently retypes what
+        // the provider reported.
+        try {
+            $json = json_encode(['interactions' => $this->interactions], \JSON_PRETTY_PRINT | \JSON_UNESCAPED_SLASHES | \JSON_UNESCAPED_UNICODE | \JSON_PRESERVE_ZERO_FRACTION | \JSON_THROW_ON_ERROR);
+        } catch (\JsonException $exception) {
+            throw new RuntimeException(\sprintf('Cannot encode cassette "%s"; the recorded result holds a value JSON cannot represent.', $this->path), previous: $exception);
+        }
+
+        $contents = $json."\n";
+        if (\strlen($contents) !== file_put_contents($this->path, $contents)) {
+            throw new RuntimeException(\sprintf('Cannot write cassette "%s".', $this->path));
+        }
     }
 }
--- a/docs/components/platform.rst
+++ b/docs/components/platform.rst
@@ -1762,7 +1762,27 @@ closure to normalize those parts before hashing::
     ));
 
 Supported result types are text, structured output (``ObjectResult``), embeddings (``VectorResult``),
-tool calls (``ToolCallResult``) and text streams; result metadata and token usage are not preserved.
+tool calls (``ToolCallResult``) and text streams.
+
+Result metadata is recorded too, so a replayed result can be asserted on for what the provider
+reported alongside the answer - most usefully token usage::
+
+    $result = $provider->invoke('claude-sonnet-4-5', $messages);
+
+    // read the result first: a DeferredResult copies the metadata up when it resolves,
+    // so getMetadata() is empty until asText() (or another as*() call) has run
+    $result->asText();
+
+    $result->getMetadata()->get('token_usage')->getTotalTokens();
+    $result->getMetadata()->get('finish_reason')->getCase();
+
+Metadata is preserved for the values a provider reports at this boundary: scalars, ``null``,
+:class:`Symfony\\AI\\Platform\\FinishReason\\FinishReason` and token usage
+(:class:`Symfony\\AI\\Platform\\TokenUsage\\TokenUsage` and
+:class:`Symfony\\AI\\Platform\\TokenUsage\\TokenUsageAggregation`), plus arrays nesting any of
+those to any depth. Values keep their type across a round trip, floats included. Anything else
+throws while recording rather than being dropped silently, so a cassette never claims to hold
+something it lost.
 
 A cassette stores the model name, the signature and the result. The input itself is never written,
 only its hash, but a recorded answer can still repeat back what the prompt contained, so treat a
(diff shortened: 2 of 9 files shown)


src/platform/src/Test/Recording/Cassette.php
```diff
@@ -114,10 +116,25 @@ private function load(): void
     private function save(): void
     {
         $directory = \dirname($this->path);
-        if (!is_dir($directory)) {
-            mkdir($directory, 0777, true);
+        if (!is_dir($directory) && !@mkdir($directory, self::DIRECTORY_MODE, true) && !is_dir($directory)) {
+            throw new RuntimeException(\sprintf('Cannot create cassette directory "%s".', $directory));
+        }
+
+        // JSON_THROW_ON_ERROR: without it an unencodable value (NAN, INF, invalid UTF-8) makes
```
wachterjohannes, line 123: To answer the question in your description: keep it here, do not split it out.

I reproduced the truncation on `main`, and it is worse than "returns false":

```
########## main ##########
after 1st recording : 277 bytes, interactions=1
record() returned normally
after 2nd recording : 1 byte, interactions=UNREADABLE (file destroyed)

########## with this PR ##########
record() threw RuntimeException
after 2nd recording : 277 bytes, interactions=1
```

A single result whose content JSON cannot encode silently destroys every interaction recorded before it, and `record()` reports nothing. That is a data-losing bug reachable today through content alone, so it belongs with the change that widens what reaches this method, not in a queue behind it.

wachterjohannes (approved):
Verified rather than taken on trust, and it holds up on every point I checked. Your red to green claim reproduces exactly. Verified locally, all green.

Round trip through real JSON, with a mixed metadata payload:

```
finish_reason   : STOP / stop
signature       : sig-abc
usage seconds   : 2.0 (float)
usage nested    : {"a":[1,2.5,null,true]}
token_usage     : TokenUsageAggregation, count=2, total=33
```

`JSON_PRESERVE_ZERO_FRACTION` earns its place, `2.0` survives as a float.

The reserved key check is recursive, which is the part I expected to be able to break:

```
top-level "#type" in metadata array   -> InvalidArgumentException
nested   "#type" in metadata array    -> InvalidArgumentException
object in metadata                    -> InvalidArgumentException
```

On throwing for unsupported metadata: I checked what bridges actually write today, and it is `finish_reason` (a `FinishReason`), `token_usage`, `signature` (string), Whisper's `usage` (array), and Perplexity's `citations`/`search_results` (arrays) through `MetadataDelta`. All covered, so the throw should not fire on any current provider while still catching a future one.

One note inline, answering the question in your description.

wachterjohannes:
Thank you @ousamabenyounes.

PR #2436, opened 2026-08-23
[Agent] Agent to return a lazy Execution instance

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | 
| License       | MIT

`AgentInterface::call()` now returns a lazy `Execution` that is also the result it produces:

```php
// eager reads and the streaming loop are unchanged
$text = $agent->call($messages)->getContent();
foreach ($agent->call($messages, ['stream' => true])->getContent() as $delta) { /* ... */ }

// new: observe the run step by step
foreach ($agent->call($messages) as $update) { /* Progress / Result updates */ }
$result = $agent->call($messages)->onProgress($callback)->getResult();
```

Userland changes:
* New: iterating the execution (or `onProgress()`/`onResult()`) surfaces every model request, tool call and streamed delta as updates
* New: typed accessors on the execution — `asText()`, `asObject()`, `asBinary()`, `asFile()`, `asDataUri()`, `asToolCalls()` and, when streamed, `asStream()`, `asTextStream()`, `asStreamedObject()`, `asPartialJsonStream()` — narrowing the result like `DeferredResult` does (shared via an internal `TypedResultTrait`)
* BC break: `call()` is lazy — the agent only runs on consumption, so side effects and exceptions surface on `getContent()`/`getResult()`/iteration, and result type-checks need `->getResult()` first
* BC break: custom `AgentInterface` implementations and decorators must return an `Execution`
* BC break: `Toolbox\StreamListener` removed; `call()` with `stream` no longer returns a `StreamResult` (its `getContent()` still yields the deltas)

Both consumption styles are demoed in `examples/agent/execution-iterable.php` and `examples/agent/execution-callbacks.php`. See the `UPGRADE.md` entry for migration details.


src/agent/src/Execution/Execution.php
```diff
@@ -0,0 +1,208 @@
...
+        private readonly \Closure $factory,
+        private readonly bool $streamed = false,
+    ) {
+        $this->metadata = new Metadata();
+    }
+
+    /**
+     * @return \Generator<int, UpdateInterface, mixed, void>
+     */
+    public function getIterator(): \Generator
```
wachterjohannes, line 80: The three consumption paths each re-implement the update dispatch, and they do not agree on what they do with an update:

| consumed via | `onProgress` | `onResult` | `$metadata` |
| --- | --- | --- | --- |
| `await()` | fires | fires | (delegates) |
| `getContent()` streamed, so `deltas()` | fires | never fires | populated |
| `foreach ($execution)`, so here | never fires | never fires | stays empty |

Two things fall out of that. `->onResult($cb)` on a streamed execution is silently dead, which is the case where I would most want it since the final metadata only exists once the deltas are drained. And iterating a streamed execution leaves `getMetadata()` empty:

```php
$e = $agent->call($messages, ['stream' => true]);
foreach ($e as $update) {}
$e->getMetadata()->get('token_usage'); // null

$e2 = $agent->call($messages, ['stream' => true]);
foreach ($e2->getContent() as $delta) {}
$e2->getMetadata()->get('token_usage'); // TokenUsage
```

Same run, two consumption styles, two answers, and the second one is what `examples/agent/execution-iterable.php` does.

Could all three funnel through one private dispatch that caches the result, merges the metadata and fires both callback lists, so each entry point only decides what it yields?

src/agent/src/Execution/Runner.php
```diff
@@ -65,47 +56,111 @@ public function __construct(
...
+
+            if ($delta instanceof TextDelta) {
+                $text .= $delta->getText();
+            }
+
+            yield new Progress('delta', 'Received a streamed delta.', $delta);
         }
 
-        return $this->handleToolCalls($budget, $agent, $messages, $options, $toolCallResult);
+        $result = [] !== $toolCalls ? new ToolCallResult($toolCalls) : new TextResult($text);
```
wachterjohannes: Every non-tool-call round ends as a `TextResult` built from the concatenated `TextDelta`s, so a streamed structured-output run loses its object. The platform already built the real `ObjectResult` at this point and exposes it through `PartialObjectStreamListener::getFinalObjectResult()` / `DeferredResult::asObject()`, but the `DeferredResult` is dropped on line 79.

With the real listener attached:

```
partial object     : {"title":"Pancakes","minutes":20}
FINAL result class : Symfony\AI\Platform\Result\TextResult
FINAL getContent() : '{"title":"Pancakes","minutes":20}'
```

So `await()` on `['stream' => true, 'response_format' => Recipe::class]` hands back raw JSON. Not a regression, there was no final result for a stream before, but it is new surface returning a wrong value, and `demo/src/Recipe/Chat.php` is exactly that shape. Keeping the `DeferredResult` around and asking it for the final result would cover it.

UPGRADE.md
```diff
@@ -18,6 +18,55 @@ Agent
...
+   foreach ($result->getContent() as $delta) { /* ... */ } // unchanged
+
+   // or observe the whole execution
+   foreach ($agent->call($messages, ['stream' => true]) as $update) { /* ... */ }
+   ```
+
+   Code relying on the `StreamResult` type specifically — `instanceof StreamResult`, `addListener()` — must be
+   reworked to consume the execution instead (as the `Chat` component's `stream()` now does).
+
+ * The `maxToolCalls` cap is now enforced across the whole agent call. Previously the counter was reset on every
```
wachterjohannes: This was fixed by 1022ca90 and is already in `main`, so nothing changes here. Measured on both branches, streamed and not, `maxToolCalls: 3` aborts after the same 4 model invocations with the same exception. As written it asks people to re-tune a setting that did not move, I would drop the bullet (and the matching one in `src/agent/CHANGELOG.md`).

src/agent/src/MultiAgent/MultiAgent.php
```diff
@@ -64,68 +65,76 @@ public function getName(): string
...
+                return;
+            }
 
-        $this->logger->debug('MultiAgent: Delegating to agent', ['agent_name' => $decision->getAgentName()]);
+            $this->logger->debug('MultiAgent: Delegating to agent', ['agent_name' => $decision->getAgentName()]);
 
-        // Call the selected agent with the original user question
-        return $targetAgent->call(new MessageBag($userMessage), $options);
+            // Call the selected agent with the original user question
+            yield new ResultUpdate($targetAgent->call(new MessageBag($userMessage), $options)->await());
```
wachterjohannes: Nit: `await()` here (and in `SpeechAgent`) swallows the inner execution's updates, so iterating a `MultiAgent` yields no progress at all, not the handoff, not the delegate's tool calls. `yield from` the inner execution and keep only its `Result` would make the handoff visible, which is the one place I would really want to watch it.

wachterjohannes (changes requested):
Nice one, folding streaming into the tool-calling loop makes the whole thing much easier to follow, and the `StreamListener` delta substitution is not missed. I checked the streamed tool call flow against `main` and it comes out identical, sources and aggregated token usage included.

Two things I would like to see addressed before this goes in, plus one stale upgrade note, details inline.

Verified locally, all green.

src/agent/src/MultiAgent/MultiAgent.php
chr-hertel: opened #2444 to have it in backlog but follow-up PRs will address parts of it anyways

chr-hertel:
Thanks @wachterjohannes - should be addressed now :+1:

demo/src/Recipe/Chat.php
```diff
@@ -77,33 +76,24 @@ public function isAwaitingRecipe(): bool
...
+            $recipe = $partial;
 
-                yield $recipe;
-            }
+            yield $recipe;
         }
 
         $assistantMessage = Message::ofAssistant($recipe->toString());
-        $assistantMessage->getMetadata()->add('recipe', $recipe);
         $messages->add($assistantMessage);
```
wachterjohannes, line 97: The switch to `asStreamedObject()` dropped `$assistantMessage->getMetadata()->add('recipe', $recipe);`, which was the only writer of that key. `Chat::getRecipe()` right above reads it and throws without it, `TwigComponent::getRecipe()` catches that to `null`, and `shareRecipe()` returns early on `null`, so the recipe card comes back empty after a reload and sharing by email silently sends nothing.

Same stream through `getRecipeStream()`, then `getRecipe()`:

```
this head          -> RuntimeException: The last message does not contain a recipe.
with the line back -> OK -> Pancakes
```

The demo suite stays green; the only coverage is `tests/E2E/RecipeTest.php`, which needs a live `OPENAI_API_KEY`.

```suggestion
        $assistantMessage = Message::ofAssistant($recipe->toString());
        $assistantMessage->getMetadata()->add('recipe', $recipe);
        $messages->add($assistantMessage);
```

The same hunk also drops the comment explaining why the `isLastMessageFrom(Role::User)` guard is there (stale or reconnecting SSE connections), worth keeping.

wachterjohannes (commented):
All three are addressed, and I verified each: the single `consume()` makes the callbacks and the metadata come out the same whichever way the execution is consumed, a streamed structured-output round now resolves to the platform's `ObjectResult`, and `maxToolCalls` still aborts at the same point as on `main`. Thanks for opening #2444 for the composite agents.

Two small things left, one of them a real (silent) regression in the demo, details inline.

Verified locally, all green.

src/chat/src/Chat.php
```diff
@@ -56,12 +58,22 @@ public function stream(UserMessage $message): \Generator
         $messages = $this->store->load();
         $messages->add($message);
 
-        $result = $this->agent->call($messages, ['stream' => true]);
+        $execution = $this->agent->call($messages, ['stream' => true]);
+
+        // the execution's deltas are the streamed answer; accumulate the text to persist it afterwards
+        $content = '';
+        foreach ($execution->asStream() as $delta) {
```
wachterjohannes, line 65: This was the only production use of `ChatStreamListener`, which is public API since `v0.12.0`, and `src/chat/CHANGELOG.md` is untouched. No behaviour change, it only had `onDelta`/`onComplete`, so an aborted stream persisted nothing before either, but the class is now orphaned in the component that owns it. Deprecate it, or add a line saying `Chat::stream()` no longer uses it.

chr-hertel: gonna remove it, good catch, thanks! 👍

wachterjohannes (approved):
Both addressed, and I checked them rather than the replies: the recipe metadata is back on the assistant message, so `getRecipe()` resolves again after a streamed turn. Removing `ChatStreamListener` outright is the better call than the changelog line I asked for, and the bookkeeping is complete.

wachterjohannes:
Thank you @chr-hertel.

PR #2437, opened 2026-08-23
[Examples][Platform] Add record-and-replay integration testing for bridges

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | -
| License       | MIT

This would enable us to
1. build more high-level tests for bridges: test:data -> fixture:http -> bridge -> test:output
2. run the examples with http fixtures and assert on output against snapshots

---

Claude says:

Builds on the `Test\Replay` cassette scaffolding from #2128 to harden result converters against real provider payloads: record once against the live API, replay deterministically in CI without credentials, through the full `Platform → ModelClient → ResultConverter` path.

- `Test\Replay\AbstractBridgeReplayTestCase`: base class for bridge replay tests with one cassette per scenario (missing cassette = skip).
- The `examples/` corpus becomes the record/replay harness: `./runner --record openai` records cassettes (credentials redacted, binary bodies elided to metadata stubs) and refreshes replay goldens, `ExamplesReplayTest` replays every recorded example offline against its golden — new `examples-replay` CI job, no keys needed.
- Pilot on OpenResponses/OpenAI: real recordings of `openai/chat.php`, `stream.php`, `stream-reasoning.php`, plus bridge `ReplayTest`s (OpenResponses, OpenAI TTS) with focused cassettes for delta-level and error-mapping assertions. `toolbox/brave.php` proves the agent tool loop offline: one cassette interleaves 15 FIFO interactions across OpenAI, Brave search and scraped PHP Foundation blog pages — and recording it surfaced that the example had printed `Array` since the GPT-5 bump (reasoning models return a `MultiPartResult`), fixed here. `openai/audio-output.php` covers binary responses: the TTS audio is elided to a 2.7 KB metadata stub and replays as a placeholder.
- `AbstractBridgeReplayTestCase` and body elision documented in the Testing section of `docs/components/platform.rst`, the workflow in `examples/README.md`; recording stays a local maintainer task (e.g. alongside the pre-release live run).


examples/tests/ExamplesReplayTest.php
```diff
@@ -0,0 +1,86 @@
...
+        );
+
+        if (null === $fixture) {
+            return;
+        }
+
+        $this->assertSame(
+            $this->normalize((string) file_get_contents($fixture)),
+            $this->normalize($process->getOutput()),
+            \sprintf('Example "%s" produced output differing from its recorded fixture, consider running `./runner for this example to record.', $examplePath),
```
wachterjohannes: The message is missing a backtick and the `--record` flag, and it is the first thing you see when a golden drifts.

```suggestion
            \sprintf('Example "%s" produced output differing from its recorded fixture, consider running "./runner --record" for it.', $examplePath),
```

src/platform/src/Test/Replay/AbstractBridgeReplayTestCase.php
```diff
@@ -0,0 +1,61 @@
...
+     */
+    protected function platformForCassette(string $scenario): PlatformInterface
+    {
+        $path = $this->cassetteDirectory().'/'.$scenario.'.json';
+
+        if (!is_file($path)) {
+            $this->markTestSkipped(\sprintf('No cassette recorded for scenario "%s".', $scenario));
+        }
+
+        return $this->createPlatform(new CassetteHttpClient(new HttpCassette($path), record: false));
```
wachterjohannes, line 60: Replay is FIFO and never looks at the request, so this covers the response half of the pipeline only. `HttpCassette` already writes a `signature` per request, but `next()` returns just `['response']` and nothing reads it.

I sent a wrong model through the recorded runs:

```php
'body' => $this->encodeJsonBody(array_merge($options, ['model' => 'totally-wrong-model'], $payload)),
```

```
OK (5 tests, 10 assertions)
```

A normalizer dropping an option or a client hitting the wrong endpoint replays just as happily. That gap is from #2128, not from this PR, but this is what puts the harness in CI as bridge integration testing. Opting into a signature check here would close it, happy to do it as a follow-up.

wachterjohannes (changes requested):
Good concept and it holds up: I injected a converter regression (`TextResult`) and a streaming one (`TextDelta`) into the OpenResponses bridge and both examples failed with a readable diff, offline and without keys. Recording locally but writing the golden from the *replay* run is the right way round. Verified locally, all green.

One thing I would change before this lands, and GitHub won't let me anchor it because the file is too large to diff: `examples/tests/fixtures/toolbox/brave.json` is 563 KB, and 300 KB of that is three verbatim pages of `thephp.foundation`, one of them a 202 KB blog index. That is third-party content committed into the repo, and it is the bulk of the ~800 KB of fixtures this PR adds. The agent only needs enough text to answer the prompt, so truncating scraped bodies on record, or eliding non-API responses the way binary bodies already are, would keep the test and lose the payload.

- The description says `Docs? yes`, but no `docs/` file changed. `AbstractBridgeReplayTestCase` is published API of the platform package and belongs in the Testing section of `docs/components/platform.rst`, next to `CassetteHttpClient`.
- `audio-output.out` is a binary file to git because of the leading NUL in the placeholder, so the golden never shows up in a diff. Worth dropping the NUL if the format detection can live without it.
- The four OpenResponses cassettes are hand-seeded (`resp_text`, `call_1`), while the base class docblock says cassettes are recorded against the real provider. Fine as is, the wording should just admit it.

src/platform/src/Test/Replay/AbstractBridgeReplayTestCase.php
wachterjohannes: Tracked as #2443, so this is not something you have to solve here.

examples/tests/ExamplesReplayTest.php
chr-hertel: hahaha, one of the few things i coded myself ... should stop doing that :joy:

chr-hertel: Applied in b06bb7dee, thanks — took the suggestion verbatim.

src/platform/src/Test/Replay/AbstractBridgeReplayTestCase.php
chr-hertel: Agreed — FIFO replay only pins the response half, and your wrong-model experiment shows it plainly. The opt-in shape you sketch in #2443 (strict for AbstractBridgeReplayTestCase, relaxed for the examples harness) sounds right to me, as does keeping the normalized body for a readable diff. Leaving it to the follow-up.

chr-hertel:
all addressed, thanks @wachterjohannes. a bit skeptical about the brave thingy but safe enough to try - it's mostly about the bridges anyways.

docs/components/platform.rst
```diff
@@ -1834,7 +1834,45 @@ cookies and provider account identifiers) are replaced with ``[redacted]`` in bo
 headers before the cassette is written, so a cassette is safe to commit. Per-request trace headers
 (``date``, ``cf-ray``, correlation and request ids, proxy latencies) are dropped on write, so that
 re-recording a cassette produces a diff of what the provider actually changed instead of noise;
-rate limiting headers are kept, because the converters read them.
+rate limiting headers are kept, because the converters read them. Binary response bodies (generated
+images, audio, ...) and scraped web pages are not stored: the cassette keeps a metadata stub (status,
+headers, byte size) and replay serves a small placeholder body, so the real converter still runs
+without committing opaque bytes or third-party content.
```
wachterjohannes: The scraped text still lands in the repository, one layer up. On record the scraper receives the real bytes and feeds them back to the model as a tool output, and that request is what gets recorded: a single `function_call_output` in `brave.json` carries 36 KB of `thephp.foundation` prose, and roughly 125 KB of the file's remaining 251 KB is page text. The response half is clean, the request half is not.

Same sentence in `examples/README.md:118`. Either weaken it, or elide on the request side too.

src/platform/src/Test/Replay/CassetteHttpClient.php
```diff
@@ -104,27 +108,55 @@ public function withOptions(array $options): static
...
+        if (str_contains($contentType, 'text/event-stream')) {
             return 'sse';
         }
 
         if (str_starts_with($body, 'data:') || str_starts_with($body, 'event:')) {
             return 'sse';
         }
 
+        // Binary bodies cannot be committed as JSON, and scraped web pages are bulky
+        // third-party content: neither belongs in the repository, so both are elided.
```
wachterjohannes: Eliding a body whose content flows into the next request makes record and replay diverge by construction: recorded is the real page, replayed is the placeholder. It is visible in the golden, where the page titles became `[202201 bytes of body elided by the cassette]`.

Nothing to solve here, but those interactions could then never pass the signature check from #2443, however often they are re-recorded.

wachterjohannes (commented):
All four addressed, thanks. Verified locally, all green.

Two notes inline, neither blocking: the new sentence about third-party content promises more than the elision delivers, and eliding `text/html` has a consequence for #2443 worth knowing about.

src/platform/src/Test/Replay/CassetteHttpClient.php
chr-hertel: you're confusing me - i only did this because of your comment :joy:

anyhow, will revert - suits me better as well ;)

chr-hertel:
so, i think we should pay attention to fixture sizes, but having a 800kb file is not that much of harm i'd say

src/platform/src/Test/Replay/CassetteHttpClient.php
wachterjohannes: sorry, thats on me :) the eliding was my own suggestion in the first round. what made me doubt it once i saw it: the text is still in the repo anyway, the scraper feeds it back to the model so it just moved from the response into the next request. and the golden got worse with the placeholders instead of the titles. so revert is good.

wachterjohannes (approved):
fixture size is fine for me too, its mostly about the bridges anyway. all good from my side!

wachterjohannes:
Thank you @chr-hertel.

PR #2440, opened 2026-08-23
[MCP Bundle] Resolve completion providers from the container

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | -
| License       | MIT

A completion provider with a constructor dependency always fails:

```php
final class CityCompletionProvider implements ProviderInterface
{
    public function __construct(private CityRepository $cities)
    {
    }

    public function getCompletions(string $currentValue, SessionInterface $session): array
    {
        return $this->cities->matching($currentValue);
    }
}
```

```php
#[McpTool(name: 'find_venue')]
public function findVenue(
    #[CompletionProvider(provider: CityCompletionProvider::class)] string $city,
): string {
}
```

Calling `completion/complete` returns *"Error while handling completion request"*.

The SDK resolves `provider:` at request time from the container the server was built with, and falls back to `new $provider()` when the container does not have it. The bundle hands each server a service locator built in `McpPass`, but it only holds element handlers — tools, prompts, resources and apps — so a provider is never in it. The fallback then instantiates it with no arguments and dies with an `ArgumentCountError`, which the SDK reports as the generic message above.

Providers with no constructor arguments work by accident, which is why this is easy to miss.

This autoconfigures every service implementing `Mcp\Capability\Completion\ProviderInterface` with an `mcp.completion_provider` tag and adds them to each server's locator, keyed by both class name and service id so either spelling of `provider:` resolves. `McpPassTest` covers the locator contents.

Found while building a demo application against the bundle: the provider worked as a plain class and broke the moment it needed a repository.

### DIFF
--- a/src/mcp-bundle/src/DependencyInjection/McpPass.php
+++ b/src/mcp-bundle/src/DependencyInjection/McpPass.php
@@ -189,9 +189,26 @@ final class McpPass implements CompilerPassInterface
         /** @var array<string, array<string, array{template: string, meta: array<string, mixed>|null}>> $appTemplates */
         $appTemplates = $container->hasParameter('mcp.servers.app_tool_templates') ? $container->getParameter('mcp.servers.app_tool_templates') : [];
 
+        // Resolved by class name at request time, so every server needs all of them.
+        $completionProviders = [];
+        foreach ($container->findTaggedServiceIds('mcp.completion_provider') as $serviceId => $tags) {
+            $definition = $container->getDefinition($serviceId);
+            if ($definition->isAbstract()) {
+                continue;
+            }
+
+            $class = $container->getParameterBag()->resolveValue($definition->getClass() ?? $serviceId);
+            $completionProviders[$class] = new Reference($serviceId);
+            $completionProviders[$serviceId] ??= new Reference($serviceId);
+        }
+
         foreach (array_keys($servers) as $server) {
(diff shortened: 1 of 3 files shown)


chr-hertel:
Closing this — the premise is wrong, and I should have checked it more carefully before opening.

`McpPass` already resolves completion providers from the container:

```php
if (\in_array($tag, ['mcp.prompt', 'mcp.resource_template'], true)) {
    foreach ($this->completionProviderClasses($class, $method) as $providerClass) {
        ...
        $serviceReferences[$server][$providerClass] ??= new Reference($providerClass);
    }
}
```

That covers `#[CompletionProvider]` on `#[McpPrompt]` arguments and `#[McpResourceTemplate]` variables — which is every place the protocol can reach one. `completion/complete` takes a `ref/prompt` or a `ref/resource` and nothing else, so a provider named on a `#[McpTool]` argument is never invoked, and excluding `mcp.tool` from that guard is correct rather than a gap.

I verified it against a real application: providers with constructor dependencies (a repository each) complete correctly on both prompt arguments and resource-template variables with this branch reverted.

What misled me was my own application annotating tool arguments with `#[CompletionProvider]` and those completions never arriving. That is a no-op by design, not a container-resolution failure.

Sorry for the noise. #2439 and #2441 stand on their own.

PR #2446, opened 2026-08-28
[Agent] Add Human-in-the-Loop tool approval and resumable checkpoints

saifulferoz:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | 
| License       | MIT

### Description

Introduces **Human-in-the-Loop (HITL)** capabilities and **Resumable State Checkpointing** for autonomous agents in `symfony/ai-agent`.

In enterprise applications, agents should not execute sensitive actions (payments, database mutations, sending emails, external mutations) without explicit human confirmation.

### Summary of Changes

1. **`#[RequiresApproval]` Attribute & Dynamic Policies (`ApprovalPolicyInterface`)**:
   - Declaratively require human approval for tools with customizable prompt templates and authorized roles.
   - Dynamic policy evaluation via `ApprovalPolicyInterface` (e.g. threshold-based approvals).

2. **Durable State Suspension & Checkpoints (`ExecutionCheckpoint`, `CheckpointSigner`)**:
   - Suspends agent execution into an immutable `ExecutionCheckpoint` and returns an `ApprovalPendingResult`.
   - Generates tamper-proof HMAC-SHA256 tokens for transport over HTTP requests, webhooks, or emails.
   - Storage abstraction via `CheckpointStoreInterface` and `InMemoryCheckpointStore`.

3. **Agent Resumption Engine (`$agent->resume()`)**:
   - Resume execution with `ApprovalDecision::approve()`, `ApprovalDecision::reject($reason)` (which feeds rejection feedback to the LLM for self-correction), or `ApprovalDecision::modify($args)`.
   - Contract added to `AgentInterface`, `Agent`, `MockAgent`, `MultiAgent`, `SpeechAgent`, and `TraceableAgent`.

4. **Documentation & Examples**:
   - Component guide in `docs/components/agent.rst`.
   - Standalone runnable example in `examples/agent/human_in_the_loop.php`.

*(Note: The AI Bundle integration and CLI commands `ai:agent:list-pending` / `ai:agent:approve` are prepared on `feature/ai-bundle-approval-commands` and will follow in a dedicated PR once this core component PR is reviewed and merged).*

### DIFF
--- a/docs/components/agent.rst
+++ b/docs/components/agent.rst
@@ -664,6 +664,51 @@ The sources can be fetched from the metadata of the result after the agent execu
 
 See `Anthropic Toolbox Example`_ for a complete example using sources with Wikipedia tool.
 
+Human-in-the-Loop (HITL) & Tool Approval
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+For sensitive operations (such as executing payments, modifying database records, or sending emails), you can require human confirmation before a tool is executed using the :class:`Symfony\\AI\\Agent\\Approval\\Attribute\\RequiresApproval` attribute::
+
+    use Symfony\AI\Agent\Approval\Attribute\RequiresApproval;
+    use Symfony\AI\Agent\Toolbox\Attribute\AsTool;
+
+    #[AsTool('transfer_funds', 'Transfers money from user account to recipient')]
+    #[RequiresApproval(prompt: 'Confirm transfer of ${amount} to account #{toAccount}', roles: ['ROLE_FINANCE_ADMIN'])]
+    final class PaymentTool
+    {
+        public function __invoke(string $toAccount, float $amount): string
+        {
+            // Execute transfer
+            return sprintf('Transferred $%0.2f to %s', $amount, $toAccount);
+        }
+    }
+
+When an agent encounters a tool that requires approval, execution is suspended and returns an instance of :class:`Symfony\\AI\\Agent\\Approval\\ApprovalPendingResult`::
+
+    use Symfony\AI\Agent\Approval\ApprovalDecision;
+    use Symfony\AI\Agent\Approval\ApprovalManager;
+    use Symfony\AI\Agent\Approval\ApprovalPendingResult;
+    use Symfony\AI\Agent\Approval\Checkpoint\CheckpointSigner;
+    use Symfony\AI\Agent\Approval\Checkpoint\InMemoryCheckpointStore;
+
+    $store = new InMemoryCheckpointStore();
+    $signer = new CheckpointSigner($secret);
+    $approvalManager = new ApprovalManager([], $store, $signer);
+
+    $agent = new Agent($platform, $model, toolbox: $toolbox, approvalManager: $approvalManager);
+
+    $result = $agent->call('Transfer $500 to ACC-99')->getResult();
+
+    if ($result instanceof ApprovalPendingResult) {
+        $checkpoint = $result->getCheckpoint();
+        $token = $result->getToken(); // Secure, cryptographically signed token
+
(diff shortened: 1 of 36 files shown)


mahono:
These are really great ideas. 🙏

Would it make sense to split this up into two PRs?

saifulferoz:
Hi @mahono, thanks for the great feedback! 🙏

Totally agree — I've updated this PR to focus strictly on the **Agent component** (`[Agent] Add Human-in-the-Loop tool approval and resumable checkpoints`).

The **AI Bundle** integration (`ai:agent:list-pending`, `ai:agent:approve`, and DI autoconfiguration) has been isolated to a dedicated branch (`feature/ai-bundle-approval-commands`) and will be submitted as a follow-up PR once this core component PR lands. 🚀

mahono:
I should have been clearer about what I had in mind. I was thinking about separating the resume/checkpoint part from the HITL part. But maybe that does not make sense.

Let's wait for more useful feedback. ;o)

mahono:
Picking my earlier comment back up — I've spent some time with the branch now,
and I do think separating the two halves is worth it.

The state half is already independent of approval: nothing in Approval/Checkpoint/ references an approval type. What ties it to HITL is essentially one thing, `AgentInterface::resume(ExecutionCheckpoint|string, ApprovalDecision)` — that
makes approval part of the core contract of every agent.

Which matters because suspension has other callers: a tool that needs to ask the user something mid-turn, a genuinely async tool, budget exhaustion, resuming after a worker crash. None of those resume with an `ApprovalDecision`. A neutral
resume input, with `ApprovalDecision` as one implementation, would leave the HITL flow reading exactly as it does today — and make the suspend/resume half a much smaller review on its own.

I also ran the branch while reading it. Two things I'd flag before it lands:

- a checkpoint containing image/audio/document content fatals — the serializer calls `getData()`/`fromData()` on `Image`/`Audio`/`Document`, which don't have them (PHPStan on that one file shows it)
- approval tokens never expire, and the same one can be resumed repeatedly, re-running the approved tool each time

There are a few smaller ones too. Happy to write them all up as a proper review if that's useful — I just didn't want to bury the design question underneath it.

(Disclaimer: I did the research and wrote this message with help of Claude Opus 5.)

chr-hertel:
Hi @saifulferoz, thanks for reaching out and working on this topic.
Did you already see the [Cookbook article about Human-in-the-Loop](https://ai.symfony.com/cookbook/human-in-the-loop)?

That event-based solution is def more low-level, but the current scope we're looking for. The Agent will become stateful soonish - not only because of HITL, and then we could revisit if something is still missing.

For now I would recommend to put your work to hold here, or switch to starting with a additional lib on top of `symfony/ai-agent`.

Cheers!

saifulferoz:
Hi @chr-hertel and @mahono,

Thank you both so much for taking the time to review and for sharing such valuable context and insights! 🙏

@chr-hertel: I appreciate the pointer to the Cookbook article! It makes complete sense to keep things aligned with the current event-driven model while the team works on the broader stateful agent roadmap. Putting this on hold for now sounds like the right call so we can revisit once the core stateful architecture takes shape.

@mahono: Really great analysis regarding separating neutral state checkpointing/resumption from the specific approval mechanism — decoupling suspension from approval decisions makes total sense for broader use cases (async tools, budget limits, worker recovery). Will definitely keep this design in mind when the stateful agent foundation lands.

Happy to pause this for now. Looking forward to seeing the stateful agent evolution in Symfony AI! 🚀

saifulferoz:
Closing this PR for now to keep the PR queue clean while the stateful agent roadmap is being designed. We can revisit once the stateful foundation lands. Thanks again everyone!

PR #2450, opened 2026-08-28
[Agent][Chat][Platform] Replay the assistant turn as the provider produced it

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| Docs?         | yes
| Issues        | n/a
| License       | MIT

Alternative proposal to #2390 — /cc @fabpot @wachterjohannes

The intent was broader than #2390: across Agent, Chat, the demo and the examples we usually
rebuild the assistant turn from its text (`Message::ofAssistant($result->asText())` and friends),
where providers expect us to hand back everything they gave us. Anthropic answers a rebuilt turn
with a 400, the Responses API needs its reasoning items back untouched when `store: false`, and
Gemini needs its thought signatures resent.

Summary:

* **Platform**: `Message::ofAssistant($result)` now takes a streamed result just like a buffered
  one, so replaying a turn no longer depends on how it arrived. `StreamResult` collects the turn as
  it is read (draining it if the caller never iterated), and the reassembly itself lives in the new
  `Result\Stream\AssistantMessageStreamListener` — in Platform, because the delta vocabulary is
  bridge knowledge. Thanks @wachterjohannes for that argument on #2390.
* **Agent**: a tool-calling round adds the turn the provider produced as one assistant message
  instead of the text and the tool calls as two, and a streamed round returns the result its
  buffered counterpart would — a thinking block survives as a `ThinkingResult` rather than being
  flattened to its text.
* **Chat**: `Chat::stream()` persists the turn whole, reading it off the execution's result
  instead of re-accumulating the text.
* **OpenResponses**: the normalizer emits output items in model order; it used to drop the
  assistant's text whenever the turn carried a tool call.
* **Mistral**: a reasoning turn is replayed as a `thinking` chunk of the content list — the
  shape the API accepts and `Llm\ResultConverter` already parses — instead of the generic
  contract's `reasoning_content`, which Mistral answers with a 422. The shared contract keeps
  emitting it for the endpoints #1855 added it for.
* **Cerebras**: the trace is dropped there instead — it rejects `reasoning_content` with a 400
  and accepts no shape of its own to put it in.
* Cleaned up the lossy call sites in `examples/` and `demo/`.

Tests: unit tests for the reassembler and `StreamResult`, an OpenResponses `AssistantReplayTest`,
an Anthropic bridge replay test over a recorded thinking+tool_use stream, and three new multi-turn
agent examples with committed cassettes. Their goldens are the proof —
`Thinking(signed) + ToolCall(clock)` in one assistant message, where `main` has just
`ToolCall(clock)`.

Rebased onto #2436, which removed both listeners this branch hooked into (`Toolbox\StreamListener`
and `ChatStreamListener`) and rewrote the tool loop, so the Agent and Chat sides were re-landed
against the lazy `Execution` rather than replayed.

Two caveats worth knowing while reading the diff:

* A `BinaryResult` next to a tool call now maps to a `File` part instead of throwing, so the turn
  is constructible — but no assistant message normalizer emits a `File` yet, so binary output still
  does not round-trip.
* `StreamResult::getAssistantMessage()` must not be called from inside a loop over the stream: a
  suspended generator is indistinguishable from an abandoned one, so the drain that finishes an
  abandoned stream runs and the loop ends on the next delta. Attach the listener to read the turn
  mid-stream.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -64,6 +64,9 @@ Agent
    Code relying on the `StreamResult` type specifically — `instanceof StreamResult`, `addListener()` — must be
    reworked to consume the execution instead (as the `Chat` component's `stream()` now does).
 
+ * A tool-calling round adds one assistant message per turn — text, thinking blocks and tool calls
+   together — instead of splitting it. Code asserting on the message bag's shape needs updating.
+
 AI Bundle
 ---------
 
@@ -91,6 +94,9 @@ Chat
    streamed deltas and persists the assistant message, including the result metadata, once the stream is complete.
    Since `AgentInterface::call()` no longer returns a `StreamResult` (see the Agent section), there is no stream to
    attach the listener to anymore.
+
+ * `Chat::stream()` persists the assistant turn whole — a `Thinking` part now survives where plain text
+   was stored. Code reading persisted history must expect structured content.
 
 MCP Bundle
 ----------
--- a/src/platform/src/Result/Stream/AssistantMessageStreamListener.php
+++ b/src/platform/src/Result/Stream/AssistantMessageStreamListener.php
@@ -1,0 +1,252 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\AI\Platform\Result\Stream;
+
+use Symfony\AI\Platform\Message\AssistantMessage;
+use Symfony\AI\Platform\Message\Content\ContentInterface;
+use Symfony\AI\Platform\Message\Content\Text;
+use Symfony\AI\Platform\Message\Content\Thinking;
+use Symfony\AI\Platform\Result\Stream\Delta\DeltaInterface;
+use Symfony\AI\Platform\Result\Stream\Delta\TextDelta;
+use Symfony\AI\Platform\Result\Stream\Delta\ThinkingComplete;
+use Symfony\AI\Platform\Result\Stream\Delta\ThinkingDelta;
+use Symfony\AI\Platform\Result\Stream\Delta\ThinkingSignature;
+use Symfony\AI\Platform\Result\Stream\Delta\ThinkingStart;
+use Symfony\AI\Platform\Result\Stream\Delta\ToolCallComplete;
+use Symfony\AI\Platform\Result\Stream\Delta\ToolCallStart;
+
+/**
+ * Rebuilds the assistant message a provider would have returned unstreamed, from its deltas.
+ *
+ * Attach it to a {@see \Symfony\AI\Platform\Result\StreamResult} and read
+ * {@see self::getAssistantMessage()} once the deltas of interest have been consumed.
+ *
+ * The delta vocabulary differs per bridge: not every bridge announces a thinking block before its
+ * first chunk or a tool call before its arguments, and the Responses API emits a reasoning item's
+ * signature only after the block has closed.
+ *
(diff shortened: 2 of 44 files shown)


src/platform/src/Result/StreamResult.php
```diff
@@ -31,6 +50,31 @@ public function __construct(
...
+     * collected so far, since the generator cannot be resumed while its body is running.
+     */
+    public function getAssistantMessage(): AssistantMessage
+    {
+        $stream = $this->stream?->get();
+
+        // A stream nobody read is drained here. One left suspended can only be finished while the
+        // caller still holds it - a fresh generator would re-read the delta it stopped on - and
+        // one that is running, ended, or already collected gives back what it collected.
+        if (!$this->consuming && !$this->finished && (null !== $stream || !$this->started)) {
```
wachterjohannes, line 72: Non-blocking: the drain only happens while the caller still holds the generator, so it misses the form callers actually write.

```
foreach ($r->getContent() as $d) { break; }           -> 'one '
$g = $r->getContent(); foreach ($g as $d) { break; }  -> 'one two three'
never consumed at all                                 -> 'one two three'
```

Nothing in the repo hits it, `onComplete` does not fire on an abandoned stream and the agent always drains. But `Message::ofAssistant($streamResult)` is documented as taking a streamed result like a buffered one, and there it returns a partial turn depending on whether a variable happens to still be alive. Worth either draining from the suspended position or saying so on the method.

src/platform/src/Result/Stream/AssistantMessageStreamListener.php
```diff
@@ -0,0 +1,255 @@
...
+        $this->content = [];
+        $this->toolCallSlots = [];
+        $this->openThinking = null;
+        $this->lastThinking = null;
+    }
+
+    /**
+     * Feeds a single delta in, for callers driving the reassembly themselves.
+     */
+    public function accumulate(DeltaInterface $delta): void
```
wachterjohannes, line 92: Non-blocking: `ChoiceDelta` falls through, so a stream carrying one reassembles to an empty message.

```
StreamResult yielding ChoiceDelta([TextDelta('candidate one')])
  -> parts: 0, asText(): NULL
```

`Gemini/ResultConverter.php:160` and `VertexAi/ResultConverter.php:177` both emit it on the multi-candidate path, which the comment there calls exotic, so the reach is small. Unwrapping the inner deltas or naming the limitation in the class docblock next to the other bridge caveats would close it.

UPGRADE.md
```diff
@@ -18,6 +18,11 @@ Agent
    `Toolbox\ToolExecutorInterface` (default: `SequentialToolExecutor`), so a custom execution strategy can be
    plugged in via the `toolExecutor` argument.
 
+ * A streamed tool call is now replayed as the single assistant message the provider produced -
```
wachterjohannes: Non-blocking: this reads as streaming-only, but `Runner.php` changes the buffered path identically, `handleToolCalls(..., Message::ofAssistant($result))` plus `$messages->add($providerResponse ?? ...)`, so every bridge now replays the full turn rather than the tool calls alone.

Also missing a `Chat` entry in this section: `ChatStreamListener` now stores a structured `AssistantMessage` (Thinking/ToolCall parts) where it stored plain text, which anything reading persisted history will notice.

wachterjohannes (approved):
One thing before merging: `review.md` is committed at the repo root, 103 lines of review notes that came in with `b05bb3ddf`.

Otherwise this is good. I checked the claim rather than the description: the golden really does carry `Thinking(signed) + ToolCall(clock)` as one assistant message where `main` has just `ToolCall(clock)`, and all eight recorded examples replay clean, the three new ones included. Putting the reassembler in Platform is the right home, and `StreamListener` losing 108 lines follows from it. Three non-blocking notes inline.

Verified locally, all green.

PR #2451, opened 2026-08-28
[MCP Bundle] Support the 2026-07-28 revision on the existing endpoint

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | -
| License       | MIT

Adds support for the 2026-07-28 revision — no `initialize` handshake, no session, every request
self-describing — **on the endpoint that is already there**.

`mcp/sdk` 0.8 builds a dispatcher per era and lets the HTTP transport classify each request before
anything else looks at it: an envelope claiming a modern revision goes to the modern dispatcher, the
handshake and its session's later requests go to the other. `Builder::build()` already carries both
legs and `McpController` already runs a `StreamableHttpTransport`, so there is no second controller,
no second route and no per-server mode to pick. STDIO carries the handshake era alone, having no
per-request envelope to classify.

What is left to configure is the modern leg itself.

```yaml
mcp:
    servers:
        api:
            protocol_versions: ['2025-11-25', '2026-07-28']
            http: { path: /mcp }
            registry: '*'
```

- **`protocol_versions`** — narrows the revisions the server answers for. Left unset it inherits the
  SDK's own support in both eras. Each era takes what belongs to it: the modern revisions listed become
  the only ones that leg answers for (`setModernVersions()`), listing none refuses the era
  (`withoutModernEra()`), and a handshake-era revision pins the handshake to it
  (`setProtocolVersion()`). The SDK either negotiates over everything or pins to one revision, so
  narrowing the handshake era to a subset is rejected at compile time.
- **`request_state`** — a handler asking for another round trip has no session to keep progress in, so
  the state travels through the client HMAC-signed. The key must reach every process that might serve
  the retry.
- **`cache`** — cache hints (`ttl_ms`, `private`/`public` scope, per-method overrides). `0` refuses
  caching.
- **`subscriptions`** — delivery for `subscriptions/listen`, which replaces the held-open GET stream.
  Under PHP-FPM the publisher is a different worker from the stream, so `memory` cannot reach it.

### Also

- `clients.<name>.roots` now points at a `RootsCallbackInterface` service instead of being a boolean
  capability flag, and connections gain `getProtocolVersion()` and `sendRootsListChanged()`.

### Notes

- Replaces #2414, which was written against an SDK branch where a stateless endpoint was a separate
  thing to mount, and could not go green before `mcp/sdk` 0.8 was released.
- **Why the `BC Break` label:** since the SDK's default answers for every modern revision, existing
  endpoints start accepting 2026-07-28 requests once the constraint moves to `^0.8`, with no config
  change. That revision removed server-initiated requests, so a handler calling `$gateway->sample()`
  or `$gateway->listRoots()` fails for clients speaking it while continuing to work for handshake-era
  clients on the same endpoint. `UPGRADE.md` covers it; listing only handshake-era revisions under
  `protocol_versions` is the opt-out, and `supportsSampling()` / `supportsRoots()` guard a handler
  that wants to serve both.
- Roots, sampling and MCP logging are deprecated by the same revision (SEP-2577, earliest removal
  2027-07-28) and `mcp/sdk` 0.8 triggers a deprecation per configured handler — including the one
  behind `clients.<name>.forward_server_logs`, which defaults to `true`. Documented, not changed.
- The tasks extension (SEP-2663) is left out: 0.8.0 ships the extension framework but no tasks
  implementation. Worth a follow-up once it lands.

190 tests / 575 assertions, PHPStan and CS clean against `mcp/sdk` v0.8.0.

### DIFF
--- a/docs/bundles/mcp-bundle.rst
+++ b/docs/bundles/mcp-bundle.rst
@@ -600,6 +600,168 @@ This wraps the configured Symfony session handler (e.g. Redis, database, filesys
 your application uses for HTTP sessions) with a JSON envelope for application-level TTL.
 Expired sessions are cleaned up lazily on read.
 
+The Modern Era
+..............
+
+The 2026-07-28 revision drops the ``initialize`` handshake and the session that went with it: every
+request describes itself, so any worker can answer any request, and no long-running connection is
+needed.
+
+Both eras are served **on the same endpoint**, with nothing to configure. The SDK builds a dispatcher
+per era, and the HTTP transport classifies every request before anything else looks at it: an envelope
+claiming a modern revision goes to the modern dispatcher, the ``initialize`` handshake and its
+session's later requests go to the other. A client picks nothing.
+
+STDIO carries the handshake era alone — it has no per-request envelope to classify — so a server
+configured for both transports simply serves the modern era over HTTP only.
+
+By default the HTTP endpoint supports every revision the SDK supports and negotiates with the client.
+Narrow the supported revisions through ``protocol_versions``:
+
+.. code-block:: yaml
+
+    # config/packages/mcp.yaml
+    mcp:
+        servers:
+            api:
+                protocol_versions: ['2025-11-25', '2026-07-28']
+                http:
+                    path: /mcp
+                registry: '*'
+
+Each era takes what belongs to it. The modern revisions listed become the only ones that leg answers
+for; listing none of them refuses the modern era, leaving the handshake one alone. A handshake-era
+revision pins the handshake to exactly that one — the SDK either negotiates over everything it knows
+or is pinned to a single revision, so narrowing that era to a subset is rejected when the container is
+compiled.
+
+The options below tune the modern-era leg. They are inert for handshake-era clients, which keep being
+served exactly as before. The converse does not hold: the modern leg is on for every server, including
+one written before the revision existed, and it drops things the handshake era offers.
+
+.. caution::
+
+    **The 2026-07-28 revision has no server-initiated requests.** Sampling and roots were removed with
+    it, so a handler calling ``$gateway->sample()`` or ``$gateway->listRoots()`` fails for a
+    2026-07-28 client — *"This protocol revision has no server-initiated requests: sampling and roots
+    were removed with it, so take what you need through tool arguments, resource URIs or server
+    configuration instead"* — while the same handler keeps working for handshake-era clients on the
+    same endpoint. Both calls are also deprecated as of ``mcp/sdk`` 0.8 (SEP-2577), as is
+    ``$gateway->log()``.
+
+    Take what such a handler needs through tool arguments, resource URIs or server configuration
+    instead, or guard the call with ``$gateway->supportsSampling()`` / ``$gateway->supportsRoots()``,
+    which report what the client of the current request declared. Elicitation is the one ask that
+    survived, as a multi round-trip request; see *Request State* below.
+
+    Until the handlers are ready, ``protocol_versions`` is the opt-out: list only handshake-era
+    revisions and the server refuses the modern era altogether.
+
+    .. code-block:: yaml
+
+        # config/packages/mcp.yaml
+        mcp:
+            servers:
+                api:
+                    protocol_versions: ['2025-11-25'] # handshake era only, no modern leg
+                    registry: '*'
+
+Request State
+^^^^^^^^^^^^^
+
+A handler that needs another round trip has nowhere to keep its progress once sessions are gone, so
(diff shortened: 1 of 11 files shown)


docs/bundles/mcp-bundle.rst
```diff
@@ -600,6 +600,101 @@ This wraps the configured Symfony session handler (e.g. Redis, database, filesys
 your application uses for HTTP sessions) with a JSON envelope for application-level TTL.
 Expired sessions are cleaned up lazily on read.
 
+The Modern Era
+..............
+
+The 2026-07-28 revision drops the ``initialize`` handshake and the session that went with it: every
+request describes itself, so any worker can answer any request.
```
chr-hertel:
```suggestion
request describes itself, so any worker can answer any request, and no long-running connection is needed.
```

docs/bundles/mcp-bundle.rst
```diff
@@ -600,6 +600,101 @@ This wraps the configured Symfony session handler (e.g. Redis, database, filesys
...
+Both eras are served **on the same endpoint**, with nothing to configure. The SDK builds a dispatcher
+per era, and the HTTP transport classifies every request before anything else looks at it: an envelope
+claiming a modern revision goes to the modern dispatcher, the ``initialize`` handshake and its
+session's later requests go to the other. A client picks nothing.
+
+STDIO carries the handshake era alone — it has no per-request envelope to classify — so a server
+configured for both transports simply serves the modern era over HTTP only.
+
+By default a server answers for every modern revision the SDK knows. Narrow that, or refuse the era
+outright with an empty list, through ``protocol_versions``:
```
chr-hertel:
```suggestion
By default the HTTP endpoint supports all versions the SDK supports, inherits the default, but negotiates with the client. Narrow the supported versions through ``protocol_versions``:
```

src/mcp-bundle/config/options.php
```diff
@@ -134,13 +135,65 @@
...
+                                ->ifTrue(static fn (array $v): bool => \count(array_filter($v, static fn (string $r): bool => !ProtocolVersion::from($r)->isModern())) > 1)
+                                ->thenInvalid('The handshake era can be pinned to a single revision or left to negotiate, not narrowed to a subset, got %s.')
+                            ->end()
+                        ->end()
+
+                        ->arrayNode('request_state')
+                            ->addDefaultsIfNotSet()
+                            ->info('Signs the state a multi-round-trip answer carries through the client, which has no session to keep progress in. Required for a modern-era server whose handlers return an InputRequiredResult, and for one whose handlers call ClientGateway::elicit() more than once: the second ask has to carry the first answer to the next round.')
+                            ->children()
+                                ->stringNode('key')->cannotBeEmpty()->defaultNull()->info('HMAC key, at least 32 bytes. The same value must reach every process that might serve the retry.')->end()
```
wachterjohannes: Non-blocking: the node only has `cannotBeEmpty()`, while this info string and `UPGRADE.md` both promise at least 32 bytes. The SDK does enforce it, but from `RequestStateCodec::__construct`, which `Builder::build()` only reaches at request time:

```
container with an 8-byte key      -> compiles
new RequestStateCodec('tooshort') -> must be at least 32 bytes, got 8
```

Every neighbouring rule here is caught at compile time, the `protocol_versions` subset check and `ttl`'s `min(1)` included, so a `->validate()` would put the failure next to the config that causes it.

wachterjohannes (approved):
Good to go. I checked the escape hatch for the BC break rather than trusting the description, and it wires up exactly as documented: unset leaves the SDK default, handshake-only produces `withoutModernEra` plus `setProtocolVersion`, modern-only produces `setModernVersions`. The multi-handshake case is refused with a clear message. Dropping `capabilities.roots` and widening `ServerConnectionInterface` look like breaks but reach nobody, the whole `clients` section landed after `v0.12.0`. The `mcp/sdk` widening is complete too, no bridge constrains it and mate's suite passes against 0.8.0.

One thing that is yours to weigh rather than mine, since you already flagged it: `forward_server_logs` defaults to `true`, and 0.8's `LoggingNotificationHandler` triggers its deprecation from the constructor, so every application with a configured client emits `MCP logging is deprecated since protocol revision 2026-07-28` without having opted into anything, and the only way out is turning the feature off. Worth considering whether the default should follow the deprecation. Not blocking.

Verified locally, all green.

src/mcp-bundle/config/options.php
chr-hertel: Good one, will patch! :+1:

PR #2452, opened 2026-08-29
[Mate] Unify the Monolog bridge's result-limit parameter to `limit`

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | yes
| Issues        | none
| License       | MIT

`monolog-tail` took `lines` while `monolog-search` and `monolog-context-search` took `limit`, for
the same "how many results" meaning. Since `tools:list` does not show parameter names, getting it
right on the first try needs a `tools:inspect` round-trip per tool.

All three take `limit` now. This renames a tool parameter, so it is a BC break: documented in
`UPGRADE.md`.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -341,6 +341,14 @@ Mate
    Symfony CLI, an unpinned Mate started on the host reads the host's runtime and reports on
    something that is not the application under test.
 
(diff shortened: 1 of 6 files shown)


chr-hertel (approved):
Ship it 👍 💪

PR #2458, opened 2026-08-29
[MCP Bundle] Publish the registry's changes to the configured notification bus

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | -
| License       | MIT

`subscriptions:` configures a notification bus, and the 2026-07-28 protocol reads it: every `subscriptions/listen` stream polls it for the list-changed notifications it agreed to carry. Nothing ever writes to it.

The SDK wires the publishing half itself — `Builder` wraps the event dispatcher in a `PublishingEventDispatcher` when a bus is configured. But it can only wrap a registry it constructs, and this bundle always supplies one (`mcp.server.<name>.registry`, built with Symfony's `event_dispatcher`), so the wrapping never happens. The two halves are each individually right and never meet.

From a client that is worse than an error: the stream opens, the acknowledgment names the types the server agreed to carry, keep-alives arrive for the configured lifetime, and it closes gracefully having carried nothing — whatever changed on the server.

Gives the registry the publishing dispatcher where it is registered, which is the only place the bundle knows both it and the bus. Per server, because the registries are per server.

Requires **mcp/sdk 0.8.1**. Below that the SDK loads a supplied registry without its `loading` guard (modelcontextprotocol/php-sdk#490), so publishing from the registry also publishes every element the load registers — this patch would turn an inert bus into a noisy one.

Found by a demo application that drives its own MCP servers with its own MCP client over a real transport ([chr-hertel/mcp-demo](https://github.com/chr-hertel/mcp-demo)).

### DIFF
--- a/src/mcp-bundle/composer.json
+++ b/src/mcp-bundle/composer.json
@@ -14,7 +14,7 @@
         }
     ],
     "require": {
-        "mcp/sdk": "^0.8",
+        "mcp/sdk": "^0.8.1",
         "php-http/discovery": "^1.20",
         "symfony/config": "^7.3|^8.0",
         "symfony/console": "^7.3|^8.0",
--- a/src/mcp-bundle/src/McpBundle.php
+++ b/src/mcp-bundle/src/McpBundle.php
@@ -39,6 +39,7 @@ use Mcp\Server\Session\Psr16SessionStore;
 use Mcp\Server\Subscription\InMemoryNotificationBus;
 use Mcp\Server\Subscription\NotificationBusInterface;
 use Mcp\Server\Subscription\Psr16NotificationBus;
+use Mcp\Server\Subscription\PublishingEventDispatcher;
 use Mcp\Server\Wire\CachePolicy;
 use Symfony\AI\McpBundle\App\McpAppRenderer;
 use Symfony\AI\McpBundle\Attribute\AsMcpApp;
@@ -358,7 +359,7 @@ final class McpBundle extends AbstractBundle
         $sessionId = $this->configureSessionStore($name, $server['session'], $container);
 
         $container->register($registryId, Registry::class)
-            ->setArguments([new Reference('event_dispatcher'), new Reference('logger')])
+            ->setArguments([$this->registryDispatcher($name, $server['subscriptions']), new Reference('logger')])
             ->addTag('monolog.logger', ['channel' => 'mcp']);
 
         $container->register($builderId, Builder::class)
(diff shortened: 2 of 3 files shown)


chr-hertel:
Flagging an interaction found while reviewing this, because it decides whether this should merge on its own.

The fix is right and the bug it fixes is real — a configured bus is read by every stream and written to by nothing. But making the registry publish also makes it publish its **load**, and that is a lot louder than it looks.

`Builder::resolve()` loads a supplied registry by calling the loader directly:

```php
if ($this->hasCustomRegistry) {
    $registry = $this->registry;
    $chainLoader->load($registry);
```

`Registry::dispatch()` suppresses list-changed events while `$this->loading` is set, but that flag is only set by `Registry::load()` — which this bypasses. So with this PR applied, every `build()` publishes one `list_changed` per registered element.

On a demo application with 28 elements and `subscriptions: { bus: cache }`:

```
cursor before building the server:   0
cursor after  building the server:  28
  notifications/tools/list_changed:      13
  notifications/resources/list_changed:  11
  notifications/prompts/list_changed:     4
```

and the cursor keeps climbing across processes — 0 → 28 → 56 → 84 → 112 over successive runs. Under PHP-FPM that means every request touching the server broadcasts its whole element list to every open `subscriptions/listen` stream, for a registry that did not change, and consumes `Psr16NotificationBus`'s 256-entry backlog in ~9 requests. Past that, a reader that fell behind silently skips genuine notifications.

`bus: memory` is mostly spared — the bus dies with the process, and `StatelessProtocol::listen()` takes its cursor after the build — but `cache` is the one the docs recommend for FPM, which is where this lands hardest.

**This is not the bundle's bug to fix.** The registry carries its own dispatcher, which the Builder never holds, so nothing on the SDK side can suspend it and nothing in a DI extension can arm it after `build()` — `Registry` exposes no `isLoaded()` either. The guard already exists in the right place; one branch reaches around it. Opened as modelcontextprotocol/php-sdk#490.

Suggest holding this until that lands, or merging it together with an `mcp/sdk` constraint that has the fix. Happy either way — say which and I'll adjust.

chr-hertel:
modelcontextprotocol/php-sdk#490 is merged and released as **mcp/sdk 0.8.1**, so the interaction above is resolved. Updated:

- `src/mcp-bundle/composer.json` requires `mcp/sdk: ^0.8.1`. Below that the SDK loads a supplied registry without its `loading` guard, so publishing from the registry also publishes every element the load registers — this PR would turn an inert bus into a noisy one.
- One predicate (`publishesNotifications()`) and one id helper (`notificationBusId()`) behind both halves, so the registry and `configureModernEra()` cannot disagree about whether there is a bus, or about what it is called. The forward reference was only safe while the two conditions happened to agree.
- `registryDispatcher()` takes `$server['subscriptions']` with the same array shape as its sibling `notificationBus()`, rather than the whole server config as `array<string, mixed>`.
- The four new tests moved into `McpBundleTest`'s existing subscriptions cluster; `McpBundleSubscriptionsTest` is gone, and with it a second copy of `buildContainer()`/`config()`.
- `assertCount(1, ...)` before indexing the `setNotificationBus` calls, so a missing call fails as an assertion rather than an undefined index.

196 tests / 586 assertions, PHPStan and CS clean against 0.8.1.

PR #2459, opened 2026-08-29
[MCP Bundle] Let request_state.key come from an environment variable

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | -
| License       | MIT

`request_state.key` signs the state a multi-round-trip answer carries through the client. It is an HMAC secret, and `docs/bundles/mcp-bundle.rst` shows it arriving the way a Symfony secret does:

```yaml
request_state:
    key: '%env(MCP_REQUEST_STATE_KEY)%'
```

That configuration cannot be compiled. `cache:clear` fails with:

> The path `mcp.servers.modern.request_state.key` cannot contain an environment variable when empty values are not allowed by definition and are validated.

`VariableNode::finalizeValue()` refuses a registered env placeholder outright on any node that is both `cannotBeEmpty()` and carries a `->validate()`, before the validator runs at all. So the only expressible way to set the key was to inline the secret in a committed file — for the one option in the bundle that is explicitly a secret.

### Why dropping `cannotBeEmpty()` is not the fix

`ValidateEnvPlaceholdersPass` re-processes the whole configuration with the placeholder **substituted for a sample value**, so the length check then measures `''` and refuses that instead. The guard the check already carries — a regex for `%…%` — cannot match at that point, because the placeholder is gone by the time the validator sees it.

A config-tree validator is structurally unable to tell a short literal from an env reference, which is exactly the distinction this check exists to make.

### The fix

The check moves to `McpBundle::configureModernEra()`, where `resolveEnvPlaceholders()` turns a registered placeholder back into `%env(NAME)%` and a reference written in the configuration is still in that form anyway. What is left to measure is exactly the literals, which is what the documentation promises.

Behaviour is otherwise unchanged: a short literal key is still refused when the container is built. It becomes the bundle's own `LogicException` rather than an `InvalidConfigurationException`, in line with the other compile-time refusals in `McpBundle` (duplicate HTTP paths, shared session stores). Not a BC break — every configuration that compiled before compiles now, and the option is unreleased.

### Tests

`testAnEnvPlaceholderKeyIsNotMeasured` passes today because it builds the container by hand, and a container built by hand never registers the placeholder — which is why the bug shipped with a test that looks like it covers it. Its new sibling runs `MergeExtensionConfigurationPass` and `ValidateEnvPlaceholdersPass`, the path a kernel takes and the only one where the bug exists. Reverting the fix turns it red.

Found by [chr-hertel/mcp-demo](https://github.com/chr-hertel/mcp-demo), which configures the key from the environment because the same value has to reach every process that might serve the retry.

### DIFF
--- a/src/mcp-bundle/config/options.php
+++ b/src/mcp-bundle/config/options.php
@@ -12,7 +12,6 @@
 namespace Symfony\Component\Config\Definition\Configurator;
 
 use Mcp\Schema\Enum\ProtocolVersion;
-use Mcp\Server\Stateless\RequestStateCodec;
 use Symfony\AI\McpBundle\McpBundle;
 use Symfony\Component\Config\Definition\Builder\ArrayNodeDefinition;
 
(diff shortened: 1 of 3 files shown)


src/mcp-bundle/config/options.php
```diff
@@ -153,15 +152,12 @@
...
-                                        // rather than measured as the placeholder string it still is here.
-                                        ->ifTrue(static fn (?string $v): bool => null !== $v && 1 !== preg_match('/^%[^%]+%$/', $v) && \strlen($v) < RequestStateCodec::MINIMUM_KEY_BYTES)
-                                        ->thenInvalid(\sprintf('The "request_state.key" must be at least %d bytes: below that the signature protecting the carried state is forgeable, and the SDK refuses it.', RequestStateCodec::MINIMUM_KEY_BYTES))
-                                    ->end()
-                                ->end()
+                                // The length is checked in McpBundle::configureModernEra() rather than
+                                // here. Only a literal can be measured, and this is the one place that
+                                // cannot tell a literal from an env variable: cannotBeEmpty() on a
+                                // validated node refuses every placeholder outright, and without it the
+                                // validator is handed a substituted sample value to measure instead.
```
chr-hertel:
```suggestion
```

src/mcp-bundle/src/McpBundle.php
```diff
@@ -469,8 +470,17 @@ private function configureModernEra(string $name, array $server, string $builder
...
         $builder = $container->getDefinition($builderId);
 
-        if (null !== $server['request_state']['key']) {
-            $builder->addMethodCall('setRequestState', [$server['request_state']['key'], $server['request_state']['ttl']]);
+        if (null !== $key = $server['request_state']['key']) {
+            // Measured here because this is the last point that still knows whether the key is a
+            // literal. resolveEnvPlaceholders() turns a registered env placeholder back into
+            // "%env(NAME)%", and a reference written in the configuration is still in that form
+            // anyway, so anything left looking like "%...%" has no value to measure until runtime —
+            // and the SDK refuses a short one there.
```
chr-hertel:
```suggestion
```

PR #2467, opened 2026-08-30
[Platform] Add provider identity to ResultConvertedEvent and ResultErrorEvent

clegginabox:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no 
| License       | MIT

A Platform can hold several providers, but a listener on `ResultConvertedEvent` / `ResultErrorEvent` can't tell which provider actually served the request.

This adds an optional `?ProviderInterface $provider` to both events (with a `getProvider(): ?ProviderInterface` getter), and `Provider` now passes itself when dispatching them.

I need this to rework #1452: capacity is acquired when routing picks a provider and a listener on these events releases it when a result resolves. This will only work if the event says which provider's capacity to release.


wachterjohannes (approved):
Correct and complete: every event dispatch site passes the provider, and the new parameter is trailing/nullable so it's fully BC-safe. Verified locally, all green. One thing: needs a rebase, #2196's merge touched the same `CHANGELOG.md` line.

clegginabox:
@wachterjohannes rebased 👍

wachterjohannes:
Thank you @clegginabox.

PR #2469, opened 2026-08-31
[Platform] Verify cassette requests before replay

### LINKED ISSUE #2443: [Platform] Verify the recorded request signature when replaying a cassette
`Test\Replay\HttpCassette` writes a `signature` for every recorded request (method, URL and body hash), but nothing ever reads it. `next()` returns only the `response` half of an interaction, so replay is pure FIFO: the cassette decides what comes back, regardless of what the bridge actually sent.

That leaves half the pipeline uncovered. Contract normalizers and `ModelClient` payload building can regress without a single replay test noticing. Sending a completely wrong model through the recorded examples of #2437 changes nothing:

```php
// src/platform/src/Bridge/OpenResponses/ModelClient.php
'body' => $this->encodeJsonBody(array_merge($options, ['model' => 'totally-wrong-model'], $payload)),
```

```
OK (5 tests, 10 assertions)
```

A dropped option, a wrong endpoint or a mangled message payload behaves the same way.

The data is already there, so the fix is small: compare the outgoing request against the recorded `signature` on replay and fail loudly on a mismatch.

Two things to decide:

- **Strictness.** Failing hard on any mismatch is the point, but it makes every cassette brittle against unrelated payload changes (a new default option would invalidate every cassette at once). An opt-in flag on `CassetteHttpClient`, defaulting to on for `AbstractBridgeReplayTestCase` and off for the examples harness, may be the better balance.
- **Granularity.** The signature is a hash, so a mismatch can only say "differs", not what differs. Keeping the normalized request body in the cassette (it already is, redacted) allows a readable diff in the failure message, which is what makes such a failure actionable.

Follow-up to #2128, which introduced the scaffolding, and #2437, which puts it in CI as bridge integration testing.

ousamabenyounes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | Fix #2443
| License       | MIT

This makes cassette replay validate the outgoing request signature before serving the recorded response. The signature covers the HTTP method, raw URL, query options and body, so changed bridge payloads fail fast instead of silently consuming stale recorded responses.

Existing cassettes without a stored request signature still replay for backward compatibility.

Tests run:

```
./run-ci.sh
```


ousamabenyounes:
Rebased onto the latest `main` to resolve the merge conflict that had made this PR `CONFLICTING`. The only conflict was in `src/platform/CHANGELOG.md`: I kept every new `0.14` bullet that landed on `main` and appended this PR's entry to the same upcoming section (no new section, no bullets dropped). The feature commit itself is unchanged — `git range-diff` confirms the only difference versus the previous head is that CHANGELOG hunk.

Validated the rebased head locally (PHP 8.5, real replay, no credentials):

- `src/platform` unit suite: **904 tests OK** (1 pre-existing skip)
- Anthropic `ReplayTest`: **145 tests OK**
- Examples replay (`ExamplesReplayTest`, `CASSETTE=replay` through the real bridge pipeline): **10 tests OK** — this exercises the new request-signature verification against the cassettes that were added to `main` after this PR branched; they carry the legacy `signature` only, so the backward-compatible fallback path is covered
- PHPStan (platform + Anthropic bridge shard): no errors
- PHP-CS-Fixer: 0 of 2233 files need fixing

Pushed as `be0b33fc873d21c72051d0dc4ff47ea32e4fec8e`.

wachterjohannes:
Heads up: #2487 (redact recorded bodies) touches `HttpCassette::redactRequest()`/`signature()` too, redacting the body before hashing it. This branch still hashes the raw body when verifying, so once #2487 lands, a live request carrying credentials or PII would fail verification against the redacted recorded signature. Whichever of these merges second needs to redact the live request the same way before comparing signatures.

src/platform/src/Test/Replay/CassetteHttpClient.php
```diff
@@ -73,7 +71,7 @@ public function __construct(
     public function request(string $method, string $url, array $options = []): ResponseInterface
     {
         if (!$this->record) {
-            return $this->replayClient->request($method, $url, $options);
+            return self::toMockClientResponse($this->cassette->nextFor($method, $url, $options), $method, $url, $options);
         }
```
wachterjohannes, line 75: #2443 suggested this default to on for bridge replay tests and off for the examples harness. This runs verification unconditionally instead, and the cost is already in this diff: five examples had to swap their `Clock` tool for a `clock_tool()` helper in `examples/bootstrap.php` that regex-extracts the recorded timestamp so the live signature matches. That's a trade that can be made deliberately, but right now it's an undocumented pattern: the next example with any non-deterministic tool output hits an opaque signature-mismatch error with no hint that this workaround is the expected fix.

Please add a short doc note for future example authors (`docs/components/platform.rst`, near the replay section), and confirm this unconditional-verification choice over #2443's suggested opt-out for examples is the deliberate call, not an oversight.

src/platform/src/Test/Replay/HttpCassette.php
```diff
@@ -232,7 +249,73 @@ private static function sanitizeHeaders(array $headers): array
...
+        if ($recordedRequest[self::REQUEST_SIGNATURE] === self::legacySignature($method, $url, $body)) {
+            return;
+        }
+
+        if (self::legacyRecordedRequestMatches($recordedRequest, $method, $url, $body)) {
+            return;
+        }
+
+        throw new RuntimeException(\sprintf('Outgoing request #%d does not match the recorded request signature in cassette "%s"; delete it to re-record.', $cursor + 1, $path));
+    }
```
wachterjohannes, line 399: #2443's second point was that keeping the redacted body in the cassette should let the failure message show what actually differed. Both branches here throw the same generic "does not match the recorded request signature" regardless of whether method, URL, query or body changed, so debugging a mismatch today means diffing the cassette JSON by hand. Worth adding what differed to the message?

wachterjohannes (changes requested):
The signature-verification mechanism itself is solid, well-tested, and matches FIFO/backward-compat behavior on existing cassettes. Verified locally, all green. Two things from #2443 that this PR was meant to resolve are inline instead of settled: the strictness trade-off, and the error-message granularity.

- Two OpenAI `text_to_speech` cassettes had their request body reshaped (dropping a stale `response_format` key) alongside the signature change. That matches what `TextToSpeech/ModelClient.php` actually sends today, so it looks like a correction, not a regression, just want to check it came from an actual re-record rather than a manual edit.

src/platform/src/Test/Replay/CassetteHttpClient.php
ousamabenyounes: Deliberate, not an oversight — and now documented, which was the part that was genuinely missing.

The reasoning: an opt-out for the examples harness reopens exactly the hole #2443 describes. The examples of #2437 are the case where a wrong model went unnoticed, so switching verification off there would turn the replay back into a fixed-response stub instead of a check of the payload the bridge actually builds.

The cost is real but bounded, and `clock_tool()` already pays it. What was missing is that the next example author had no way to discover the pattern — they would meet an opaque signature mismatch with no hint that pinning the recorded timestamp is the expected fix. `docs/components/platform.rst` now states that verification is unconditional, why, and shows the helper with its actual call site, plus the rule that an example whose tool output cannot be pinned does not belong in the replayed harness.

src/platform/src/Test/Replay/HttpCassette.php
ousamabenyounes: Done in `ae366c83`. The message now names what diverged — `(body differs)`, `(method differs, url differs)`, and so on — so a mismatch no longer means diffing the cassette JSON by hand.

Two constraints shaped it:

- The body is compared through `signature()`'s own encoding, so the comparison can never contradict the hash it explains: `1.0` vs `1`, or a raw JSON string vs its decoded array, stay distinct. A looser comparison would let a body change hide and then be reported as a query change — worse than the generic message it replaces.
- `query` is only named by elimination, since it is folded into the signature rather than stored. That inference is drawn only when every other part was genuinely comparable; otherwise the message stays generic rather than guessing. Values are never interpolated, since a cassette body holds request data.

Making that comparison trustworthy surfaced two pre-existing bugs in `save()`, both fixed here:

- It wrote without `JSON_PRESERVE_ZERO_FRACTION` while `signature()` hashes with it, so a recorded `1.0` was persisted as `1`: the cassette could not reproduce its own signature and the stored body silently retyped the payload.
- It also lacked `JSON_THROW_ON_ERROR`. An unencodable value — the audio bridges (Cartesia, ElevenLabs, Decart, Deepgram) upload a `body` opened as a resource — made `json_encode()` return `false`, and the cassette was overwritten with a bare newline, destroying every interaction recorded so far. It now fails loudly and leaves the file intact.

`Test/Recording/Cassette.php` already pairs both flags for these exact documented reasons; this brings `HttpCassette` in line. Covered by regression tests, including one asserting the cassette survives an unencodable interaction.

ousamabenyounes:
On the `text_to_speech` cassettes: re-recorded, not hand-edited — and the diff shape is the evidence rather than my memory of it.

The body went from a JSON **string** (`"{\"model\":...,\"response_format\":\"mp3\"}"`) to a structured object with the keys in a different order (`voice` before `model`). That is what the writer produces once the request body is stored decoded; it is not a shape anyone reaches by editing the file by hand.

`response_format` disappearing is a correction, as you suspected: `TextToSpeech/ModelClient.php` builds `['json' => array_merge($options, ['model' => ..., 'input' => ...])]` and never sends `response_format`, so the key was a stale leftover in the recording rather than something the client emits today.

CI on `ae366c83` is green apart from `Integration / Store / Milvus`, which fails on `Failed to connect to 127.0.0.1 port 19530` — the Milvus service did not come up on the runner. Unrelated to this branch, which touches only `Test/Replay`. Notably `Integration / Examples (replay)` and the full Examples matrix are green, so the `save()` flag changes replay the committed cassettes unchanged.

ousamabenyounes:
Pushed `b07c6e3d`. Two things beyond the review items, one of which I got wrong first.

**A retyped `tool_use.input` in `multi-turn-thinking-stream.json`.** The recorded body carried `"input":[]` where the live request sends `"input":{}` — an empty JSON object, not an empty array. It is the trace of a decode/encode cycle (in PHP an empty array decoded from `{}` re-encodes as `[]`), not of the writer: `redactRequest()` stores the body verbatim, and `signature()`, `legacySignature()` and `normalizeJsonBody()` each sort a local copy.

It stayed invisible because that fixture carries `signature` but no `signature_v2`, so verification falls to the legacy path — and there `legacyRecordedRequestMatches()` decodes and ksorts both sides, where `json_decode('{}', true)` and `json_decode('[]', true)` are both the empty array. The lenient path cannot tell them apart, which is why the suite stayed green over a cassette that no longer matched the wire.

I first restored the whole body from before this PR, which was wrong: the re-record had also legitimately fixed a duplicated `thinking` block, and throwing that away broke `Integration / Examples (replay)`. The pushed fix touches only the retyped value, keeps every other byte of the recorded body, and recomputes the signature by invoking `redactRequest()` itself rather than writing it by hand. No `signature_v2` is introduced, so the verification path is unchanged.

Added a test that recomputes every committed interaction's signature from its own stored body, so a cassette that cannot reproduce its own signature fails instead of replaying through the lenient path.

**A finding I have deliberately not acted on.** In `examples/tests/fixtures/toolbox/brave.json`, 7 of 15 interactions have a stored signature that does not reproduce their own stored body — exactly the 7 OpenAI requests carrying a body; the 8 bodyless GETs are consistent. Their bodies differ entirely from the pre-PR version, so this looks like a real re-record whose signatures were not updated. I did not touch it: I cannot establish which body is authoritative, and recomputing the hashes blind would only prove internal consistency, which is the very thing this change argues against. Happy to fix it here, in a follow-up, or leave it — your call.

CI on `b07c6e3d` is green except `Integration / Store / Milvus`, which fails on `Failed to connect to 127.0.0.1 port 19530` (10 tests, 9 skipped, 0 assertions — the service never came up). It fails identically on every push of this branch, and this PR touches no store code.

ousamabenyounes:
One separable thing in this branch, and it is a data-loss bug that already affects `main`.

`HttpCassette::save()` encodes without `JSON_THROW_ON_ERROR`. On an unencodable body, `json_encode()` returns `false` and the file is overwritten with a bare newline. Measured against `main` with no patches applied:

```
cassette after first record:     595 bytes
second record (resource body):   no error raised
cassette after:                  1 byte
```

It is reachable in practice: `Cartesia/CartesiaClient.php:95` and `ElevenLabs/ElevenLabsClient.php:71` pass an open `fopen()` handle as the request body, so recording either bridge silently destroys whatever the cassette already held. `Test/Recording/Cassette.php` already guards against exactly this and documents why, so this brings `HttpCassette` in line with its sibling.

The same commit also adds `JSON_PRESERVE_ZERO_FRACTION`, which `signature()` already hashes with: without it a recorded `1.0` is written as `1`, so a cassette cannot reproduce its own signature and the stored body silently retypes the payload.

Both are three lines plus tests and depend on nothing else in this PR. If it helps your merge queue, I am happy to pull them into a separate PR so they can land while this one is still in review — just say the word and I will open it.

ousamabenyounes:
**Update on `b07c6e3d`** — two fixture findings and one offer.

**1. A retyped `tool_use.input`.** `multi-turn-thinking-stream.json` recorded `"input":[]` where the request sends `"input":{}`. It is the trace of a decode/encode cycle — in PHP an empty array decoded from `{}` re-encodes as `[]` — not of the writer, which stores the body verbatim.

It hid because that fixture has `signature` but no `signature_v2`, so verification falls to the legacy path, where both sides are decoded and ksorted — and `json_decode('{}', true)` and `json_decode('[]', true)` are both the empty array. The lenient path cannot tell them apart.

My first attempt restored the whole pre-PR body, which was wrong: the re-record had also fixed a duplicated `thinking` block, and dropping that broke `Integration / Examples (replay)`. The pushed fix changes only the retyped value and recomputes the signature through `redactRequest()` itself. Also added a test that recomputes every interaction's signature from its own stored body.

**2. `brave.json`, left untouched on purpose.** 7 of its 15 interactions have a signature that does not reproduce their own body — exactly the 7 OpenAI requests carrying a body. I cannot establish which body is authoritative, and recomputing the hashes blind would only prove internal consistency, which is what this change argues against. Fix here, follow-up, or leave — your call.

**3. A data-loss bug that already affects `main`.** `HttpCassette::save()` encodes without `JSON_THROW_ON_ERROR`, so an unencodable body makes `json_encode()` return `false` and the file is overwritten with a bare newline. Measured on `main`, unpatched:

```
cassette after first record:     595 bytes
second record (resource body):   no error raised
cassette after:                  1 byte
```

Reachable via `Cartesia/CartesiaClient.php:95` and `ElevenLabs/ElevenLabsClient.php:71`, which pass an open `fopen()` handle as the body. `Test/Recording/Cassette.php` already guards against this and documents why. The commit also adds `JSON_PRESERVE_ZERO_FRACTION`, which `signature()` already hashes with.

That part is three lines plus tests and depends on nothing else here — happy to split it into its own PR so it can land while this one is in review.

CI is green except `Integration / Store / Milvus` (`Failed to connect to 127.0.0.1 port 19530`, 9 of 10 tests skipped); it fails identically on every push and this PR touches no store code.

wachterjohannes (approved):
Both points addressed, verified locally: the docs now explain why verification is unconditional, and the mismatch message names what diverged, confirmed via `HttpCassetteTest`, including the regression test for the data-loss fix. The two extra fixes found along the way (retyped `tool_use.input`, `JSON_THROW_ON_ERROR` guard) check out too.

`brave.json`'s signature mismatch is real but predates this PR and is unrelated to it. Opening a follow-up issue for it, this can merge as-is.

wachterjohannes:
Follow-up for brave.json: #2518

wachterjohannes:
Thank you @ousamabenyounes.

PR #2471, opened 2026-08-31
[Agent] Add execution cancellation

fabpot:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Docs?         | yes 
| Issues        | n/a
| License       | MIT

My main use case is when using fibers, but not mentioned in the docs as nothing does that in Symfony AI core. Signals is also a god use case, used in the docs.


chr-hertel (approved):
🔥

chr-hertel:
Thank you @fabpot.

PR #2473, opened 2026-08-31
[Platform] Record the model name when TraceablePlatform is invoked with a Model

### LINKED ISSUE #2472: [Platform] TraceablePlatform's PlatformCallData phpstan-type declares "model: string", but invoke() stores a string|Model value
### Description

`TraceablePlatform`'s class-level docblock declares the `@phpstan-type PlatformCallData` shape as:

```php
/**
 * @phpstan-type PlatformCallData array{
 *     model: string,
 *     input: array<mixed>|string|object,
 *     options: array<string, mixed>,
 *     result: DeferredResult,
 * }
 */
```

(`src/platform/src/TraceablePlatform.php`, currently lines 25-30 on `main`)

However, `invoke()` accepts `string|Model $model` (line 49) and stores the argument **unnormalized**:

```php
$this->calls[] = [
    'model' => $model,
    // ...
];
```

(lines 61-62)

Since the `0.10` release, `Platform::invoke()`/`Provider::invoke()` officially accept a fully-defined `Model` instance instead of a name string (see `src/platform/CHANGELOG.md`, "0.10" section: *"Add support for passing a fully defined `Model` instance to `Platform::invoke()`"*). When a caller invokes with a `Model` instance, `TraceablePlatform::getCalls()` returns a call entry whose `model` key is a `Model` object at runtime, contradicting the `model: string` type. `Model` has no `__toString()`, so naive string usage of that field throws a `TypeError` rather than producing a wrong string.

Because PHPStan trusts the stale docblock, it cannot flag this at analysis time — the failure only surfaces at runtime.

### Reproduction

```php
use Symfony\AI\Platform\Model;
use Symfony\AI\Platform\TraceablePlatform;

$traceable = new TraceablePlatform($platform);
$traceable->invoke(new Model('gpt-4o'), 'Hello');

foreach ($traceable->getCalls() as $call) {
    // PlatformCallData says $call['model'] is `string`, PHPStan is happy...
    $modelName = strtoupper($call['model']);
    // ...but at runtime: TypeError: strtoupper(): Argument #1 ($string)
    // must be of type string, Symfony\AI\Platform\Model given
}
```

### Impact

Any downstream consumer that trusts the `PlatformCallData`/`CollectedPlatformCallData` type (e.g. cost-tracking, logging, or profiler integrations built on `TraceablePlatform::getCalls()`) can hit an undetectable `TypeError` at runtime as soon as any part of the call chain invokes the platform with a `Model` instance instead of a model-name string. The same stale `model: string` shape is duplicated in `src/ai-bundle/src/Profiler/DataCollector.php`'s `CollectedPlatformCallData` phpstan-type, so the Bundle's own profiler/data-collector layer inherits the same inaccurate contract.

### Suggested fix

Two options, either resolves the type/runtime mismatch:

**(a) Widen the docblock type to `model: string|Model`.**
Non-breaking (doesn't change runtime behavior), and preserves the full `Model` object (capabilities, options) for consumers that want it. Downside: every consumer that currently assumes `model` is a plain string (e.g. for display/logging) now has to handle both cases, and the duplicated `CollectedPlatformCallData` type in `ai-bundle`'s `DataCollector` would need the same widening for consistency (similar in spirit to #1177, which fixed the `result`/`metadata` fields of that same type but left `model` untouched).

**(b) Normalize in `invoke()`:** store `$model instanceof Model ? $model->getName() : $model`, keeping `getCalls()`'s `model` field a plain string as the docblock already promises.
Simpler for existing consumers (matches what most currently expect and what the profiler template presumably renders), but silently discards the `Model` object's capabilities/options for anyone who might want to inspect them via the recorded calls.

I'd lean towards (b) for minimal downstream disruption, but happy to defer to maintainer preference — either fixes the discoverability gap.

lazerg:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #2472
| License       | MIT

`TraceablePlatform::invoke()` accepts `string|Model`, but stored the argument as-is, so `getCalls()` could return a `Model` object under the `model` key while the `PlatformCallData` phpstan-type (and the `CollectedPlatformCallData` copy in the profiler's data collector) promises a `string`. PHPStan trusts the docblock, so the mismatch only shows up at runtime, and `Model` has no `__toString()`, so the profiler template's `{{ call.model }}` breaks as soon as a call passes a model instance.

This normalizes the value at the point where the call is recorded, keeping the recorded `model` a plain string as both docblocks already state.

### DIFF
--- a/src/platform/src/TraceablePlatform.php
+++ b/src/platform/src/TraceablePlatform.php
@@ -59,7 +59,7 @@ final class TraceablePlatform implements PlatformInterface, ResetInterface
         }
 
         $this->calls[] = [
-            'model' => $model,
+            'model' => $model instanceof Model ? $model->getName() : $model,
             'input' => \is_object($input) ? clone $input : $input,
             'options' => $options,
             'result' => $deferredResult,
--- a/src/platform/tests/TraceablePlatformTest.php
+++ b/src/platform/tests/TraceablePlatformTest.php
@@ -12,6 +12,7 @@
 namespace Symfony\AI\Platform\Tests;
 
 use PHPUnit\Framework\TestCase;
+use Symfony\AI\Platform\Model;
 use Symfony\AI\Platform\PlainConverter;
 use Symfony\AI\Platform\PlatformInterface;
 use Symfony\AI\Platform\Result\DeferredResult;
@@ -41,5 +42,18 @@ final class TraceablePlatformTest extends TestCase
         $this->assertCount(0, $traceablePlatform->getCalls());
         $this->assertNotSame($oldCache, $traceablePlatform->getResultCache());
         $this->assertInstanceOf(\WeakMap::class, $traceablePlatform->getResultCache());
+    }
+
+    public function testInvokeWithModelInstanceRecordsModelName()
+    {
+        $platform = $this->createStub(PlatformInterface::class);
+        $traceablePlatform = new TraceablePlatform($platform);
+        $result = new TextResult('Assistant response');
+
+        $platform->method('invoke')->willReturn(new DeferredResult(new PlainConverter($result), $this->createStub(RawResultInterface::class)));
+
+        $traceablePlatform->invoke(new Model('gpt-4o'), 'Hello');
+
+        $this->assertSame('gpt-4o', $traceablePlatform->getCalls()[0]['model']);
     }
 }


wachterjohannes (approved):
Correct and minimal, and the test asserts the recorded value rather than a flag.

Option (b) is also the one the existing consumers require, which I think settles the choice you left open, @lunetics: the bundle's own `AiAssertionsTrait::getCalledModels()` maps over this field with a `: string` return type, so a `Model` instance throws a `TypeError` there, not only in the profiler template. Reproduced that on main and confirmed it is gone with this patch. Verified locally, all green.

wachterjohannes:
Thank you @lazerg.

PR #2474, opened 2026-08-31
[Store][ChromaDb] Use default ChromaDB query limit

### LINKED ISSUE #812: [Store][ChromaDB] query method should return more than 4 results
Hi,

I'm using ChromaDB bridge and the query method return only 4 results. Why this value has been hardcoded ?

Here the expected code:

```php
/**
 * @param array{where?: array<string, string>, whereDocument?: array<string, mixed>, nResults?: int} $options
 */
public function query(Vector $vector, array $options = []): array
{
    $collection = $this->client->getOrCreateCollection($this->collectionName);
    $queryResponse = $collection->query(
        queryEmbeddings: [$vector->getData()],
        nResults: $options['nResults'] ?? 4, // if no limit (aka nResults) passed, use the hardcoded value
        where: $options['where'] ?? null,
        whereDocument: $options['whereDocument'] ?? null,
    );

    $documents = [];
    for ($i = 0; $i < \count($queryResponse->metadatas[0]); ++$i) {
        $documents[] = new VectorDocument(
            id: Uuid::fromString($queryResponse->ids[0][$i]),
            vector: new Vector($queryResponse->embeddings[0][$i]),
            metadata: new Metadata($queryResponse->metadatas[0][$i]),
        );
    }

    return $documents;
}
````

What do you think about that? I'm able to create the PR. Just tell me.

### PR DESCRIPTION
**Pull request by ousamabenyounes**
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #812
| License       | MIT

ChromaDB queries currently fall back to 4 results when no `limit` option is passed. The underlying ChromaDB PHP client default is 10, and the bridge already exposes `limit` for callers that need a custom value.

This changes the default query limit to 10 for vector and text queries while keeping explicit `limit` handling intact.

Tests:

```text
RED: php vendor/bin/phpunit --filter "testQueryWithoutFilters|testQueryWithConfiguredLimitOption|testQueryWithTextQuery" --no-coverage
FAILURES! Tests: 4, Assertions: 23, Failures: 2.
Failed asserting that 4 matches expected 10.
```

```text
GREEN: ./run-ci.sh
OK (36 tests, 244 assertions)
[OK] No errors
OK (36 tests, 244 assertions)
```

`./run-ci.sh` covers the ChromaDB bridge unit suite, PHPStan, and coverage text. The integration group was excluded because it requires a live ChromaDB service; the same environment-only failure was present before this patch.

### DIFF
--- a/src/store/src/Bridge/ChromaDb/Store.php
+++ b/src/store/src/Bridge/ChromaDb/Store.php
@@ -33,6 +33,7 @@ use Symfony\AI\Store\StoreInterface;
 final class Store implements ManagedStoreInterface, StoreInterface
 {
     private const BATCH_SIZE = 1000;
+    private const DEFAULT_QUERY_LIMIT = 10;
 
     public function __construct(
         private readonly Client $client,
@@ -180,7 +181,7 @@ final class Store implements ManagedStoreInterface, StoreInterface
         $collection = $this->client->getOrCreateCollection($this->collectionName);
         $queryResponse = $collection->query(
             queryEmbeddings: [$query->getVector()->getData()],
-            nResults: $options['limit'] ?? 4,
+            nResults: $options['limit'] ?? self::DEFAULT_QUERY_LIMIT,
             where: $options['where'] ?? null,
             whereDocument: $options['whereDocument'] ?? null,
             include: $include,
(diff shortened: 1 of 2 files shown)


wachterjohannes (approved):
Matches the underlying ChromaDB client's own default, and issue #812 already has chr-hertel's sign-off on making this configurable. Verified by reading and the existing test updates; the bridge's dev dependency wasn't installable locally to run it directly.

wachterjohannes:
Follow-up: got the bridge's dev dependency installed after all (a stale local lock file was the blocker, not a real conflict) and ran the suite directly. Verified locally, all green.

wachterjohannes:
Thank you @ousamabenyounes.

PR #2478, opened 2026-09-01
[Platform] Unwrap a JSON code fence before decoding structured output

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | -
| License       | MIT

Replaces #2114. Models without native structured output often answer with their JSON inside a Markdown code fence, and `StructuredOutput\ResultConverter` handed that string straight to the decoder, so the call failed with `Cannot json decode the content.` even though the payload itself was correct.

The converter now unwraps the fence before decoding, but only when the fenced body parses as JSON on its own, so a text answer that merely contains a fence is passed through unchanged and fails exactly as before. A fence surrounded by prose is deliberately not unwrapped either; there is a test pinning that boundary, so widening it stays a conscious decision.

#2114 solved the same problem with a normalizer chain on `ResultEvent`. That approach also stripped Unicode format characters from every result by default, which corrupts bidi text, Indic ligatures and emoji sequences, and its subscriber resolved the deferred result eagerly, which suppresses `ResultConvertedEvent` and `ResultErrorEvent` for every call. Doing the unwrapping where the decode happens avoids both.

### DIFF
--- a/src/platform/src/StructuredOutput/ResultConverter.php
+++ b/src/platform/src/StructuredOutput/ResultConverter.php
@@ -80,6 +80,8 @@ final class ResultConverter implements ResultConverterInterface
 
     private function convertTextToObject(TextResult $textResult, RawResultInterface $result): ObjectResult
     {
+        $content = $this->unwrapCodeFence($textResult->getContent());
+
         try {
             $context = [];
             if (null !== $this->objectToPopulate) {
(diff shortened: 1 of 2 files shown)


chr-hertel (approved):
Fair enough, ran into that already as well - thanks 👍

chr-hertel:
Thank you @wachterjohannes.

PR #2480, opened 2026-09-01
[Chat][Platform] Declare the PHP 8.3 polyfill where json_validate() is used

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | -
| License       | MIT

The Gemini, Vertex AI and Pogocache packages call `json_validate()`, which is PHP 8.3, while declaring `php: ">=8.2"`. Nothing breaks today only because `symfony/http-client` requires `symfony/polyfill-php83` from v7.3.3 on, and all three require http-client. A resolution that lands on http-client 7.3.0, 7.3.1 or 7.3.2 on PHP 8.2 gets a fatal error instead. They now declare the polyfill themselves, the way `symfony/ai-store` already does for the same function in the Redis store.

The Bedrock tests used `json_validate()` to assert a request body and now use PHPUnit's `assertJson()`, which drops the version dependency rather than moving it.

Worth knowing for whoever looks at this next: the reason nothing flagged it is that PHPStan runs without a `phpVersion`, so it assumes the 8.5 it runs on and treats every post-8.2 function as available. Pinning the declared floor does not work today: any `phpVersion` setting makes PHPStan fall back to its version stubs, which have no `Pdo\Sqlite` at all, so the Sqlite store bridge and `examples/rag/sqlite-vec.php` light up with `class.notFound` even though the class exists on 8.4 and 8.5. A user stub does not help, PHPStan will not invent an internal class for a version. So the gap stays open for now.

### DIFF
--- a/src/chat/src/Bridge/Pogocache/composer.json
+++ b/src/chat/src/Bridge/Pogocache/composer.json
@@ -28,6 +28,7 @@
         "php": ">=8.2",
         "symfony/ai-chat": "^0.13",
         "symfony/http-client": "^7.3|^8.0",
+        "symfony/polyfill-php83": "^1.32",
         "symfony/serializer": "^7.3|^8.0"
     },
     "require-dev": {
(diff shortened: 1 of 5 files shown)


chr-hertel (approved):

chr-hertel:
Thank you @wachterjohannes.

PR #2483, opened 2026-09-02
[Agent] Forward Progress updates from delegated executions in MultiAg…

### LINKED ISSUE #2444: [Agent] MultiAgent and SpeechAgent swallow the updates of the delegated execution
Since #2436, `AgentInterface::call()` returns a lazy, iterable `Execution` reporting every model request, tool call and streamed delta as a `Progress` update.

The decorating agents resolve the execution of the agent they delegate to eagerly (`->getResult()`) and only yield the final `Result` update themselves:

- `MultiAgent`: the orchestrator's routing call, the selected agent's execution and the fallback path
- `SpeechAgent`: the inner agent's execution before the text-to-speech step

So iterating a `MultiAgent` or `SpeechAgent` yields no progress at all — not the handoff, not the delegate's tool calls, not its streamed deltas — although the handoff is exactly the thing one would want to observe.

Proposal: iterate the delegated execution and forward every update except its `Result`, then yield the decorator's own `Result` (after iteration, `getResult()` on the delegated execution is the cached value). A dedicated `handoff` progress stage naming the selected agent would additionally make the routing decision itself visible.

Raised by @wachterjohannes in the review of #2436: https://github.com/symfony/ai/pull/2436#discussion_r3866174308

saifulferoz:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| Docs?         | yes
| Issues        | Fix #2444
| License       | MIT

### Description

Iterating a `MultiAgent` or `SpeechAgent` previously yielded no `Progress` updates because both
decorating agents eagerly resolved the delegated execution via `->getResult()`.

This PR fixes that by:

1. Forwarding the `Progress` updates of the delegated executions in both `MultiAgent` and
   `SpeechAgent` (model requests, tool calls, streamed deltas), with cancellation still propagating.
2. Emitting a `handoff` `Progress` stage on `MultiAgent` carrying the orchestrator's
   `Handoff\Decision` as payload, so consumers see which agent answers and why. `Decision` is
   therefore no longer `@internal`.

The orchestrator's own routing round is excluded from delta forwarding: those deltas spell out the
`Decision`, not the answer, and would otherwise leak into the `delta` stage that consumers render as
the answer.

Docs: a new step in the multi-agent cookbook recipe, with `examples/multi-agent/orchestrator-iterable.php`
as its runnable counterpart.

### DIFF
--- a/ai.symfony.com/templates/cookbook/content/multi-agent-orchestration.html
+++ b/ai.symfony.com/templates/cookbook/content/multi-agent-orchestration.html
@@ -316,6 +316,99 @@ the <code translate="no" class="notranslate">MultiAgent</code> constructor to se
 </div>
 </div>
 <div class="section">
+<h2 id="step-7-show-which-agent-answers">
+    Step 7: Show Which Agent Answers
+    
+</h2>
+<p>Routing is invisible from the outside: the caller asks a question and gets an answer, with no clue
+which specialist produced it. Iterating the multi-agent instead of reading its result lets you show
+that, because the delegated agents report their own steps into the very same execution. The routing
+itself arrives as a <code translate="no" class="notranslate">Progress</code> update of the <code translate="no" class="notranslate">handoff</code> stage, carrying the orchestrator's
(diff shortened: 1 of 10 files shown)


chr-hertel (approved):
Thanks @saifulferoz - will rebase while merging 👍

chr-hertel:
Thank you @saifulferoz.

PR #2486, opened 2026-09-03
[Agent] Keep server tools passed in the tools option

### LINKED ISSUE #2485: [Agent] cannot send server tools in options
Both OpenAI and Anthropic APIs I tested requires setting `tools` to enable server tools provided by them (web search for example).

```php
$agent->call(
    ...,
    ...,
    'tools' => [
        [
            'type' => 'web_fetch_20260318',
            'name' => 'web_fetch',
        ]
    ]
)
```

However, this does not work if I already have tools defined on my side. [Runner.php](https://github.com/symfony/ai/blob/f101f14ffb5662de2925058d5c5a1de91ece22be/src/agent/src/Execution/Runner.php#L240) seems to override `tools` option without a way to introduce more server tools.

I also tried defining this with the `MemoryToolFactory` which did not work.

lazerg:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #2485
| License       | MIT

When a toolbox is registered, `Runner::exposeTools()` replaces the whole `tools` option with the toolbox tool map, so provider-side server tools passed by the caller (`['type' => 'web_fetch_20260318', 'name' => 'web_fetch']`) never reach the API.

The option now keeps both: string entries still filter the toolbox by tool name, everything else is appended after the resulting tool map.

### DIFF
--- a/src/agent/src/Execution/Runner.php
+++ b/src/agent/src/Execution/Runner.php
@@ -232,22 +232,21 @@ final class Runner
             return $options;
         }
 
-        // only filter tool map if a list of strings is provided as option
-        if (isset($options['tools']) && \is_array($options['tools']) && $this->isFlatStringArray($options['tools'])) {
-            $toolMap = array_values(array_filter($toolMap, static fn (Tool $tool): bool => \in_array($tool->getName(), $options['tools'], true)));
+        $serverTools = [];
+
+        if (isset($options['tools']) && \is_array($options['tools'])) {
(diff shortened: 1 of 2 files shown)


wachterjohannes (approved):
Fix is correct: a raw server-tool array in `tools` now survives alongside a registered toolbox instead of being silently dropped, and the empty-array edge case (`tools => []`) is handled too. Verified locally, all green.

wachterjohannes:
Thank you @lazerg.

PR #2487, opened 2026-09-03
[Platform] Redact recorded request bodies in HTTP cassettes

### LINKED ISSUE #2409: [Platform] Redact recorded prompts in HTTP cassettes
Follow-up from @MikiBuilder's review in https://github.com/symfony/ai/pull/2129#issuecomment-5114825333.

HTTP-boundary cassettes (#2128) commit the request body verbatim, so a recorded prompt carries whatever the user sent: emails, phone numbers, occasionally an API key pasted by mistake. Only credential headers are redacted today. Redaction of the body should be the default rather than opt-in, on the assumption that nobody reads the configuration reference of a dev dependency.

One trap to avoid, from the same review: if the body is redacted on write but matched against the raw text, a request carrying a real email never matches its own cassette, so it fails precisely on the requests that contain personal data. Replay currently matches FIFO, so this does not bite yet, but the stored request signature would have to be computed over the redacted form.

Provider-level cassettes (#2129) are not affected: they store a hash of the input, never the input itself.

Prior art offered by the reporter, MIT licensed: [mikibuilder/llm-vcr](https://github.com/MikiBuilder/llm-vcr).

MikiBuilder:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | Fix #2409
| License       | MIT

Adds `BodyRedactor` and wires it into `HttpCassette` so recorded request bodies are sanitized before they are committed. Follows the shape agreed in #2409.

Three tiers of rules, as discussed there:

- **credentials** (provider keys, JWTs, bearer tokens, cloud keys) — always applied
- **personal data** (emails, phone numbers) — applied by default, switchable off with `BodyRedactor::credentialsOnly()` for the cases where the personal data is the thing under test
- **user patterns** — for domain-specific identifiers

The redactor is an optional constructor argument (`?BodyRedactor $redactor = null`), so existing callers and the static helpers are unchanged.

Redaction happens before `signature()` is computed, so the stored signature describes what the cassette actually contains. Existing cassettes keep replaying, since matching is FIFO today and the signature is informational.

The default patterns are deliberately conservative. `BodyRedactorTest` includes legitimate payloads — timestamps, unix timestamps, token counts, order numbers, UUIDs, model names, hashes — asserting nothing is touched. An over-eager pattern corrupts the recording it is meant to protect.

One docs change worth flagging: the headers paragraph in `docs/components/platform.rst` claimed a cassette is safe to commit because headers are sanitized. That sentence is removed and replaced by a paragraph on body redaction, since the guarantee only holds once bodies are covered too.

Platform suite green (909 tests, 1921 assertions, 2 skipped). PHPStan clean on the touched files.


src/platform/src/Test/Replay/BodyRedactor.php
```diff
@@ -0,0 +1,127 @@
...
+        '/(?<![\w.])\+\d{1,3}[ .-]?\(?\d{1,4}\)?(?:[ .-]?\d{2,4}){2,4}(?![\w.])/' => '[redacted-phone]',
+    ];
+
+    /** @var array<string, string> */
+    private readonly array $patterns;
+
+    /**
+     * @param bool                  $pii           whether to redact personal data on top of credentials
+     * @param array<string, string> $extraPatterns additional `pattern => replacement` pairs, applied last
+     *
```
wachterjohannes, line 77: `applied last` here, but PHP's `+` union means `$extraPatterns` entries keep their earlier position when `redactString()` iterates `$this->patterns`, so they're actually applied first. Verified: an extra pattern matching a credential-shaped string never fires, the credential pattern wins. Doesn't affect this PR's own examples, but the doc is backwards.

src/platform/src/Test/Replay/HttpCassette.php
```diff
@@ -151,13 +159,16 @@ public function next(): array
...
         $headers = self::sanitizeHeaders(self::normalizeRequestHeaders($options));
 
         $request = ['method' => $method, 'url' => $url];
 
-        $body = $options['json'] ?? $options['body'] ?? null;
+        // Redact before signing: a committed cassette has to stay reproducible from what it
+        // actually contains, and hashing the raw body would describe something the file no
+        // longer holds.
+        $body = $this->redactor->redact($options['json'] ?? $options['body'] ?? null);
         $request['signature'] = self::signature($method, $url, $body);
```
wachterjohannes: #2469 touches this same method to verify a live request's signature against the recorded one, but its branch still hashes the raw, unredacted body. Whichever of these lands second needs to redact the live request the same way before comparing signatures, otherwise verification fails specifically on requests carrying the data this PR protects. Worth flagging to that PR's author rather than something to fix here.

wachterjohannes (commented):
Matches the shape from #2409 exactly: optional collaborator, redact-before-sign, three-tier defaults, and a test proving legitimate payloads survive untouched. One nit inline, and a heads-up that needs coordinating with #2469 rather than a code change here.

src/platform/src/Test/Replay/BodyRedactor.php
MikiBuilder: You are right that the doc is backwards. `+` keeps the left operand's positions, so `redactString()` iterates extras first.

It is a behaviour bug rather than only a doc one, though. I ran it: with an extra pattern `/sk-[A-Za-z0-9_\-]{16,}/ => '[mine]'`, the string `sk-proj-AbCdEf0123456789XyZ` comes out as `[mine]`. The extra fires first and the credential pattern then finds nothing left to match, so a user pattern can currently override the tier this class documents as non-negotiable.

That is the opposite of what you observed, so one of us is testing something different. Worth reconciling before you take my word for it.

Fixed the order rather than the sentence, in 91345f0f:

    $this->patterns = self::CREDENTIAL_PATTERNS + ($pii ? self::PII_PATTERNS : []) + $extraPatterns;

Credentials first, then PII, then user patterns. The class docblock now states the order and why it matters, and `itNeverLetsAUserPatternOverrideACredential` fails on the old arrangement.

Happy to go back to just fixing the sentence if you would rather user patterns keep priority, there is a case for the domain-specific rule being the most specific one. But then the class docblock has to stop promising credentials cannot be switched off.

src/platform/src/Test/Replay/HttpCassette.php
MikiBuilder: Agreed, and the collision is silent rather than a merge conflict, which is what makes it worth flagging now. #2469 hashes the live request and compares it to the stored signature; this PR stores the signature of the redacted body. Verification would then fail exactly on the requests that carried a key or an email, which is the opposite of what either PR is trying to do.

#2469 is older and further along, so I am happy to be the one who adapts. If it lands first I will rebase and run the live body through the redactor before comparing. If this lands first I will comment there with the change needed.

@ousamabenyounes heads-up, so this does not surprise you later.

wachterjohannes:
#2469 merged today. Its signature check on main still hashes the raw request body, so the redaction/verification mismatch we talked about is now live, not hypothetical. Whenever you get to the rebase, redacting the live body before comparing should close it.

wachterjohannes:
On 1: the premise doesn't hold, `redactRequest()` already computes both signatures from one `$body` in the same call, so there's no scenario where a freshly written cassette has v1 but not v2. Redact `$body` once before both signature calls and this is a one-liner, no permanent v1/v2 split needed. That also removes your "lesser problem": a freshly redacted cassette verifies fine against itself through v1, since v1 was hashed over the same redacted body that's stored.

The real gap is on the verify side, not the write side: `assertRequestSignatureMatches()` always hashes the raw live body. Once cassettes store redacted bodies, a live request with a real credential won't match through either path. Redacting the live body unconditionally would risk breaking any already-committed cassette that happens to have credential-shaped text in it, so: try the raw live body first, exactly like today, and only redact and retry on a mismatch before falling through to `mismatchMessage()`. No version flag, can't regress an existing cassette since the first attempt is unchanged.

`assertRequestSignatureMatches()` is `private static`, so it'll need the same instance-method treatment you already gave `redactRequest()` to reach the redactor.

On 2: agreed, redact the live body for the message comparison too, that one's just display and carries no correctness risk either way.

wachterjohannes:
Design's landed and verified, just a merge conflict left. Can you rebase when you get a chance?

wachterjohannes:
The remaining red `Integration / Examples (replay)` check is unrelated: it's a stale run against a base before #2531 (merged 2026-09-16), which fixed object-ID drift in the dumped example output. A rebase on current main should turn it green.

wachterjohannes:
Thank you @MikiBuilder.

PR #2488, opened 2026-09-04
[Mate] Show tool arguments in tools:list, point tools:call errors at tools:inspect

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | n/a
| License       | MIT

`tools:list` already collected each tool's `input_schema`, but its table output dropped it and truncated the description to 50 characters. The table now shows an `Arguments` column (required names first, optional ones marked with `?`), and the description column is no longer truncated, so a tool's shape is visible without a separate `tools:inspect` call.

`tools:call` already reports unknown or missing parameter names via `ArgumentCaster`'s existing message; that part was not broken. A parameter error now additionally points at `tools:inspect <tool-name>`.

Measured against an external benchmark (a `monolog-search` scenario, 240 runs total across the batch): no movement in that scenario's error rate, for two independent reasons. `tools:list` was never called in that flow, so the new column had no path to the agent. And the observed failures were all missing-required-argument errors that `ArgumentCaster`'s message already named correctly, with 100% self-correction on retry even before this change. Noting that here rather than overselling it: this is a plausible hygiene improvement for flows that do read `tools:list` or a `tools:call` error, not a fix validated to move a measured number.

### DIFF
--- a/src/mate/src/Command/ToolsListCommand.php
+++ b/src/mate/src/Command/ToolsListCommand.php
@@ -215,12 +215,13 @@ HELP
         }
 
         $table = new Table($io);
-        $table->setHeaders(['Tool Name', 'Description', 'Handler', 'Extension']);
+        $table->setHeaders(['Tool Name', 'Description', 'Arguments', 'Handler', 'Extension']);
 
         foreach ($tools as $toolName => $toolData) {
             $table->addRow([
                 $toolName,
-                $this->truncate($toolData['description'] ?? '', 50),
+                $toolData['description'] ?? '',
+                $this->formatArguments($toolData['input_schema']),
                 $toolData['handler'],
                 $toolData['extension'],
             ]);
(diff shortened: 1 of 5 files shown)


src/mate/src/Command/ToolsListCommand.php
```diff
@@ -215,12 +215,13 @@ private function outputTable(array $tools, SymfonyStyle $io): void
         }
 
         $table = new Table($io);
-        $table->setHeaders(['Tool Name', 'Description', 'Handler', 'Extension']);
+        $table->setHeaders(['Tool Name', 'Description', 'Arguments', 'Handler', 'Extension']);
```
chr-hertel, line 218: just a thought: could use a stress test how useful the table formatting is with longer description and larger set of arguments.

but LLMs most likely don't care :D

chr-hertel (approved):

chr-hertel:
Thank you @wachterjohannes.

src/mate/src/Command/ToolsListCommand.php
wachterjohannes: @chr-hertel this was already part of a benchmark :) the command itself is not really often called! but i am currently testing how we could improve the call efficiency and that is a small part of that

PR #2470, opened 2026-08-31
 Feature/hosted web search assistant replay

fabpot:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Docs?         | yes/no 
| Issues        | n/a
| License       | MIT

PR #2492, opened 2026-09-05
[Platform] Replay hosted web searches across turns

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | n/a
| License       | MIT

Alternative to #2470, rebased on current `main` and reconciled with the `server_tools` work that landed since. The first commit is @fabpot's, with the merge resolved.

A provider-hosted web search now survives into the next turn. Anthropic takes one back only as the pair of blocks it sent — the `server_tool_use` call and its `web_search_tool_result` — and rejects either alone, so `WebSearchResult` carries both and `Contract\AssistantMessageNormalizer` replays them in order. OpenResponses replays its `web_search_call` item the same way. A search whose blocks came from another provider is dropped rather than re-sent.

Kept `main`'s result shape: one `WebSearchResult` per search, carrying query, id and status. Streaming now emits one `WebSearchComplete` delta per search, once its result block arrives, so a streamed turn replays like a buffered one.

Verified against the live API: `examples/anthropic/server-tools-web-search-roundtrip.php` searches on turn 1 and names the source on turn 2. The recorded cassette shows both blocks going back on turn 2 and Anthropic answering 200, and `ExamplesReplayTest` re-runs it offline in CI.

Not covered: a web search initiated from Anthropic code execution still throws on replay, because its surrounding code execution blocks are not converted. Search hits (URL, title, page age) round-trip inside the replay payload but are still not exposed on `WebSearchResult`.


fabpot:
I've tested this branch and it works well for me. Thank you.

chr-hertel:
Thanks for testing Fabien!

PR #2500, opened 2026-09-06
[Mate] Resolve the composer-plugin project root from Composer's config instead of getcwd()

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | n/a
| License       | MIT

**Problem**: `MatePlugin::onPostInstallOrUpdate()` derived the project root from `getcwd()`, but the current working directory at the moment Composer fires its post-install/post-update script is not guaranteed to be the actual project root — it's whatever directory `composer` happened to be invoked from (a monorepo, CI running from a subfolder, or `composer --working-dir=...` while the shell's own cwd stays elsewhere). This caused Mate's generated files (`AGENTS.md`, `CLAUDE.md`, `mate/`, skill folders) to be written into the wrong directory when the invocation cwd didn't match the real project root. Running `vendor/bin/mate discover` manually always worked because cwd is the root by construction there, which is why the bug was invisible until a nested-directory Composer invocation exposed it.

**Fix**: derive the root from `dirname($event->getComposer()->getConfig()->get('vendor-dir'))` instead of `getcwd()`. Composer's `Config::get('vendor-dir')` always resolves to an absolute path relative to the real project root regardless of invocation cwd (confirmed in Composer's own `Config::realpath()`/`baseDir` handling), so this is robust to any cwd/root mismatch. Everything else in the method (the `mate/extensions.php` existence check, the `vendor/bin/mate` binary check, the `proc_open` invocation and its cwd argument) stays the same, just anchored to the corrected root.

**Tests**: added `testUsesComposerVendorDirInsteadOfCurrentWorkingDirectory`, which reproduces the bug by `chdir()`-ing into an unrelated nested subdirectory before invoking the plugin, with the real project root (containing `mate/extensions.php` and a fake `vendor/bin/mate`) supplied only via the mocked `Composer`/`Config`. It asserts the plugin ran discovery against the real root and did not fall back to the bogus cwd-derived one. The two pre-existing tests were adapted to the new `Composer`/`Config` mock plumbing and still pass.


chr-hertel:
somehow reminded me of #1857 - how about that one?

chr-hertel (approved):

wachterjohannes:
@chr-hertel pushed a new commit: switched to `Composer\InstalledVersions::getRootPackage()['install_path']` after your #1857 pointer showed `vendor-dir` isn't a reliable anchor either. Mind another look?

chr-hertel:
Thank you @wachterjohannes.

Issue #2280, opened 2026-07-07
Improving DX on Symfony AI: A Journey Through "Error-Driven Development"


lyrixx:
I recently started building a foundational CLI tool using the new Symfony AI component, and I deliberately chose an **"Error-Driven Development"** approach—navigating the implementation without looking at the documentation beforehand to test the intuitiveness of the DX.

While the final result is elegant, the onboarding friction for a task as simple as "send a prompt and get a response" highlighted several critical pitfalls. Here is the breakdown of the DX journey and where we can improve it.

---

## The Goal

Create a simple Symfony command that takes a user question, sends it to a LLM platform (like Gemini), and displays the output.

### Initial Intuitive Implementation

Based on typical IDE autocompletion and intuitive naming, one might naturally write this:

```php
#[AsCommand(
    name: 'app:ai:ping',
    description: 'Ping an AI model with a simple question',
)]
class AiPingCommand extends Command
{
    public function __construct(
        private readonly PlatformInterface $platform
    ) {
        parent::__construct();
    }

    protected function configure(): void
    {
        $this
            ->addArgument('question', InputArgument::REQUIRED, 'The question to ask the AI')
            ->addOption('model', null, InputOption::VALUE_REQUIRED, 'The model to use', 'gemini-2.5-flash')
        ;
    }

    protected function execute(InputInterface $input, OutputInterface $output): int
    {
        // Intuitively passing the string argument directly
        $result = $this->platform->invoke($input->getOption('model'), $input->getArgument('question'));

        dd($result);

        return Command::SUCCESS;
    }
}

```

---

### 1. Misleading API Signatures (String vs. Array)

Passing a raw string failed. Attempting to wrap it in an array (`[$input->getArgument('question')]`) to satisfy potential iterable expectations moved us to the next error, but the signature didn't clearly guard against this at the PHP level.

```text
I have no name! /var/www bin/console app:ai:ping "salut, comment ca va?"

In ModelClient.php line 52:
                                                                                                               
  Payload must be an array, but a string was given to "Symfony\AI\Platform\Bridge\Gemini\Gemini\ModelClient".  
                                                                                                               

app:ai:ping [--model MODEL] [--] <question>

```

### 2. Side-Effects when Debugging with `dd()`

When trying to dump the `$result` (which returns a `DeferredResult`), Symfony's internal `HttpClient` threw a cryptic streaming error:

```text
In AsyncResponse.php line 291:
[...]

PR #2501, opened 2026-09-06
[Platform] Make DeferredResult and RawHttpResult safe to dump()/dd()

alireza-aminzadeh:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Contributes to #2280
| License       | MIT

Every HTTP-based `ModelClient` wraps its injected client with
`EventSourceHttpClient` (needed to decode SSE streaming responses), so the
response held by `RawHttpResult` is always a Symfony HttpClient
`AsyncResponse`, itself wrapping a `TraceableResponse` whenever HTTP client
profiling is enabled, which is the default in a full-stack app's dev
environment. Neither class has a dedicated VarDumper caster, and both keep
their network stream alive until it is explicitly consumed.

`dump()`'ing or `dd()`'ing a `DeferredResult` right after
`PlatformInterface::invoke()`, before the result is ever read, therefore
reflects into that live, not-yet-consumed response instead of a safe
snapshot: closures bound to the response, its HTTP client and its internal
chunk-accounting state all get expanded, and poking at that state from
outside the client's own bookkeeping is exactly what the
`AsyncResponse`/`TraceableResponse` pair is not designed to tolerate.

This PR adds `__debugInfo()` to `RawHttpResult` so it never exposes the live
response in a dump, replacing it with a short, static summary of its type
instead of letting the dumper walk into its internals. It also adds
`__debugInfo()` to `DeferredResult` so a dump surfaces the conversion state
(`pending`, `converted` or `failed`) at a glance, without ever calling
`getResult()` itself, so inspecting a pending result cannot trigger a
conversion as a side effect.

Tests assert that `__debugInfo()` never calls any method on the wrapped
response, that the `httpStream` property stays visible, and that
`DeferredResult`'s reported state matches pending/converted/failed without
triggering a conversion.

No `CHANGELOG.md`/`UPGRADE.md` changes, since this is a bug fix only.

### DIFF
--- a/src/platform/src/Result/DeferredResult.php
+++ b/src/platform/src/Result/DeferredResult.php
@@ -65,7 +65,13 @@ final class DeferredResult
      * conversion or otherwise touch $rawResult. That property is already safe to dump on its own, see
      * {@see RawHttpResult::__debugInfo()}, which is what keeps a plain dump() or dd() of a DeferredResult safe.
      *
-     * @return array<string, mixed>
+     * $options is a constructor-injected, always-initialized array and never holds $rawResult or anything
+     * derived from it. $metadata starts out empty and, once conversion succeeds, only ever holds what a
+     * ResultConverter or a stream listener explicitly added to it (e.g. token usage) — not the raw response
+     * itself. Reading $conversionFailure's message does not touch its (potentially deep) stack trace. None
+     * of the four can therefore drive the underlying HTTP stream forward as a side effect of dumping.
+     *
+     * @return array{state: 'pending'|'converted'|'failed', options: array<string, mixed>, metadata: array<string, mixed>, error: string|null}
      */
     public function __debugInfo(): array
     {
@@ -75,6 +81,9 @@ final class DeferredResult
                 $this->isConverted => 'converted',
                 default => 'pending',
             },
+            'options' => $this->options,
+            'metadata' => $this->getMetadata()->all(),
+            'error' => $this->conversionFailure?->getMessage(),
         ];
     }
 
--- a/src/platform/src/Result/DeferredResult.php
+++ b/src/platform/src/Result/DeferredResult.php
@@ -58,6 +58,27 @@ final class DeferredResult
     }
 
     /**
+     * Adds an at-a-glance conversion-state summary to dumps, on top of the normal properties.
+     *
+     * The underlying HTTP response is not read here, directly or indirectly: $isConverted and
+     * $conversionFailure are plain, always-initialized properties, so computing "state" cannot trigger a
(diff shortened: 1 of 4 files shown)


chr-hertel (changes requested):
Nice idea, thanks @alireza-aminzadeh! Haven't tested it yet, but I think for the `DeferredResult` it would be nice to keep more of the instance's state, e.g. error message, metadata, options

alireza-aminzadeh:
Thanks for the feedback! Pushed a commit that extends `DeferredResult::__debugInfo()` with the three pieces of state you mentioned, none of which touch `$rawResult` any more than the existing `state` entry does:

- `options` — the array the `DeferredResult` was constructed with (a plain, readonly, always-initialized constructor argument, e.g. the model options a call site passed to `invoke()`).
- `metadata` — `$this->getMetadata()->all()`; empty until conversion succeeds, and even then only ever holds what a `ResultConverter` or a stream listener explicitly added to it (e.g. `token_usage`), never the raw response itself.
- `error` — `$conversionFailure?->getMessage()`, `null` unless the state is `failed`; reading the message doesn't walk the exception's stack trace.

Also widened the `@return` docblock to an explicit array shape and added one test per new entry (`testDebugInfoIncludesOptionsWithoutConverting`, `testDebugInfoReportsEmptyMetadataAndNullErrorWhilePending`, `testDebugInfoIncludesMetadataAfterGetResult`, `testDebugInfoIncludesErrorMessageAfterConversionFailure`), covering the pending/converted/failed states.

Rebased on the latest `main` while at it. The full `platform` suite (908 tests) and PHPStan pass locally. Ready for another look whenever you have time!

src/platform/src/Result/DeferredResult.php
```diff
@@ -57,6 +57,36 @@ public function __construct(
...
+    public function __debugInfo(): array
+    {
+        return [
+            'state' => match (true) {
+                null !== $this->conversionFailure => 'failed',
+                $this->isConverted => 'converted',
+                default => 'pending',
+            },
+            'options' => $this->options,
+            'metadata' => $this->getMetadata()->all(),
```
chr-hertel: just realizing that the API design here is not that great because the `getMetadata()` has the side-effect of creating metadata if not existing. usually not an issue, but maybe misleading while debugging.

how about:
```suggestion
            'metadata' => $this->metadata?->all(),
```
keeps the state more honest

chr-hertel (approved):
Will patch that minor finding while merging - thanks @alireza-aminzadeh!

PR #2503, opened 2026-09-07
[Mate] Keep non-ASCII characters intact when init rewrites composer.json

mpysiak:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | -
| License       | MIT

`mate init` rewrites the project's `composer.json` to add the `extra.ai-mate` block and the `Mate\` autoloader. It re-encodes the whole file with `JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES`, so every non-ASCII character already present in the file comes out as a `\uXXXX` escape. An author entry such as `Michał Pysiak` becomes `Micha\u0142 Pysiak` after `init`; the same happens to any non-English `description`, author name or keyword.

Adding `JSON_UNESCAPED_UNICODE` keeps those characters as written, which matches what Composer itself does when it writes `composer.json`.

Includes a regression test: a `composer.json` with a non-ASCII author name keeps it after `init`, with no `\u` escapes in the output.

### DIFF
--- a/src/mate/src/Command/InitCommand.php
+++ b/src/mate/src/Command/InitCommand.php
@@ -353,7 +353,7 @@ class InitCommand extends Command
         if ($modified) {
             file_put_contents(
                 $composerJsonPath,
-                json_encode($composerJson, \JSON_PRETTY_PRINT | \JSON_UNESCAPED_SLASHES)."\n"
+                json_encode($composerJson, \JSON_PRETTY_PRINT | \JSON_UNESCAPED_SLASHES | \JSON_UNESCAPED_UNICODE)."\n"
             );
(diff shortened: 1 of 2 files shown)


wachterjohannes (approved):
Correct, minimal fix — matches Composer's own unicode-preserving write behavior. Verified locally: the regression test fails on `main` and passes with this change, all green otherwise. The one red check (`PHP 8.5 / Symfony 8.0`) is a network flake fetching sqlite-vec, unrelated to this PR.

wachterjohannes:
Thank you @mpysiak.

PR #2506, opened 2026-09-09
[MCP Bundle] Document RequestContext in tools

### LINKED ISSUE #2200: [MCP Bundle] Support client-conditional tool result shaping (structuredContent-only vs. text content)
### Summary

When an MCP tool returns an array, the SDK emits the result **twice** — a serialized-JSON `TextContent` block (`ToolReference::formatResult()`) **and** `structuredContent` (`ToolReference::extractStructuredContent()`, wired in `CallToolHandler`). Different clients want **opposite** shapes, and there's no first-class way to vary the result per client.

### The problem

- **ChatGPT connectors**: when a result carries *both* a JSON text block and `structuredContent`, ChatGPT treats it as a large payload, truncates it (*"Response output was truncated … to fit the tool response budget"*), stores it as a resource, and chains internal "Read resource" paging that **elides fields** — so identifiers (ids, etc.) never reach the model. Returning **only `structuredContent`** (empty `content`) fixes it. (See OpenAI community report [1383071](https://community.openai.com/t/tool-response-truncation-on-mcp-connector-responses-that-previously-worked/1383071).)
- **Claude Desktop / Claude Code / Cursor**: read the **text** `content` block; dropping it leaves the model with nothing.

So the correct `content` shape is genuinely client-dependent, but the framework gives no hook for it.

### What we had to do (works, but boilerplate + a footgun)

For every tool: declare a `?RequestContext $context` parameter, read `$context->getSession()->get('client_info')`, and hand-build a `CallToolResult` choosing `content: []` + `structuredContent` for OpenAI clients vs. text + `structuredContent` for the rest.

Footgun encountered: our first attempt used a parameter named `$_session`. `SchemaGenerator` **throws** on reserved parameter names (`_session`/`_request`) during discovery, which aborts discovery and registers **zero** tools — while the DI container still compiles, so it looks healthy but the live server returns "no tools". `RequestContext` is the correct injection, but this isn't obvious from the docs.

### Feature request

First-class support for client-conditional structured results, e.g. one or more of:

1. Honor the negotiated client capabilities / `client_info` so clients that consume `structuredContent` don't also receive the redundant JSON text dump (and vice-versa) — ideally automatic.
2. A per-tool toggle (attribute/option, e.g. `structuredContentOnly`) and/or a small documented helper to shape the result from the injected `RequestContext`.
3. Docs for the `RequestContext` injection pattern and the reserved `_session`/`_request` parameter names (so the silent "zero tools" discovery failure is avoidable).

### Environment

- `symfony/mcp-bundle` v0.10.0
- `mcp/sdk` v0.6.0
- PHP 8.4.20

andreybolonin:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Docs?         | yes
| Issues        | Fix #2200
| License       | MIT

Documents how to use `Mcp\Server\RequestContext` in MCP tools.

The documentation covers:

- injecting `RequestContext` into a tool;
- accessing the current request and session;
- retrieving the MCP protocol version;
- accessing client capabilities and trace context;
- using the client logger.

It also clarifies that `RequestContext` is request-specific and must not be
reused between requests.


docs/bundles/mcp-bundle.rst
```diff
@@ -693,6 +693,62 @@ next round, and carrying them is what needs the key. A handler that asks exactly
...
+between requests.
+
+It provides access to request-specific information through methods such as:
+
+* ``getRequest()`` — the current JSON-RPC request;
+* ``getSession()`` — the current MCP session;
+* ``getProtocolVersion()`` — the MCP protocol revision used for the request;
+* ``getClientCapabilities()`` — capabilities declared by the client;
+* ``getTraceContext()`` — W3C trace context carried by the request;
+* ``getClientLogger()`` — a logger for sending log messages to the client.
```
wachterjohannes: `RequestContext` in `mcp/sdk` (v0.7.1, what this bundle requires) only has `getRequest()`, `getSession()`, `getClientGateway()` and `getClientLogger()`. `getProtocolVersion()`, `getClientCapabilities()` and `getTraceContext()` don't exist anywhere in the SDK. Both code examples above call one of them and would fail with "Call to undefined method".

docs/bundles/mcp-bundle.rst
```diff
@@ -693,6 +693,62 @@ next round, and carrying them is what needs the key. A handler that asks exactly
 state and works without one. Missing the key, the ask is answered with a JSON-RPC internal error and
 the reason is logged on the ``mcp`` channel.
 
+Request Context
+^^^^^^^^^^^^^^^
```
wachterjohannes: This sits under "The Modern Era", but `RequestContext` works the same regardless of session mode. Might read better under "Creating MCP Capabilities → Tools", where someone writing a tool would actually look for it.

wachterjohannes (commented):
Good addition, the `RequestContext` injection pattern is genuinely undocumented right now. Details inline, one blocks: two of the methods and both code examples reference an API that doesn't exist on `RequestContext`.

- #2200's actual ask (point 3) was documenting that `_session`/`_request` are reserved parameter names that `SchemaGenerator` rejects during discovery, not the general `RequestContext` injection pattern. Worth adding if this PR is meant to close that issue.

wachterjohannes (approved):
Fabricated `RequestContext` methods are gone, the doc now matches the real vendored class, and the section moved where suggested. Approving.

wachterjohannes:
Thank you @andreybolonin.

PR #2510, opened 2026-09-11
Add #[MapToolArguments] for flat DTO tool arguments #2509

### LINKED ISSUE #2509: [Agent] Add #[MapToolArguments] for flat DTO tool arguments
I propose adding support an opt-in `#[MapToolArguments]` parameter attribute, similar to HttpKernel's `#[MapRequestPayload]`.

## Why it's useful

Imagine `read(path, limit, offset)` tool.
Some parts can be validate by `#[Schema()]` attribute, like simple checks for not blank, range for limit etc.

But for proper validation we also need to check for example
 - is file exist/readable
 - content type for example
 - is it opens up/not corruptes
 - offset is not larger then file

In this case Schema is not enough. 

We could use DTO as parameter, and put validations there:
```php
public function __invoke(
	ReadFileArgumentsDTO $arguments,
): string
```
But then our model need to send this schema:
```json
{"arguments":{"path":"src/Example.php","offset":10,"limit":30}}
```

Which is not what we want, as we are confusing model with nested schema, making harder to work with our tool, adding tokens etc.

## Proposed format
Instead we can map arguments to DTO in the same manner HTTP requests mapped
```php
use Symfony\AI\Agent\Toolbox\Attribute\MapToolArguments;

public function __invoke(
	#[MapToolArguments]
	ReadFileArgumentsDTO $arguments,
): string
```

This gives us flat structure for model, and DTO comfort with proper validation layer
```json
{"path":"src/Example.php","offset":10,"limit":30}
```

## Why existing options do not cover this

Scalar method parameters produce flat schemas. [PR #2362](https://github.com/symfony/ai/pull/2362), released in v0.13.0, also adds runtime validation for supported scalar and array `#[Schema]` constraints. That suits independent fields, but does not replace DTO class-level constraints or custom validators that inspect several fields together.

A raw array with a custom schema preserves flat arguments, but leaves us responsible for DTO conversion and object validation.

A nested DTO parameter preserves native typing and validation, but adds the parameter-name envelope to the model-facing contract.

## Proposed implementation

Like `#[MapRequestPayload]`, `#[MapToolArguments]` makes payload mapping explicit on the receiving parameter. It would:

- Expose the DTO properties as the root tool schema.
- Denormalize the entire tool-call argument object into that DTO using the configured Serializer.
- Preserve the resolved parameter map, such as `['arguments' => $dto]`, for `ToolCallArgumentsResolved` and the existing validation listener.
- Leave tools without the attribute unchanged.

The initial implementation requires exactly one non-nullable concrete DTO parameter. Invalid attributed signatures fail during tool registration.

The exact attribute name is open for discussion.

ineersa:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes 
| Docs?         | yes 
| Issues        | Fix #2509 
| License       | MIT

- Implements new `#[MapToolArguments]` attribute for exposing DTO properties at the schema root.
- Updates factories and argument resolvers to handle the new mapping functionality.
- Includes extensive test coverage for valid and invalid use cases.
- Documents usage and configuration of the new attribute.


src/agent/src/Toolbox/ToolCallArgumentResolver.php
```diff
@@ -67,7 +67,14 @@ classDiscriminatorResolver: new ClassDiscriminatorFromClassMetadata($classMetada
...
     {
-        $method = new \ReflectionMethod($metadata->getReference()->getClass(), $metadata->getReference()->getMethod());
+        $reference = $metadata->getReference();
+        $method = new \ReflectionMethod($reference->getClass(), $reference->getMethod());
+        $mapped = MappedToolArgument::forMethod($method);
+        if (null !== $mapped) {
+            return [
+                $mapped->parameter->getName() => $this->denormalizer->denormalize($toolCall->getArguments(), $mapped->className, 'json'),
+            ];
+        }
```
wachterjohannes, line 84: The rest of this method wraps failures as `ToolException` (the class docblock promises `@throws ToolException`). This path denormalizes straight from the DTO and lets `MissingConstructorArgumentsException` escape raw instead, confirmed by the PR's own test expecting that exact class. Doesn't matter through `Toolbox::execute()` today, both get wrapped into `ToolExecutionException` the same way, but it's an inconsistent contract for anyone calling the resolver directly.

src/agent/src/Toolbox/ToolFactory/MemoryToolFactory.php
```diff
@@ -79,4 +80,17 @@ public function getTool(object|string $reference): iterable
...
+     */
+    private function buildParameters(string $className, string $methodName): ?array
+    {
+        $mapped = MappedToolArgument::forMethod(new \ReflectionMethod($className, $methodName));
+        if (null === $mapped) {
+            return $this->factory->buildParameters($className, $methodName);
+        }
+
+        return $this->factory->buildProperties($mapped->className);
+    }
```
wachterjohannes: This `buildParameters()` helper is copy-pasted verbatim into `ReflectionToolFactory.php` too. Could live once, e.g. as a static helper on `MappedToolArgument` itself.

wachterjohannes (commented):
Solid feature: reuses `buildProperties()` so nullable/enum/nested/`#[Schema]` handling comes for free, and the tests assert the actual flattened schema and DTO round-trip, not just presence. Two nits inline, plus one CI fix.

- Fabbot's only red check is the commit title using Conventional Commits (`feat(agent): ...`). This repo doesn't use that convention, worth amending the title before merge.

src/agent/src/Toolbox/ToolCallArgumentResolver.php
ineersa: Yeah, but below https://github.com/symfony/ai/blob/258041af00637c8056c6809a84cb36e9b60c1be4/src/agent/src/Toolbox/ToolCallArgumentResolver.php#L115-L117 same thing happens when denormalize() being called.

Should I just add docblock to show that multiple exceptions possible or catch in both cases and produce `ToolException` ?

src/agent/src/Toolbox/ToolFactory/MemoryToolFactory.php
ineersa: Fixed

src/agent/src/Toolbox/ToolCallArgumentResolver.php
wachterjohannes: Good catch, that gap predates this PR. For this PR, please wrap just the new mapped-DTO denormalize call in a try/catch and rethrow as `ToolException`. That keeps the new code matching the docblock's promise. The pre-existing gap at the per-parameter path is real too, but it's not part of this diff. A separate follow-up issue for that works fine, no need to fix it here.

ineersa: Fixed

src/agent/src/Toolbox/MappedToolArgument.php
chr-hertel, line 1: Can we transform this into `ObjectDescriberInterface` implementation and encapsulate within the Factory basically?

chr-hertel (changes requested):
Thanks for working on this @ineersa - I think it's a great feature!

Please revisit how services in ToolFactory implementations, JsonSchema\Factory and `MappedToolArgument` work together. There are a couple of symptoms that raise a flag for me here:
* in both `ToolFactory` implementations the `JsonSchema\Factory` is injected, but only forwarded _as argument_ into a static method
* it is an optional feature but feels like taking over the steering here
* Is `MappedToolArgument` a state object or a service? maybe we need to decouple - not sure

src/agent/src/Toolbox/MappedToolArgument.php
ineersa: Yeah, this looks better now.

One thing to notice is factory wiring, if we want to provide ability to use this argument attribute by default, we need to wire describer into factories.

That was IMHO a bit too invasive, I've left previous commit, so you could check how it was looking.

As a tradeoff, application side now needs to configure factory with describer to use this feature properly - https://github.com/symfony/ai/pull/2510/changes#diff-e04a1b45e14764524023222d3e64d8f72f52e20b4d3ae6d7c2691c49fe787a83R528

chr-hertel: No, i think that was a fair change - I'm in favor of having this active by default 👍

chr-hertel (approved):
Will just drop the second commit & merge it - cleverly done 😎👍

chr-hertel:
Thank you @ineersa.

PR #2512, opened 2026-09-11
[Mate] Break LogReader's mtime tie deterministically

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | n/a
| License       | MIT

`LogReader::collectLogFiles()` sorts log files newest-first by `filemtime()`, which only has 1-second resolution. Two rotated files created in the same second (plausible during rotation, easy to hit in fast CI) tie there, and PHP 8's stable `usort()` then falls back silently to `glob()`'s incidental filesystem order, not rotation order. A caller that assumes `collectLogFiles()[0]` is the current file can then read a stale rotated one instead.

On an mtime tie, sorting now falls back to comparing `basename()` descending. Monolog's own rotation naming (`app.log`, `app.1.log`, `app-2024-01-01.log`, ...) keeps the unsuffixed/current file lexically greatest, so it still sorts first regardless of tie-break source.

Found by an external benchmark building a fixture that seeds rotated log files quickly, not a synthetic edge case.


src/mate/src/Bridge/Monolog/Tests/Service/LogReaderTest.php
```diff
@@ -257,6 +258,36 @@ public function testGetKernelContextDoesNotMatchSiblingDirectoryWithSharedPrefix
...
     }
 
+    /**
+     * @return iterable<string, array{string, string}>
+     */
+    public static function provideMtimeTieCreationOrders(): iterable
+    {
+        yield 'current file created first' => ['app.log', 'app.1.log'];
+        yield 'rotated file created first' => ['app.1.log', 'app.log'];
+    }
```
chr-hertel: this doesn't make sense like this unfortunately and should be extended to cover the monolog rotation used as example before as well, or at least more diverse examples.

you're using `glob` above which by default sorts alphanumerical => glob() returns the same array to be used in `usort` for both rows now

wachterjohannes:
You're right, thanks. `glob()` already returns entries alphabetically sorted, so touching `app.log` before `app.1.log` or the other way round fed `usort()` the exact same array both times, the "creation order" data provider was testing the same case twice.

Replaced it with the two rotation naming schemes from the PR description (`app.1.log` and `app-2024-01-01.log`), both of which glob() lists before `app.log` alphabetically, so the test now actually starts from the "wrong" order and proves the tie-break fixes it.

chr-hertel (approved):

chr-hertel:
Thank you @wachterjohannes.

PR #2513, opened 2026-09-11
[Mate] Make monolog-search's term parameter optional

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | yes
| Issues        | Found while hand-testing `monolog-search` against the demo app
| License       | MIT

`monolog-search`'s `term` was required, so filtering purely by level, channel or date, e.g. "just show me the errors" (`--level=ERROR`), failed outright. The documented workaround (pass an empty string) still ran the text-match code path, so a real `ERROR` entry could be missed if its message didn't literally contain the word "error".

`$term` is now `?string $term = null` and drops out of the generated schema's required list. When it's omitted or empty, `null` is passed to `SearchCriteria` so the text-match check is skipped entirely instead of matching against an empty string. `regex: true` without `term` now throws a clear `InvalidArgumentException` instead of failing on a missing required argument. Docs and skill files updated to describe `term` as optional. `CHANGELOG.md` is untouched since required-to-optional is additive, not a BC break.


chr-hertel (approved):

chr-hertel:
Thank you @wachterjohannes.

PR #2514, opened 2026-09-11
[Demo] Fix has_api_key falsely reporting true for unset env vars

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | none
| License       | MIT

`demo/mate/src/SymfonyAiFeaturesTool.php` computed `has_api_key` as `isset($settings['api_key'])`, but that value is almost always the raw YAML placeholder string `%env(OPENAI_API_KEY)%`, not the resolved secret. `isset()` on that string is always true, so the tool reported a key as present even when the real environment variable was empty or missing, actively misleading for anyone debugging a dead AI pipeline.

`has_api_key` now measures the real environment: it resolves `%env(NAME)%` placeholders (stripping processor prefixes like `string:`) against `$_SERVER`/`$_ENV`/`getenv()` and reports whether that resolved value is non-empty. A literal (non-placeholder) config value is checked directly. The secret value itself is never included in the output, only the boolean fact.


chr-hertel:
did you do a functional test here? maybe my claude is off, but the test that got executed failed with the standard demo setup:
image

wachterjohannes:
Good catch, thanks. The demo's `mate/config.php` never loaded any `.env` file (`mate.env_file` defaults to `null`), so `has_api_key` was reading a process environment that never saw your `.env.local` key. Pushed a fix that sets `mate.env_file` to `'.env'` there.

While tracing it I found the same bug one level up: the `mate init` template shipped `->set('mate.env_file', ['.env'])`, an array, but `ContainerFactory` only accepts a string and silently no-ops otherwise. Fixed the template and the matching docs example too, so this doesn't reintroduce itself on the next `init` or `discover`.

Verified against the demo with a real key in `.env.local`, `has_api_key` now correctly reports true.

chr-hertel (approved):

chr-hertel:
Thank you @wachterjohannes.

PR #2517, opened 2026-09-12
[Platform] Expose Albert's carbon footprint as result metadata

### LINKED ISSUE #2511: [Platform][Albert] How to retrieve the last stream data containing metadata after upgrading to 0.13?
After upgrading to Symfony AI 0.13, I am having trouble retrieving the last data sent by the provider when using a streaming response.

With the previous version, I was able to access the raw result and retrieve the final response data.

With 0.13, AgentInterface::call() now returns a lazy Execution, and I use:

```php
$execution = $agent->call($messages, [
    'stream' => true,
]);

foreach ($execution->asStream() as $delta) {
    // Process the streamed response
}
```
This works correctly for the text content: I receive TextDelta instances.

However, the provider sends additional information in the last SSE event, after the text chunks. For example, the Albert API sends:
```json
{
    "choices": [],
    "usage": {
        "prompt_tokens": 79,
        "completion_tokens": 6,
        "total_tokens": 85,
        "carbon": {
            "kWh": {
                "min": 5.7095659415205885e-06,
                "max": 5.7095659415205885e-06
            },
            "kgCO2eq": {
                "min": 4.5809521528648614e-07,
                "max": 4.5809521528648614e-07
            }
        }
    }
}
```
The problem is that:
```php
foreach ($execution->asStream() as $delta) {
    // ...
}
```
only seems to expose TextDelta instances in my case. I cannot access the final usage / metadata event.
Question

What is the recommended way in Symfony AI 0.13 to retrieve the final streamed data containing metadata such as usage?

Is there a specific Delta type, method, or API that should be used to access the last event of the stream?

More generally, how can I retrieve the complete response, including the final metadata, while still processing the response incrementally through asStream()?

Thank you in advance for your help!

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | Fix #2511
| License       | MIT

Albert reports the environmental footprint of a call next to the token usage, which the
OpenAI-compatible converter dropped so far. It is now exposed as `carbon` result metadata,
on buffered results and - via a `MetadataDelta` on the trailing chunk - on streamed ones:

```php
$result = $platform->invoke('openweight-small', $messages, ['stream' => true]);

foreach ($result->asStream() as $delta) { /* ... */ }

$carbon = $result->getMetadata()->get('carbon');
$carbon['kWh']['min'];     // 5.7095659415205885e-06
$carbon['kgCO2eq']['max']; // 4.5809521528648614e-07
```

* Adds a `CompletionsConversionTrait::yieldChunkMetadata()` extension point, so a bridge can
  promote the provider-specific payload of a stream chunk to result metadata
* Albert's gateway appends its own usage event after the one of the inference server it routes
  to, and the two disagree - only the last one is reported now, instead of both aggregating
  into a roughly doubled token count
* New `examples/albert/stream-carbon.php`, with a recorded cassette and replay golden

https://claude.ai/code/session_01BMZ5s8F9afcg4JipuJYPw7


OskarStark (approved):

gundesli:
Thank's @chr-hertel :)

PR #2519, opened 2026-09-13
[Store][Milvus] Fix minio image pull failing CI

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | n/a
| License       | MIT

The Milvus integration job fails on every PR with `pull access denied for minio/minio, repository does not exist`. Docker Hub no longer serves `minio/minio`, so the minio service container never starts, and the whole job fails on a connection refused to Milvus itself rather than anything Milvus-specific.

Pointed the image at `quay.io/minio/minio`, which mirrors the same release tags and is already how this same file pulls etcd. Fixed the identical reference in `examples/compose.yaml` too, used for local Milvus example runs.

Verified locally: pulled and started the full etcd/minio/milvus stack from quay.io and ran the bridge's integration suite against it, 10/10 passing.


wachterjohannes:
@chr-hertel FYI ...

PR #2527, opened 2026-09-14
[AI Bundle][Platform] Make the structured output validation groups configurable

marco-jouwweb:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | -
| License       | MIT

`ValidatorSubscriber` always validates the populated structured output object in the `Default` validation group. When the target class is an existing domain object that already carries `Default` constraints for other purposes (forms, API input), the only ways out are decorating the subscriber or moving every existing constraint to custom groups. As discussed with @chr-hertel on Slack, this PR makes the groups configurable instead.

### Platform

`ValidatorSubscriber` gets an optional `groups` argument accepting the same values as `ValidatorInterface::validate()` (a group name, a list of groups, or a `GroupSequence`). It is passed through `ValidatorResultConverter` to the buffered validation call and to `PartialObjectStreamListener::setValidator()`, so the final object of a streamed structured output is validated in the same groups.

```php
$dispatcher->addSubscriber(new PlatformSubscriber());
$dispatcher->addSubscriber(new ValidatorSubscriber(groups: ['ai']));
```

A single call can override them with the `validation_groups` option. Like `PlatformSubscriber` does for `response_format`, the subscriber consumes the option on `InvocationEvent` so it is never merged into the provider request, and applies it on `ResultEvent`:

```php
$platform->invoke('gpt-4o', $messages, [
    'response_format' => Invoice::class,
    'validation_groups' => ['ai', 'strict'],
]);
```

### AI Bundle

New `structured_output.validation_groups` option, applied to the `ai.platform.structured_output.validator_subscriber` service as the default groups. It is a top-level node (not under `platform`, whose children are the bridge types). An empty list keeps the current behavior; configuring groups while `symfony/validator` is missing throws the same kind of "requires package" exception as the other optional integrations.

```yaml
ai:
    structured_output:
        validation_groups: ['ai']
```

```php
final class Invoice
{
    #[Assert\NotBlank(groups: ['ai'])]
    public string $customer = '';

    #[Assert\Positive] // "Default" group only, skipped for model output
    public int $total = 0;
}
```

### Notes

- All new parameters are optional and last, so there is no BC break; `null` keeps validating in the validator's default group.
- Out of scope on purpose: `ValidatorConstraintsDescriber` still describes constraints of every group in the generated JSON schema.


src/platform/src/StructuredOutput/Validator/ValidatorSubscriber.php
```diff
@@ -23,21 +26,53 @@
...
 {
+    public const VALIDATION_GROUPS = 'validation_groups';
+
     private readonly ValidatorInterface $validator;
 
+    /** @var string|GroupSequence|array<string|GroupSequence>|null */
+    private string|GroupSequence|array|null $invocationGroups = null;
+
+    /**
+     * @param string|GroupSequence|array<string|GroupSequence>|null $groups The validation groups to validate the structured output in unless the "validation_groups" option is passed, or null for the validator's default group
```
marco-jouwweb, line 37: Not very happy with this union type.. WDYT?

chr-hertel: fine by me - it's honest: it's overloading. we take a bullet for DX

src/platform/src/StructuredOutput/Validator/ValidatorSubscriber.php
```diff
@@ -50,6 +85,7 @@ public function processResult(ResultEvent $event): void
         $converter = new ValidatorResultConverter(
             $deferred->getResultConverter(),
             $this->validator,
+            $this->invocationGroups ?? $this->groups,
```
chr-hertel, line 88: why do we need that `$invocationGroups` property - we have `$event->getOptions()` here as well

chr-hertel (changes requested):
API design looks good, only one question about state handling in the subscriber

src/platform/src/StructuredOutput/Validator/ValidatorSubscriber.php
```diff
@@ -23,21 +26,53 @@
...
+        $groups = $options[self::VALIDATION_GROUPS];
+
+        if (null !== $groups && !\is_string($groups) && !\is_array($groups) && !$groups instanceof GroupSequence) {
+            throw new InvalidArgumentException('The "validation_groups" option must be a string, an array or a GroupSequence.');
+        }
+
+        $this->invocationGroups = $groups;
+
+        // Consume the option, so it is not forwarded to the provider
+        unset($options[self::VALIDATION_GROUPS]);
```
marco-jouwweb, line 72: @chr-hertel question about this mechanic in general: I wonder if unsetting to not forward them is not going to bite in long term. I could imagine being able to properly distinguish between options & provider parameters would be cleaner. What do you think about introducing typed option keys, or dividing the array in 2 keys (e.g. a separate "provider" key in the array which contains all options that should be forwarded to the provider) at a later stage?

Just a nitpick but interested in your opinion:) I can imagine this does bring additional complexity because each provider has its own options, but I think it should be doable at one point.

chr-hertel: I def see where you're coming from, and that's always a bit painful with those array-wildcards or metadata bags, that buy us flexibility in abstractions. I think for now I'd prefer to be lazy on that and we can re-evaluate when it bites us - but "good instinct" like Claude would write :D

src/platform/src/StructuredOutput/Validator/ValidatorSubscriber.php
marco-jouwweb: `processInput()` unsets `validation_groups` so it is not sent to the provider. `Provider::invoke()` then builds the `ResultEvent` from those stripped options, so `processResult()` never sees the key on `$event->getOptions()`.

`$invocationGroups` keeps the per-call groups between the two events. `$this->groups` is only the subscriber default.

(This is the same split `PlatformSubscriber` makes for `response_format`: the invocation listener rewrites the options into something the provider understands, and keeps the original meaning in instance state for the result listener.)

PR #2528, opened 2026-09-14
[Agent] Suspend the execution while its model request is in flight

marco-jouwweb:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | yes
| Issues        | -
| License       | MIT

`Platform::invoke()` returns a `DeferredResult` backed by a lazy HTTP response, so several results created back to back have their requests in flight at the same time. `Agent::call()` returns a lazy `Execution`, which suggests the same at the agent level, but `Runner::run()` yields the `model_request` update *before* the request exists and then sends the request and reads the response in the very same generator step. Executions driven side by side (advancing each one by one update per round) therefore never have more than one request in flight: the first one sends and fully reads its response before the second one gets to send.

Two executions advanced in lock-step, against a fake endpoint that takes one second per request:

| Scenario                                          | before | after |
| ------------------------------------------------- | ------ | ----- |
| Platform: 2x `invoke()`, then 2x `getResult()`     | 1.0s   | 1.0s  |
| Agent: 2x `call()`, advanced in lock-step          | 2.0s   | 1.0s  |
| Agent: 2x `call()`, then 2x `getResult()`          | 2.0s   | 2.0s  |

The fix yields the `model_request` update *after* `invoke()` and the cancellation was activated, inside the existing `try`, so the generator suspends while the request is in flight but not read yet. The last row is unchanged on purpose: `getResult()` drives one execution to completion, which is what a lazy `call()` means.

Two details of the changed semantics:

- `model_request` now reports that the request was sent rather than that it is about to be sent. Cancelling from that update aborts the in-flight HTTP response through the already activated `Cancellation` instead of skipping the request.
- If `invoke()` throws (routing, contract), no `model_request` update precedes the exception.

The regression test drives two runs in lock-step against deferred results that record when they are resolved, and the docs get a short paragraph on running executions side by side.


chr-hertel (approved):

chr-hertel:
Thank you @marco-jouwweb.

PR #2428, opened 2026-08-22
[Platform][OpenAI] Add Realtime sessions support for bidirectional audio & voice agents

saifulferoz:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | no
| Issues        | Fix #2427
| License       | MIT

This PR introduces support for long-lived, bidirectional **Realtime sessions** (for speech-to-speech voice agents, live audio streaming, and WebRTC / WebSocket ephemeral token generation) in `symfony/ai`.

### Motivation
Instead of proxying high-bandwidth WebRTC audio streams through PHP, Symfony acts as the secure **session broker & ephemeral credential minter**, generating short-lived client tokens and session configurations (`client_secret`, instructions, voice settings, and modalities) for direct client-to-provider WebRTC connections.

### Key Changes
- **Capability**: Added `Capability::REALTIME_SESSION` to `Symfony\AI\Platform\Capability`.
- **Result DTO**: Added `RealtimeSessionResult` (`getId()`, `getClientSecret()`, `getExpiresAt()`, `getModel()`, `getVoice()`, `getModalities()`, `getContent()`).
- **OpenAI Model**: Added `Symfony\AI\Platform\Bridge\OpenAi\Realtime` model class and registered `gpt-4o-realtime-preview` & `gpt-4o-mini-realtime-preview` in `ModelCatalog`.
- **ModelClient & ResultConverter**:
  - `Symfony\AI\Platform\Bridge\OpenAi\Realtime\ModelClient` for `POST /v1/realtime/client_secrets` (OpenAI GA API).
  - `Symfony\AI\Platform\Bridge\OpenAi\Realtime\ResultConverter` converting responses into `RealtimeSessionResult`.
- **Factory**: Registered Realtime ModelClient and ResultConverter in `Bridge\OpenAi\Factory`.
- **Tests**: Comprehensive unit tests added for `ModelClient`, `ResultConverter`, and `Realtime` model.

### Example Usage
```php
use Symfony\AI\Platform\Bridge\OpenAi\Realtime;
use Symfony\AI\Platform\Bridge\OpenAi\Factory;

$platform = Factory::createPlatform($_ENV['OPENAI_API_KEY']);

$result = $platform->invoke(
    new Realtime('gpt-4o-realtime-preview'),
    'You are a friendly voice support assistant.',
    [
        'voice' => 'alloy',
        'modalities' => ['text', 'audio'],
    ]
);

[...]

Issue #123, opened 2025-07-15
`Expected array, received object` error in PHPStorm when following the MCP example tool


tom-hart-sky-uk:
Hi, I'm trying to add `mcp-bundle` to my Symfony 7.2 application . What I've done so far:

1. Added `"symfony/mcp-bundle": "@dev",`
2. Created `mcp.yaml`
```yaml
mcp:
    app: MyApp
    version: 0.0.1
    client_transports:
        stdio: true
```
3. Created the tool
```php
class CurrentTimeTool implements MetadataInterface, ToolExecutorInterface
{
    public function call(ToolCall $input): ToolCallResult
    {
        $format = $input->arguments['format'] ?? 'Y-m-d H:i:s';

        return new ToolCallResult(
            (new DateTime('now', new DateTimeZone('UTC')))->format($format)
        );
    }

    public function getName(): string
    {
        return 'now-time';
    }

    public function getDescription(): string
    {
        return 'Returns the current time in UTC';
    }

    public function getInputSchema(): array
    {
        return [
            'type' => 'object',
            'properties' => [
                'format' => [
                    'type' => 'string',
                    'description' => 'The format of the time, e.g. "Y-m-d H:i:s"',
                    'default' => 'Y-m-d H:i:s',
                ],
            ],
            'required' => [],
        ];
    }
}
```
4. Wired in the `ToolChain`
```yaml
    Symfony\AI\McpSdk\Capability\Tool\ToolExecutorInterface:
        class: Symfony\AI\McpSdk\Capability\ToolChain
        arguments:
            - [ '@App\CurrentTimeTool' ]

    Symfony\AI\McpSdk\Capability\Tool\CollectionInterface:
        alias: Symfony\AI\McpSdk\Capability\Tool\ToolExecutorInterface
```
5. Added the server to `mcp.json`
```json
{
    "servers": {
        "symfony": {
            "command": "php",
            "args": [
                "/Users/thr15/Development/Tom/HiveTrack/bin/console",
                "mcp:server"
            ]
        }
    }
}
```

Then, when trying a prompt of "Call the `now-time` tool", it tries to call it, but this is what I see in the MCP log (from the point of adding the server)
```
[2025-07-15 09:03:57.247][info]    [symfony]Stopping server symfony
[2025-07-15 09:03:57.247][info]    [symfony]Connection state: Stopped
[2025-07-15 09:03:57.247][info]    [symfony]Starting server symfony
[...]

PR #2531, opened 2026-09-16
[Examples] Drop object ids from dumped example output

chr-hertel:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | -
| License       | MIT

The `{#123}` handles VarDumper prints next to an object are `spl_object_id()` values, so they shift whenever anything allocates one more object earlier in the process. Adding `Capability::REALTIME_SESSION` in #2428 was enough to break the replay goldens of 21 examples across Cerebras, Anthropic, Gemini, Mistral, HuggingFace, OpenAI, OpenRouter, OVH and Scaleway - `main` is red today for that reason alone.

Instead of resyncing the ids, `bootstrap.php` now dumps with `withRefHandles(false)`, so the handles never enter the output. Repeated objects still get one, but a dump-local sequential index rather than a global handle.

The goldens are regenerated from the existing cassettes - no re-recording, no cassette touched - and every changed line is a handle line.

Verified by re-adding two `Capability` enum cases, the exact change that broke the goldens: suite stays green.

https://claude.ai/code/session_01QAbXZU6bQd34FvhzkCWNHi


examples/bootstrap.php
```diff
@@ -38,12 +38,35 @@
...
 use Symfony\Contracts\HttpClient\HttpClientInterface;
 
 require_once __DIR__.'/vendor/autoload.php';
 // Replay runs use the placeholder values of .env.test, which are also what recordings store.
 (new Dotenv())->loadEnv(__DIR__.'/.env', defaultEnv: 'replay' === ($_SERVER['CASSETTE'] ?? null) ? 'test' : 'dev');
 
+// The "{#123}" handles VarDumper prints next to an object are spl_object_id() values: they count
+// every object PHP allocated before, so adding an enum case or a service anywhere in the codebase
+// shifts them and breaks the recorded goldens of unrelated examples. They say nothing about the
+// example, so dumps drop them - repeated objects still get a handle, but a dump-local one.
```
chr-hertel:
```suggestion
// Configure VarDumper to drop ID handles like `{#123}` while dumping object instances
```

PR #2535, opened 2026-09-16
[Agent][MCP] Show an MCP server's tool error to the model

### LINKED ISSUE #2534: [Agent][MCP] A tool error reported by an MCP server never reaches the model
Follow-up to #2497, split out from [my comment there](https://github.com/symfony/ai/pull/2497#issuecomment-5698023202) as suggested.

**Problem.** When an MCP server reports a tool execution error (a `CallToolResult` with `isError: true`, which is what the SDK sends when a tool throws `Mcp\Exception\ToolCallException`), the model never sees the server's message. It only gets:

```
An error occurred while executing tool "demo_movie_details".
```

MCP puts tool errors inside the result so the model can see them and fix the call. Here the text is lost.

**Cause.** `McpToolbox::invoke()` throws `ToolCallException::returnedError()`. That exception doesn't implement `ToolExecutionExceptionInterface`, so `AbstractToolbox` wraps it in `ToolExecutionException::executionFailed()`, and `FaultTolerantToolbox` replaces it with the generic message.

**Expected.** The server's error text is the tool call result the model receives. Transport and connection failures (`callFailed()`, `listFailed()`) should stay hidden from the model, as they are now.

**Reproducer.** In the demo, give an agent `{ mcp_server: ... }` pointing at a server whose tool throws `Mcp\Exception\ToolCallException('No movie with slug "nope".')`. The model receives the generic message instead.

**A correction to my #2497 comment.** The cases I tested there (a missing required argument, a wrong argument type) take a different path. The SDK rejects schema violations as a JSON-RPC `-32602` error rather than an `isError` result, so they arrive as `callFailed()` and are masked for that reason. Whether invalid-params errors should also reach the model is a separate question, and I've left it out of the fix for this issue.

tacman:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #2534
| License       | MIT

An MCP server reports a tool execution error inside the result (`isError: true`) so the model can see it and correct the call. `McpToolbox` raised it as a plain `ToolCallException`. `AbstractToolbox` wrapped that in `ToolExecutionException`, and `FaultTolerantToolbox` then replaced it with `An error occurred while executing tool "..."`, so the server's message never reached the model.

`McpToolbox` now throws a `ToolErrorException`, which extends `ToolCallException` and implements `ToolExecutionExceptionInterface`. `AbstractToolbox` rethrows it as it is, and its tool call result is the server's message: the text content, or the structured payload as JSON, exactly what went into the exception message before.

Transport and protocol failures (`ToolCallException::callFailed()` and `listFailed()`) are unchanged and still hidden from the model. A test covers both sides through a `FaultTolerantToolbox`.

`ToolCallException::returnedError()` moves to `ToolErrorException::returnedError()`. The bridge is unreleased (added in #2497 for 0.14), so this has no BC impact and needs no changelog entry.


src/agent/src/Bridge/Mcp/Exception/ToolErrorException.php
```diff
@@ -0,0 +1,41 @@
...
+{
+    private string $detail;
+
+    public static function returnedError(string $serverName, string $toolName, string $detail): self
+    {
+        $exception = new self(\sprintf('Tool "%s" on MCP server "%s" returned an error: %s', $toolName, $serverName, $detail));
+        $exception->detail = $detail;
+
+        return $exception;
+    }
```
wachterjohannes: `$detail` currently has no default and is only set after construction. A constructor keeps it always initialized:

```suggestion
    public function __construct(string $message, private readonly string $detail)
    {
        parent::__construct($message);
    }

    public static function returnedError(string $serverName, string $toolName, string $detail): self
    {
        return new self(\sprintf('Tool "%s" on MCP server "%s" returned an error: %s', $toolName, $serverName, $detail), $detail);
    }
```

wachterjohannes (commented):
Fixes exactly the described problem: the server's error message now reaches the model instead of degrading into a generic string. Verified locally, all green. One nit inline.

chr-hertel (approved):

chr-hertel:
Thank you @tacman.

PR #2537, opened 2026-09-16
[AI Bundle] Keep an mcp_server-only tool list explicit

### LINKED ISSUE #2536: [AI Bundle] An mcp_server-only tool list still injects every #[AsTool] service
Follow-up to #2497, split out from [my comment there](https://github.com/symfony/ai/pull/2497#issuecomment-5698023202) as suggested.

**Problem.** An agent whose `tools:` list names only MCP servers still gets every service tagged `ai.tool`:

```yaml
ai:
    agent:
        mcp:
            tools:
                - { mcp_server: 'remotes.weather' }
```

In the demo, the `mcp` agent offers `similarity_search, clock, wikipedia_search, wikipedia_article, movie_search` alongside its remote tools. The demo's AGENTS.md says the opposite: "The `mcp` agent lists them as `mcp_server` tools, so it has no local tools at all."

**Why it matters.** It changes which tool the model picks. In a real `gpt-5-mini` chat, an agent meant to use its app's MCP tools answered a question about the app's movie catalog with the local `#[AsTool]` `movie_search` instead of the MCP `demo_movie_search`. There's also no way to configure an agent that has only MCP tools.

**Cause.** In `AiBundle`, the local toolbox only has its tools replaced when the list contains non-MCP entries (`if ([] !== $services)`). Otherwise it keeps the default argument of `ai.toolbox.abstract`, which is every tagged tool.

This looks intentional: `testMcpToolboxesSitNextToTheTaggedToolsOfTheLocalToolbox` asserts it, and the docs say remote tools are "also combined with `tools: true`". But it contradicts the documented rule right above it ("Tools are opt-in: if undefined, the agent gets no tools; use `tools: true` to inject all tools"), under which a list is explicit.

**Proposal.** A list naming only `mcp_server` entries gets only those servers' tools. To combine local and remote tools, list the local ones too, as you already can with `testMcpToolboxesSitNextToAnExplicitToolList`.

tacman:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | yes
| Issues        | Fix #2536
| License       | MIT

An agent whose `tools:` list named only `mcp_server` entries still received every service tagged `ai.tool`. The local toolbox only had its tools replaced when the list contained local entries, so otherwise it kept the tagged iterator of `ai.toolbox.abstract`.

That contradicts the rule the config states ("Tools are opt-in: if undefined, the agent gets no tools; use `tools: true` to inject all tools") and the demo's AGENTS.md, which says the `mcp` agent "has no local tools at all". It currently offers `similarity_search, clock, wikipedia_search, wikipedia_article, movie_search` too. In practice, a `gpt-5-mini` agent meant to use its app's MCP tools picked the local `movie_search` over the MCP `demo_movie_search` (details in #2536).

A list naming only MCP servers now injects no tagged tools. To combine local and remote tools, list the local ones next to the servers, which already works. `tools: true` is unchanged.

This reverses a deliberate choice from #2497, which a test and one docs sentence ("also combined with `tools: true`") described. Both are updated here. The bridge is unreleased, so no released behavior changes. If you'd rather keep "all tagged tools + MCP" available, an explicit flag for it would work too, and I'm happy to rework this that way.


wachterjohannes (approved):
Correctly scopes the fix: an mcp_server-only list no longer pulls in the tagged local tools, while tools: true and mixed lists are untouched, verified by config normalization and by a test that fails without the fix. All green locally.

wachterjohannes:
Thank you @tacman.

PR #2540, opened 2026-09-16
[AI Bundle] Give an agent the tools of the application's own MCP server

### LINKED ISSUE #2539: [AI Bundle] Let an agent use the tools of the application's own MCP server, in-process
Follow-up to #2497, split out from [my comment there](https://github.com/symfony/ai/pull/2497#issuecomment-5698023202) as suggested. This is the case behind #2003.

**Use case.** An application exposes its tools as an MCP server (`#[McpTool]` services under `mcp.servers`) so external agents can use them, and it also has its own chat page. The chat agent should use the same tools. Writing them a second time as `#[AsTool]` duplicates names, descriptions and schemas, and they drift. The demo has this duplication today: `movie_search` exists as both.

**What works today.** Since #2497, the agent can reach its own server as if it were remote:

```yaml
mcp:
    clients:
        self:
            servers:
                demo:
                    transport: stdio
                    command: ['php', '%kernel.project_dir%/bin/console', 'mcp:server', 'demo']
ai:
    agent:
        chat:
            tools:
                - { mcp_server: 'self.demo' }
```

It works, but the application spawns a copy of itself as a child process (about 0.3 s to start and list tools, paid again per request under php-fpm), or makes an HTTP request back into itself.

**Proposal.** Let `mcp_server` name one of the application's own servers directly, without a client: `{ mcp_server: 'demo' }`. The tools are then called in-process through the server's own stateless protocol (`Builder::buildStateless()`), so schema validation, result formatting and tool errors are exactly what external clients get, with no transport. A prototype in the demo returns identical results at 0.00-0.03 s per call, and a real `gpt-5-mini` chat works through it.

tacman:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | Fix #2539
| License       | MIT

An `mcp_server` tool entry naming a plain server, without a client, now references one of the application's own `mcp.servers`. Its tools are called in-process:

```yaml
# config/packages/mcp.yaml
mcp:
    servers:
        archive:
            registry: 'App\Mcp\'

# config/packages/ai.yaml
ai:
    agent:
        archive_chat:
            tools:
                - { mcp_server: 'archive' } # tools reach the model as "archive_<name>"
```

The agent gets exactly the tools, schemas and errors that external MCP clients see, with no second `#[AsTool]` layer and no connection. Before this, the only option was a client pointing back at the application itself: a stdio child spawning `mcp:server`, or an HTTP request into itself.

* `LocalServerToolset` implements the agent bridge's `ToolsetInterface` over the server's `Builder::buildStateless()`. Requests go through the same stateless protocol an HTTP client hits, so schema validation, result formatting and MCP App handlers behave identically. JSON-RPC errors (invalid params, unknown tool) surface as `ToolCallException::callFailed()` and `listFailed()`, like `ClientToolset`, and `isError` results pass through as results.
* The builder's assembled parts are memoized and shared with `build()`, which the MCP bundle's profiler and `debug:mcp` already call on the same service, so serving the server over HTTP and in-process in one request does not load its elements twice.
* The config validation now accepts `<server>` or `<client>.<server>`. `<client>.<server>` behaves as before.

Tested in the demo: `{ mcp_server: 'demo' }` on an agent, with no `mcp.clients` entry, gives it `demo_movie_search`, `demo_movie_details` and `demo_current-time`, and a real `gpt-5-mini` chat answered through them in-process.

Note: #2537 edits the paragraph directly above the new docs section, so whichever merges second will need a trivial rebase.


src/ai-bundle/src/Mcp/LocalServerToolset.php
```diff
@@ -0,0 +1,125 @@
...
+            'id' => ++$this->requestId,
+            'method' => $method,
+            'params' => $params,
+        ], \JSON_THROW_ON_ERROR), $headers);
+
+        $message = json_decode($result->toJson(), true, flags: \JSON_THROW_ON_ERROR);
+
+        if (\is_array($message) && \is_array($message['error'] ?? null)) {
+            throw new RuntimeException(\sprintf('%s (%s)', json_encode($message['error']['message'] ?? null), json_encode($message['error']['code'] ?? null)));
+        }
```
wachterjohannes, line 117: For the error case, `StatelessResult` already exposes `isError()` and a typed `message` object, decoding `toJson()` back into an array is an unnecessary detour here (the success path still needs it, since `$body` has no getter).

wachterjohannes (commented):
Solid design: sharing the Builder between the real MCP server and the in-process toolset avoids double-loading, and the request/client-name validation now rejects the ambiguous multi-part names it used to let through silently. Verified locally, all green. One nit inline.

@chr-hertel, tagging you for a second look: this adds new mcp_server config semantics (a bare name now means the app's own server, in-process) on top of the client/server config from #2497. Given you shaped that config, worth confirming the bare-name meaning doesn't collide with anything you had planned there.

chr-hertel:
Let's sort the purpose of this first in the issue please :pray:

tacman:
Purpose written up in #2539 — the concrete duplication this removes, and the syntax question is open there. Holding this PR until that's settled.

PR #2541, opened 2026-09-16
[MCP Bundle] Fix profiler tools panel for parameters without a top-level type

simoheinonen:
| Q             | A
  | ------------- | ---
  | Bug fix?      | yes
  | New feature?  | no
  | Docs?         | no
  | Issues        | -
  | License       | MIT

  The MCP profiler panel crashes when a tool has an input property without a top-level `type` key:

  > Key "type" for sequence/mapping with keys "maxLength, anyOf" does not exist in @Mcp/tools.html.twig at line 30.

  `tools.html.twig` read `prop.type is iterable` before the `?? 'any'` fallback applied. With `strict_variables` on (the default when `kernel.debug` is true, so on every dev profiler), any such property throws. The MCP server itself keeps working; only the panel breaks.

  Property schemas without a top-level `type` are valid JSON Schema and common. For example, API Platform's MCP integration describes nullable inputs like this:

  ```json
  {
      "type": "object",
      "properties": {
          "query": {
              "maxLength": 320,
              "anyOf": [{"type": "string"}, {"type": "null"}]
          }
      }
  }
  ```

  `oneOf`, `$ref`, `const` and plain `enum` schemas have no top-level `type` either.

  The template now:
  - checks that `type` exists before reading it,
  - for `anyOf`/`oneOf`, collects the member types (flattening nested type arrays and removing duplicates), so the example above shows `string|null` instead of crashing,
  - falls back to `any` when there's no type (e.g. schemas that are only `enum` or `$ref`).

  Added `ToolsTemplateTest`, which renders `@Mcp/tools.html.twig` with `strict_variables` enabled and checks the Type badge for plain, union, `anyOf`, `oneOf`, nested, enum-only and `$ref`-only properties. It fails on the old template with the error above.

  Found on 0.13.0; still present on `main`.

  ---

  Fix and test written with Claude; I reproduced the bug on our app and reviewed the change.


src/mcp-bundle/tests/Profiler/ToolsTemplateTest.php
chr-hertel, line 1: i wonder if it makes sense to join that with the DataCollectorTest --- basically: when we have a registry like _this_ (setup), we should have a rendered template like _that_ (assert). WDYT?

chr-hertel:
Thanks for reporting and fixing this right away @simoheinonen! can you share how your tool looks like - the parameter profile? do you actually have nullable? what's your experience how harnesses use that?

chr-hertel (approved):
reproduced with

```php
<?php

namespace App\Mcp\Tools;

use Mcp\Capability\Attribute\Schema;

class AnyOfProbeTool
{
    /**
     * Probe tool mimicking API Platform's nullable input schema.
     */
    #[McpTool(name: 'anyof-probe')]
    public function search(
        #[Schema(definition: ['maxLength' => 320, 'anyOf' => [['type' => 'string'], ['type' => 'null']]])]
        ?string $query = null,
    ): string {
        return (string) $query;
    }
}
```

and confirmed the patch works

chr-hertel:
Thank you @simoheinonen.

PR #2543, opened 2026-09-16
[Platform][Generic] Fix tool calls reported with a stop finish reason

### LINKED ISSUE #2542: [Platform][Generic] convertChoice throws a TypeError when a choice carries a null content
### Description

`Bridge/Generic/Completions/CompletionsConversionTrait::convertChoice()` documents the incoming payload as `content: ?string` in its own `@param` block, then passes that value straight into `TextResult::__construct(string $content)`. Any OpenAI compatible server answering `finish_reason: "stop"` with `"content": null` makes the bridge throw a `TypeError` instead of returning a result.

Servers do that when the assistant turn is a tool call: they populate `message.tool_calls`, leave `message.content` null, and report `finish_reason: "stop"` rather than `"tool_calls"`. Reported downstream against a self hosted gateway running GPT OSS 120B, where another model on the same gateway completed fine.

Two problems sit in that one method:

1. the null content is unguarded;
2. `tool_calls` is only read when `finish_reason === 'tool_calls'`, so guarding the content alone is not enough. I tried that in isolation: the crash goes away, the caller gets `TextResult('')`, and the tool call is silently dropped.

The streaming path in the same trait already detects tool calls by presence rather than by finish reason:

```php
protected function streamIsToolCall(array $data): bool
{
    return isset($data['choices'][0]['delta']['tool_calls']);
}
```

The buffered path does not, and that asymmetry is what this report is about.

### Reproduction

```php
use Symfony\AI\Platform\Bridge\Generic\Completions\ResultConverter;
use Symfony\AI\Platform\Result\RawHttpResult;
use Symfony\Component\HttpClient\MockHttpClient;
use Symfony\Component\HttpClient\Response\JsonMockResponse;

$payload = [
    'id' => 'chatcmpl_1',
    'choices' => [[
        'index' => 0,
        'message' => [
            'role' => 'assistant',
            'content' => null,
            'tool_calls' => [[
                'id' => 'call_1',
                'type' => 'function',
                'function' => ['name' => 'record_vulnerability', 'arguments' => '{"severity":"high"}'],
            ]],
        ],
        'finish_reason' => 'stop',
    ]],
];

$response = (new MockHttpClient(new JsonMockResponse($payload)))->request('POST', 'https://gateway.example/v1/chat/completions');

(new ResultConverter())->convert(new RawHttpResult($response));
```

### Actual

```
TypeError: Symfony\AI\Platform\Result\TextResult::__construct(): Argument #1 ($content) must be of type string, null given, called in vendor/symfony/ai-generic-platform/Completions/CompletionsConversionTrait.php on line 248
```

The same payload with `"content": ""` converts fine, and so does the same payload with `finish_reason: "tool_calls"`.

### Expected

A `ToolCallResult` carrying the single tool call. With no tool calls and a null content, a `TextResult('')` rather than a `TypeError`.

### Suggested fix

Key the tool call branch on the payload instead of the finish reason, and coalesce the content:

```php
if ([] !== ($choice['message']['tool_calls'] ?? [])) {
    return $this->withFinishReason(new ToolCallResult(array_map([$this, 'convertToolCall'], $choice['message']['tool_calls'])), FinishReasonMapper::map($choice['finish_reason']));
}

if (\in_array($choice['finish_reason'], ['stop', 'length'], true)) {
    return $this->withFinishReason(new TextResult($choice['message']['content'] ?? ''), FinishReasonMapper::map($choice['finish_reason']));
}
```

`FinishReasonMapper::map()` already accepts every reason and keeps the raw string, so the `finish_reason` metadata on that result still reads `stop` rather than being rewritten to `tool_calls`.

I ran that patch against v0.13.0 over five payloads, before and after:

* `stop`, null content, with tool_calls: `TypeError`, then `ToolCallResult` with finish_reason `stop`
* `stop`, null content, no tool_calls: `TypeError`, then `TextResult('')`
* `tool_calls`, with tool_calls: `ToolCallResult` both ways
* `stop`, `"hello"`: `TextResult('hello')` both ways
* `length`, `"truncated"`: `TextResult('truncated')` both ways

### Versions

`symfony/ai-generic-platform` v0.13.0 with `symfony/ai-platform` v0.13.0 on PHP 8.4, and current `main`, where the unguarded call sits at `src/platform/src/Bridge/Generic/Completions/CompletionsConversionTrait.php:266`.

Downstream report: https://github.com/vinceAmstoutz/symfony-security-auditor/issues/365

lazerg:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #2542
| License       | MIT

`convertChoice()` read `message.tool_calls` only when `finish_reason` was `tool_calls`, and passed `message.content` straight into `TextResult`, which requires a string. OpenAI-compatible servers routinely answer a tool call with `finish_reason: "stop"`, a populated `message.tool_calls` and a null `message.content`, and that combination threw a `TypeError` instead of returning a result. Reported downstream against a self-hosted gateway running GPT OSS 120B.

The tool-call branch now keys on the payload rather than on the finish reason, the way `streamIsToolCall()` already does on the streaming side of the same trait, and the text branch coalesces a null content to an empty string. `FinishReasonMapper::map()` keeps the raw reason, so such a result still reports `stop` rather than being rewritten to `tool_calls`.

A tool call with a null content and `finish_reason: "stop"` now converts to a `ToolCallResult`; a null content with no tool calls converts to `TextResult('')`. Payloads that already worked are unchanged.


chr-hertel (approved):
Thanks @lazerg - will add some commit with a follow up PR to get some things straight for streams and related bridges, see #2551

chr-hertel:
Thank you @lazerg.

PR #2548, opened 2026-09-18
[MCP Bundle] Stop the profiler collector from logging on every request

### LINKED ISSUE #2547: [MCP Bundle] Profiler collector floods the dev log with element registration logs on every request
In `dev`, every web request writes one `Registered manual tool … from handler …` debug line per MCP element, including requests that have nothing to do with MCP. With a handful of tools this fills the log faster than the application's own messages.

The cause is the profiler collector. `McpBundle::configureProfiler()` registers `mcp.data_collector` whenever `kernel.debug` is true, and `DataCollector::lateCollect()` calls `build()` on every server builder so the panel shows the full capability set on any profiled request. That build re-runs the loaders, and the SDK's `ArrayLoader` logs one debug line per registered element.

There is a workaround today: the builder definition carries the `mcp` monolog channel, so `channels: ["!mcp"]` on the main handler silences it. That also hides the logs of actual MCP requests, which is not what you want.

A better fix is to let the collector build with a null logger. That build only exists to populate the registry for the panel, so its log output is a duplicate of what a real MCP request writes. A failed registration still throws `ConfigurationException`, so nothing gets swallowed. Happy to send a PR.

wachterjohannes:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Docs?         | no
| Issues        | Fix #2547
| License       | MIT

The profiler collector builds every MCP server so the panel shows the full capability set on any profiled request. That build re-runs the loaders, and the SDK's `ArrayLoader` logs one debug line per registered element. So in `dev` every web request writes a `Registered manual tool …` line per tool, including the requests that have nothing to do with MCP.

The collector now builds through a clone of the builder with a null logger. The clone matters because the builder is not reset between requests: mutating the shared one would silence the real MCP requests in a worker as well. A failed registration still throws `ConfigurationException`, so nothing is swallowed.


chr-hertel (approved):

chr-hertel:
Thank you @wachterjohannes.

PR #2549, opened 2026-09-19
[AI Bundle][Platform] Add TypeSafe bridge

### LINKED ISSUE #2546: Support TypeSafe Jev
https://typesafe.ai/

Laravel already ships with it: https://github.com/laravel/ai/pull/1010

yoanbernabeu:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Docs?         | yes
| Issues        | Fix #2546
| License       | MIT

This adds a bridge for [TypeSafe](https://typesafe.ai/) and its Jev models, as requested in #2546.

Jev is not a chat model: you send it a *state* (text or structured data) and typed questions, and it answers with calibrated probabilities instead of generated text. There are three question types: `noul` (yes/no), `choice` (one option out of a set) and `score` (a rating along ordered levels).

```php
use Symfony\AI\Platform\Bridge\TypeSafe\Evaluation;
use Symfony\AI\Platform\Bridge\TypeSafe\Factory;
use Symfony\AI\Platform\Bridge\TypeSafe\Question\ChoiceQuestion;
use Symfony\AI\Platform\Bridge\TypeSafe\Question\NoulQuestion;
use Symfony\AI\Platform\Bridge\TypeSafe\Question\ScoreQuestion;

$result = $platform->invoke('jev-latest', new Evaluation($ticket, [
    'department' => new ChoiceQuestion('Which team should handle this?', [
        'billing' => 'Payment or subscription issues',
        'technical' => 'Bugs or integration problems',
    ]),
    'frustration' => new ScoreQuestion('How frustrated does the customer appear?', ['Calm', 'Frustrated', 'Very angry']),
    'is_urgent' => new NoulQuestion('Does the message convey urgency?'),
]));

$answers = $result->asObject();

$answers->getChoice('department')->getChoice();        // 'billing'
$answers->getChoice('department')->getProbabilities(); // ['billing' => 0.65, 'technical' => 0.35]
$answers->getChoice('department')->getConfidence();    // 0.48
$answers->getScore('frustration')->getScore();         // 1.0
$answers->getNoul('is_urgent')->isTrue();              // true
```

With the AI Bundle:

```yaml
ai:
    platform:
        typesafe:
            api_key: '%env(TYPESAFE_API_KEY)%'
```

### Design

Everything lives in the bridge, and the Platform component is left untouched: no new `Capability`, no new result class. The typed answers are bridge-local objects returned through an `ObjectResult`, the same way the HuggingFace bridge returns its `Output\*` objects. The input follows the Cohere reranker: the `ModelClient` takes a plain array payload, and `Evaluation` and the `*Question` objects are `JsonSerializable`, so the default `Contract` normalizes them without any custom normalizer. A raw `['state' => ..., 'questions' => [...]]` array works as input too.

I kept the question and answer types inside the bridge, like the Mistral OCR or HuggingFace classification results are today, because TypeSafe is the only provider exposing this so far. Reranking landed in Platform once several providers offered it. That said, this is only a proposal, and you know the direction of the project far better than I do: if you would rather have a generic capability like the one in the prototype linked from #2546, I will gladly step aside: I can close this PR, or reduce it to whatever remains useful on top of it (the bundle wiring, the example, the error handling validated against the live API). Whatever suits you best.

### Tested against the live API

The bridge was developed against the real API with a paid key, which shaped a few details:

* the catalog ships `jev-latest`, `jev-preview` and `jev-1.13.0`; a bare `jev` is rejected by the API with `Unknown model: jev`;
* errors are nested under a `detail` key, in three shapes: an `{error_type, message}` object, a plain string, and a list of violations for a 422. The `ResultConverter` reads all three, so exceptions carry the message of the API (`body.questions: Field required`) rather than a generic one;
* maps with numeric keys (question identifiers, choice options) are kept as JSON objects;
* the model that answered (`jev-1.13.0` for an alias) is reported through `TokenUsage::getModel()`;
* an answer missing a required key raises an exception instead of being silently read as `0.0`, since `confidence` is meant to gate decisions.

`examples/typesafe/evaluate.php` was recorded with `./runner --record typesafe`, so CI replays it without credentials.

One limitation worth knowing: Jev takes no messages, calls no tools and does not stream, so it cannot be the model of an `Agent`. The docs say so, and suggest using it next to one.

This bridge was developed with the help of Claude Code (Fable 5.1 model); I tested it against the live API and I stand behind its content.


src/platform/src/Bridge/TypeSafe/ResultConverter.php
```diff
@@ -0,0 +1,139 @@
...
+
+        $data = $result->getData();
+
+        if (!isset($data['answers']) || !\is_array($data['answers'])) {
+            throw new RuntimeException('Response does not contain answers.');
+        }
+
+        $answers = [];
+        foreach ($data['answers'] as $id => $answer) {
+            $answers[$id] = $this->convertAnswer((string) $id, $answer);
```
chr-hertel, line 69: crashes with `TypeError` if `$answer` is not an array - please check for that

yoanbernabeu: Good catch, fixed. A non-array answer now throws a `RuntimeException` instead of a `TypeError`.

While at it, I hardened the other spots that could fail the same way on a malformed payload: `probabilities`/`legend` must be arrays and `choice` a string before reaching the `fromArray()` factories, and a 422 violation with a non-array `loc` is now skipped instead of crashing `implode()`. All covered by new tests in `ResultConverterTest`.

chr-hertel (approved):
Thanks @yoanbernabeu - will land some minor tweaks while merging, but works well - thank you!

chr-hertel:
Thank you @yoanbernabeu.

RamyHakam:
Thank you, @yoanbernabeu! I was trying to use Jév yesterday with Symfony AI, and that’s what made me think about it.

